PR #75, opened 2015-03-24
Hardcode a manual into a subtopic.

alext:
This is a one-off until we have the ability to add items in the content-store to a subtopic.

The link title needs to be confirmed before this is merged.


KushalP:
I don't want to give this the thumbs up, but I will say that I am okay for this.

rboulton:
Looks functionally correct.

PR #77, opened 2015-04-23
Use shared mustache template for subsections

tijmenb:
The subsections template now uses the pre-existing mustache template used by the JS, reducing duplication.


KushalP:
:clap:

rboulton:
Top.  :sparkles:

PR #78, opened 2015-04-23
Use related topics from content-store for 'detailed guidance'

tijmenb:
Trello: https://trello.com/c/lTPNcdzI/18-collections-fe-uses-content-store-to-get-topic-links-for-mainstream-browse

Currently entries under "Detailed guidance" on the browse pages come from the whitehall API.

This change will first check content-store for related topics. When content-store doesn't have anything, it will fall back to using the content from whitehall.

The duplication will be removed when content-store serves all related content. Once that happens, we can clean up by removing the whitehall fallback in:

- `RelatedTopicList` model
- browse-columns.js

JS tests (which are outdated and don't run in CI) will be handled in a separate PR.

### Testing

Use the gds-content-schema dummy store. 

http://www.dev.gov.uk/browse/benefits/families

**Before** 

Detailed guidance lists "Support services for military and defence personnel and their families" (content from whitehall).

**After**

Detailed guidance lists "Universal credit" (content from content-store)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -25,6 +25,8 @@ end
 
 gem 'plek', '1.7.0'
 
+gem 'govuk-content-schema-test-helpers', '~> 1.1.0'
+
 group :test do
   gem 'capybara-webkit', '1.1.1'
   gem 'minitest-spec-rails', '4.7.6'
--- a/app/assets/javascripts/browse-columns.js
+++ b/app/assets/javascripts/browse-columns.js
@@ -171,12 +171,17 @@
     showSubsection: function(state){
       state.title = this.getTitle(state.slug);
 
+      var relatedTopics = state.relatedTopicsData.links && state.relatedTopicsData.links.related_topics
+      if (!relatedTopics || relatedTopics.length == 0) {
+        relatedTopics = state.detailedGuideData.results;
+      }
+
       this.setTitle(state.title);
       this.$subsection.mustache('browse/_section', {
         title: state.title,
         options: state.sectionData.results,
-        "detailed_guide_categories_any?": !!state.detailedGuideData.results,
-        detailed_guide_categories: state.detailedGuideData.results,
+        "detailed_guide_categories_any?": !!relatedTopics,
+        detailed_guide_categories: relatedTopics,
         showDescription: false
       });
       this.highlightSection('section', state.path);
@@ -250,6 +255,8 @@
       this.$el.attr('aria-busy', 'false');
       this.$el.find('a.loading').removeClass('loading');
     },
+
+    // FIXME: Once the content has been migrated to content store, remove whitehall code.
     getDetailedGuideData: function(state){
       var data = this.sectionCache('detailed', state.slug);
       var url = "/api/specialist/tags.json?type=section&parent_id=";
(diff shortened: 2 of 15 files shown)


edds:
The JS and markup changes look sane to me. Haven't had time to follow through the new ruby.

app/assets/javascripts/browse-columns.js
```diff
@@ -272,6 +286,28 @@
...
     },
 
+    getRelatedTopicsData: function(state){
+      var data = this.sectionCache('related', state.slug);
+      var url = "/api/content" + state.path;
+
+      var out = new $.Deferred()
+      if(typeof state.detailedGuideData !== 'undefined'){
+        out.resolve(state.detailedGuideData);
+      }else if(typeof data !== 'undefined'){
```
rboulton: really picky, sorry: spacing after } here

Only commenting because it's so neat elsewhere.

app/models/related_topic_list.rb
```diff
@@ -0,0 +1,38 @@
+# CLEANUP: 22/4 -  Related topics are currently in both the content-store and
```
rboulton: I'd skip the "22/4" here - we can use git to get the date (and that way we don't have to guess the data format)

I'd also say "FIXME" rather than "CLEANUP", since that's the more commonly used convention in our code (I think).

app/views/browse/sub_section.html.erb
```diff
@@ -6,9 +6,8 @@
     <%= render_mustache('browse/section',
       title: sub_section_tag.title,
       options: sub_section_artefacts,
-      detailed_guide_categories_any?: detailed_guide_categories.any?,
-      detailed_guide_categories: detailed_guide_categories,
-      showDescription: false
```
rboulton: Why don't we need the `showDescription: false` any more?

rboulton: Ah, I see it's the default.  Ok.

test/models/related_topic_list_test.rb
```diff
@@ -0,0 +1,72 @@
+require 'test_helper'
+
+describe RelatedTopicList do
+  describe "#related_topics_for" do
+    it "returns related topics for a subsection" do
+      content_store = content_store_with_response(links: {related_topics: [{ title: 'From Content Store', web_url: 'http://example.org/whatever'}]})
+
+      topic_list = RelatedTopicList.new(content_store, nil)
```
rboulton: I think it would be good to check here that, if the content_store service returns something, the whitehall service isn't even called.

rboulton:
A few very minor things to take a quick look at, but this is great work, thanks. :100:

test/models/related_topic_list_test.rb
tijmenb: If the code calls whitehall, the first test would fail because the `whitehall` service is nil for that case.

Gemfile
```diff
@@ -25,6 +25,8 @@ end
 
 gem 'plek', '1.7.0'
 
+gem 'govuk-content-schema-test-helpers', '~> 1.1.0'
```
jamiecobbett: I can see you configure it in test/test_helper.rb, but I can't see where you're actually using the gem.

rboulton: Thanks - this was temporarily used in development but backed out later. We'll move this to the test group, and add schema tests, in a subsequent PR.

PR #79, opened 2015-04-24
Use content schema examples in tests

tijmenb:
We now use the [content schema examples](https://github.com/alphagov/govuk-content-schemas/blob/master/formats/mainstream_browse_page), so that when the content schema of content-store changes, the tests will start failing.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -25,8 +25,6 @@ end
 
 gem 'plek', '1.7.0'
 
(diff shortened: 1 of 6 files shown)


jamiecobbett:
\o/ let me know if you want help sorting out the build.

rboulton:
Excellent work.

PR #80, opened 2015-04-24
Add jenkins-schema for contract testing

tijmenb:
Adding jenkins-schema allows the CI to run the tests in this project whenever the contract (govuk-content-schemas) changes.

Since the tests run in less than 10 seconds, we choose not to restrict the run to only tests that use the content schemas, like other projects do.

More about setting up the CI:

https://github.com/alphagov/govuk-content-schemas/blob/master/docs/contract-testing-howto.md

### DIFF
--- a/jenkins-schema.sh
+++ b/jenkins-schema.sh
@@ -1,0 +1,3 @@
+#!/bin/bash
+export REPO_NAME="alphagov/govuk-content-schemas"
+./jenkins.sh
--- a/jenkins.sh
+++ b/jenkins.sh
@@ -3,7 +3,7 @@ set -x
 export DISPLAY=:99
 export GOVUK_APP_DOMAIN=test.gov.uk
 export GOVUK_ASSET_ROOT=http://static.test.gov.uk
-export REPO_NAME="alphagov/collections"
+export REPO_NAME=${REPO_NAME:="alphagov/collections"}
 env
 
 function github_status {


jenkins-schema.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash
+./jenkins.sh
```
jamiecobbett: I think this needs a line like: https://github.com/alphagov/finder-frontend/blob/master/jenkins-schema.sh#L10

Otherwise how will the govuk-content-schemas commit on GitHub get the build status?

rboulton: jenkins.sh has these lines in it, already.

rboulton: Ah, scratch that, the significant feature is calling gh-status with `alphagov/govuk-content-schemas` as the repo name, rather than `alphagov/collections`.

So, we should probably make jenkins.sh read the "REPO_NAME" environment variable, use `alphagov/collections` as a default, and make jenkins-schema.sh set `REPO_NAME` to `alphagov/govuk-content-schemas`

jamiecobbett: @rboulton are you referring to this line? https://github.com/alphagov/collections/blob/master/jenkins.sh#L13 I believe that sets the build status for the collections commit, not the govuk-content-schemas commit.

PR #82, opened 2015-04-29
Refactor routing for email signups

tijmenb:
Rails 4.2 changes the way parameter segments are escaped: https://github.com/rails/rails/commit/5460591f0226a9d248b7b4f89186bd5553e7768f

This conflicts with `email_signup_path` using the subcategory slug (like `oil-and-gas/wells`) in a single `subtopic` parameter.

For example, `email_signup_path("foo/bar")` will generate the URL `/foo%2Fbar/email-signup`, not `/foo/bar/email-signup`.

This commit refactors the email signup routes to use the same URL structure as other subtopic-related routes.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -11,7 +11,6 @@ Collections::Application.routes.draw do
   get "/:sector/:subcategory/latest", as: "latest_changes", to: "subcategories#latest_changes"
   get "/:sector/:subcategory", as: "subcategory", to: "subcategories#show"
   get "/:sector", to: "specialist_sectors#show"
-
-  resources :email_signups, path: "/:subtopic/email-signups", only: [:create], subtopic: %r{[^/]+/[^/]+}
-  get "/:subtopic/email-signup", as: "email_signup", to: "email_signups#new", subtopic: %r{[^/]+/[^/]+}
+  get "/:sector/:subcategory/email-signup", to: "email_signups#new", as: "email_signup"
+  post "/:sector/:subcategory/email-signup", to: "email_signups#create"
 end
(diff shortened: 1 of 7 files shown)


config/routes.rb
```diff
@@ -11,7 +11,6 @@
   get "/:sector/:subcategory/latest", as: "latest_changes", to: "subcategories#latest_changes"
   get "/:sector/:subcategory", as: "subcategory", to: "subcategories#show"
   get "/:sector", to: "specialist_sectors#show"
-
-  resources :email_signups, path: "/:subtopic/email-signups", only: [:create], subtopic: %r{[^/]+/[^/]+}
-  get "/:subtopic/email-signup", as: "email_signup", to: "email_signups#new", subtopic: %r{[^/]+/[^/]+}
+  get "/:sector/:subcategory/email-signup", to: "email_signups#new", as: "email_signup"
+  post "/:sector/:subcategory/email-signup", to: "email_signups#create"
```
alext: Note: previously the GET route ended `/email-signups` (plural) whereas the POST route ended `/email-signup`.  I'm not sure there's a good reason for that, but if we're going to change it, I think that should be in a separate PR (and include a redirect from the old location...)

Another point.  While we're renaming the params, it's probably worth changing to use the correct naming (ie topic/subtopic).  They could be `topic_slug` and `subtopic_slug` to make it very obvious what's expected in them.

alext: > Note: previously the GET route ended /email-signups (plural) whereas the POST route ended /email-signup. I'm not sure there's a good reason for that, but if we're going to change it, I think that should be in a separate PR (and include a redirect from the old location...)

Although I've just noticed that your change makes this consistent with the routes above, so probably makes sense as is.

PR #89, opened 2015-05-20
Link topic title on subtopic page

tijmenb:
Links the topic title above the page title to the topic page.

## Before/after

_No change_

normal

## Hover state

hover

Trello:

https://trello.com/c/REqNtQec/94-user-facing-on-topic-and-subtopic-pages-remove-breadcrumbs


edds:
I don't think it is a great idea to have content which we are expecting users to be able to use as secondary text colour. It also goes against most of the patterns of how we use this context link. I can't see the trello ticket as I don't have permission but is there no other way of letting users find the hierarchy?

tijmenb:
Thanks for reviewing @edds. Not sure, but I believe user research showed that people are actually expecting the title to be linked, despite it also being in the breadcrumb.

Maybe @alextea @rboulton have something to add here? 

We're working on opening up the Trello board btw.

alextea:
@edds @tijmenb this was to go hand in hand with removing breadcrumb navigation from topics and subtopics (which may have been parked?). These are small incremental changes, and we will be monitoring the analytics/usage to see how they perform. There is an ongoing and more considered look at the navigation and orientation within topics and subtopics.

tijmenb:
For posterity: we noticed that [document pages also contain a link back to the topic page](https://www.gov.uk/government/publications/maritime-and-coastguard-agency-mca-safety-bulletins). 

Added in https://github.com/alphagov/whitehall/pull/1373

rboulton:
Given the precedent, and that we're actively reviewing this, and that I think this is a minor improvement, I'm going to merge this.  Agree it's not great, though.

PR #90, opened 2015-05-21
Hide "Home" breadcrumb on topic page

tijmenb:
From Trello: 

> ... topics have a single breadcrumb of "Home" which isn't any help as it doesn't give any context.

https://trello.com/c/REqNtQec/94-user-facing-on-topic-and-subtopic-pages-remove-breadcrumbs

### Before
before

### After
after

Not sure if we want to hide the breadcrumb on the subtopic page as well, @rboulton @alextea.

### DIFF
--- a/app/views/topics/show.html.erb
+++ b/app/views/topics/show.html.erb
@@ -1,6 +1,10 @@
 <% content_for :title do %><%= @topic.title %> - GOV.UK<% end %>
 <% content_for :page_class, "topics-page" %>
 
+<div class="header-context">
+  <%# This empty div makes the breadcrumbs dissapear. %>
+</div>
+
 <header class="page-header group">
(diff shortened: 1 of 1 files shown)


app/views/topics/show.html.erb
```diff
@@ -1,6 +1,10 @@
 <% content_for :title do %><%= @topic.title %> - GOV.UK<% end %>
 <% content_for :page_class, "topics-page" %>
 
+<div class="header-context group">
```
edds: You don't actually need the 'group' class here. The 'group' class was a way we used to use to make sure elements 'wrap' any floated elements contained within them.

tijmenb: Thanks! Rebased.

PR #91, opened 2015-05-26
Use an endpoint on collections app for mainstream browse calls

tijmenb:
Explained well [on the Trello ticket](https://trello.com/c/QYYN4uWY/129-use-an-endpoint-on-collections-app-for-mainstream-browse-ajax-calls):
 
> Currently, the Javascript enhancement for mainstream browse in the collections app uses the content API, whitehall content API, and content store API directly. Instead, make it use a new API on the collections app, so that we only have the logic for combining the results from these APIs in one place.

This PR achieves this in the following fashion:

1. Commit b08fdd0 sets up some partials that can be reused later.
2. Commit 1992297 makes the javascript request the partials from the app instead of going out to the APIs.
3. We reap the rewards in d3d6086 and d65fac3: integration tests for the javascript-enhanced functionality is suddenly fairly easy.
4. Further rewards in a523d13 (replace mustache with erb) and  0b7bef0 (clean up the views by extracting erb to partials).

Best reviewed per-commit.


rboulton:
We've had several offline but inconclusive discussions about this, so I thought I'd attempt to summarise them to allow us to make progress here.  TL;DR - unless anyone makes a strong objection now, I'll merge this tomorrow morning.

The changes here could be thought of as doing two things:

 - changing the javascript to use a single API endpoint, served by this app, instead of 3 different endpoints served by other apps (and thus removing duplication of logic across ruby and javascript for combining the responses from these endpoints).  Everyone seems to be in consensus that this is a good thing.

 - making the javascript obtain rendered fragments of HTML, rather than lists of values which need to be rendered to HTML.  There is general unease about this, but no concrete reasons for the unease.  Partly the unease is because this isn't what we've done so far, partly it's because the "received wisdom" is that this isn't the right thing to do - but we've not been able to come up with any good reasons why.

The reasons that we've implemented the rendering this way are:

 - it was easier to do it this way (rather than designing a JSON format for the data, writing code in the server to serve that format, and then writing code in JS to serve it).

 - it reduces the complexity of the resulting code.

The arguments I've read about which of these techniques are better all seem to be based on performance: whether the JSON or HTML form is bigger, whether the JS rendering is quick enough, etc.  I believe these arguments to be mostly or completely irrelevant: we're only transmitting fragments, so I'd expect the size difference to be minimal (no boilerplate, etc), and the speed of the JS rendering is essentially unnoticeable on modern browsers.  It is conceivable that an old browser with very slow JS might benefit from avoiding doing the rendering in JS.  It it also possible that we need to transmit more data using the HTML approach, so slow mobile connections might suffer from this (though as long as we have transport compression, this seems unlikely - there's no easy way to check the relative sizes without implementing the alternative approach).

We do not believe there are any accessibility implications of this, since the javascript is making exactly the same changes to the DOM as before; we're just performing the rendering to HTML fragments on the server end, rather than on the client.

Therefore, I plan to merge this tomorrow, and move on.

PR #99, opened 2015-06-04
Show curated lists for mainstream browse pages

tijmenb:
The links on a mainstream browse page can be curated into lists (see  https://github.com/alphagov/collections-publisher/pull/71).

This PR makes the change to make those lists visible on subsection pages:

screen shot 2015-06-04 at 13 42 39

Note that when the links are not curated into lists, the links will be displayed as usual, in alphabetical order. 

Trello: https://trello.com/c/dkaGFxTT/123


app/assets/stylesheets/views/_browse.scss
```diff
@@ -189,14 +189,22 @@ main.browse {
...
         }
+
+        .pane-inner.curated-list {
+          padding-left: 30px;
+        }
       }
       h1 {
         padding-left: 0;
       }
+      h2.list-header {
```
edds: I would be tempted to drop the tag from this selector.

app/models/sub_section.rb
```diff
@@ -0,0 +1,70 @@
...
+
+      OpenStruct.new(name: group.name, links: links)
+    end
+  end
+
+  def fallback_list_of_uncurated_links
+    @a_to_z = true
+
+    [OpenStruct.new(
+      name: 'A&#8202;to&#8202;Z',
```
edds: It would be good to keep the comment about what the `&#8202;` entity is next to this as it isn't immediately obvious

app/models/sub_section.rb
```diff
@@ -0,0 +1,70 @@
...
+    @sub_section_slug = sub_section_slug
+  end
+
+  def lists
+    @lists ||= curated_lists || fallback_list_of_uncurated_links
+  end
+
+  def a_to_z?
+    lists
+    @a_to_z
```
edds: I think this would be clearer having the logic to work out if it is an a_to_z list in here rather than using a return at the top of curated_lists, so something like

```
def a_to_z?
  !(subsection_from_content_store.details && subsection_from_content_store.details.groups)
end
```

That would also enable to you make the `lists` method slightly easier to read as currently the double `||` is slightly confusing. Something like

```
def lists
  if a_to_z?
    @lists ||= fallback_list_of_uncurated_links
  else
    @lists ||= curated_lists
  end
end
```

It is then very obvious as to what lists is going to contain and why it would contain it.

tijmenb: Yeah, I like that!

app/assets/stylesheets/views/_browse.scss
tijmenb: The "Detailed guidance" heading is also a h2.

tijmenb:
Thanks @edds! https://github.com/alphagov/collections/commit/fbdcf72afda199a5cf4a5e9a0663567905fb4084 addresses the feedback. I've inverted `a_to_z?` to make it easier to read.

edds:
That looks great, for some reason github hasn't updated this pull requests diff.

Is it worth also adding a test for `curated_links?` and `exists?`?

tijmenb:
Good point, those were tested in `lists?` and indirectly via a controller test. I've pulled them apart and added to `sub_section_test.rb`.

Rebased & force pushed.

app/models/sub_section.rb
```diff
@@ -0,0 +1,67 @@
...
+    tag_from_content_api.present?
+  end
+
+private
+
+  def curated_lists
+    subsection_from_content_store.details.groups.map do |group|
+      links = group.contents.map do |url|
+        tagged_items_from_content_api.find { |artefact| artefact.id == url }
+      end
```
alext: I think we need a `.compact` call here to remove any nil items.  This could happen if the group contains any items that are no longer tagged to this subtopic.

rboulton:
Nice work - and thanks for the reviews, @edds

edds:
:+1:

tijmenb:
Thanks @edds !

app/models/sub_section.rb
```diff
@@ -0,0 +1,72 @@
...
+  def lists
+    if curated_links?
+      @lists ||= curated_lists
+    else
+      @lists ||= fallback_list_of_uncurated_links
+    end
+  end
+
+  def curated_links?
+    subsection_from_content_store.details && subsection_from_content_store.details.groups
```
alext: This will return true if the subsection contains an empty list of groups.  We need it to return false in that case.

PR #102, opened 2015-06-12
Clean up controller and integration tests

tijmenb:
This PR improves the tests in two ways:

- Remove view testing from controller tests. Controller tests should not test the generated HTML or instance variables (the `assigns` thing). That's more easily done in integration tests - and in this case it's already tested by cucumber features.

The controller tests now only verify HTTP headers and response codes.

- Also removes unnecessary helpers for integration tests. In some cases helpers are being over-used in the integration tests. Causes the reader to look up a step definition, and then a helper, to understand what a cucumber step is doing. This commit removes the indirection of examples where helpers are only used once, or where the actual implementation is easier to understand than the helper.

Also adds simplecov gem to verify test overage (it stayed the same).


Gemfile
```diff
@@ -35,6 +35,10 @@ group :test do
   gem 'govuk-content-schema-test-helpers', '~> 1.1.0'
 end
 
+group :test do
+  gem 'simplecov'
```
alext: We'll want simplecov-rcov as well so that we can get Jenkins to graph the output etc.

test/test_helper.rb
```diff
@@ -16,6 +16,15 @@
 require 'gds_api/test_helpers/content_store'
 require 'gds_api/test_helpers/rummager'
 
+# Note that simplecov only reports test coverage for the /tests, not /features.
+if ENV["USE_SIMPLECOV"]
+  require "simplecov"
+
+  SimpleCov.start do
```
alext: I think you can just do `SimpleCov.start 'rails'`, and it'll set up all the necessary filters etc. for a Rails app.

PR #103, opened 2015-06-12
Repair Jasmine javascript tests

tijmenb:
The Javascript tests are currently not run in CI and therefore not maintained. Most of them have become out of date in the last year or so.

This commit repairs the tests to get them green again.

Run with `rake spec:javascript`.


edds:
:+1: This is good.

PR #107, opened 2015-06-16
Revert hardcoding a manual into a topic

rboulton:
This was implemented for a specific manual in
2a11b1f15b95cf7c1f05543735a4a4554a43b5a0 but is no longer required
(instead, an overview collection has been made, and the manual is linked
to from there).  Putting the manual into the topic was always a
temporary hack, and removing the hack makes our code simpler and more
maintainable.

Trello: https://trello.com/c/EKBlsx9w/188-remove-a-hardcoded-manual-from-a-topic-page


rboulton:
This reverts https://github.com/alphagov/collections/pull/75

PR #108, opened 2015-06-16
Refactor browsing

tijmenb:
This commit refactors the code for the browse pages. This hopefully makes upcoming changes easy, in particular the move to getting the data from the content store instead of the content api.

Things of note:

- Each controller method has its own model class (`IndexBrowsePage`, `TopLevelBrowsePage`, `SecondLevelBrowsePage`) with all information needed to render the page.
- There is some duplication between the three classes. Some of that will be removed when we move to displaying data from the content-store. Didn't want to use inheritance right now to make the code clearer.
- A lot of renaming of actions, methods and variables to be consistent with nomenclature in the content schemas.

Trello: https://trello.com/c/4iq0DhfM/111-in-collections-frontend-use-information-from-content-store-to-render-first-two-columns-of-a-mainstream-browse-page

Best reviewed in **Split view**.


app/models/second_level_browse_page.rb
```diff
@@ -0,0 +1,40 @@
+class SecondLevelBrowsePage
+  attr_reader :top_level_slug, :second_level_slug
+
+  delegate :title, :curated_links?, :lists, to: :browse_page_content_iten
+
+  def initialize(top_level_slug, second_level_slug)
+    @top_level_slug, @second_level_slug = top_level_slug, second_level_slug
```
edds: Think these would be clearer split out onto two lines.

tijmenb: Fixed in https://github.com/alphagov/collections/pull/109, thanks.

PR #112, opened 2015-06-18
jenkins.sh fixes

alext:
Various fixes to the script.  See individual commits for details.


bradleywright:
If we're approaching a good setup here, could you consider adding these to fix https://github.com/alphagov/govuk-rails-app-template/issues/6 ?

alext:
> If we're approaching a good setup here, could you consider adding these to fix alphagov/govuk-rails-app-template#6?

Sounds like a good plan.  I want to have this running in the wild for a little bit before doing so though, just to iron out any issues.

PR #113, opened 2015-06-18
Add note about routing changes

tijmenb:
The routes for this app are defined in two places - here and in collections-publisher. This note reminds us to change both. 

Would've prevented https://github.com/alphagov/collections-publisher/pull/93


jackscotti:
:+1:

PR #114, opened 2015-06-18
Add namespaced routes for topics

tijmenb:
Topics will be moving to the `/topic` namespace. To keep the content available when this app has been deployed, but collections-publisher has not, we'll serve topics from both the namespaced and non-namespaced URLs.

Trello: https://trello.com/c/EoWPp4qZ/187-move-topics-to-topic

After collections-publisher has deployed and the content has been migrated, we can remove the non-namespaced routes.

Related: https://github.com/alphagov/collections-publisher/pull/94


KushalP:
It would be good to have some (at least one) tests for this functionality. Even if the route switchover is going to happen in the very near future.

tijmenb:
Good idea @KushalP, thanks. Added routing tests.

PR #115, opened 2015-06-18
Get second level browse page title from content store

tijmenb:
Currently the second level browse page gets its title from the content-api. This creates an unnecessary request because we already have the title from the content store object, and we don't need the
call to content-api for anything else.

### DIFF
--- a/app/models/browse_page_content_item.rb
+++ b/app/models/browse_page_content_item.rb
@@ -1,8 +1,6 @@
 class BrowsePageContentItem
   attr_reader :slug, :item_from_content_store
 
-  delegate :title, to: :tag_from_content_api
-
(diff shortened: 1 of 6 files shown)


rboulton:
I'm surprised this doesn't require test changes - is this functionality tested anywhere?

tijmenb:
It's tested by this integration test: 

https://github.com/alphagov/collections/blob/master/test/integration/topic_browsing_test.rb#L76

After we've moved to the content-store we should probably review the tests to see if there's anything missing / better tested in unit tests.

alext:
> It's tested by this integration test:
> https://github.com/alphagov/collections/blob/master/test/integration/topic_browsing_test.rb#L76

That's a topic page test, not a mainstream browse page test...

PR #116, opened 2015-06-19
Remove all usage of collections-api

alext:
This updates this app to call content-store, contentapi and rummager directly as needed.  This will allow collections-api to be deprecated.

This may be easier to review commit-by-commit.


alext:
Note:  With these changes as they stand, this won't support topics under the new "/topic" prefix.  I'd like to address that in a follow-up PR as this one is already big enough.

I think there are additionally some existing issues with the /topic support due to generated URLs pointing at the old namespace regardless of which route it's accessed by.  This will cause problems for the email signup form submissions in particular.

app/controllers/subtopics_controller.rb
```diff
@@ -1,57 +1,45 @@
 class SubtopicsController < ApplicationController
   before_filter { validate_slug_param(:topic_slug) }
   before_filter { validate_slug_param(:subtopic_slug) }
-  before_filter :send_404_if_not_found
   before_filter :set_slimmer_format
 
+  rescue_from GdsApi::HTTPNotFound, :with => :cacheable_404
```
tijmenb: Should this be a `error_404` instead of a `cacheable_404`? I thought `cacheable_404` is for URLs that can't exist (malformed slugs).

alext: Good point, I'll update.

PR #118, opened 2015-06-22
Use content store for browse pages

tijmenb:
This commit makes the first two browse pages request its data from the content-store instead of content-api.

This **cannot be merged** until we've [republished all browse pages](https://trello.com/c/v1B0aJxW/185). 

Trello: https://trello.com/c/4iq0DhfM/111

Previous PRs:
- https://github.com/alphagov/collections/pull/109
- https://github.com/alphagov/collections/pull/108

Depends on https://github.com/alphagov/collections-publisher/pull/96 being merged.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -1,15 +1,9 @@
 module ApplicationHelper
   def browsing_in_top_level_page?(top_level_page)
-    params[:top_level_slug].present? && params[:top_level_slug] == top_level_page.slug
+    request.path.starts_with?(top_level_page.base_path)
   end
 
   def browsing_in_second_level_page?(section)
-    params[:second_level_slug].present? && full_slug == section.slug
-  end
-
-private
-
-  def full_slug
-    [params[:top_level_slug], params[:second_level_slug]].join('/')
+    request.path.starts_with?(section.base_path)
   end
 end
--- a/app/models/content_item.rb
+++ b/app/models/content_item.rb
@@ -1,0 +1,5 @@
+class ContentItem
+  def self.find!(base_path)
+    Collections.services(:content_store).content_item!(base_path)
+  end
+end
--- a/app/models/index_browse_page.rb
+++ b/app/models/index_browse_page.rb
@@ -14,8 +14,6 @@ class IndexBrowsePage
 private
 
   def content_store_item
-    @content_store_item ||= begin
-      Collections.services(:content_store).content_item!('/browse')
-    end
+    @content_store_item ||= ContentItem.find!('/browse')
   end
 end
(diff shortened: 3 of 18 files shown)


app/helpers/application_helper.rb
```diff
@@ -1,15 +1,9 @@
 module ApplicationHelper
   def browsing_in_top_level_page?(top_level_page)
-    params[:top_level_slug].present? && params[:top_level_slug] == top_level_page.slug
+    request.fullpath.starts_with?(top_level_page.base_path)
```
alext: Any reason to use `fullpath` over just `path` here?

config/initializers/extensions.rb
```diff
@@ -0,0 +1,8 @@
+class Array
+  # Use one_and_only instead of first when dealing with arrays that
+  # should only ever have one element. 
+  def one_and_only
+    raise "Not an array with a single element" unless size == 1
+    first
+  end
```
alext: I'm not convinced by this.  I think we should be making the frontend more defensive against unexpected data.  This will lead to 500 errors.  There are cases (particularly while migrating) where the "singular" links items could be unset, or set to empty array.

Also I think adding methods to core classes should be avoided if at all possible.

tijmenb: For context, it is used here:

https://github.com/alphagov/collections/blob/use-content-store-for-browse-pages/app/models/second_level_browse_page.rb#L28

Elements in the `links` hash from the content-store are _always_ arrays, even if it makes no sense for them to be arrays and there can only ever be one of them.

If we write `links.active_top_level_browse_page.first` it reads very awkwardly because, there can't be a first or last of singular thing. It tells the developer that we can expect more items, and that they may be used somewhere else. There is also a risk that elements are added to the array. When that happens, the `active_top_level_browse_page` becomes meaningless and we should be warned about it.

`Array#one_and_only` tries to communicate clearly to the reader that only one element will be ever expected in the array. Ideally, this problem should be solved by the content-store by returning just a hash and not an array. But from discussions I've gathered that that is not an option. That's why I'm proposing this solution. 

Open to other viewpoints!

You're right about the zero-length array, it could be `unless size > 1`

rboulton: We don't want frontend to return `500` errors if we could manage to return something reasonable.  If there were 2 items in the links field here, we could still serve the page (but might end up highlighting the wrong column, or similar).

I do think making our code clear that we don't expect the array to have more than one entry is good, though.  How about making it log an error to errbit in this case, but still return the first value in the array?  That way, we get reports in errbit, but the users still get their content. We've used that pattern in the past a few times, and it's been useful.

I think we should also make the schemas validate that there is at most one item in these arrays, too.

alext: > I think we should also make the schemas validate that there is at most one item in these arrays, too.

Good idea: https://github.com/alphagov/govuk-content-schemas/pull/86

tijmenb:
Because we need to republish all browse pages first, this is waiting for https://github.com/alphagov/collections-publisher/pull/96, to be deployed, among others. When that is deployed and `rake send_published_tags` is ran, this can be merged and deployed.

PR #120, opened 2015-06-23
Disable sessions

alext:
This app doesn't use sessions, and therefore has no need to set a
session cookie.  Disabling the session_store prevents the cookie from
being set.


tijmenb:
:+1:

PR #126, opened 2015-06-26
Clean up routes file

tijmenb:
Cleans up the routes file. Seen individual commits for changes.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1,14 +1,14 @@
 Collections::Application.routes.draw do
-
-  mount JasmineRails::Engine => '/specs' if defined?(JasmineRails)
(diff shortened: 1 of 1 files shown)


config/routes.rb
```diff
@@ -1,15 +1,15 @@
 Collections::Application.routes.draw do
+  # Note that this app only receives requests for routes registered with the
+  # content-store (by collections-publisher) - whenever the routes below
+  # change, also change the routes claimed by collections-publisher.
 
   mount JasmineRails::Engine => '/specs' if defined?(JasmineRails)
```
alext: While you're tidying this up, it's probably worth wrapping this in an `unless Rails.env.production?` just to make sure we never expose this accidentally.

PR #129, opened 2015-06-30
Enable pages to be used in iframes

dsingleton:
See https://github.com/alphagov/manuals-frontend/pull/123

Rails 4+ sets the X-Frame-Options header to SAMEORIGIN by default. This means
that the resource can only be put into an iframe on the same domain.

One problem this causes is that the side-by-side browser - which is a tool used
during transition - can't work correctly.

In the past we have decided that, with the exception of transaction start
pages, we want to allow content on GOV.UK to be iframed. Transaction start
pages are exceptional because we are particularly concerned about clickjacking
of the start button.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -22,5 +22,10 @@ module Collections
     # config.i18n.load_path += Dir[Rails.root.join('my', 'locales', '*.{rb,yml}').to_s]
     # config.i18n.default_locale = :de
     config.assets.prefix = "/collections/"
+
+    # Override Rails 4 default which restricts framing to SAMEORIGIN.
+    config.action_dispatch.default_headers = {
+      'X-Frame-Options' => 'ALLOWALL'
+    }
   end
 end


fofr:
Is the side by side browser still being used? Do we track usage of that tool?

dsingleton:
This change was a direct result of user feedback about the side-by-side browser not working.

fofr:
:+1:

Summary of conversation in meatspace: Sites are still transitioning and using this tool. It's probably worth adding analytics to https://github.com/alphagov/side-by-side-browser/blob/master/public/index.html to measure the "low but bursty" activity, although deploying the side by side browser is _tricky_.

dsingleton:
It's deployed in a non-standard way, but documented in the [Ops Manual](https://github.gds/pages/gds/opsmanual/infrastructure/side-by-side.html?highlight=side%20side).

PR #134, opened 2015-07-06
Remove slug validation code.

alext:
All routes to this application are now created as exact routes in the
router. This means that requests for pages with invalid or malformed
slugs will never make it to this application because the router will
return a 404 for them.


rboulton:
I like this removal of code. :+1:

PR #135, opened 2015-07-06
[refactoring] Unify mainstream browse models

alext:
This replaces the 3 existing models for mainstream browse pages with a single model to represent a browse page from the content-store. This reduces the duplication, and I think makes the encapsulation clearer.

There's scope for further reduction of duplication by having the Topic and MainstreamBrowsePage models share code, but that can wait for a future PR.

This is probably easiest to review commit at a time.


test/models/mainstream_browse_page_test.rb
```diff
@@ -0,0 +1,197 @@
...
+          "base_path"=>"/browse/bar",
+        },
+      ]
+
+      assert_equal 'Foo', @page.related_topics[0].title
+      assert_equal '/browse/bar', @page.related_topics[1].base_path
+      assert_equal 'All about foo', @page.related_topics[0].description
+    end
+
+    #it "returns empty array with no items" do
```
rboulton: Did you mean for these tests to be included but commented out?

alext: I was on the fence here.  These will replace the block below when we're no longer supporting detailed guide categories from whitehall.

alext: I've added a `FIXME` comment and rebased.

PR #136, opened 2015-07-06
Fix analytics on browse pages.

alext:
Since the switch to universal analytics, the GA events were being sent
with the wrong path (not taking the pushState stuff into account).

This fixes it by passing the path explicitly to analytics on these
events.

Trello: https://trello.com/c/7vxnLYkL/217


rboulton:
:+1:

Though having merged, I'm now wondering if there's a way to test this.  We have javascript tests for this app...

PR #138, opened 2015-07-13
Fix browse pages without guidance categories.

alext:
The Whitehall API returns 404, not empty array when there are no
detailed guidance categories for a given parent browse section.  This
was bubbling up to the controller, and being rescued causing the whole
page to 404.

This also updates the controller handling to rescue a specific error relating
to content items not being found, which should avoid masking errors like this in future.


rboulton:
Nice robust fix.

PR #144, opened 2015-07-16
Remove calls to detailed guide API from collections

tijmenb:
Now that no mainstream browse pages are linking to detailed guide categories, we do not need to call the whitehall API to find out about these links.

:fire:

Trello: https://trello.com/c/9wIGlKYV/251-remove-calls-to-detailed-guide-api-from-collections


alext:
:+1: Yay, good to see this dependency gone.

PR #145, opened 2015-07-17
Add `/topic` page

tijmenb:
Since we have introduced the `/topic` namespace for topics, an index page has been sorely missing. This page shows all topics in a list, comparable to the topic view.

## Before

screen shot 2015-07-17 at 14 26 43

## After

screen shot 2015-07-17 at 16 32 28

Trello: https://trello.com/c/L84pTDgJ/233-create-page-at-topic

:no_entry_sign: Don't merge. https://github.com/alphagov/collections-publisher/pull/124 has to be merged, deployed and republished first.

### DIFF
--- a/test/integration/topic_browsing_test.rb
+++ b/test/integration/topic_browsing_test.rb
@@ -15,6 +15,28 @@ class TopicBrowsingTest < ActionDispatch::IntegrationTest
     base.merge(params)
   end
 
+  it "is possible to visit the topic index page" do
+    content_store_has_item("/topic", {
+      base_path: "/topic",
+      title: "Topics",
+      format: "topic",
+      public_updated_at: 10.days.ago.iso8601,
+      details: {},
+      links: {
+        children: [
+          {
+            title: "Oil and Gas",
+            base_path: "/topic/oil-and-gas",
+          },
+        ],
+      }
+    })
+
+    visit "/topic"
+
+    assert page.has_content?("Oil and Gas")
+  end
+
   it "renders a topic tag page and list its subtopics" do
     content_store_has_item("/topic/oil-and-gas", oil_and_gas_topic_item.merge({
       :links => {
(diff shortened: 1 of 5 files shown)


test/integration/topic_browsing_test.rb
```diff
@@ -15,6 +15,23 @@ def oil_and_gas_topic_item(params = {})
...
+            "title" => "Oil and Gas",
+            "base_path" => "/topic/oil-and-gas",
+          },
+        ],
+      }
+    }))
+
+    visit "/topic"
+
+    assert page.has_content?("Oil and gas")
```
alext: This feels wrong. It would be better if this test could use more representative data.

tijmenb: True. Updated.

rboulton:
This is shiny.  But please don't merge until we have checked the copy for the beta label, too.

tijmenb:
Updated code + screenshot with new copy from John.

rboulton:
Great.  This is ready to merge, but there's a lot of outstanding things to be deployed on `collections-publisher`, and I don't want to create a deploy dependency while it's in that state.  We'll get this out early next week.

alext:
I may be missing something, but I don't think there's anything preventing this being merged and deployed before the collections-publisher stuff is merged.  This page will just 404 until collections-publisher sends the content item to the publishing-api and the routes therefore get created.

rboulton:
Good point @alext: I can't see any problem, either.

Also, we won't be able to deploy these today - I've booked a slot for deploying both of them tomorrow.  I won't merge this until @tijmenb has confirmed this analysis, though, in case we've both missed something.

rboulton:
On second thoughts, since Tijmen is away today, I've tested this out on preview - before the /topic page is published to the content store, this simply results in a 404 (the same as it does on production currently).  After the /topic page is published, this works nicely.

Therefore, I'm merging this now - there's no deploy dependency.

PR #146, opened 2015-07-21
Don't use the Content API to fetch subtopics

KushalP:
This PR stops using the Content API for fetching subtopics and instead uses Rummager. It should be read commit-by-commit and is split into the following sections:

1. 17eba07, 23199c0, and a765c4d refactor some test helpers so that they can be reused in later commits below
2. 8f7265f, 15edb2f and 6e893f0 introduce a new `ContentTaggedToTopic` class which will search Rummager for content tagged to a given topic
3. 3a34512, 4d2bb97, and 9b9a5b6 update the Rummager test stubs to conform to the kind of data that Rummager produces
4. fcb2f2a and 783da34 are the effective changes for this PR. They stop the use of the Content API and switch to using Rummager
5. a51d781 and 8b6c471 reduce duplication between tests and common `Struct` classes that were being used
6. 4ff6dfd, cab5043, f15e20a, and b9c0d00 rename the `ListSet::Specialist` and `ListSet::Section` classes that were created as part of this PR and make it explicit where they're creating their lists from

It's also worth saying that in the (very near) future Collections will stop using the Content API. The code has been written in a way that makes doing that easier. With the intention of having `ListSet::FromRummager` superceding the `ListSet` class. There's also an explicit comment in: https://github.com/alphagov/collections/pull/146/files#diff-56314c0207488ac5f06567d358c0d86eR33

Story: https://trello.com/c/28EDxi37

/cc @jennyd

### DIFF
--- a/app/models/list_set/from_rummager.rb
+++ b/app/models/list_set/from_rummager.rb
@@ -1,4 +1,4 @@
-class ListSet::Specialist
+class ListSet::FromRummager
   include Enumerable
 
   def initialize(slug, groups)
--- a/test/models/list_set_test.rb
+++ b/test/models/list_set_test.rb
@@ -1,8 +1,8 @@
 require "test_helper"
 
 describe ListSet do
-  it "uses ListSet::Section by default for lists" do
-    ListSet::Section.expects(:new)
+  it "uses ListSet::FromContentAPI by default for lists" do
+    ListSet::FromContentAPI.expects(:new)
       .with(anything(), anything(), anything())
       .returns([])
 
--- a/test/models/list_set_test.rb
+++ b/test/models/list_set_test.rb
@@ -1,131 +1,19 @@
 require "test_helper"
 
 describe ListSet do
+  it "uses ListSet::Section by default for lists" do
+    ListSet::Section.expects(:new)
+      .with(anything(), anything(), anything())
+      .returns([])
 
-  describe "for a curated subtopic" do
-    setup do
-      @group_data = [
-        {"name"=>"Paying HMRC",
-         "contents"=> [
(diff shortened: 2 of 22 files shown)


test/models/list_set_test.rb
```diff
@@ -1,131 +1,19 @@
...
-      assert_equal "A to Z", @list_set.first.title
-    end
+    assert_equal [], ListSet.new("", "", []).to_a
   end
 
-  describe "filtering out some formats" do
-    setup do
-      @list_set = ListSet.new('specialist_sector', 'business-tax/paye', [])
-    end
+  it "uses ListSet::Specialist for specialist_sector content" do
```
tijmenb: Should this be `ListSet::FromRummager`?

KushalP: Yes. I've pulled apart the previous commit which seemed to have got squashed together for this. It's now made clearer in b9c0d00.

tijmenb:
:+1: :dancer: Looks good!

app/models/list_set/from_rummager.rb
```diff
@@ -0,0 +1,49 @@
...
+          content.base_path + ".json" == URI.parse(api_url).path
+        end
+      end.compact
+
+      ListSet::List.new(group["name"], contents) if contents.any?
+    end.compact
+  end
+
+  def content_tagged_to_topic
+    Topic::ContentTaggedToTopic.new(@slug)
```
alext: This only returns a single page of results, not everything tagged to the topic.

KushalP: This is fixed by 54fbfcc.

PR #147, opened 2015-07-22
Adjust line spacing for curated lists

alextea:
The spacing between the subheadings and list items on curated browse topics wasn’t looking quite right.

This tightens the spacing between list items and balances the spacing with the list headers.

## Before

browse-before

## After

browse-after


tijmenb:
:+1:

PR #148, opened 2015-07-23
Handle base_path's in curated list data.

alext:
Currently the curated list data contains an array of contentapi API urls
that are used to match up agains the data returned from contentapi. We
want to start pulling this data from Rummager, and it will be easier to
do so if this data is sent as an array of base_path's. This therefore
updates collections to handle this data sent in either format so that we
can make the switch on the publisher end.

### DIFF
--- a/app/models/list_set/from_content_api.rb
+++ b/app/models/list_set/from_content_api.rb
@@ -36,8 +36,8 @@ class ListSet::FromContentAPI
 
   def build_curated_lists
     @group_data.map do |group|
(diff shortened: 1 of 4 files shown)


KushalP:
This looks good, but as per a discussion in person we're going to hold merging this once #146 is merged.

tijmenb:
This needs rebasing since https://github.com/alphagov/collections/pull/146 was merged.

alext:
Rebased...

PR #151, opened 2015-07-27
Serve mainstream browse links from rummager

tijmenb:
Rummager contains the links for mainstream browse pages. We should use those to render the pages instead of querying the content-api.

This commit removes collections' dependency on content-api.

Trello: https://trello.com/c/Xc9F0bfk/262-collections-reads-ms-browse-content-from-rummager


KushalP:
:heart_eyes: :+1:

app/models/topic/content_tagged_to_topic.rb
```diff
@@ -45,8 +45,16 @@ def search_params
     {
       start: 0,
       count: PAGE_SIZE_TO_GET_EVERYTHING,
-      filter_specialist_sectors: [@topic_slug],
+      filter_name => [@topic_slug],
       fields: %w(title link public_timestamp format),
     }
```
alext: Can we not mix hash styles within one hash please. (Another reason the fat-arrow syntax is superior - you avoid this sort of thing)

app/models/list_set.rb
```diff
@@ -30,16 +30,36 @@ def curated?
...
+          content.base_path == base_path
+        end
+      end.compact
+
+      ListSet::List.new(group["name"], contents) if contents.any?
+    end.compact
+  end
+
+  def content_tagged_to_topic
+    Topic::ContentTaggedToTopic.new(@tag_type, @tag_slug)
```
alext: This class is no longer named correctly because it now applies to browse pages as well as topics.

PR #152, opened 2015-07-27
Use RummagerSearch class for tag requests

tijmenb:
This is a follow up to https://github.com/alphagov/collections/pull/151.

RummagerSearch is used by Document `Topic::ChangedDocuments` and `ListSet` to request the taggings from rummager. This refactoring removes the reliance on inheritance for ChangedDocuments and clears up the code.

Trello: https://trello.com/c/Xc9F0bfk/262-collections-reads-ms-browse-content-from-rummager


test/models/list_set_test.rb
```diff
@@ -147,4 +147,66 @@
       assert_equal "A to Z", @list_set.first.title
     end
   end
+
+  describe "with a single page of results available" do
```
alext: This description doesn't really make sense in the new context.

PR #154, opened 2015-07-28
Comparing rummager to contentapi

bishboria:
In order to remove `contentapi` from the system, we need to ensure `rummager` has the same data as `contentapi`

See: https://trello.com/c/h161ilsH/261-validate-data-in-rummager


alext:
Superseded by https://github.com/alphagov/collections-publisher/pull/129

PR #165, opened 2015-09-28
Use `Services` module for external services

tijmenb:
This commit gets rid of the two different ways API clients are constructed in this application: via
`Collections.services(:api_name)` and `Collections::Application.config.api_name` (called in two different
ways too).

The Services module makes stubbing easier, means less typing and uses normal Rails conventions. It's also consistent across applications: `Services.content_store` means the same everywhere.

`RUMMAGER_HOST` used to be used to override which search instance to use in testing and development. This is not used anymore and can be safely removed.


mobaig:
Nice, like this pattern. Going to adopt this in email-alerts-api as well.

Merging!

PR #168, opened 2015-10-15
Add content ID to payload sent to email-alert-api on email signup

mobaig:
The email-alert-api has been updated to support both slugs and content
IDs. This change moves us closer to handling email subscriptions and
notifications solely via the content ID.

Trello: https://trello.com/c/ZD65QUBN/326-email-alerts-based-on-links-hash


mobaig:
cc: @elliotcm @rboulton

elliotcm:
:+1:

PR #171, opened 2015-10-29
Modify how formats are excluded from 'A to Z' lists

bishboria:
/browse and /topic pages may exclude different formats, so they require to be dealt with separately.

Ticket: https://trello.com/c/COmww1sJ/392-do-not-show-travel-advice-on-browse-pages

### DIFF
--- a/app/models/list_set.rb
+++ b/app/models/list_set.rb
@@ -2,12 +2,19 @@ class ListSet
   include Enumerable
   delegate :each, to: :lists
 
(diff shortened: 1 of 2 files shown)


rboulton:
This change looks good, but could do with some tests.

bishboria:
Yes, `list_set.rb` has no tests at all for filtering…

rboulton:
... eek :)

PR #172, opened 2015-11-13
Sort related topics alphabetically

tijmenb:
The ordering of items in the `links` hash is undefined (https://github.com/alphagov/content-store/blob/master/doc/content_item_fields.md#links). It's best if they are sorted alphabetically in the UI.

## Before

screen shot 2015-11-13 at 11 11 47

## After

screen shot 2015-11-13 at 11 11 02

### DIFF
--- a/app/models/mainstream_browse_page.rb
+++ b/app/models/mainstream_browse_page.rb
@@ -40,7 +40,7 @@ class MainstreamBrowsePage
   end
 
   def related_topics
-    linked_items("related_topics")
+    linked_items("related_topics").sort_by(&:title)
   end
 
   def slug
--- a/test/models/mainstream_browse_page_test.rb
+++ b/test/models/mainstream_browse_page_test.rb
@@ -126,9 +126,22 @@ describe MainstreamBrowsePage do
         },
       ]
 
-      assert_equal 'Foo', @page.related_topics[0].title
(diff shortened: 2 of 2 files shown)


jamiecobbett:
I seem to recall having a conversation with @rboulton and he said that the order of items in a links array was significant. This was in relation to identifying a content item's primary browse page.

tijmenb:
Hmm, confused by the content-store docs:

> be careful of relying on the order of links to imply a nuance or distinction in a relationship (such as a primary organisation or section). Use a separate link_type instead.

> The list_of_links is an array of content items, order is preserved

I'll investigate further.

tijmenb:
@jamiecobbett - you're right, the order is currently preserved. The reason the links in the screenshot aren't alphabetised is because they were made by the new tagging-app, which didn't send them alphabetised. So this is actually a issue in that app, not this one. Thanks for the heads up!

tijmenb:
Reopening this. 

Currently [collections-publisher sorts items by title](https://github.com/alphagov/collections-publisher/blob/37830fd561b9cd8c212a9c63b126ed93bb655dc1/app/presenters/mainstream_browse_page_presenter.rb#L15) before sending it to the publishing-api.

1. This means that when a topic title changes, the links will not be in alphabetical order in this app.
2. We'd like to get rid of the implicit ordering of links (an RFC for this is underway).

PR #182, opened 2016-07-07
Sort 2nd level browse pages

jackscotti:
content-store does not retain the order of the items that are sent in the links hash, therefore in order to allow custom 2nd level browse pages ordering we have to read the links order from the details hash.

Extra: 
- Added a test to make sure that top-level-browse-pages are always alphabetically ordered
- Upgraded Slimmer

Depends on: https://github.com/alphagov/collections-publisher/pull/206

paired with @rubenarakelyan

### DIFF
--- a/app/models/mainstream_browse_page.rb
+++ b/app/models/mainstream_browse_page.rb
@@ -20,7 +20,15 @@ class MainstreamBrowsePage
   end
 
   def second_level_browse_pages
-    linked_items("second_level_browse_pages")
+    links = linked_items("second_level_browse_pages")
+
+    if second_level_pages_curated?
+      links.sort_by do |link|
+        details["ordered_second_level_browse_pages"].index(link.content_id)
+      end
+    else
+      links
+    end
   end
 
   def second_level_pages_curated?
--- a/features/step_definitions/viewing_browse_steps.rb
+++ b/features/step_definitions/viewing_browse_steps.rb
@@ -1,33 +1,9 @@
-Given(/^there is an? (\w+) browse page set up with links$/) do |second_level_ordering|
-  top_level_browse_pages = [{ title: 'Crime and justice', base_path: '/browse/crime-and-justice' }]
+Given(/^there is an alphabetical browse page set up with links$/) do
   second_level_browse_pages = [{ title: 'Judges', base_path: '/browse/crime-and-justice/judges' }]
 
-  content_store_has_item '/browse', { links: {
-    top_level_browse_pages: top_level_browse_pages
-    }
-  }
-
-  content_store_has_item('/browse/crime-and-justice', {
-    base_path: '/browse/crime-and-justice',
-    links: {
-      top_level_browse_pages: top_level_browse_pages,
-      second_level_browse_pages: second_level_browse_pages,
-    },
-    details: {
-      second_level_ordering: second_level_ordering,
-    },
-  })
-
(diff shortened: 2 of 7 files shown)


app/models/mainstream_browse_page.rb
```diff
@@ -19,12 +19,16 @@ def active_top_level_browse_page
...
 
+  def ordered_second_level_browse_pages
+    details["ordered_second_level_browse_pages"]
+  end
+
   def second_level_browse_pages
     links = linked_items("second_level_browse_pages")
 
-    if second_level_pages_curated?
+    if second_level_pages_curated? and ordered_second_level_browse_pages
```
tijmenb: Prefer to use `&&` instead of `and` (govuk-lint would've told you, but it's not running on this repo).

If you're worried about this being `nil` (perhaps unnecessarily?) you could also do:

```ruby
details["ordered_second_level_browse_pages"].to_a.index(link.content_id)
```

jackscotti: The worry it's probably unnecessary given that collections-publisher is the only app able to publish mainstream browse pages. Might remove this commit after product review.

features/step_definitions/viewing_browse_steps.rb
```diff
@@ -77,3 +82,43 @@
...
+  ]
+end
+
+def add_browse_pages
+  content_store_has_item '/browse', { links: {
+    top_level_browse_pages: top_level_browse_pages
+  }}
+end
+
+def add_first_level_browse_pages(second_level_browse_pages, order_type)
```
MatMoore: `add_first_level_browse_pages` takes `second_level_browse_pages` as input.... 😖 

I'm wondering if there's a better name to use for this, or maybe you could make `second_level_browse_pages` a keyword argument so it's obvious this is actually second level?

features/step_definitions/viewing_browse_steps.rb
```diff
@@ -77,3 +82,43 @@
...
+  content_store_has_item('/browse/crime-and-justice', {
+    base_path: '/browse/crime-and-justice',
+    links: {
+      top_level_browse_pages: top_level_browse_pages,
+      second_level_browse_pages: second_level_browse_pages,
+    },
+    details: {
+      second_level_ordering: order_type,
+      ordered_second_level_browse_pages: second_level_browse_pages.map { |page| page[:content_id] }
```
MatMoore: Can `ordered_second_level_browse_page` ever be a subset of the the full list instead of the full thing? If so we should have a test for it. I would expect non curated pages to be ignored.

jackscotti: Not, it will never be a subset. Full list.

features/step_definitions/viewing_browse_steps.rb
jackscotti: I understand the pain in reading this. What if I rename the signature to:

`def add_first_level_browse_pages(links, order_type)`

features/step_definitions/viewing_browse_steps.rb
MatMoore: OK, I'm happy as long as collections publisher enforces this.

features/step_definitions/viewing_browse_steps.rb
MatMoore: If `links` is the full links hash then I think this is a lot better 👍 

Or `def add_first_level_browse_pages(links:, order_type:)`

PR #183, opened 2016-08-02
Use 12-factor Airbrake config

tijmenb:
We'll soon add environment variables to puppet to configure this. Until then Airbrake keeps working because alphagov-deployment overwrites this file.

https://trello.com/c/dKfYzDNx


rubenarakelyan:
👍

PR #185, opened 2016-08-10
Remove browse page from breadcrumbs

rubenarakelyan:
The browse index page at /browse currently displays a breadcrumb as “Home > Browse”, thereby linking back to itself. For consistency with breadcrumbs elsewhere on GOV.UK, this commit removes this link, so that the breadcrumb on the index page is simply “Home”.

The decision was taken not to display the breadcrumb as "Home > Browse" when a user clicks on a top-level browse page, because direct visits to the browse index page are very low and most users end up on browse pages from the GOV.UK home page, by clicking a link to go directly to a top level browse page. Therefore, the breadcrumb only changes once a user clicks on a second level browse page link.

Trello: https://trello.com/c/LCUGEogs/47-fix-breadcrumbs-on-browse-pages

With @jackscotti 

Before:
screen shot 2016-08-10 at 10 25 14

After:
screen shot 2016-08-10 at 10 25 30

### DIFF
--- a/app/controllers/browse_controller.rb
+++ b/app/controllers/browse_controller.rb
@@ -4,11 +4,7 @@ class BrowseController < ApplicationController
 
   def index
     @page = MainstreamBrowsePage.find("/browse")
-    set_slimmer_artefact_headers({
-      title: "browse",
-      section_name: @page.title,
-      section_link: @page.base_path,
-    })
+    set_slimmer_artefact_headers
   end
 
   def top_level_browse_page


tijmenb:
@rubenarakelyan is going to update the commit message with some backstory, why we made the choice to make https://www.gov.uk/browse & https://www.gov.uk/browse/births-deaths-marriages have the same breadcrumb.

rubenarakelyan:
@tijmenb: I've updated the commit and PR with a description of the reasoning behind not changing the breadcrumb for top level browse pages.

app/controllers/browse_controller.rb
```diff
@@ -4,11 +4,7 @@ class BrowseController < ApplicationController
 
   def index
     @page = MainstreamBrowsePage.find("/browse")
-    set_slimmer_artefact_headers({
-      title: "browse",
-      section_name: @page.title,
-      section_link: @page.base_path,
-    })
+    set_slimmer_artefact_headers
```
tijmenb: Does this method call still do something without args?

rubenarakelyan: Yup, it calls `set_slimmer_headers(format: 'browse')` regardless of presence/absence of args, but it sends an empty hash to `set_slimmer_dummy_artefact` (which sets the data to be used for breadcrumbs in Slimmer),.

tijmenb: 👍 excellent

PR #190, opened 2016-09-01
Properly create subscriberlists

tijmenb:
These parameters tell email-alert-api to create a "subscriberlist" with certain parameters. The parameters determine when the user receives an email. This used to be done by saying specifying the URL of a topic the user wants to subscribe to. Because URLs always change, this is being moved to be based on `content_id` in the `links` hash.

This app was still sending both `links` and `tags`, when email-alert-api only accepts one of these. Since a deploy this morning email-alert-api has been suddenly rejecting the payloads (as if the validation just got introduced) - we don't know why.

https://trello.com/c/Nmyl1NkF

### DIFF
--- a/app/models/email_signup.rb
+++ b/app/models/email_signup.rb
@@ -28,13 +28,10 @@ private
 
   def subscription_params
     {
-      title: subtopic.combined_title,
-      tags: {
-        topics: [subtopic.slug]
-      },
-      links: {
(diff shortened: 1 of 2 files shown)


elliotcm:
One thing that went out in the deploy was a move to reading from JSON columns instead of HSTORE.  Maybe the validation didn't work properly before.

PR #191, opened 2016-09-21
Add meta descriptions

rubenarakelyan:
This commit adds meta description tags to all email alert pages, the contents of which are set to the summary associated with the relevant alert. This adds descriptions to external search engine results when the pages are re-indexed.

Trello: https://trello.com/c/bbAwFAQ1/108-add-meta-descriptions-to-pages-missing-them


Davidslv:
do you happen to have a screenshot ?

rubenarakelyan:
screen shot 2016-09-21 at 13 51 17

tijmenb:
I think this would cause the email alert pages to have the same description as their topic page. Could we change it somehow like we did for that other description?

rubenarakelyan:
Yeah, good point. I've changed the meta description for the email signup page using some of the text already on the page.

PR #192, opened 2016-10-12
Filter topics and specialist sectors via their content IDs

carvil:
This PR changes how `collections` filters data using Rummager.

Instead of filtering both `specialist_sectors` (aka `topics`) and `mainstream_browse_pages` via their slugs, it now filters using their content IDs.

Rummager supports this via new fields called `topic_content_ids` and `mainstream_browse_page_content_ids`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,6 @@
 /db/*.sqlite3-journal
 
 # Ignore all logfiles and tempfiles.
-/log/*.log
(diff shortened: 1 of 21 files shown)


carvil:
@tijmenb yes, that wasn't supposed to be there. Please see https://github.com/alphagov/collections/pull/193

PR #194, opened 2016-10-26
Pick a random secret key base

tijmenb:
I'd like to move this repo to only use environment variables. To make that transition easier this commit sets the secret key base to a random string, so we don't have to move it into hieradata.

WDYT @alexmuller @boffbowsh?


boffbowsh:
This makes sense to me.

tijmenb:
https://github.gds/gds/alphagov-deployment/pull/1267 removes the key from alphagov-deployment.

PR #196, opened 2016-11-09
Add govuk_navigation_helpers gem

klssmith:
Add the `govuk_navigation_helpers` gem to collections so that breadcrumbs can be shown correctly when services and information pages are moved from whitehall.

Trello: https://trello.com/c/5kK0YosJ/276-move-services-and-information-pages-to-collections


rubenarakelyan:
Looks good

PR #197, opened 2016-11-18
Migrate services and information pages from whitehall to collections

rubenarakelyan:
Following on from https://github.com/alphagov/whitehall/pull/2818, this commit adds support in `collections` for rendering “services and information pages (for example, https://www.gov.uk/government/organisations/hm-revenue-customs/services-information), using content items published by `whitehall`.

Trello: https://trello.com/c/5kK0YosJ/276-move-services-and-information-pages-to-collections


app/controllers/services_and_information_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+    set_slimmer_analytics_headers
+  end
+
+private
+
+  def set_slimmer_analytics_headers
+    set_slimmer_headers(organisations: "<#{@organisation['analytics_identifier']}>")
+  end
+
+  def parsed_grouped_links
```
tijmenb: Can we pull this method (+ what it calls) into a separate class somewhere?

app/views/services_and_information/index.html.erb
```diff
@@ -0,0 +1,17 @@
+<% content_for :title do %>Services and information - <%= @organisation["title"] %> - GOV.UK<% end %>
```
tijmenb: This would probably read nicer as `content_for :title, "foo bar"`.

test/support/services_and_information_helpers.rb
```diff
@@ -0,0 +1,21 @@
+module ServicesAndInformationHelpers
+  def stub_services_and_information_content_item
+    content_store_has_item("/government/organisations/hm-revenue-customs/services-information",
```
tijmenb: Can you use the schemas gem for to generate this?

https://alphagov.github.io/govuk_schemas_gem/GovukSchemas/RandomExample.html#merge_and_validate-instance_method

It'll validate the data you put in the content store, so we can be certain that the stubbed data is real.

app/controllers/services_and_information_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+    @navigation_helpers = GovukNavigationHelpers::NavigationHelper.new(content_item)
+    @organisation = content_item.dig("links", "parent", 0)
+    @grouped_links = parsed_grouped_links
+    set_slimmer_analytics_headers
+  end
+
+private
+
+  def set_slimmer_analytics_headers
+    set_slimmer_headers(organisations: "<#{@organisation['analytics_identifier']}>")
```
tijmenb: We should look into using the analytics component for this (in a separate PR).

rubenarakelyan: Yup, I copied over the functionality but I'll raise a separate PR to update this.

app/controllers/services_and_information_controller.rb
rubenarakelyan: Yup, now in `ServicesAndInformationHelper`

app/views/services_and_information/index.html.erb
rubenarakelyan: Done

app/controllers/services_and_information_controller.rb
```diff
@@ -0,0 +1,18 @@
+class ServicesAndInformationController < ApplicationController
+  include ServicesAndInformationHelper
+
+  def index
+    base_path = "/government/organisations/#{params[:organisation_id]}/services-information"
+    content_item = Services.content_store.content_item!(base_path).to_hash
+    @navigation_helpers = GovukNavigationHelpers::NavigationHelper.new(content_item)
+    @organisation = content_item.dig("links", "parent", 0)
+    @grouped_links = parsed_grouped_links
```
tijmenb: This is better as class you can instantiate, instead of a helper. That makes it clear where the method comes from.

rubenarakelyan: I've moved it to a separate class which is instantiated from the controller

PR #198, opened 2016-11-22
Change default govuk-content-schemas branch

cbaines:
Change the branch from master to deployed-to-production. This will help
prevent changes being merged in to the Collections app that don't work
with the currently deployed govuk-content-schemas (this changes comes
from GOV.UK RFC 59).


Davidslv:
💯

PR #202, opened 2016-12-13
Fix flaky integration test

suzannehamilton:
MainstreamBrowsingTest was failing intermittently. It would pass if it happened to be run after one of the other integration tests, but would fail otherwise (example seed for failure case: 424).

I think the problem was that the test was using the Rails 'get' helper to get the page, rather than the Capybara 'visit' function. This meant that the Capybara 'page' variable was only defined if another integration test had already run. This means that the breadcrumb assertions were not testing the page they were intended to.


tijmenb:
Nice!

PR #203, opened 2016-12-13
Fix all Rubocop linter warnings

suzannehamilton:
This is in preparation for migrating the build to Jenkins 2. Jenkins 2 will require
us to add new build scripts which could optionally include Rubocop linting, so this
seems like a convenient point to add linting to this project.

Most of the Rubocop errors were trivial and could be fixed automatically.

One fix used Ruby's keyword arguments with no defaults, which are only supported by
Ruby 2.1 and above, so the Rubocop language level has been raised to 2.3 to be
consistent with the language level of the project.

I'll comment on the lines that needed manual updates, in case you want to review them in more detail.

### DIFF
--- a/app/models/list_set.rb
+++ b/app/models/list_set.rb
@@ -27,7 +27,7 @@ class ListSet
     @group_data.any?
   end
 
-  private
+private
 
   def lists
     if @group_data.any?
@@ -40,31 +40,30 @@ class ListSet
   def a_to_z_list
     [ListSet::List.new(
       "A to Z",
-      content_tagged_to_tag.reject do |content|
-        excluded_formats.include? content.format
-      end.sort_by(&:title)
+      content_tagged_to_tag
+        .reject { |content| excluded_formats.include? content.format }
+        .sort_by(&:title)
     )]
   end
 
   def curated_list
-    @group_data.map do |group|
+    curated_data = @group_data.map do |group|
       contents = group["contents"].map do |base_path|
-        content_tagged_to_tag.find do |content|
-          content.base_path == base_path
-        end
-      end.compact
+        content_tagged_to_tag.find { |content| content.base_path == base_path }
+      end
 
-      ListSet::List.new(group["name"], contents) if contents.any?
-    end.compact
+      ListSet::List.new(group["name"], contents.compact) if contents.any?
+    end
+
+    curated_data.compact
   end
 
   def content_tagged_to_tag
-    @content_tagged_to_tag ||= RummagerSearch.new({
+    @content_tagged_to_tag ||= RummagerSearch.new(
       :start => 0,
       :count => RummagerSearch::PAGE_SIZE_TO_GET_EVERYTHING,
       filter_name => [@tag_content_id],
-      :fields => %w(title link format),
-    })
+      :fields => %w(title link format))
   end
 
   def filter_name
--- a/features/step_definitions/browsing_topics_steps.rb
+++ b/features/step_definitions/browsing_topics_steps.rb
@@ -1,12 +1,12 @@
-Given /^there are documents in a subtopic$/ do
+Given(/^there are documents in a subtopic$/) do
   stub_topic_lookups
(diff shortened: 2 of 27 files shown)


features/step_definitions/browsing_topics_steps.rb
```diff
@@ -1,12 +1,12 @@
-Given /^there are documents in a subtopic$/ do
+Given(/^there are documents in a subtopic$/) do
```
suzannehamilton: These are now consistent with the other step definition files.

features/step_definitions/viewing_browse_steps.rb
```diff
@@ -115,28 +113,25 @@ def top_level_browse_pages
...
 
 def add_browse_pages
-  content_store_has_item '/browse', { links: {
+  content_store_has_item '/browse', links: {
     top_level_browse_pages: top_level_browse_pages
-  }}
+  }
 end
 
-def add_first_level_browse_pages(child_pages: second_level_browse_pages, order_type: order_type)
```
suzannehamilton: I have to admit, I'm not sure what this was originally trying to do, because neither `second_level_browse_pages` nor `order_type` were defined. But none of the tests were relying on the default values, so deleting them seemed to be the safest thing to do.

app/models/list_set.rb
```diff
@@ -40,31 +40,29 @@ def lists
...
 
       ListSet::List.new(group["name"], contents) if contents.any?
-    end.compact
+    }.compact
   end
 
   def content_tagged_to_tag
-    @content_tagged_to_tag ||= RummagerSearch.new({
-      :start => 0,
+    @content_tagged_to_tag ||= RummagerSearch.new(:start => 0,
```
Davidslv: would you mind moving `:start => 0` to another line?
Then change all the `:key =>` to `key: `

app/models/list_set.rb
```diff
@@ -40,31 +40,29 @@ def lists
...
+        content_tagged_to_tag.find { |content|
           content.base_path == base_path
-        end
-      end.compact
+        }
+      }.compact
 
       ListSet::List.new(group["name"], contents) if contents.any?
-    end.compact
+    }.compact
```
MatMoore: I'm not sure if it's actually in the styleguide, but the convention I've seen on GOV.UK is to use `do..end` for any multiline expression, and braces for single line ones, but only single line ones would be chained. (Hopefully someone else can confirm that!) 

Some of these could fit on one line, but for this one I think it would be clearer to extract methods/variables, as it's still quite hard to read as one expression.

features/step_definitions/viewing_browse_steps.rb
Davidslv: Are you saying that `add_first_level_browse_pages` is not in use?

Davidslv: This method is being used on line 11 and 45

Inside this method, `second_level_browse_pages` becomes `child_pages` (used in line: 125 of this commit) and `order_type`is in use on line 128

app/models/list_set.rb
suzannehamilton: I've moved `:start` to another line.

I've left the key syntax as-is because one of the keys is generated dynamically (line 65), so we're forced to use the `=>` syntax:

```
filter_name => [@tag_content_id],
```

```filter_name``` is actually a method call, because the filter name depends on the tag type.

I might have overlooked a different way to solve this, though. :)

app/models/list_set.rb
suzannehamilton: Thanks - I've introduced a variable and moved one of the `compact` calls inline.

(As an aside, I personally quite like the convention that you should use braces for expressions which are functional (i.e. map or filter something with no side effects), and `do..end` for expressions which, say, modify state or perform validation, because it makes the latter stand out as potentially dangerous. But I'll stick with the single/multiline convention if that's what you normally use. 🙂)

PR #204, opened 2016-12-14
Handle link groups with no title

rubenarakelyan:
Currently, if the “services and information” controller encounters a link group without a title then it returns a 500 error. This usually happens when links are not expanded by rummager due to unsynced data between rummager and the publishing-api. This commit removes any such groups to prevent such issues from affecting the “services and information” pages.

Trello: https://trello.com/c/aDBf32Dl/363-fix-crash-on-collections


rubenarakelyan:
Awaiting https://github.com/alphagov/gds-api-adapters/pull/632

tijmenb:
We would still know these cases because of the rummager check right?

rubenarakelyan:
Yeah, the rummager rake task reports all of these cases when run

PR #205, opened 2016-12-14
Add Jenkinsfile

suzannehamilton:
This adds the Jenkins configuration to allow pipeline builds on the Jenkins 2 CI server.

Currently blocked while we add phantomJS to the CI servers (alphagov/govuk-puppet#5273).

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,77 @@
+#!/usr/bin/env groovy
+
+REPOSITORY = 'collections'
+
+node {
+  def govuk = load '/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy'
+
+  properties([
+    buildDiscarder(
+      logRotator(
+        numToKeepStr: '50')
+      ),
+    [$class: 'RebuildSettings', autoRebuild: false, rebuildDisabled: false],
+    [$class: 'ThrottleJobProperty',
+      categories: [],
+      limitOneJobWithMatchingParams: true,
+      maxConcurrentPerNode: 1,
+      maxConcurrentTotal: 0,
+      paramsToUseForLimit: 'collections',
+      throttleEnabled: true,
+      throttleOption: 'category'],
+  ])
+
+  try {
+    stage("Checkout") {
+      checkout scm
+    }
+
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -0,0 +1,77 @@
+#!/usr/bin/env groovy
+
+REPOSITORY = 'collections-publisher'
```
carvil: Should this be `collections`?

Jenkinsfile
```diff
@@ -0,0 +1,77 @@
...
+      logRotator(
+        numToKeepStr: '50')
+      ),
+    [$class: 'RebuildSettings', autoRebuild: false, rebuildDisabled: false],
+    [$class: 'ThrottleJobProperty',
+      categories: [],
+      limitOneJobWithMatchingParams: true,
+      maxConcurrentPerNode: 1,
+      maxConcurrentTotal: 0,
+      paramsToUseForLimit: 'collections-publisher',
```
carvil: Same as above, should we change this to `collections`?

Jenkinsfile
```diff
@@ -0,0 +1,77 @@
...
+    stage("Set up content schema dependency") {
+      govuk.contentSchemaDependency()
+      govuk.setEnvar("GOVUK_CONTENT_SCHEMAS_PATH", "tmp/govuk-content-schemas")
+    }
+
+    stage("bundle install") {
+      govuk.bundleApp()
+    }
+
+    stage("rubylinter") {
```
carvil: Nice! Do you think we should also run the `sass` linter as well?

Jenkinsfile
suzannehamilton: Argh, copy-pasta. I've fixed it, thanks.

Jenkinsfile
suzannehamilton: Fixed.

Jenkinsfile
carvil: Ah, ok. That's fine, we should keep it in mind and try to fix those warnings at some point!

carvil:
:shipit:

PR #206, opened 2016-12-15
Fix SASS linter warnings

carvil:
This commit fixes most SASS linter warnings, more precisely 154
warnings.  There are 2 types of issues not yet solved, which are a bit
more complicated:

- IdSelector: Avoid using id selectors
- NestingDepth: Nesting should be no greater than 5, but was 6

These might have impact in the layout, so they need a bit more testing.
I will try to fix those in a separate commit later.

The aim is to then add `govuk-lint-sass` to the Jenkinsfile and make
sure we keep out SASS files consistent with the guidelines.


app/assets/stylesheets/helpers/_reset.scss
```diff
@@ -1,29 +1,115 @@
...
-small, strike, strong, sub, sup, tt, var,
-b, u, i, center,
-dl, dt, dd, ol, ul, li,
-fieldset, form, label, legend,
-table, caption, tbody, tfoot, thead, tr, th, td,
-article, aside, canvas, details, embed,
-figure, figcaption, footer, header, hgroup,
-menu, nav, output, ruby, section, summary,
-time, mark {
+// Borrowed from http://meyerweb.com/eric/tools/css/reset/
```
fofr: Its odd that collections has its own reset CSS, as we provide something similar through the govuk_template. I wonder what a wraith visual regression test would reveal if the reset was removed.

carvil: I will setup a test and see what happens

carvil: @fofr I've removed the `_reset.scss` file and I can see differences in some pages (resolution used: 1280x1500). Here is an example for https://www.gov.uk/browse

Live:

1280_phantomjs_live

Dev (without `_reset.scss`):

1280_phantomjs_dev

fofr: Ok. It looks like the `ul` margins aren't reset – feels like the problem that was had with the breadcrumbs component. Must be somewhere else in static that's doing that part of the reset.

Let's not worry about it for now. Thanks for looking into it.

app/assets/stylesheets/views/_topics.scss
```diff
@@ -118,11 +123,12 @@
   }
 
   .browse-container {
-    background-color: #fff;
+    background-color: $white;
```
carvil: Thanks @fofr I've added a new commit with that change.

fofr:
LGTM :+1:

Issue #210, opened 2017-01-04
Breadcrumbs are not wrapped in a container


nickcolley:
On smaller viewports the breadcrumbs do not have the correct padding.

![screen shot 2017-01-04 at 10 44 28](https://cloud.githubusercontent.com/assets/2445413/21639942/c956cbee-d26a-11e6-836a-12c9c12465ce.png)

tijmenb:
Closed per https://github.com/alphagov/collections/pull/213.

PR #213, opened 2017-01-11
Remove custom reset file and fix CSS throughout

carvil:
In this PR, we have:

- deleted the custom `_reset.scss` file;
- changed the layout to be the default `core_layout`;
- fixed the styles throughout in order to keep it working as expected;
- fixed an existing issue with breadcrumbs in tablet and mobile versions
  (since breadcrumbs weren't nested under a wrapper, they would not
  align properly - https://github.com/alphagov/collections/issues/210);
- fixed a double blue line that was in between the main body and the
  footer.

There are wraith tests to verify this here:

https://carvil.github.io/collections-wraith-tests/shots/gallery.html

As you can see from the gallery above, there are differences between most pages, but those differences are the fixes mentioned above.

Trello:
https://trello.com/c/ECPy2bbv/318-remove-custom-reset-scss-file-from-collections-in-favour-of-the-one-already-in-static

The wraith configuration can be found here: https://github.com/carvil/collections-wraith-tests

Paired with @carolinegreen on this


Davidslv:
Awesome work 👍

app/views/layouts/application.html.erb
```diff
@@ -14,9 +14,11 @@
 <body>
   <div class="wrapper" id="wrapper">
     <%= yield :breadcrumbs %>
-    <main id="content" role="main" class="content <%= yield :page_class %>">
-      <%= yield %>
-    </main>
+    <div class="grid-row">
```
fofr: I'm not sure the `grid-row` should be here.

app/assets/stylesheets/helpers/_layouts.scss
```diff
@@ -1,8 +1,8 @@
 .wrapper {
-  border-bottom: 10px solid $govuk-blue;
-
   .content {
+    @extend %contain-floats;
+
     display: block;
-    @include outer-block;
+    padding: 0 15px;
```
nickcolley: We should be using the measurement variables here I think

https://github.com/alphagov/govuk_frontend_toolkit/blob/master/stylesheets/_measurements.scss

app/views/topics/index.html.erb
```diff
@@ -8,7 +8,12 @@
 
 <% content_for :breadcrumbs do %>
   <%= render partial: 'govuk_component/breadcrumbs', locals: @navigation_helpers.breadcrumbs %>
-  <div class="header-context"><!-- Deliberately empty. This will prevent Static from adding a placeholder breadcrumb. --></div>
+  <div class="header-context">
+    <!--
```
fofr: Could this use ERB comments to prevent the text rendering in the page?

fofr:
Really happy to see this :100:
I've added some minor comments after merge, sorry.

app/assets/stylesheets/views/_topics.scss
```diff
@@ -123,7 +123,7 @@
   .browse-container {
     background-color: $page-colour;
     min-height: 500px;
-    padding: 0 30px 45px 30px;
+    padding: 0 0 45px 0;
```
nickcolley: Nit: We should use long-hand where possible so it's clear what's happening here.

This is the equivalent of an a function with many arguments, whereas the long hand is easy to understand at a glance.

http://csswizardry.com/2016/12/css-shorthand-syntax-considered-an-anti-pattern/

carvil:
@fofr we will address your comments in a new PR. 
I thought `grid-row` would be the expected markup there. Should it be something else, or should we remove it entirely?

app/assets/stylesheets/helpers/_layouts.scss
carvil: Will do in a new PR.

app/views/topics/index.html.erb
carvil: Yup, will change.

tijmenb: After https://github.com/alphagov/static/pull/879 is deployed we can get rid of this line entirely (and the others - https://github.com/search?q=org%3Aalphagov+%22Deliberately+empty.+This+will+prevent%22&type=Code)

fofr:
`grid-row` should be used when we want to divide a "row" of items, eg two-thirds one-third columns. It's a wrapper for grid classes.

Example in use:
https://www.gov.uk/government/case-studies/cash-transfers-help-for-those-who-need-it-most
screen shot 2017-01-11 at 13 44 54

app/assets/stylesheets/views/_topics.scss
carolinegreen: I'm not sure I agree with it being an anti-pattern.  Even post you've linked to it mentions things being less DRY as a caveat of using longhand.  I think it is easier to look at first glance if it's shorter, and if this were a function, four arguments wouldn't necessarily be that many.

fofr: I agree with @carolinegreen, as long as the zeroes for all those values need to be explicitly set. See also: https://github.com/alphagov/govuk-lint/commit/dcdd0a97307fd8f982c9bb5ccbdc1d11ee193526

nickcolley: I guess the way I see it is it's the difference between.

`setPadding(0, 0, 45px, 0)` and `setPadding({
  top: 0,
  left: 0,
  bottom: 45px,
  right: 0
})`

```css
padding: 0;
padding-bottom: 45px;
```

vs

```css
padding: 0 0 45px 0;
```

If you can write the explicit long hand it makes it much easier to know what's happening without having to remember the ordering.

Not worth pushing for this PR though, wider discussion. 👍

PR #215, opened 2017-01-11
Follow up css updates

carvil:
This is a follow up PR from https://github.com/alphagov/collections/pull/213

cc/ @fofr @nickcolley @tijmenb @carolinegreen

### DIFF
--- a/app/assets/stylesheets/helpers/_layouts.scss
+++ b/app/assets/stylesheets/helpers/_layouts.scss
@@ -3,6 +3,6 @@
     @extend %contain-floats;
 
     display: block;
-    padding: 0 15px;
+    padding: 0 $gutter-half;
   }
 }
--- a/app/assets/stylesheets/helpers/_layouts.scss
+++ b/app/assets/stylesheets/helpers/_layouts.scss
@@ -3,6 +3,5 @@
     @extend %contain-floats;
 
     display: block;
-    padding: 0 $gutter-half;
   }
 }
--- a/app/assets/stylesheets/views/_email-signups.scss
+++ b/app/assets/stylesheets/views/_email-signups.scss
@@ -1,6 +1,4 @@
 .wrapper .content.email-signup-new {
-  @extend %site-width-container;
-
   @include media(tablet) {
     margin-top: 30px;
   }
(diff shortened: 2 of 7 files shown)


fofr:
Does the wraith output look the same as before?

carvil:
@fofr Yes, here is the link: https://carvil.github.io/collections-wraith-tests/shots/gallery.html

The new commit is here: https://github.com/carvil/collections-wraith-tests/commit/39f9ea7439e14a1a829a3d7d07c527e63676dacd

fofr:
The email sign up content is indented strangely:

screen shot 2017-01-11 at 15 02 14

carvil:
@fofr that's true. I didn't know if that was on purpose, though. The latest page also has that indentation. Should the alignment be like in the other pages (i.e. aligned to the breadcrumbs)?

fofr:
Yep, breadcrumb and content should align on all viewports.

carvil:
@fofr We've updated the CSS to fix the email signup page.

In the process, we also fixed the `latest` page and came across other issues. Together with @alextea we changed a few things in topic and latest pages (e.g. the width of the `from` organisations). We introduced the `grid-row` class there and it does look a bit different, but now uses the new CSS we should be using.

We have also started using standard measurements in some places (e.g. padding in titles) instead of hardcoded pixels. This will explain some of the differences you will see in the wraith output.

Here are the latest tests: https://carvil.github.io/collections-wraith-tests/shots/gallery.html

What do you think?

nickcolley:
Looks like the 'is anything wrong' is misaligned now

screen shot 2017-01-11 at 17 27 03

fofr:
@nickcolley That's a known issue with government-frontend too. It's a tricky one that I think is acceptable for this PR.

carvil:
@fofr / @nickcolley the issue with the "Is there anything wrong with this page?" div comes from static and govuk_frontend_toolkit I think.

The [wrapper](https://github.com/alphagov/static/blob/9e8795b262c948a65b798029864cbe04f193262a/app/assets/stylesheets/helpers/_report-a-problem.scss#L88-L91) div extends `%site-width-container`, which in turn adds left/right margin of size `$gutter` in the tablet viewport (more details [here](https://github.com/alphagov/govuk_frontend_toolkit/blob/e03e7f56e1b70194310ee970a3ccf00bd9376b36/stylesheets/_grid_layout.scss#L27-L29)).

I don't think this issue should be fixed in this PR.

Are you ok to merge this change?

PR #216, opened 2017-01-17
Delete all 'page is in beta' messages from topic pages

suzannehamilton:
Topic pages are no longer in beta, and they will be deprecated soon when the new taxonomy-based navigation is added. This means that the 'BETA' tag is out-of-date, and can be deleted.

https://trello.com/c/oW7gSMoB/328-remove-beta-tags-from-topics-in-collections


carvil:
👍

PR #217, opened 2017-01-17
New navigation pages

carvil:
This is the first iteration on the new navigation pages for GOV.UK.

I've paired with @carolinegreen on this and this recreates the work @alextea did on the navigation prototype repo.

**NOTE FOR REVIEWERS**

This PR does not include all the changes needed to release the new navigation pages. However, as one can see from the code, these pages are behind a feature flag. We have enabled it on integration with https://github.com/alphagov/govuk-puppet/pull/5357 so we can have product reviews and at the same time allow multiple developers to work on top of these changes.

There is a comprehensive list of things left to do in Trello, which I encourage you to read too: https://trello.com/c/kFvjCgDX/306-as-a-developer-i-want-to-start-implementing-navigation-pages-in-production-to-understand-and-prepare-for-the-process-of-building

Some things are left as WIP on purpose as there are other trello cards to fix them. For example, some styles need rework, we disabled some lint checkers for the new styles until that work happens, and we need to filter down the tagged content to be guidance only.

The research will continue and it is likely things will change going forward.

**What the changes are**

We have built a new taxonomy for the Education theme and this will be the first area we will cover. Within education, we will show guidance content. The Education theme has been subdivided into multiple taxons, which can have parent taxons and child taxons as well.

The navigation pages are meant to help us drill down from high level taxons to more specific taxons.

The initial page for the top level taxon looks like this:

screen shot 2017-01-17 at 14 30 47

The page can be seen here: https://www-origin.integration.publishing.service.gov.uk/alpha-taxonomy/2b5d0d82-a099-41fc-b40b-a71f2404dcde-education-training-and-skills

This page displays a grid of child taxons because the current taxon has grandchildren. When a taxon doesn't have grandchildren, we display an accordion view:

screen shot 2017-01-17 at 15 34 06

The page can be seen here: https://www-origin.integration.publishing.service.gov.uk/alpha-taxonomy/6f6f5609-9075-4762-9d98-3d649a26348b-primary-curriculum-key-stage-1-and-key-stage-2

The initial set of tasks needed to get this up and running are described here:

https://trello.com/c/LyldAS1J/317-display-taxon-details-as-per-the-prototype-in-collections


.env
```diff
@@ -0,0 +1 @@
+ENABLE_NEW_NAVIGATION=yes
```
tijmenb: Won't this enable the new pages in production too?

carvil: No, because `dotenv` is only included in development and test in the Gemfile. This file will be ignored in all other environments.

app/controllers/taxons_controller.rb
```diff
@@ -0,0 +1,37 @@
+class TaxonsController < ApplicationController
+  before_action :redirect_to_www, unless: :new_navigaton_enabled?
+
+  def show
+    render :show, locals: {
```
tijmenb: Could we do this with an instance var? In content-tagger we do things in both ways, and the inconsistency isn't great. 

Perhaps it would be nice to create a single `NavigationPage` presenter, which exposes the methods from this hash.

config/routes.rb
```diff
@@ -18,4 +18,6 @@
   post "/topic/:topic_slug/:subtopic_slug/email-signup", to: "email_signups#create"
 
   get "/government/organisations/:organisation_id/services-information", to: "services_and_information#index", as: :services_and_information
+
+  get '/alpha-taxonomy/:taxon_base_path', to: 'taxons#show'
```
tijmenb: Is the namespacing intentional?

test/integration/taxon_browsing_test.rb
```diff
@@ -0,0 +1,164 @@
...
+      },
+      {
+        'title' => 'Content item 2',
+        'description' => 'Description of content item 2',
+        'link' => 'content-item-2'
+      },
+    ]
+  end
+
+  describe 'Browsing taxon with grandchildren' do
```
tijmenb: Have you considered using the cucumber-method-style from content-tagger integration tests for this?

test/support/taxon_helpers.rb
```diff
@@ -0,0 +1,82 @@
+module TaxonHelpers
+  # This taxon has grandchildren
+  def funding_and_finance_for_students(params = {})
+    {
```
tijmenb: If you wrap these items in a random example (`merge_and_validate`), they'll be more robust against changes in the schemas.

For example see: https://github.com/alphagov/contacts-frontend/pull/69/commits/885c2733501ccf1060ec2b8334eafdc6e9af95d7#diff-93c10073ef45397d34e244e5d41cc6df

.env
tijmenb: Ah, of course. Thanks!

app/controllers/taxons_controller.rb
carvil: I can see how this looks like a large set of variables being passed around. I don't think instance variables are a better approach, it's often confusing and if you miss setting it we get a `nil` rather than an actual error.

I'm happy to only send the `taxon` to the view and call out to the several pieces of information from within the view. What do you think?

I expanded these locals so people know what's really needed to render a show page. Wrapping it in another presenter could be less explicit I think.

config/routes.rb
carvil: I wanted to avoid doing a wildcard here. However, we will have to do that now that taxon URLs have changed. When that content tagger change is live we can start fixing this. There is a trello card for it here: https://trello.com/c/AgKA9XaM/333-render-taxons-and-their-new-base-paths-in-collections

test/integration/taxon_browsing_test.rb
carvil: I thought about it and I think it would be nice. Collections doesn't use rspec right now, so we would have to do that change before we could work on this. Would it make sense to do it in a separate PR, or would you rather have it here?

test/support/taxon_helpers.rb
carvil: Ah, that's nice, I didn't know we could do that. Ok, will change.

app/controllers/taxons_controller.rb
tijmenb: I totally missed that it's mostly from `taxon` that are being passed around, so calling those directly from the view would be nice.

I'm still troubled by the introduction of a new pattern here. In content-tagger the same happened, but we haven't followed through on moving everything to it, causing inconsistency.  Any thoughts around that?

test/integration/taxon_browsing_test.rb
tijmenb: Where would it need rspec?

carvil: Ah, sorry, you mean just calling out to ruby methods? Yup, we can do that here.

app/controllers/taxons_controller.rb
carvil: Ok, we can reduce the number of locals being passed.

I am in general against instance variables being set in controllers and then used in views. It's better to explicitly pass variables need so we 1) detect errors when we don't 2) avoid having partials inside partials relying on those instance variables.

If you agree with this, I think we can make an effort to migrate things as we touch them and also review PRs taking this into account. What do you think?

carvil: @tijmenb this commit reduces the locals being passed to the view: https://github.com/alphagov/collections/pull/217/commits/f6676fe42050f2f06e4c73f49560481da920ba3f

test/integration/taxon_browsing_test.rb
carvil: @tijmenb I've pushed a commit to change the tests to use this cucumber-like syntax: https://github.com/alphagov/collections/blob/644efc494c5281d069dbf6304bc1ce204877676a/test/integration/taxon_browsing_test.rb

carvil:
As discussed IRL, we are merging this now so the work can continue in other PRs.

tijmenb:
🎈

PR #219, opened 2017-01-19
Fetch breadcrumbs from the server in browse pages

carvil:
We have recently enabled click tracking on breadcrumbs via data
attributes. These data attributes are added by `static` and are needed
to know where people click. The old functionality in `collections` meant
that we would be adding list items to the breadcrumbs ourselves in the
app and therefore not including any of the necessary data attributes.

This commit re-fetches the breadcrumbs from either cache or the server
rather than modifying it.

This way, we always render breadcrumbs that have the expected attributes
set.

The tests are now simpler as well, since the logic of fetching
components from the server is already tested.

Trello: https://trello.com/c/mShtuEW6/340-bug-add-correct-data-attributes-to-the-breadcrumbs-in-collections

### DIFF
--- a/app/assets/javascripts/browse-columns.js
+++ b/app/assets/javascripts/browse-columns.js
@@ -12,7 +12,7 @@
     this.$root = this.$el.find('#root');
     this.$section = this.$el.find('#section');
     this.$subsection = this.$el.find('#subsection');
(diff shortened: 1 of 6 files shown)


tijmenb:
💯 💯 💯 💯 💯 💯 💯 💯 💯 💯  Excellent solution.

PR #222, opened 2017-01-23
Render taxons with new base path

carvil:
https://trello.com/c/AgKA9XaM/333-render-taxons-and-their-new-base-paths-in-collections

### DIFF
--- a/app/controllers/taxons_controller.rb
+++ b/app/controllers/taxons_controller.rb
@@ -19,10 +19,7 @@ private
   end
 
   def taxon
-    @taxon ||= begin
-      taxon_base_path = params[:taxon_base_path]
-      Taxon.find('/alpha-taxonomy/' + taxon_base_path)
-    end
+    @taxon ||= Taxon.find(request.path)
   end
 
   def navigation_helpers
--- a/app/controllers/taxons_controller.rb
+++ b/app/controllers/taxons_controller.rb
@@ -1,5 +1,5 @@
 class TaxonsController < ApplicationController
-  before_action :redirect_to_www, unless: :new_navigaton_enabled?
+  before_action :return_404, unless: :new_navigaton_enabled?
 
   def show
     render :show, locals: {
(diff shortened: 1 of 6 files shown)


app/controllers/taxons_controller.rb
```diff
@@ -21,7 +21,7 @@ def redirect_to_www
   def taxon
     @taxon ||= begin
       taxon_base_path = params[:taxon_base_path]
-      Taxon.find('/alpha-taxonomy/' + taxon_base_path)
+      Taxon.find("/" + taxon_base_path)
```
alecgibson: Doesn't the base path already include the leading slash?

carvil: I've realised I need to change the routes for that to happen. Let me try that.

carvil: It does not seem to work for the root path unfortunately. The controller params are:

```
{"controller"=>"taxons", "action"=>"show", "taxon_base_path"=>"alpha-taxonomy/ca860c58-665d-491d-88ed-26bf2e76fe19-school-education-5-to-18"}
```

Hence the extra `/` at the beginning.

alecgibson: How confusing. Ah well.

tijmenb: content-store solves the confusion by calling it `base_path_without_root`:

https://github.com/alphagov/content-store/blob/master/config/routes.rb#L4 

We can also use `request.path` like topics controller:

https://github.com/alphagov/collections/blob/master/app/controllers/topics_controller.rb#L3

carvil: Ok, I will use the `request.path` then in order to avoid manipulating the param, otherwise we would still need to do this: https://github.com/alphagov/content-store/blob/master/app/controllers/application_controller.rb#L26-L28

carvil:
@tijmenb / @alecgibson PR updated

alecgibson:
Looks good to me.

PR #227, opened 2017-01-25
Remove reliance on .env

alecgibson:
A change to puppet makes this environment variable available across all applications, so rely on that instead.

Waiting on this PR to be deployed: https://github.com/alphagov/govuk-puppet/pull/5395


cbaines:
Looks like this change will change the default behaviour for local development. You might want to consider adding a default value in to the app?

alecgibson:
@cbaines the development behaviour should remain the same once the Puppet change (linked) has deployed?

tijmenb:
Do we still need the gem now?

alecgibson:
@tijmenb we need the gem for tests, where we don't have the environment variable, but we need it set to display new content in integration tests.

cbaines:
> @cbaines the development behaviour should remain the same once the Puppet change (linked) has deployed?

I'm guessing that this is predicated on all development being only done on machines that govuk-puppet has been run on?

Regardless, I'm not aware of any specific guidance on how dependent applications should be on govuk-puppet, so I'm happy for this to be merged.

tijmenb:
In the tests we can just set `ENV['BLA']`, we don't need a dependency for that probably.

alecgibson:
> I'm guessing that this is predicated on all development being only done on machines that govuk-puppet has been run on?

Yes. Which is why @carvil suggested merging this a few days after the Puppet change has been deployed.

alecgibson:
@tijmenb @carvil The Puppet branch (https://github.com/alphagov/govuk-puppet/pull/5395) was merged 5 days ago - is that enough time to merge this change now?

PR #229, opened 2017-01-25
WIP - Rework css

carolinegreen:
WIP  - This isn't ready for review.  I'm also closing this as I don't agree with the WIP pull requests as an approach.  I haven't seen any other team do this, and I find them really confusing and that they act as noise, as you can't really review them, but they are there as if they are a pull request to be merged.

I'm going to put this on a trello card, and put down the branch this is on instead.  This is how I have worked in the past in other teams, and I thought this was how git was supposed to work.

Done:
- Using frontend toolkit mixins and fonts where possible.
-  Making measurements more consistent and remove measurements unsupported by older browsers.
-  Unnesting CSS
-  Remove unused CSS from collections for breadcrumbs.
-  Cut down CSS and unused selectors.
-  Added tab index for now to subheading, but I'm not sure about the markup around this currently.

Still to do:
- Ensure markup is semantic - I'm not sure the subheading is being handled the best way.  I'm not even sure if it should be referred to as a subheading as it is kind of a subheading/navigation element and breadcrumb/orientation element at the same time.  Added in a tab index to the span for the time being, apparently this helps with screenreaders but isn't completely ideal so I want to rethink this.
- Make sure images are supported by older browsers - We are using SVGs but don't have png fallbacks currently which are required for ie8.
- Stop the grid from breaking without nth selectors - Nth selectors are not supported by ie 8, so I need to find another way of making sure the grid doesn't break..
-  Check with Alex how the spacing should be now, as this has changed recently in the prototype.
-  Work needs to go into this on different viewports.

### DIFF
--- a/app/views/taxons/_page_header.html.erb
+++ b/app/views/taxons/_page_header.html.erb
@@ -1,17 +1,16 @@
 <div class="grid-row">
   <div class="column-two-thirds">
     <div class="page-header">
+      <span  role=”link” tabindex=”0″ class="sub-heading">
+        <% if parent_taxon.present? %>
+          <%= link_to parent_taxon.title, parent_taxon.base_path %>
+        <% else %>
+          <%= link_to 'Home', Plek.find('www') %>
+        <% end %>
+      </h2>
       <h1>
-        <span class="sub-heading">
-          <% if parent_taxon.present? %>
-            <%= link_to parent_taxon.title, parent_taxon.base_path %>
-          <% else %>
-            <%= link_to 'Home', Plek.find('www') %>
-          <% end %>
-          </span>
         <%= taxon.title %>
       </h1>
-
       <p><%= taxon.description %></p>
     </div>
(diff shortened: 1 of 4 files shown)


app/views/taxons/_page_header.html.erb
```diff
@@ -1,17 +1,16 @@
 <div class="grid-row">
   <div class="column-two-thirds">
     <div class="page-header">
+      <span  role=”link” tabindex=”0″ class="sub-heading">
```
fofr: It looks like this line is using the wrong type of quotes.

carolinegreen: This is why I closed it and didn't want it reviewed.  I find the idea of putting out a pull request specifically for it not to be reviewed, really weird.  I don't even know if I am keeping this.  Which I think is a waste of your time as a reviewer as well.  Thanks for pointing it out though.

PR #230, opened 2017-01-31
Rework css

carolinegreen:
There are no screenshots attached as the changes are not visual.  Whitespace is slightly different as this hasn't yet been finalised by Alex and will have to be tweaked.

This pull request covers:

-  Using frontend toolkit mixins and fonts where possible.
-  Making measurements more consistent and remove measurements unsupported by older browsers.
-  Unnesting CSS
-  Remove unused CSS from collections for breadcrumbs.
-  Cut down CSS and unused selectors.
-  Added tab index for now to subheading, but I'm not sure about the markup around this currently. 

Future pull requests will cover:

-  Ensure markup is semantic - I'm not sure the subheading is being handled the best way.  I'm not even sure if it should be referred to as a subheading as it is kind of a subheading/navigation element and breadcrumb/orientation element at the same time. Added in a tab index to the span for the time being, apparently this helps with screenreaders but isn't completely ideal so I want to rethink this.
-  Make sure images are supported by older browsers - We are using SVGs but don't have png fallbacks currently which are required for ie8.
-  Stop the grid from breaking without nth selectors - Nth selectors are not supported by ie 8, so I need to find another way of making sure the grid doesn't break..
-  Work needs to go into this on different viewports.

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -8,7 +8,6 @@
 
 @import "helpers/layouts";
 @import "helpers/pagination";
-@import "helpers/govuk-component-helpers";
 
 @import "views/browse";
 @import "views/email-signups";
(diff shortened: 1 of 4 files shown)


carolinegreen:
Yes, usually I would do that, but thought you wanted the tiniest commits possible so didn't.  I will change it now though.

PR #231, opened 2017-02-01
Fix the taxonomy breadcrumbs for taxon pages

carvil:
The taxonomy breadcrumbs use a different method from the navigation
helpers and therefore can't rely on the default breadcrumb data.

This commit re-adds the taxonomy breadcrumbs.


rubenarakelyan:
👍

PR #232, opened 2017-02-03
Redirect Education browse page in B variant of navigation A/B test

suzannehamilton:
If the user has been assigned to the B variant of the Education Navigation A/B test, redirect them from /browse/education to the new navigation page at /education.

Ensure that this redirect response is cached separately from the original A version.

https://trello.com/c/D7QsJmAy/329-redirects-to-the-new-navigation

This doesn't cover second-level browse pages such as https://www.gov.uk/browse/education/student-finance. I've added that to the Trello card.

Dependencies:

- [x] alphagov/govuk_ab_testing#3

### DIFF
--- a/app/controllers/browse_controller.rb
+++ b/app/controllers/browse_controller.rb
@@ -15,7 +15,29 @@ class BrowseController < ApplicationController
 
     respond_to do |f|
       f.html do
-        render :show, locals: { page: page }
+        is_page_under_ab_test = false
+
+        dimension = Rails.application.config.navigation_ab_test_dimension
+        ab_test = GovukAbTesting::AbTest.new("EducationNavigation", dimension: dimension)
+        ab_test_variant = ab_test.requested_variant(request)
+
+        if new_navigation_enabled? && params[:top_level_slug] == "education"
+          ab_test_variant.configure_response(response)
+
+          if ab_test_variant.variant_b?
+            return redirect_to controller: "taxons",
+              action: "show",
+              taxon_base_path: "education"
+          end
+
+          is_page_under_ab_test = true
+        end
+
+        render :show, locals: {
+          page: page,
+          is_page_under_ab_test: is_page_under_ab_test,
+          ab_test_variant: ab_test_variant,
+         }
       end
       f.json do
         render json: {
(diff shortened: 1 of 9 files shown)


tijmenb:
This will be updated to use the A/B testing gem.

suzannehamilton:
This now uses the gem, so it's finally ready for review. 🎆

suzannehamilton:
This is now using a released version of the A/B gem. The tests can be simplified a little bit once @tijmenb has tidied up the dimensions parameters, but I raise another PR for that.

test/controllers/browse_controller_test.rb
```diff
@@ -29,13 +33,80 @@
...
+
+      describe "with the new navigation not enabled" do
+        ["A", "B"].each do |variant|
+          it "returns the original version of the page for variant #{variant}" do
+            setup_ab_variant("EducationNavigation", variant)
+
+            get :show, top_level_slug: "education"
+
+            assert_response 200
+            assert_unaffected_by_ab_test
```
MatMoore: Not related to this PR, but I find the name of this method a bit confusing - I had to look at the source to work out what it was asserting. Could we give it a different name? Something like `assert_response_not_modified_for_ab_test`?

app/controllers/browse_controller.rb
```diff
@@ -15,7 +15,29 @@ def show
...
     respond_to do |f|
       f.html do
-        render :show, locals: { page: page }
+        is_page_under_ab_test = false
+
+        dimension = Rails.application.config.navigation_ab_test_dimension
+        ab_test = GovukAbTesting::AbTest.new("EducationNavigation", dimension: dimension)
+        ab_test_variant = ab_test.requested_variant(request)
+
+        if new_navigation_enabled? && params[:top_level_slug] == "education"
```
MatMoore: Will this direct all child browse pages of education to the root of the new taxonomy?

MatMoore: Oops, nevermind, it's answered in the description.

test/controllers/browse_controller_test.rb
suzannehamilton: Yeah, that name would be clearer. I'll update the gem - we're still working out the API, and I think I'm the only person using that method.

suzannehamilton:
Closing this without merging because all the redirects are now being reviewed together in #249.

PR #235, opened 2017-02-06
[DO NOT MERGE] Enable A/B education redirects from JavaScript-powered browse pages

suzannehamilton:
Ensure that when a user clicks on 'Education and learning' from /browse, they are redirected to the new taxon page if they are in the B bucket.

https://trello.com/c/D7QsJmAy/329-redirects-to-the-new-navigation

Before merging:
- [ ] Use the [A/B testing gem](https://github.com/alphagov/govuk_ab_testing) to simplify the `HTTP_GOVUK_ABTEST_EDUCATIONNAVIGATION` checks
- [ ] Merge #232 and rebase this PR on master

I'm not very keen on all the hard-coded "education" strings in this code - even once the A/B test is removed, we'll (presumably?) be leaving the redirects in place from the old to the new navigation. At that point, it seems wrong for "education" to be hard-coded into the controller and "/browse/education" to be hard-coded into the JS which handles the browse page links.

Would it be better to add another type of link to the content store, named something like `new_navigation_redirect`? Then we could use it store the redirect from /browse/education to /education. @tijmenb and I talked about doing this for topic pages, but it might make sense here, too. It would make it simple to start redirecting other themes as they're added to the taxonomy, because we'd just make a config change and push it to the content store rather than updating hard-coded theme names in collections.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -33,6 +33,7 @@ group :test do
   gem 'minitest-spec-rails', '~> 5.3.0'
   gem 'mocha', '~> 1.1.0', require: false
   gem 'webmock', '~> 1.21.0', require: false
+  gem 'climate_control', '~> 0.1.0', require: false
   gem 'cucumber-rails', '~> 1.4.2', require: false
   gem 'govuk-content-schema-test-helpers', '~> 1.3.0'
   gem 'govuk_schemas', '~> 2.1'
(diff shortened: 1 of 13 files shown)


suzannehamilton:
Closing: replaced by #249.

PR #236, opened 2017-02-06
Remove nth selector from grid CSS

alecgibson:
This change removes the usage of the `:nth-child` CSS selector in the Taxon grid. This selector is [not supported in IE8](http://caniuse.com/#feat=css-sel3).

In particular, the `clear: left` being applied is needed for very "uneven" content in the grid (ie if one of the descriptions is much longer than the others).

I've tested this in BrowserStack, and it looks fine.

Trello: https://trello.com/c/g8rg70Oa/359-find-a-way-of-making-sure-the-grid-doesn-t-break-without-using-nth-selectors-which-are-not-supported-in-ie8

### DIFF
--- a/app/views/taxons/_child_taxons_grid.html.erb
+++ b/app/views/taxons/_child_taxons_grid.html.erb
@@ -1,8 +1,8 @@
 <nav role="navigation" class="child-topics-list">
   <h2>In this section</h2>
   <ul>
-    <% child_taxons.each do |child_taxon| %>
-      <li>
+    <% child_taxons.each_with_index do |child_taxon, index| %>
+      <li class="<%= 'leftmost-row-cell' if index % 3 == 0%>">
         <h3>
           <%= link_to child_taxon.title, child_taxon.base_path %>
         </h3>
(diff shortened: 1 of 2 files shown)


carolinegreen:
Mostly looks good.  I'm not sure if the & selector is adding anything in this case.  Also really minor nitpick, but could you describe more why the class is there, like with what it is and what it's doing, like left-grid-element-clear or something.

app/views/taxons/_child_taxons_grid.html.erb
```diff
@@ -1,8 +1,8 @@
 <nav role="navigation" class="child-topics-list">
   <h2>In this section</h2>
   <ul>
-    <% child_taxons.each do |child_taxon| %>
-      <li>
+    <% child_taxons.each_with_index do |child_taxon, index| %>
+      <li<%= ' class="leftmost-row-cell"'.html_safe if index % 3 == 0%>>
```
tijmenb: This perhaps reads a little easier like:

```ruby
<li class="<%= 'leftmost-row-cell' if index % 3 == 0 %>">
```

PR #242, opened 2017-02-08
Add beta label to taxonomy pages

mobaig:
We will release the new navigation pages to production in Beta mode. In order to do that, we need to show the beta label on all of those new pages.

https://trello.com/c/r7fux83w/377-add-beta-label-to-the-new-taxonomy-pages-in-collections

screen shot 2017-02-08 at 15 22 16

### DIFF
--- a/app/views/taxons/show.html.erb
+++ b/app/views/taxons/show.html.erb
@@ -7,6 +7,7 @@
 } %>
 
 <% content_for :breadcrumbs do %>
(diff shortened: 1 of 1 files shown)


alextea:
I think the beta banner should go above the breadcrumb. Is that possible?

alextea:
The service manual does it that way: https://www.gov.uk/service-manual/design

mobaig:
@alextea yep, updated! How about the message? Fine as it is or prefer different wording?

PR #247, opened 2017-02-10
Render the current taxon's content after the accordion

mobaig:
Remove the currently displayed taxon and its tagged content from the
accordion. Render the content list after the accordion instead.

In cases where a taxon doesn't have any children , don't render an
accordion at all. Render its content list only.

## Taxon with children:
screen shot 2017-02-10 at 15 41 31

## Taxon with no children: 
screen shot 2017-02-10 at 15 41 58


app/views/taxons/show.html.erb
```diff
@@ -25,4 +25,9 @@
     child_taxons: taxon.child_taxons,
     tagged_content: taxon.tagged_content
   } %>
+
+  <%= render partial: 'content_list_for_current_taxon', locals: {
```
carvil: This seems to be the same in both the `if` and `else` clauses, should we move it outside of the `if` statement?

mobaig: 🤦‍♂️ done!

carvil:
🚀

PR #251, opened 2017-02-14
Track clicks on navigation page accordions in Google Analytics

suzannehamilton:
Merge all the recent changes from the accordion code in service-manual-frontend except for the BEM CSS selectors.

These changes include Google Analytics tracking for the accordion, which we will use to investigate how users use the accordion in the education navigation A/B test.

The long term plan is to extract the accordion to a common module, but we will only do this once the final design of the taxonomy navigation is complete.

https://trello.com/c/27U7LDI0/395-track-accordion-in-google-analytics

### DIFF
--- a/app/assets/javascripts/modules/accordion-with-descriptions.js
+++ b/app/assets/javascripts/modules/accordion-with-descriptions.js
@@ -1,3 +1,7 @@
+// Most of this is taken from the service manual and at some point needs to be
+// separated out into a component.  We are currently holding off on this until
+// it is tested more within the current navigation changes.
+
 window.GOVUK.Modules = window.GOVUK.Modules || {};
 
 (function (Modules) {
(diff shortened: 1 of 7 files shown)


app/assets/javascripts/modules/accordion-with-descriptions.js
```diff
@@ -148,33 +168,39 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
+            var $button = $subsection.find('.js-subsection-button');
+            var $subsectionContent = $subsection.find('.js-subsection-content');
 
-          // show/hide content
-          $subsectionHeader.each(function( index ) {
             if (action == 'open') {
-              openSection($(this).next());
-              showOpenIcon($(this));
+              $button.attr("aria-expanded", "true");
+              $subsectionContent.removeClass('js-hidden');
```
alecgibson: (I personally think these add/remove classes should have been combined into a [`.toggleClass()`](http://api.jquery.com/toggleclass/), but I realise this is just a direct read-across so fine to leave)

app/assets/javascripts/modules/accordion-with-descriptions.js
```diff
@@ -188,71 +214,121 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
-        $node.parent().removeClass('subsection-is-open');
-        $node.parent().addClass('subsection');
+      this._clickedOnHeading = function () {
+        return that.$target.hasClass('js-subsection-button');
       }
 
-      function setExpandedState($node, state) {
-        $node.attr("aria-expanded", state);
+      this._iconType = function () {
+        return (subsectionView.isClosed() ? 'Minus' : 'Plus');
```
alecgibson: (This is potentially susceptible to a race condition, but again I recognise that this is a read-across)

app/assets/javascripts/modules/current-location.js
```diff
@@ -0,0 +1,8 @@
+(function(root) {
```
alecgibson: Could we move this into `static` or similar?

spec/javascripts/helpers/jasmine-jquery.js
```diff
@@ -0,0 +1,841 @@
+/*!
```
alecgibson: Did you mean to commit this file?

spec/javascripts/modules/accordion-with-descriptions_spec.js
```diff
@@ -0,0 +1,357 @@
...
+    $element = $(html);
+  });
+
+  afterEach(function() {
+    $(document).off();
+    sessionStorage.clear();
+  });
+
+  it("has a class of js-accordion-with-descriptions to indicate the js has loaded", function () {
+    accordion.start($element);
```
alecgibson: (This could have been moved into `beforeEach`)

spec/javascripts/helpers/jasmine-jquery.js
suzannehamilton: Yes - it's used to power the jQuery assertions like `expect($foo).toHaveClass('bar')`. The JS test dependencies are just checked in - we're currently not using a package manager to add them.

alecgibson: Okay. A bit weird - maybe we should investigate integrating npm at some point (obviously not as part of this PR)

suzannehamilton:
@alecgibson - I think I agree with your other comments, but I've deliberately left the code as similar as possible to service-manual-frontend to make it easier to ultimately commonise them.

If people strongly disagree with this then I'm happy to do a bit of tidying, but I don't think it's worth the effort at this point.

PR #252, opened 2017-02-15
Remove parent taxon link from taxon navigation pages

suzannehamilton:
In user research, this link helped some users but was completely misunderstood by others. Given that users can still navigate to the parent page using the breadcrumbs, we've decided to remove this link.

https://trello.com/c/A2mVxBwh/376-update-collections-to-show-the-latest-parent-taxon-link-styles-from-the-prototype


carvil:
👍 

@suzannehamilton could you please also delete `back.svg`?

suzannehamilton:
Good spot! I've deleted it.

PR #255, opened 2017-02-20
Remove content in collections from nav pages

carvil:
Document collections should be surfaced on navigation pages, but not the content tagged to them.

We have seen from lab research that surfacing all the content makes the list view harder to navigate.

Some document collections will be removed as they are now covered by taxons, but other that group the same type of thing will remain.

This PR makes sure we filter out all content that already belongs to a document collection.

Trello: https://trello.com/c/2KDTKdo6/413-only-show-collections-on-list-pages-not-content-tagged-to-them

### DIFF
--- a/test/services/tagged_content_test.rb
+++ b/test/services/tagged_content_test.rb
@@ -38,6 +38,10 @@ describe TaggedContent do
         filter_content_store_document_type: guidance_document_types
       )
     end
+
+    it 'filters out content that belongs to a document collection' do
+      assert_includes_params(filter_document_collections: '_MISSING')
+    end
   end
 
 private
--- a/test/services/tagged_content_test.rb
+++ b/test/services/tagged_content_test.rb
@@ -1,0 +1,63 @@
+require 'test_helper'
+require './test/support/custom_assertions.rb'
+
+describe TaggedContent do
+  describe '#fetch' do
+    it 'returns the documents from search' do
+      search_results = mock(documents: ['doc1'])
+
+      RummagerSearch.stubs(:new).returns(search_results)
+
+      assert(tagged_content.fetch, ['doc1'])
+    end
+
+    it 'starts from the first page' do
+      assert_includes_params(start: 0)
+    end
+
+    it 'requests all results' do
+      assert_includes_params(count: RummagerSearch::PAGE_SIZE_TO_GET_EVERYTHING)
+    end
+
+    it 'requests a limited number of fields' do
+      assert_includes_params(fields: %w(title description link))
+    end
+
+    it 'orders the results by title' do
+      assert_includes_params(order: 'title')
+    end
+
+    it 'filters the results by taxon' do
+      assert_includes_params(filter_taxons: [taxon_content_id])
+    end
+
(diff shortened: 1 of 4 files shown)


test/support/custom_assertions.rb
```diff
@@ -0,0 +1,14 @@
+module MiniTest::Assertions
+  # Behaves a bit like `hash_including`.
+  # Given a slice of the original hash, it verifies that the original hash
+  # includes the sub-hash given. Useful when testing partial arguments/params to
+  # methods.
+  def assert_includes_subhash(expected_sub_hash, hash)
+    assert_equal(
+      expected_sub_hash.values,
+      hash.slice(*expected_sub_hash.keys).values
```
MatMoore: Might be worth adding a couple of tests for this.

I'm assuming that both values arrays follow the insertion order of `expected_sub_hash` here, but I had to think about it a bit.

carvil: @MatMoore That's a good point. Would it make sense to be a bit more verbose and change it to:

```
expected_sub_hash.each do |key, value|
  assert_equal(value, hash[key], "Expected #{hash} to include #{key}=#{value}")
end
```

instead?

test/services/tagged_content_test.rb
```diff
@@ -0,0 +1,67 @@
...
+
+    it 'filters out content that belongs to a document collection' do
+      assert_includes_params(filter_document_collections: '_MISSING')
+    end
+  end
+
+private
+
+  def assert_includes_params(expected_params)
+    search_results = mock(documents: %w(doc1 doc2))
```
MatMoore: I would consider using webmock to stub at the application boundary here, instead of stubbing the RummagerSearch service. I don't think it's important to specify that `#fetch` calls `RummagerSearch`, and what parameters it should pass, since `RummagerSearch` could be refactored and the application would still work.

test/support/custom_assertions.rb
MatMoore: Yeah, I like that better. Has the added advantage of nice failure messages too 👍

test/services/tagged_content_test.rb
carvil: I see your point. The reason I stub `RummagerSearch` rather than using webmock is because this is a unit test and I think its boundaries should be described by its immediate interfaces. We could, though, use webmock in our integration tests (https://github.com/alphagov/collections/blob/master/test/integration/taxon_browsing_test.rb#L80-L83 and https://github.com/alphagov/collections/blob/c71b20cb7ca3604239aa9f7f9314da9ad0f90d1a/test/support/rummager_helpers.rb#L2-L15) to make sure the end-to-end system makes sense.

If you agree, what I could do instead is to stub `Services.rummager.search` in the unit tests and assert that it gets the expected parameters (rather than `RummagerSearch`), and in a separate PR work on using webmock in our integration tests.

What do you think?

carvil:
@MatMoore I've added 2 more commits to address your comments. The only thing I haven't done is use webmock, which I will do in a separate PR.

MatMoore:
👍 LGTM.

PR #258, opened 2017-02-22
Use content store instead of Rummager in topic and browse pages

suzannehamilton:
Content is linked to parent topics and mainstream browse pages in the publishing API and content store, so the content store can be used to populate the subtopic and second-level browse pages rather than Rummager.

This reduces the number of requests to Rummager from collections, and is a step towards removing it from collections entirely.

Dependencies:

- [x] Add `topic_content` and `mainstream_browse_content` to frontend schemas (alphagov/govuk-content-schemas#535)
- [x] Add `topic_content` and `mainstream_browse_content` reverse links to the publishing API (alphagov/publishing-api#793)
- [x] Re-present all topics and mainstream browse pages to the content store

Trello: https://trello.com/c/D8Heon8X/410-use-the-content-store-to-serve-pages-in-collections-frontend

### DIFF
--- a/app/models/content_item.rb
+++ b/app/models/content_item.rb
@@ -10,7 +10,7 @@ class ContentItem
     @content_item_data = content_item_data
   end
 
-  %i[base_path title description content_id].each do |field|
+  %i[base_path title description content_id document_type].each do |field|
     define_method field do
       @content_item_data[field.to_s]
     end
--- a/app/models/mainstream_browse_page.rb
+++ b/app/models/mainstream_browse_page.rb
@@ -10,6 +10,14 @@ class MainstreamBrowsePage
     to: :content_item
   )
 
+  DOCUMENT_TYPES_TO_EXCLUDE = %w(
+    fatality_notice
+    news_article
+    speech
+    world_location_news_article
+    travel_advice
+  ).to_set
+
   def self.find(base_path)
     content_item = ContentItem.find!(base_path)
     new(content_item)
@@ -44,7 +52,10 @@ class MainstreamBrowsePage
   end
 
   def lists
-    @lists ||= ListSet.new("section", @content_item.content_id, details["groups"])
+    @lists ||= ListSet.new(
+      @content_item.linked_items("mainstream_browse_content"),
+      details["groups"],
+      DOCUMENT_TYPES_TO_EXCLUDE)
   end
 
   def related_topics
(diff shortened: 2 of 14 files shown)


app/models/content_item.rb
```diff
@@ -10,7 +10,7 @@ def initialize(content_item_data)
     @content_item_data = content_item_data
   end
 
-  %i[base_path title description content_id].each do |field|
+  %i[base_path title description content_id schema_name].each do |field|
```
tijmenb: We should probably be using `document_type` for this. Schema describes the structure, not the content.

app/models/mainstream_browse_page.rb
```diff
@@ -10,6 +10,14 @@ class MainstreamBrowsePage
     to: :content_item
   )
 
+  FORMATS_TO_EXCLUDE = %w(
```
tijmenb: These formats correspond to `format` in rummager, but we're comparing them with `schema_name` (to be `document_type`). We'll need to map them to the values in the content store (for example, `travel-advice` -> `travel_advice`).

features/support/topic_helper.rb
```diff
@@ -52,9 +38,17 @@ def stub_topic_lookups
         ],
       },
       links: {
-        "parent" => [
-          "title" => "Oil and Gas",
-          "base_path" => "/oil-and-gas",
+        parent: [
+          { title: "Oil and Gas", base_path: "/oil-and-gas" }
+        ],
+        topic_content: [
```
tijmenb: Now that we're only using content store data, we can use the `govuk_schemas` gem to validate fixtures like this, so they won't get out of date. Could you see if https://alphagov.github.io/govuk_schemas_gem/GovukSchemas/RandomExample.html#merge_and_validate-instance_method would work?

app/models/mainstream_browse_page.rb
suzannehamilton: Done. `travel-advice` was the only one that needed to be changed.

app/models/content_item.rb
suzannehamilton: Fixed, thanks.

features/support/topic_helper.rb
suzannehamilton: Yes, that works, thanks! I've updated the topic content item, but I haven't done it for mainstream browse page tests so I'll do that now.

suzannehamilton:
The tests are currently failing on Jenkins, but they should pass once the schema changes have been deployed to production.

suzannehamilton:
All topics and mainstream browse pages have been re-presented to the content store, so this is ready to go once it's been reviewed.

PR #262, opened 2017-02-23
Track accordion subsection clicks

alecgibson:
Adds tracking of links within expanded accordion subsections.

### Trello

https://trello.com/c/WPb0WgPo/427-track-click-events-in-the-new-accordion-pages

### DIFF
--- a/test/integration/taxon_browsing_test.rb
+++ b/test/integration/taxon_browsing_test.rb
@@ -25,6 +25,7 @@ class TaxonBrowsingTest < ActionDispatch::IntegrationTest
     then_i_can_see_the_breadcrumbs
     and_i_can_see_the_title_and_description
     and_i_can_see_links_to_the_child_taxons_in_an_accordion
+    and_the_accordion_has_tracking_attributes
     and_i_can_see_tagged_content_to_the_taxon
   end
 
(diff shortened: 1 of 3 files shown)


test/integration/taxon_browsing_test.rb
```diff
@@ -152,6 +153,23 @@ def and_i_can_see_links_to_the_child_taxons_in_an_accordion
...
+    tracked_links.size.must_equal @child_taxons.size * search_results.size
+
+    tracked_links.each_with_index do |link, index|
+      section_number = (index / search_results.size).floor + 1
+      item_number = index % search_results.size + 1
+
+      link[:'data-track-action'].must_equal "#{section_number}.#{item_number}"
+      link[:'data-track-label'].must_equal "#{search_results[item_number - 1]['link']}"
+      link[:'data-track-value'].must_equal "#{search_results.size}"
+      link[:'data-track-dimension'].must_equal "#{search_results[item_number - 1]['title']}"
```
suzannehamilton: I wonder whether it would be better to test a couple of links with real values rather than extracting values from `search_results`? The logic in the test is very similar to the logic in the code it's testing.

test/integration/taxon_browsing_test.rb
```diff
@@ -152,6 +153,34 @@ def and_i_can_see_links_to_the_child_taxons_in_an_accordion
...
+      link[:'data-track-dimension-index'].must_equal '29'
+      link[:'data-module'].must_equal 'track-click'
+    end
+
+    # Test an example free from logic
+    assert page.has_css?(
```
suzannehamilton: Hmm. The only issue with this is that the failure message won't be very meaningful - it'll just say "Can't find element with <all this stuff>". If there's an easy way of grabbing the relevant element and asserting all the properties individually then that might be better. Otherwise, I think this is good. 👍

alecgibson: That's essentially what the looped check should do (which should fail first, and give you a more meaningful failure).

suzannehamilton: Ahh, I see. Sorry, I thought this was a replacement and that you'd left both in for comparison in the review. 👍

PR #264, opened 2017-02-24
Track the type of taxon page being visited in Google Analytics

suzannehamilton:
- Add integration test of leaf node taxon page in preparation for testing new `meta` tag on all versions of the taxon page
- Make taxon `grandchildren?` and `children?` methods consistent
- Add meta tag to taxon pages which identifies whether the page is a grid, an accordion or a leaf node. This will help us analyse whether users are able to find what they need quickly, or if they navigate in circles because they are struggling to find content in the new navigation.

This won't actually report anything to Google Analytics until alphagov/static#910 is merged.

Trello: https://trello.com/c/SCsj2A9D/425-add-custom-dimension-to-page-view-tracking-on-new-navigation-pages

### DIFF
--- a/app/views/taxons/show.html.erb
+++ b/app/views/taxons/show.html.erb
@@ -12,7 +12,7 @@
     locals: @navigation_helpers.taxon_breadcrumbs %>
 <% end %>
 
-<% if taxon.has_grandchildren? %>
+<% if taxon.grandchildren? %>
   <%= render partial: 'child_taxons_grid',
     locals: { child_taxons: taxon.child_taxons } %>
 <% else %>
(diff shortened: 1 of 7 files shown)


app/views/taxons/show.html.erb
```diff
@@ -2,6 +2,7 @@
 <% content_for :page_class, "taxon-page" %>
 <% content_for :meta_tags do %>
   <%= ab_variant.analytics_meta_tag.html_safe %>
+  <meta name="govuk:navigation-page-type" content="<%=taxon_page_rendering_type(taxon)%>" />
```
carvil: How will we submit this information to GA?

suzannehamilton: I've added it to the list of GA meta tags in static: alphagov/static#910

carvil: Cool, thanks for linking! LGTM

PR #267, opened 2017-02-27
Track all nav pages as navigation pages in Google Analytics

suzannehamilton:
Add meta tag which marks taxon, browse, topic and services & information pages as being in the navigation stage of a user journey, as opposed to the content stage.

This will enable us to analyse how users navigate the site so that we can work out whether the new navigation pages are helping users find what they need.

The JS for converting the meta tags to the GA dimension was added in alphagov/static#910.

Trello: https://trello.com/c/SCsj2A9D/425-add-custom-dimension-to-page-view-tracking-on-new-navigation-pages

### DIFF
--- a/app/views/browse/index.html.erb
+++ b/app/views/browse/index.html.erb
@@ -1,5 +1,8 @@
 <% content_for :title, "All categories - GOV.UK" %>
 <% content_for :page_class, "browse" %>
+<% content_for :meta_tags do %>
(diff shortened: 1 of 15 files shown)


suzannehamilton:
Closing this PR because we're taking a different approach: defining a document supertype in [govuk_document_types](https://github.com/alphagov/govuk_document_types).

PR #269, opened 2017-02-28
Show new navigation based on A/B testing variants

carvil:
This change makes sure collections behaves like any other repos that use the new navigation by allowing us to switch between A/B variants.

### DIFF
--- a/test/controllers/browse_controller_test.rb
+++ b/test/controllers/browse_controller_test.rb
@@ -45,8 +45,8 @@ describe BrowseController do
         content_store_has_item("/browse/benefits", base_path: '/browse/benefits')
       end
 
-      describe "with the new navigation not enabled" do
-        ["A", "B"].each do |variant|
+      describe 'with the feature flag off' do
+        %w(A B).each do |variant|
           it "returns the original version of the page for variant #{variant}" do
             setup_ab_variant("EducationNavigation", variant)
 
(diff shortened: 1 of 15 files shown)


test/controllers/browse_controller_test.rb
```diff
@@ -60,33 +60,27 @@
 
       describe "with the new navigation enabled" do
         it "returns the original version of the page as the A variant" do
-          with_new_navigation_enabled do
-            with_variant EducationNavigation: "A" do
-              get :show, top_level_slug: "education"
+          with_new_navigation_disabled do
```
suzannehamilton: This contradicts the `describe` statement on line 61. I think it's just a case of wording - perhaps this should just be called `with_A_variant`?

test/controllers/browse_controller_test.rb
```diff
@@ -60,33 +60,27 @@
...
+            assert_response 302
+            assert_redirected_to controller: "taxons", action: "show", taxon_base_path: "education"
           end
         end
 
         ["A", "B"].each do |variant|
           it "does not change a page outside the A/B test when the #{variant} variant is requested" do
             setup_ab_variant("EducationNavigation", variant)
 
-            with_new_navigation_enabled do
```
suzannehamilton: Is this test still testing the right thing? It looks like it no longer sets the environment variable, so it would pass even if the `get` request was to an education page, which isn't right.

Same goes for the JSON format test below.

suzannehamilton:
Does this actually change the behaviour, or is it just to make the code consistent with other applications? I couldn't spot any differences, but I'm not sure if I've missed something subtle.

carvil:
@suzannehamilton we currently can't use the Chrome extension to turn the A/B test on/off in collections, so it will change the behaviour.

suzannehamilton:
Sorry, I think I'm just missing something. I've just spotted that this fixes the 404 bug (it now returns a 404 if the user is in the A variant). Is that what you mean by fixing the behaviour in the Chrome extension?

suzannehamilton:
That makes sense, thanks!

test/controllers/browse_controller_test.rb
carvil: Good spot, I will fix that now.

test/controllers/browse_controller_test.rb
carvil: Those tests are no longer valid, I will remove them.

carvil:
@suzannehamilton I've made the changes and have also upgraded the gem, which has now been released. This is ready for another review.

test/controllers/browse_controller_test.rb
suzannehamilton: There are still a few inconsistencies - the `describe` now contradicts the `with_new_navigation_enabled` on line 71, and the difference is between `with the new navigation not enabled` on line 48 and `with the new navigation enabled` on line 61.

The same thing applies to the other controller tests below.

test/controllers/browse_controller_test.rb
```diff
@@ -45,8 +45,8 @@
         content_store_has_item("/browse/benefits", base_path: '/browse/benefits')
       end
 
-      describe "with the new navigation not enabled" do
-        ["A", "B"].each do |variant|
+      describe 'without the feature flag off' do
```
suzannehamilton: Should be `with the feature flag off`?

carvil: Yup, fixed below.

suzannehamilton: 👍

PR #273, opened 2017-02-28
Differentiate between grid and leaf content clicks

alecgibson:
Previously, there was no differentiation between clicks on content links on grid vs. those on a "leaf" taxon page.

This change will track all clicks on grid-level content items with a category of `navGridLeafLinkClicked`, and content items at the leaf level as `navLeafLinkClicked`.

Note that it is assumed that all content tagged at the accordion level will be contained within the accordion itself, and hence we don't need a separate tracking category for that case.

### Trello

https://trello.com/c/WJDdtMs4/430-track-click-events-in-taxon-leaf-pages

### DIFF
--- a/app/views/taxons/_content_list_for_current_taxon.html.erb
+++ b/app/views/taxons/_content_list_for_current_taxon.html.erb
@@ -1,3 +1,5 @@
+<% is_grid ||= false %>
+
 <div class="grid-row">
   <div class="column-two-thirds">
     <div class="parent-topic-contents">
@@ -15,7 +17,7 @@
                   content_item.title,
                   Plek.new.website_root + content_item.base_path,
                   data: {
-                    track_category: 'navLeafLinkClicked',
+                    track_category: is_grid ? 'navGridLeafLinkClicked' : 'navLeafLinkClicked',
                     track_action: index + 1,
                     track_value: tagged_content.size,
                     track_label: content_item.base_path,
--- a/app/views/taxons/show.html.erb
+++ b/app/views/taxons/show.html.erb
@@ -20,6 +20,7 @@
   <%= render partial: 'content_list_for_current_taxon', locals: {
   taxon: taxon,
   tagged_content: taxon.tagged_content,
+  is_grid: true,
   } %>
 <% else %>
   <% if taxon.children? %>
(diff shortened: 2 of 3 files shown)


suzannehamilton:
This looks related to dimension 32 (which marks taxon pages as grid/accordion/leaf): https://trello.com/c/SCsj2A9D/425-add-custom-dimension-to-page-view-tracking-on-new-navigation-pages

This is more of an analytics spec question, but could we use that dimension rather than applying this change?

If we can't just use that dimension, I wonder if we could share some of the concepts with the meta tag that powers it e.g. `TaxonHelper`. Do you know why we care about grid/accordion/leaf when we track page views, but only grid and leaf when tracking link clicks?

alecgibson:
@suzannehamilton Tara suggested this on [this Trello card](https://trello.com/c/WJDdtMs4/430-track-click-events-in-taxon-leaf-pages#comment-58b5911c00deee3fbd639caa). She did also mention dimension 32 - happy to do whichever?

suzannehamilton:
I guess it depends which is easier to analyse in GA. The code would be simpler if we just used dimension 32, so we should only make this change if it makes things easier for Tara.

It would also be good to clarify this comment: "The same would apply to accordion pages, if there's still a list outside the accordion." I wonder if that means "accordions should have a third value like `navAccordionLeafLinkClicked`" rather than "accordions should have the same value as one of the other pages"?

alecgibson:
@suzannehamilton I've already left a comment on the Trello card regarding accordions. If I recall, we should be merging everything tagged at that level into the accordion itself (although I don't think that's yet been implemented)

alecgibson:
@suzannehamilton Tara has [confirmed that she wants this solution](https://trello.com/c/WJDdtMs4/430-track-click-events-in-taxon-leaf-pages#comment-58b5a679c0fa53039a08c7db)

However, I'm marking this as "Do not merge" for now, until we hear the results of the user research this week, as the accordion design in particular may change (as could the leaf pages!).

alecgibson:
@suzannehamilton after user research, we've determined:

- Leaf page tracking staying the same
- Accordions won't have leaves

So this should be good to re-review, please.

suzannehamilton:
🚀

PR #277, opened 2017-03-02
Add Overview section with content to accordion

carvil:
This commit makes sure the first section of the accordion view is an
Overview section with content tagged to the accordion taxon level.

When there is no content, we omit this section and only show child
taxons and their content.

The copy might change, but for now we should go ahead with this one.

Trello: https://trello.com/c/xSvrtvcG/467-move-content-tagged-at-an-accordion-level-to-the-first-position-in-the-accordion

### DIFF
--- a/app/views/taxons/show.html.erb
+++ b/app/views/taxons/show.html.erb
@@ -16,15 +16,22 @@
 <% if taxon.grandchildren? %>
   <%= render partial: 'child_taxons_grid',
     locals: { child_taxons: taxon.child_taxons } %>
-<% else %>
-  <%= render partial: 'child_taxons_list', locals: {
-    taxon: taxon,
-    child_taxons: taxon.child_taxons,
-    tagged_content: taxon.tagged_content
-  } %>
-<% end %>
 
-<%= render partial: 'content_list_for_current_taxon', locals: {
+  <%= render partial: 'content_list_for_current_taxon', locals: {
   taxon: taxon,
   tagged_content: taxon.tagged_content,
-} %>
+  } %>
+<% else %>
+  <% if taxon.children? %>
+    <%= render partial: 'child_taxons_list', locals: {
+      taxon: taxon,
+      child_taxons: taxon.child_taxons,
+      tagged_content: taxon.tagged_content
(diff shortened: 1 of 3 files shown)


app/views/taxons/show.html.erb
```diff
@@ -16,15 +16,22 @@
...
-<% end %>
 
-<%= render partial: 'content_list_for_current_taxon', locals: {
+  <%= render partial: 'content_list_for_current_taxon', locals: {
   taxon: taxon,
   tagged_content: taxon.tagged_content,
-} %>
+  } %>
+<% else %>
+  <% if taxon.child_taxons.empty? %>
```
suzannehamilton: Pretty trivial, but there's a `taxon.children?` method you can use here.

test/integration/taxon_browsing_test.rb
```diff
@@ -216,9 +228,34 @@ def and_i_can_see_links_to_the_child_taxons_in_an_accordion
...
+  def then_there_is_no_overview_section_in_the_accordion
+    subsection = first('.subsection')
+
+    assert_equal(
+      false,
+      subsection.has_selector?('.subsection-title', text: /overview/i)
+    )
+  end
+
+  def and_there_is_no_content_presented_after_the_accordion
```
suzannehamilton: Is `parent-topic-contents` used elsewhere in a way that could cause a regression? I'm wondering if this step is really necessary - would we have added it if the first iteration had put the content into the accordion?

app/views/taxons/show.html.erb
carvil: Ok, will change.

test/integration/taxon_browsing_test.rb
carvil: The idea behind this step is so no one by mistake changes the conditionals in the taxon show page in a way that would show the tagged content underneath the accordion. The content was there before and this step was failing until I fixed those conditionals.

Does this make sense?

carvil:
@suzannehamilton updated!

suzannehamilton:
🚀

PR #278, opened 2017-03-02
Update padding bottom of taxon grid elements

carvil:
### Desktop

screen shot 2017-03-02 at 16 44 39

### Tablet

screen shot 2017-03-02 at 16 45 25

### Mobile

screen shot 2017-03-02 at 16 45 46

Trello: https://trello.com/c/yTDsFUrm/468-style-changes-to-taxon-grid-pages-in-collections

### DIFF
--- a/app/assets/stylesheets/views/_taxons.scss
+++ b/app/assets/stylesheets/views/_taxons.scss
@@ -57,6 +57,7 @@
         padding-bottom: $gutter;
 
         @include media (tablet) {
+          padding-bottom: $gutter * 1.5;
           min-height: (5 * $gutter);
         }
       }


app/assets/stylesheets/views/_taxons.scss
```diff
@@ -54,11 +54,15 @@
 
       li {
         @include grid-column(1 / 3);
-        padding-bottom: $gutter;
+        padding-bottom: $gutter * 1.5;
```
alextea: I think if you kept this as `$gutter` and added `padding-bottom: $gutter * 1.5;` into the `tablet` block you wouldn't need the `mobile` block.

carvil: I was trying to make it clearer where the difference would be (mobile only), but that's perhaps too verbose. Will make the change.

mobaig:
🌮

PR #279, opened 2017-03-02
Update accordion styles to match prototype

suzannehamilton:
Four small changes:

- Under links inside the accordion sections
- Remove margin on subsections
- Update subsection title and description margins to 45px
- Reduce padding at bottom of accordion

cc @alextea 

https://trello.com/c/HvOu1E7e/469-style-changes-to-the-accordion-pages-in-collections

### DIFF
--- a/app/assets/stylesheets/views/_taxons.scss
+++ b/app/assets/stylesheets/views/_taxons.scss
@@ -133,7 +133,7 @@
 
    // When JavaScript is enabled, create accordion
   .js-accordion-with-descriptions {
-    padding-bottom: $gutter * 2;
+    padding-bottom: $gutter;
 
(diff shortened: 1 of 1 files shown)


app/assets/stylesheets/views/_taxons.scss
```diff
@@ -132,7 +127,7 @@
     }
 
     .subsection-list-item a {
-      text-decoration: none;
+      text-decoration: underline;
```
alecgibson: What happens if you remove this block of CSS altogether? Presumably the default styling is `underline` ?

suzannehamilton: Good point! That works. 👍

PR #281, opened 2017-03-03
Add document collection filtering logic

carvil:
Most of the changes in this PR related to the 3rd commit.

This commit adds rules around filtering document collections and content
tagged to document collections.

The reason for these rules is that document collections are all
currently under review. Most of them should disappear and its content
solely tagged to taxons, but others will stay on. Some of the existing
document collections have dozens or hundreds of content items tagged to
them. If we decided to show those content items without any control, the
navigation pages would become unusable.

In other cases, the document collections are manageable, which means
it's fine to show the content tagged to them.

The rules are as follows:
- if a document collection is whitelisted (with `surface_collection:
  true`), then that document collection should show up;
- if a document collection is not supposed to be surfaced (with
  `surface_collection: false`), then it should not show up in the
  navigation pages;
- if documents tagged to a collection should be surfaced (with
  `surface_content: true`), then the content tagged to a document
  collection should show in the navigation pages;
- finally, if documents tagged to a collection should not be surfaced
  (with `surface_content: false`), then the content tagged to a document
  collection should not show in the navigation pages;

Finally, these rules also allow us to make sure there are no taxons
without content showing up. There are cases where the only content
tagged to a taxon is in a document collections, which wouldn't show. By
manually allowing to control this, we make sure that doesn't happen.

Trello: https://trello.com/c/YToEtjhb/483-add-new-rules-on-document-collections-and-tagged-content-to-the-collections-app

### DIFF
--- a/app/services/filter_tagged_content.rb
+++ b/app/services/filter_tagged_content.rb
@@ -1,0 +1,62 @@
+class FilterTaggedContent
+  def valid?(document)
+    return valid_document_collection?(document) if document.document_collection?
+
+    valid_content_item?(document)
+  end
+
+private
+
+  def valid_document_collection?(document)
+    return true unless all_document_collections.include?(document.base_path)
+
+    valid_document_collections.include?(document.base_path)
+  end
+
+  def valid_content_item?(document)
+    return true unless document.tagged_to_document_collection?
+
+    document.document_collections.all? do |document_collection|
+      valid_content_items_in_document_collections.include?(
+        document_collection['link']
+      )
+    end
+  end
+
+  def valid_document_collections
+    guidance_document_collections.reduce([]) do |list, document_collection|
+      if document_collection['surface_collection']
+        list << document_collection['base_path']
+      end
+
+      list
+    end
+  end
+
+  def all_document_collections
+    guidance_document_collections.map do |document_collection|
+      document_collection['base_path']
+    end
+  end
+
+  def valid_content_items_in_document_collections
+    guidance_document_collections.reduce([]) do |list, document_collection|
+      if document_collection['surface_content']
+        list << document_collection['base_path']
+      end
+
+      list
+    end
+  end
+
+  def guidance_document_collections
+    @guidance_document_collections ||= JSON.parse(
+      File.read(
+        Rails.root.join(
+          "config",
+          "guidance_document_collections.json"
+        )
+      )
+    )
+  end
+end
--- a/config/guidance_document_collections.json
+++ b/config/guidance_document_collections.json
@@ -6,8 +6,8 @@
   },
   {
     "base_path": "/government/collections/ofsted-inspections-of-maintained-schools",
-    "surface_collection": true,
-    "surface_content": false
+    "surface_collection": false,
+    "surface_content": true
   },
   {
     "base_path": "/government/collections/further-education-commissioner-intervention-reports",
--- a/config/guidance_document_collections.json
+++ b/config/guidance_document_collections.json
@@ -1,0 +1,12 @@
+[
+  {
+    "base_path": "/government/collections/national-curriculum-assessments-information-for-parents",
+    "surface_collection": false,
+    "surface_content": true
+  },
+  {
+    "base_path": "/government/collections/ofsted-inspections-of-maintained-schools",
+    "surface_collection": true,
+    "surface_content": false
+  }
+]
--- a/config/guidance_document_collections.json
+++ b/config/guidance_document_collections.json
@@ -8,5 +8,680 @@
     "base_path": "/government/collections/ofsted-inspections-of-maintained-schools",
     "surface_collection": true,
     "surface_content": false
+  },
+  {
+    "base_path": "/government/collections/further-education-commissioner-intervention-reports",
+    "surface_collection": true,
+    "surface_content": false
+  },
+  {
+    "base_path": "/government/collections/fe-choices-information-for-providers",
+    "surface_collection": true,
+    "surface_content": false
+  },
+  {
+    "base_path": "/government/collections/send-pathfinders",
+    "surface_collection": true,
+    "surface_content": false
+  },
+  {
+    "base_path": "/government/collections/ofsted-inspections-of-residential-holiday-schemes-for-disabled-children",
+    "surface_collection": true,
+    "surface_content": false
+  },
+  {
+    "base_path": "/government/collections/functional-skills-qualifications-requirements",
+    "surface_collection": true,
+    "surface_content": false
+  },
+  {
+    "base_path": "/government/collections/sfa-financial-management-helping-colleges-to-account-for-funding",
+    "surface_collection": true,
+    "surface_content": false
+  },
+  {
+    "base_path": "/government/collections/sfa-financial-assurance-demonstrating-appropriate-use-of-funds",
+    "surface_collection": true,
+    "surface_content": false
+  },
+  {
+    "base_path": "/government/collections/success-rates-2012-to-2013",
+    "surface_collection": true,
+    "surface_content": false
+  },
+  {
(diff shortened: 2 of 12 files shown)


alecgibson:
> By manually allowing to control this, we make sure that doesn't happen.

Is this necessarily true? You could set both of these flags to `false` and you'd still get empty taxons? Is this important and do we need to be doing anything about it? (Ie ensuring that at least one of the flags is `true`?)

app/services/filter_tagged_content.rb
```diff
@@ -0,0 +1,62 @@
+class FilterTaggedContent
+  def valid?(document)
```
alecgibson: Nitpick: might be more readable as:

```ruby
if document.document_collection?
  valid_document_collection?(document)
else
  valid_content_item?(document)
end
```

carvil:
@alecgibson I see your point. I would say this is highly unlikely because this list has been manually curated to avoid 2 scenarios:

- empty taxons;
- navigation pages with way too much content.

This code won't be here for long, because these lists are being reviewed right now, but that review won't be ready before the release.

app/services/filter_tagged_content.rb
```diff
@@ -0,0 +1,62 @@
...
+    return valid_document_collection?(document) if document.document_collection?
+
+    valid_content_item?(document)
+  end
+
+private
+
+  def valid_document_collection?(document)
```
alecgibson: Nitpick:

```ruby
all_document_collections.exclude?(document.base_path)
  || valid_document_collections.include?(document.base_path)
```

app/services/filter_tagged_content.rb
```diff
@@ -0,0 +1,62 @@
...
+    return true unless document.tagged_to_document_collection?
+
+    document.document_collections.all? do |document_collection|
+      valid_content_items_in_document_collections.include?(
+        document_collection['link']
+      )
+    end
+  end
+
+  def valid_document_collections
```
alecgibson: Could this be memoized? (I don't know how often this will be accessed?)

app/services/filter_tagged_content.rb
```diff
@@ -0,0 +1,62 @@
...
+    guidance_document_collections.reduce([]) do |list, document_collection|
+      if document_collection['surface_collection']
+        list << document_collection['base_path']
+      end
+
+      list
+    end
+  end
+
+  def all_document_collections
```
alecgibson: Could this also be memoized?

app/services/filter_tagged_content.rb
```diff
@@ -0,0 +1,62 @@
...
+    guidance_document_collections.map do |document_collection|
+      document_collection['base_path']
+    end
+  end
+
+  def valid_content_items_in_document_collections
```
alecgibson: Memoize?

app/services/filter_tagged_content.rb
```diff
@@ -0,0 +1,62 @@
+class FilterTaggedContent
```
alecgibson: Could this class name be improved? It's currently phrased as a verb, which isn't very typical. I'm not 100% sure myself of what this class is intended to do (partly because of the naming), but given the amount of methods that start with `valid`, maybe something like `TaggedContentValidator` is more descriptive?

config/guidance_document_collections.json
```diff
@@ -0,0 +1,687 @@
+[
+  {
+    "base_path": "/government/collections/national-curriculum-assessments-information-for-parents",
+    "surface_collection": false,
```
alecgibson: (I assume `false` isn't technically required, but is here for clarity/emphasis?)

app/services/filter_tagged_content.rb
```diff
@@ -0,0 +1,62 @@
...
+
+    document.document_collections.all? do |document_collection|
+      valid_content_items_in_document_collections.include?(
+        document_collection['link']
+      )
+    end
+  end
+
+  def valid_document_collections
+    guidance_document_collections.reduce([]) do |list, document_collection|
```
alecgibson: Given that this is used a lot with `include?` would it be more efficient/semantically correct to use a `Set` rather than a `List`?

test/services/filter_tagged_content_test.rb
```diff
@@ -0,0 +1,87 @@
...
+      refute(
+        FilterTaggedContent.new.valid?(document),
+        "It should not show content items tagged to unknown document collections"
+      )
+    end
+
+    it 'is not valid for a document collection that should not be surfaced' do
+      document = Document.new(
+        content_store_document_type: 'document_collection',
+        base_path: '/government/collections/national-curriculum-assessments-information-for-parents'
```
alecgibson: Are these tests tied to the live `.json` file? So if that file changes, these tests could break?

app/services/filter_tagged_content.rb
carvil: I will rename if if that's clearer.

We have tried both ways of naming service objects (e.g. in content tagger: https://github.com/alphagov/content-tagger/tree/master/app/services/bulk_tagging) and I'm still not sure which one is more better, but happy to change it here.

app/services/filter_tagged_content.rb
carvil: I tend to avoid if/else clauses because I think there is a bigger chance of the code becoming too nested/hard to read over time. I personally find ruby's postfix conditionals a nice way of reading code.

alecgibson: Yeah I get that this is always a personal preference (which is why I haven't actively "requested changes"). Fine to leave.

app/services/filter_tagged_content.rb
carvil: That's a nice way of writing it (I've done it quite a bit in the past), but perhaps a bit less explicit, which might lead to confusion.

app/services/filter_tagged_content.rb
carvil: It's only being used in one place, I'm tempted to leave it as is for now if that's ok.

app/services/filter_tagged_content.rb
carvil: As above, it's only being used in one place, I'm tempted to leave it as is for now if that's ok.

app/services/filter_tagged_content.rb
carvil: Same as above: https://github.com/alphagov/collections/pull/281/files/4e4f581203769f414857cf102ff7462f3154585c#r104374707

app/services/filter_tagged_content.rb
alecgibson: Sure.

app/services/filter_tagged_content.rb
alecgibson: Sure

app/services/filter_tagged_content.rb
alecgibson: Sure

config/guidance_document_collections.json
carvil: That's correct.

test/services/filter_tagged_content_test.rb
carvil: Yes, unfortunately they are. I can stub the bit that reads from the file and returned mocked results instead, what do you think?

alecgibson: I personally think that tests being tied to live data is surprising, and can lead to unnecessary breakages later (and are less explicit than setting up a mock with the conditions being tested), so I think this is worth doing.

app/services/filter_tagged_content.rb
carvil: I can make this change. We are not expecting duplicates and even if we have, `include?` will still work, but it's semantically more correct.

PR #282, opened 2017-03-03
Remove the feature flag from collections

carvil:
This means we will be able to A/B test this in all environments. For
now, in production it's only possible to see the B variant with the
chrome extension.

Trello: https://trello.com/c/940kowVa/484-remove-enable-new-navigation-check-from-repos-and-from-puppet


suzannehamilton:
Delete ~all~ some of the things! 🎉 

Looks good to me. 👍

PR #285, opened 2017-03-06
Accordion hash

alecgibson:
This change is primarily to update the accordion element to store (some of) its state in the URL, rather than in [`sessionStorage`](https://developer.mozilla.org/en/docs/Web/API/Window/sessionStorage). This is to address the following observations from user research:

- When expanding an accordion, then continuing their journey and (much later) coming back to the accordion, users were confused as to why parts of it were expanded
- When navigating "back" in their browser from content to the accordion, users expect the section they came from to still be expanded

`sessionStorage` addressed the second issue, but gave rise to the first issue, since the state lingered long after a user has navigated away from the page.

As an alternative approach, state is instead stored in the URL hash, using [`history.replaceState`](https://developer.mozilla.org/en-US/docs/Web/API/History_API) to avoid clobbering browser history.

For example, when a user expands "Section 1", their address bar looks like `gov.uk/some-taxon#section-1`. If they then collapse the section, the hash is removed. If they open another section, the hash is **replaced** rather than appended.

Notably, this JavaScript has [some limitations on compatibility](https://developer.mozilla.org/en-US/docs/Web/API/History_API#Browser_compatibility), so a fallback of not storing state in the URL has also been added. In this case, when navigating back to the accordion, users will find the accordion completely collapsed, but this was deemed to be less confusing than coming back to a pre-expanded accordion after an unrelated journey.

The following edge cases have been identified, but will not be addressed in advance of an identified user need:

- Opening "Section 1", then opening "Section 2" will make my hash `#section-2`. If I click on content from Section 1, and then click "back" in my browser, I see Section 1 expanded and **not** Section 2 (where I came from)
- Clicking "expand all" will clear the hash, so when a user clicks content, then clicks "back", they will be returned to a collapsed accordion

### Trello

- https://trello.com/c/ZBbVh8r0/480-add-a-url-fragment-when-you-open-an-accordion-section
- https://trello.com/c/MLMgltu5/475-switching-off-the-option-that-keeps-the-accordion-open-when-moving-from-a-piece-of-content-selected-from-one-of-the-topic-child-

### DIFF
--- a/app/controllers/taxons_controller.rb
+++ b/app/controllers/taxons_controller.rb
@@ -1,6 +1,8 @@
 class TaxonsController < ApplicationController
   before_action :return_404, unless: :new_navigation_enabled?
 
+  helper_method :taxon_overview_and_child_taxons
+
   def show
     setup_content_item_and_navigation_helpers(taxon)
 
@@ -37,5 +39,21 @@ private
 
   def taxon
     @taxon ||= Taxon.find(request.path)
+  end
+
+  def taxon_overview_and_child_taxons(taxon)
+    accordion_items = taxon.child_taxons
+    return [] if taxon.child_taxons.empty?
+
+    if taxon.tagged_content.count > 0
+      accordion_items.unshift(
+        taxon.merge(
+          'title' => 'Overview',
+          'description' => '',
+        )
+      )
+    end
+
+    accordion_items
   end
 end
--- a/app/controllers/taxons_controller.rb
+++ b/app/controllers/taxons_controller.rb
@@ -50,6 +50,7 @@ private
         taxon.merge(
           'title' => 'Overview',
           'description' => '',
+          'base_path' => 'overview',
         )
       )
     end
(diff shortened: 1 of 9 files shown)


app/controllers/taxons_controller.rb
```diff
@@ -38,4 +40,20 @@ def return_404
   def taxon
     @taxon ||= Taxon.find(request.path)
   end
+
+  def taxon_overview_and_child_taxons(taxon)
+    accordion_items = taxon.child_taxons
+    return [] if taxon.child_taxons.empty?
```
carvil: Minor thing, but you could use `accordion_items` here, instead of `taxon.child_taxons`.

alecgibson: I know, but I thought it would be more explicit this way.

alecgibson: (Plus any opportunity to prevent mutation of objects is a good thing in my book)

PR #287, opened 2017-03-07
Add non-guidance document collections to the list

carvil:
We should not show non-guidance document collections. Instead, we should
show their content, in case any of it is guidance.

Trello: https://trello.com/c/s9kXmGQS/459-document-collections-navigation-rules


alecgibson:
Again, if this is an important rule should we be enforcing it in code rather than JSON?

carvil:
@alecgibson this isn't a hard rule and will need product review before we should enforce it in code. I don't think we know enough about it yet.

These hardcoded document collections will have to be deleted from the code once the review is finished, so this is only temporary.

PR #289, opened 2017-03-07
Configure nav helpers with statsd

carvil:
Currently, collections doesn't use any sidebar, so nothing will be recorded in statsd. But this commit does the setup for consistency with other apps that render education content.

Trello: https://trello.com/c/S72dddmO/488-configure-navigation-helpers-with-statsd


carvil:
I've re-done these changes based on feedback from https://github.com/alphagov/frontend/pull/1157

cbaines:
> I've re-done these changes based on feedback from alphagov/frontend#1157

Ok, still looks good :+1:

PR #292, opened 2017-03-08
Move "- GOV.UK" suffix to application template

tijmenb:
The new navigation pages are currently missing the "- GOV.UK" suffix in the title. This fixes that by moving the suffix to the application template, which will add it for all pages from now on.


suzannehamilton:
LGTM 👍

PR #293, opened 2017-03-08
Remove duplicate entries for document collection

carvil:
We should not show the document collection, and instead show the
content tagged to it. This commit makes sure that's the only entry left
in the file.

Trello: https://trello.com/c/mXa55bSX/507-document-collections-showing-as-false-but-still-appearing-2-examples

### DIFF
--- a/config/guidance_document_collections.json
+++ b/config/guidance_document_collections.json
@@ -10,11 +10,6 @@
     "surface_content": true
   },
   {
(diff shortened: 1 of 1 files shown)


suzannehamilton:
:shipit: 

Just to check: have you looked for other duplicates in the file?

carvil:
@suzannehamilton there are other inconsistencies on the file, I will merge this specific one in and do the other ones separately if that's ok. I needed confirmation from Holly on those.

PR #298, opened 2017-03-08
Add padding bottom to the grid

carvil:
This is to ensure the spacing between the `ol` and the "is there something
wrong with this page" link or tagged content is always at 60px.

cc/ @alextea 

### Before

screen shot 2017-03-08 at 16 40 37

### After

screen shot 2017-03-08 at 16 40 25


alextea:
:+1:

PR #300, opened 2017-03-09
Fix accordion for pages with no JS history object

alecgibson:
In a previous refactor I missed moving the check for JavaScript `history` support. This caused a console error in older browsers which don't support `history.replaceState`.

### Trello

https://trello.com/c/Jj6bq3Sl/55-js-error-reported-in-ie8


alecgibson:
cc @fofr

suzannehamilton:
LGTM 👍

PR #304, opened 2017-03-09
Accordion behaviour fixes

alecgibson:
This Pull Request addresses the following issues with the accordion:
- The full content of the accordion is visible before it's collapsed by JavaScript (FOUC)
- Ctrl + click or Cmd + click didn't open accordion links in new tabs
- When using the browser history to navigate, the accordion scroll position is not set correctly

### Flash of unstyled content (FOUC)

There was a previous fix for FOUC in place that has been removed through refactoring. This was fixed by replacing the `js-hidden` class on the accordion.

### Ctrl + click / Cmd + click

Solution used from here: https://gist.github.com/andreaslof/4ef2935dee9a03267692

### Browser scroll position

When navigating back in browser history to the accordion, the browser will try to be "clever" and return the user to their previous scroll position. However, since we collapse all but the currently-anchored subsection, the content length changes and the user is returned to the wrong position (often the footer).

In order to correct this behaviour, as the user leaves the page, we anticipate the correct height we wish the user to return to by forcibly scrolling them to that height, which becomes the height the browser will return them to.

If we can't find an element to return them to, then reset the scroll to the top of the page. This handles the case where the user has expanded all sections, so they are not returned to a particular section, but still could have scrolled a long way down the page.

### Trello

https://trello.com/c/QPSaKsg3/12-when-navigating-via-the-back-button-from-content-back-to-the-accordion-sometimes-i-jumped-back-to-the-footer-not-the-expanded-le

### DIFF
--- a/app/assets/javascripts/modules/accordion-with-descriptions.js
+++ b/app/assets/javascripts/modules/accordion-with-descriptions.js
@@ -116,7 +116,7 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
 
       function bindToggleForSubsections(accordionTracker) {
         $element.find('.js-subsection-header').click(function (event) {
-          event.preventDefault();
+          preventLinkFollowingForCurrentTab(event);
 
           var subsectionView = new SubsectionView($(this).closest('.js-subsection'));
           subsectionView.toggle();
@@ -126,6 +126,16 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
 
           setOpenCloseAllText();
         });
+      }
+
+      function preventLinkFollowingForCurrentTab(event) {
+        // If the user is holding the ⌘ or Ctrl key, they're trying
+        // to open the link in a new window, so let the click happen
+        if (event.metaKey || event.ctrlKey) {
+          return;
+        }
+
+        event.preventDefault();
       }
 
       function bindToggleOpenCloseAllButton(accordionTracker) {
--- a/app/assets/javascripts/modules/accordion-with-descriptions.js
+++ b/app/assets/javascripts/modules/accordion-with-descriptions.js
@@ -22,6 +22,8 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
 
     this.start = function ($element) {
 
+      $(window).unload(storeScrollPosition);
+
       // Indicate that js has worked
       $element.addClass('js-accordion-with-descriptions');
 
@@ -46,6 +48,24 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
 
       bindToggleForSubsections(accordionTracker);
       bindToggleOpenCloseAllButton(accordionTracker);
+
+      // When navigating back in browser history to the accordion, the browser will try to be "clever" and return
+      // the user to their previous scroll position. However, since we collapse all but the currently-anchored
+      // subsection, the content length changes and the user is returned to the wrong position (often the footer).
+      // In order to correct this behaviour, as the user leaves the page, we anticipate the correct height we wish the
+      // user to return to by forcibly scrolling them to that height, which becomes the height the browser will return
+      // them to.
+      // If we can't find an element to return them to, then reset the scroll to the top of the page. This handles
+      // the case where the user has expanded all sections, so they are not returned to a particular section, but
+      // still could have scrolled a long way down the page.
+      function storeScrollPosition() {
+        closeAllSections();
+        var $subsection = getSubsectionForAnchor();
+
+        document.body.scrollTop = $subsection && $subsection.length
+          ? $subsection.offset().top
+          : 0;
+      }
 
       function addOpenCloseAllButton() {
         $element.prepend('<div class="subsection-controls js-subsection-controls"><button aria-expanded="false">' + bulkActions.openAll.buttonText + '</button></div>');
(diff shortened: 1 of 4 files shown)


alecgibson:
As discussed in real life, I think the zoom case is a corner case that we probably don't need to worry about, because:

- How common is it to actually change your zoom whilst navigating?
- Even in this case, we "fail safe", because you'll be sent higher up the page than where you started (which is much better than being sent to the footer).

alecgibson:
@carvil what actually needs changing here?

app/assets/javascripts/modules/accordion-with-descriptions.js
```diff
@@ -45,6 +47,24 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
       bindToggleForSubsections();
       bindToggleOpenCloseAllButton();
 
+      // When navigating back in browser history to the accordion, the browser will try to be "clever" and return
```
fofr: When the user hits back, shouldn't the same accordion parts be open – wouldn't that put the user in the right place?

app/assets/javascripts/modules/accordion-with-descriptions.js
```diff
@@ -45,6 +47,24 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
+      // the user to their previous scroll position. However, since we collapse all but the currently-anchored
+      // subsection, the content length changes and the user is returned to the wrong position (often the footer).
+      // In order to correct this behaviour, as the user leaves the page, we anticipate the correct height we wish the
+      // user to return to by forcibly scrolling them to that height, which becomes the height the browser will return
+      // them to.
+      // If we can't find an element to return them to, then reset the scroll to the top of the page. This handles
+      // the case where the user has expanded all sections, so they are not returned to a particular section, but
+      // still could have scrolled a long way down the page.
+      function storeScrollPosition() {
```
fofr: Users should be taken back to an open accordion at the right part of the content.

app/assets/javascripts/modules/accordion-with-descriptions.js
```diff
@@ -126,6 +146,16 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
         });
       }
 
+      function preventLinkFollowingForCurrentTab(event) {
```
fofr: What's the user benefit of cmd+clicking a link that acts only as a toggle?

fofr:
This PR covers 3 fixes, I'd expect 3 commits for each, detailing the fix and why it was needed. The single commit on this PR has no detail.

app/assets/javascripts/modules/accordion-with-descriptions.js
fofr: An example on GOV.UK where cmd+clicking acts as a toggle: https://www.gov.uk/guidance/the-highway-code/rules-for-pedestrians-1-to-35

app/assets/javascripts/modules/accordion-with-descriptions.js
fofr: An example on GOV.UK where cmd+clicking acts as a toggle: https://www.gov.uk/guidance/the-highway-code/rules-for-pedestrians-1-to-35

app/assets/javascripts/modules/accordion-with-descriptions.js
alecgibson: It would. I think ideally we'd restore all the expanded sections again (like we did when using session storage). The problem with session storage was that if a user followed a circular journey that indirectly brought them back to the accordion, we found that users in testing were very confused as to why some sections were already open.

The problem of detecting if a user has just hit "Back" in their browser vs. (say) loading a bookmark is quite hard, but I'm open to suggestions.

app/assets/javascripts/modules/accordion-with-descriptions.js
alecgibson: (see above comment)

app/assets/javascripts/modules/accordion-with-descriptions.js
alecgibson: It doesn't act only as a toggle. If you open the link in a new tab, it will scroll you to the correct expanded accordion section (same as if you bookmark the section, or share it)

app/assets/javascripts/modules/accordion-with-descriptions.js
alecgibson: (I think the "real" solution to this is to keep the state perfectly in sync with the address bar, with something like a query string. This tends to give the most predictable and consistent behaviour. However this got vetoed by Alex, because it would make the URL ugly)

alecgibson: Although now I think about it, perhaps a "middle ground" solution could be something along the lines of:

- If a user navigates to `/accordion`, show all sections collapsed
- If a user navigates to `/accordion#section1`, jump to section1 and expand just that section
- If a user navigates to `/accordion#`, restore session from session storage
- Toggling links in the accordion will simply set the address bar to `/accordion#`. If a user wants to share a link, they'll have to copy the link from the accordion header (which will have `href=/accordion#section1` - this will give good middle-click behaviour, and allow said sharing)

cc @carvil

alecgibson:
@fofr done (cc @carvil)

alecgibson:
@carvil, @alextea says he's alright with this behaviour for now.

PR #308, opened 2017-03-14
Use document supertype to fetch guidance content

carvil:
This commit switches from using a hard-coded list of document types that
describe guidance content, to using a new field recently added to
search. This new field encapsulates the same logic and is being
populated during indexing in rummager and will be always present.

I deployed this branch to integration and ran wraith on all navigation pages against production to see if the content shown on each of the pages would change. There were no major changes between both environments.

Trello: https://trello.com/c/PtQXkimK/520-switch-from-using-hardcoded-guidance-document-types-into-using-document-supertypes


suzannehamilton:
This makes sense. 👍 

Were the minor changes between integration and production just things which had been updated since data replication?

PR #312, opened 2017-03-17
Add heatmap tracking for benchmarking

carvil:
### Background

We are going to conduct a 2nd round of Benchmarking in GOV.UK. The idea is that we will try it with the new Education Navigation enabled across GOV.UK and see how it performs against the current navigation patterns (mainstream browse pages, topics, etc).

Besides checking if people actually complete the tasks, we would also like to know more details about how they use the new navigation pages and also about the new orientation in content pages. In order to do that, we will track mouse movements and gather the results in a heatmap. We are using a software called mouseflow to do this.

The tracking code will only be present for the people doing the benchmarking test (around 60 people). It will be behind a new A/B test so the tracking doesn't run for other GOV.UK users. These 60 users will be presented with a consent form where this will be explained.

### What this PR does

This PR adds the Mouseflow tracking code behind a new A/B test (`GOVUK-ABTest-Benchmarking`). With this new A/B test, we will be able to enable Mouseflow for all 60 users in the benchmarking round. This will allow us to collect data on how the navigation pages are used and generate a heatmap in the end in order to help us understand what areas of the site will be improved.

Trello: https://trello.com/c/NJ6O76BR/281-heatmap-tracking

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -11,6 +11,8 @@
   <%= yield :meta_tags %>
   <%= yield :meta_section %>
 
+  <%= benchmarking_ab_test.requested_variant.analytics_meta_tag.html_safe %>
+
   <%= render partial: 'govuk_component/analytics_meta_tags',
     locals: { content_item: @content_item } %>
 </head>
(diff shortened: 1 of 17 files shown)


app/views/layouts/application.html.erb
```diff
@@ -11,6 +11,10 @@
   <%= yield :meta_tags %>
   <%= yield :meta_section %>
 
+  <% if should_track_mouse_movements? %>
```
carvil: I need to remove this conditional, but first `govuk_ab_testing` needs to allow for it.

carvil:
This PR is now ready for review.

PR #318, opened 2017-03-27
Fix JavaScript-disabled accordion

alecgibson:
The use of display: none in the CSS prevents browsers without
JavaScript enabled from seeing any content under accordion
sections. This fix removes the use of this styling to hide
the content, and instead toggles the js-hidden class

### DIFF
--- a/app/assets/stylesheets/views/_taxons.scss
+++ b/app/assets/stylesheets/views/_taxons.scss
@@ -101,8 +101,7 @@
     }
   }
 
-  .subsection,
-  .subsection-is-open {
+  .subsection {
     width: 100%;
 
     .subsection-description {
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/views/_taxons.scss
```diff
@@ -243,14 +242,4 @@
...
     }
   }
-
-  .subsection-content {
-    display: none;
-  }
-
-  .subsection-is-open {
-    .subsection-content {
-      display: block;
```
suzannehamilton: Is `subsection-content` still a `block` in the new version? Is there an equivalent of this line?

alecgibson: As discussed, this is still a `block`, as I incorrectly was using this to show/hide content (rather than set the display property).

PR #327, opened 2017-04-19
Enable Heroku review apps

carvil:
This commit allows us to deploy the application to Heroku and therefore
enable review apps for each PR.

App is currently running on http://govuk-collections.herokuapp.com/education

With this in the repo I will be able enable review apps.

### DIFF
--- a/app.json
+++ b/app.json
@@ -1,0 +1,40 @@
+{
+  "name": "collections",
+  "repository": "https://github.com/alphagov/collections",
+  "env": {
+    "GOVUK_APP_DOMAIN": {
+      "value": "www.gov.uk"
+    },
+    "GOVUK_ASSET_ROOT": {
+      "value": "https://govuk-collections.herokuapp.com"
+    },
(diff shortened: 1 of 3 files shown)


app.json
```diff
@@ -0,0 +1,40 @@
+{
+  "name": "collections",
+  "repository": "https://github.com/alphagov/collections",
+  "env": {
+    "GOVUK_APP_DOMAIN": {
+      "value": "www.gov.uk"
+    },
+    "GOVUK_ASSET_ROOT": {
+      "value": "https://govuk-collections.herokuapp.com"
```
tijmenb: I think this needs to point to the review app itself? Otherwise you'll get the assets from master instead of the PR?

carvil: I think you're right, but this is only the initial configuration. I think I need to change the Procfile in order to override this value:

https://github.com/alphagov/government-frontend/blob/master/Procfile#L1

carvil: @tijmenb I've added a new commit for this: b9917422b71501db69cfb6287ac3453204add419

PR #329, opened 2017-04-20
Remove float logic from child taxons grid view

Rosa-Fox:
- Control float logic within css

### DIFF
--- a/app/assets/stylesheets/views/_taxons.scss
+++ b/app/assets/stylesheets/views/_taxons.scss
@@ -58,6 +58,10 @@
           min-height: (5 * $gutter);
         }
       }
+
+      li:nth-child(3n+1) {
+        clear: left;
+      }
     }
 
     p {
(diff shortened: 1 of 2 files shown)


alextea:
We will need to test this on browsers. `nth-child` isn't supported in IE8 and lower.

alextea:
We had the same functionality before (@397c29) and removed it because of browser incompatibility issues.

alecgibson:
Here's the original Trello card for context: https://trello.com/c/g8rg70Oa/359-find-a-way-of-making-sure-the-grid-doesn-t-break-without-using-nth-selectors-which-are-not-supported-in-ie8

One day, we'll live in a beautiful, IE-free utopia of `nth` selectors and maybe even `css-grid`. But sadly today is not that day.

Feel free to raise a separate PR to comment on this CSS if you think it would help.

Rosa-Fox:
Thanks both!

PR #330, opened 2017-04-27
Upgrade to rails 5.0.2

Davidslv:
I'll rebase and clean up the commits after people had the chance to review.

Trello: https://trello.com/c/GYeuyMtX/69-upgrade-collections-to-rails-5


Gemfile
```diff
@@ -1,13 +1,13 @@
 source 'https://rubygems.org'
 
-gem 'rails', '4.2.7.1'
+gem 'rails', '5.0.0'
```
carvil: We could use 5.0.2: https://rubygems.org/gems/rails

config/cable.yml
```diff
@@ -0,0 +1,9 @@
+development:
+  adapter: async
+
+test:
+  adapter: async
+
+production:
+  adapter: redis
+  url: redis://localhost:6379/1
```
carvil: Do we need to setup action cable? I'm not sure we use it...

config/environments/production.rb
```diff
@@ -29,59 +25,44 @@
...
   config.assets.compile = false
 
-  # Generate digests for assets URLs.
-  config.assets.digest = true
-
-  # Version of your assets, change this if you want to expire all your assets.
-  config.assets.version = '1.0'
+  # `config.assets.precompile` and `config.assets.version` have moved to config/initializers/assets.rb
 
-  if ENV['RAILS_SERVE_STATIC_ASSETS'] == 'yes'
```
carvil: This seems to have been deleted, was it on purpose?

config/environments/test.rb
```diff
@@ -26,8 +28,6 @@
   # Disable request forgery protection in test environment.
   config.action_controller.allow_forgery_protection = false
 
-  # Print deprecation notices to the stderr.
-  config.active_support.deprecation = :stderr
-
-  config.active_support.test_order = :random
```
carvil: Was this deleted on purpose?

config/environments/test.rb
```diff
@@ -26,8 +28,6 @@
   # Disable request forgery protection in test environment.
   config.action_controller.allow_forgery_protection = false
 
-  # Print deprecation notices to the stderr.
-  config.active_support.deprecation = :stderr
```
carvil: Same question as below.

config/puma.rb
```diff
@@ -0,0 +1,47 @@
+# Puma can serve each request in a thread from an internal thread pool.
```
carvil: I think we can delete this file. Puma isn't in the Gemfile.

carvil: More details: https://github.com/alphagov/govuk-developer-docs/blob/master/source/manual/upgrade-rails.html.md

config/environments/production.rb
Davidslv: It seems the option no longe exists under that name:

> config.public_file_server.enabled configures Rails to serve static files from the public directory. This option defaults to true, but in the production environment it is set to false because the server software (e.g. NGINX or Apache) used to run the application should serve static files instead. If you are running or testing your app in production mode using WEBrick (it is not recommended to use WEBrick in production) set the option to true. Otherwise, you won't be able to use page caching and request for files that exist under the public directory.

http://guides.rubyonrails.org/v5.0.0/configuring.html#configuring-assets

config/environments/test.rb
Davidslv: The default is now `:random`

> config.active_support.test_order sets the order that test cases are executed. Possible values are :random and :sorted. This option is set to :random in config/environments/test.rb in newly-generated applications. If you have an application that does not specify a test_order, it will default to :sorted, until **Rails 5.0, when the default will become :random.**

But I'll add it back

Davidslv:
@carvil I believe I've addressed the comments.
I forgot to refer that airbrake upgrade will not work with our current errbit - although that was stated in here https://github.com/alphagov/govuk-developer-docs/blob/master/source/manual/upgrade-rails.html.md#airbrake

PR #333, opened 2017-05-08
We are making the Analytics tracking interface more explicit in which

alecgibson:
methods are used external to the module. This involves:

- removing the methods that set custom dimensions at a session level,
  and instead explicitly setting custom dimensions on the pageview
- removing the `callOnNextPage` method, which allowed arbitrary running
  of any public methods on the analytics interface, where we just need
  to set options for the next page


alecgibson:
This isn't working properly - pulling and will reopen when fixed.

PR #335, opened 2017-05-09
Update accordion tracking

alecgibson:
## Update accordion tracking label

When clicking an accordion heading, we send a tracking event. We
previously tracked where on the header we clicked:

- on the heading link
- on the +/- icon
- anywhere else (whitespace or description)

In analysis, this amount of information in the tracking label was too
verbose, so we are removing it.

Instead, we now return just the heading, and its position in the
accordion. For example, if we have a taxon entitled “Phonics” as the
fourth taxon in the accordion, if that section is expanded the tracking
label will be `4. Phonics`.

## Track extra event for accordion heading clicks

This change adds an extra event to accordion heading clicks that aligns
with the event submitted when a content item is clicked in an accordion
(see: https://github.com/alphagov/collections/blob/b8b3313abdd960f8ad6fa99eb7107a9fe12d8a94/app/views/taxons/_content_list_for_child_taxon.html.erb#L7-L14).

This behaviour is also more consistent with the taxonomy sidebar
tracking, where the sidebar heading is tracked as well as the listed
items.

## Trello

https://trello.com/c/udNckXa7/117-amend-accordion-analytics-tracking

### DIFF
--- a/app/assets/javascripts/modules/accordion-with-descriptions.js
+++ b/app/assets/javascripts/modules/accordion-with-descriptions.js
@@ -212,6 +212,7 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
       var shouldUpdateHash = true;
 
       this.title = $subsectionElement.find('.js-subsection-title').text();
+      this.href = $titleLink.attr('href');
       this.element = $subsectionElement;
 
       this.open = open;
@@ -221,6 +222,7 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
       this.isOpen = isOpen;
       this.isClosed = isClosed;
       this.preventHashUpdate = preventHashUpdate;
+      this.numberOfContentItems = numberOfContentItems;
 
       function open() {
         setIsOpen(true);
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/modules/accordion-with-descriptions.js
```diff
@@ -212,6 +212,7 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
       var shouldUpdateHash = true;
 
       this.title = $subsectionElement.find('.js-subsection-title').text();
+      this.element = $subsectionElement;
```
carvil: What do you think of setting this above and then use `this.element` instead of `$subsectionElement` in this function?

alecgibson: `this` is really confusing in JavaScript, so I prefer to avoid where possible. Also, `this.element` wouldn't work in the `function`s in this object, so you'd _still_ need the `$subsectionElement` variable in places and I'd rather use that than the confusing `this.element` call.

PR #339, opened 2017-05-11
Track link clicks on browse pages

alecgibson:
This change introduces click event tracking identical to the tracking
already implemented on taxonomy navigation pages (grid, accordion, etc.),
where we track an event every time a link is clicked with:

- the title of the link
- the `href` of the link
- the position of the link in its list
- the number of other links in its list

### DIFF
--- a/app/views/second_level_browse_page/_links.html.erb
+++ b/app/views/second_level_browse_page/_links.html.erb
@@ -1,7 +1,7 @@
 <div class="pane-inner <%= page.lists.curated? ? 'curated-list' : 'a-to-z' %>">
   <h1 tabindex="-1"><%= page.title %></h1>
 
-  <% page.lists.each do |list| %>
+  <% page.lists.each_with_index do |list, section_index| %>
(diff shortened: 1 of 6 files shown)


app/views/second_level_browse_page/_links.html.erb
```diff
@@ -19,8 +33,22 @@
...
-        <% page.related_topics.each do |related_topic| %>
-          <li><%= link_to related_topic.title, related_topic.base_path %></li>
+        <% page.related_topics.each_with_index do |related_topic, index| %>
+          <li>
+            <%= link_to(
+              related_topic.title,
+              related_topic.base_path,
+              data: {
+                track_category: 'thirdLevelBrowseLinkClicked',
+                track_action: "#{page.lists.count + 1}.#{index + 1}",
```
carvil: Does the `count` call need to be incremented?

alecgibson: Yes, because if there are 3 lists before this one, this is the 4th list.

PR #340, opened 2017-05-11
Test email alert link and refactor logic

carvil:
This commit adds tests for the email subscription link on navigation
pages. It also refactors logic around when to show the email
subscription link on accordion pages so it's clearer. On accordion
pages, the "General Information and Guidance" section should not have
the email subscription link on it, so I've renamed the method to
indicate that.

### DIFF
--- a/app/models/taxon.rb
+++ b/app/models/taxon.rb
@@ -1,5 +1,7 @@
 class Taxon
   attr_reader :content_item
+  attr_accessor :can_subscribe, :tagged_content
+
   delegate(
     :content_id,
     :base_path,
@@ -53,5 +55,11 @@ class Taxon
 
   def merge(to_merge)
     Taxon.new(content_item.merge(to_merge))
+  end
+
+  def can_subscribe?
+    return @can_subscribe if defined?(@can_subscribe)
+
+    true
   end
 end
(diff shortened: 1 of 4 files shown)


app/models/taxon.rb
```diff
@@ -54,4 +56,10 @@ def grandchildren?
   def merge(to_merge)
     Taxon.new(content_item.merge(to_merge))
   end
+
+  def can_subscribe
```
alecgibson: `can_subscribe?`?

app/models/taxon.rb
```diff
@@ -54,4 +56,10 @@ def grandchildren?
...
+    return @can_subscribe if defined?(@can_subscribe)
```
alecgibson: Should probably `return !!@can_subscribe`

app/models/taxon.rb
carvil: Will change 👍

app/models/taxon.rb
carvil: In this case, `nil` or `false` will be "falsy", so it's ok to leave it I think.

alecgibson: I guess I just find it weird that people can set this value to be things that aren't `boolean`. Because you can still do `taoxn.can_subscribe = []` for example, and then when you ask `can_subscribe?` and get an array, it's a bit weird.

PR #348, opened 2017-06-14
Add associated taxons support

gpeng:
We have added a new link type `associated_taxons` which allows a taxon to be associated with another. The effect of this association is that content tagged to the associated taxon(s) is included in the content of the taxon that declares the association.

e.g.

* `usa` taxon has `child_taxons` of `[travelling-in-the-usa]`
* `travelling-in-the-usa` has `associated_taxons` of `[travelling-abroad]`

The taxon page for `usa` would display a 'Travelling in the USA' heading. The 'Travelling in the USA' heading would contain content tagged to `travelling-in-the-usa` and also any content tagged to `travelling-abroad` .

This PR adds support for this feature.

[Trello](https://trello.com/c/r3jl4SE4/178-allow-contents-of-associated-taxons-to-be-displayed-on-a-taxon-navigation-page)

Discussed with @carvil how we can include this associated taxon content in the `#most_popular_content` results. We don't necessarily require this currently but it makes sense for it to be included in the results. Any suggestions welcome.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,11 @@ Collections serves the GOV.UK browse and topic pages.
   `child_taxons` links. None of those child taxons' links have `child_taxons`,
   in which case we display an accordion view:
   [gov.uk/education/school-governance](https://www-origin.integration.publishing.service.gov.uk/education/school-governance)
+- **Taxon with associated taxons**: a content item of type taxon that has
+  `associated_taxons` links. In this case the tagged content of the taxon will
+  include content that is directly tagged to it and also content that has been
+  tagged to any of the associated taxons.
+  
 
 ## Technical documentation
 
--- a/app/models/taxon.rb
+++ b/app/models/taxon.rb
@@ -79,8 +79,6 @@ private
 
   def fetch_tagged_content
     taxon_content_ids = [content_id] + associated_taxons.map(&:content_id)
-    taxon_content_ids.inject([]) do |content, taxon_content_id|
-      content + TaggedContent.fetch(taxon_content_id)
-    end
+    TaggedContent.fetch(taxon_content_ids)
   end
 end
--- a/app/models/taxon.rb
+++ b/app/models/taxon.rb
@@ -79,8 +79,8 @@ private
 
   def fetch_tagged_content
     taxon_content_ids = [content_id] + associated_taxons.map(&:content_id)
-    taxon_content_ids.inject([]) do |result, content_id|
-      result + TaggedContent.fetch(content_id)
+    taxon_content_ids.inject([]) do |content, taxon_content_id|
+      content + TaggedContent.fetch(taxon_content_id)
     end
   end
 end
--- a/app/models/taxon.rb
+++ b/app/models/taxon.rb
@@ -17,7 +17,7 @@ class Taxon
   end
 
   def tagged_content
-    @tagged_content ||= TaggedContent.fetch(content_id)
+    @tagged_content ||= fetch_tagged_content
   end
 
   def most_popular_content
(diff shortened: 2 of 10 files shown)


app/models/taxon.rb
```diff
@@ -68,4 +74,13 @@ def can_subscribe?
 
     true
   end
+
+private
+
+  def fetch_tagged_content
+    taxon_content_ids = [content_id] + associated_taxons.map(&:content_id)
+    taxon_content_ids.inject([]) do |content, taxon_content_id|
+      content + TaggedContent.fetch(taxon_content_id)
```
bevanloon: Might be worth exploring whether Rummager allows multiple `content_id`'s in a single request for `filter_taxons` - it looks like it is passing in an array (`filter_taxons: [content_id],`) - https://github.com/alphagov/collections/blob/master/app/services/tagged_content.rb#L30

gpeng: Thanks @bevanloon I've had a stab at this in 77c6592

app/models/taxon.rb
```diff
@@ -68,4 +74,11 @@ def can_subscribe?
 
     true
   end
+
+private
+
+  def fetch_tagged_content
+    taxon_content_ids = [content_id] + associated_taxons.map(&:content_id)
+    TaggedContent.fetch(*taxon_content_ids)
```
bevanloon: What does the splat do here? Does it turn the array into individual multiple arguments that get collected by the receiver's catch all?

app/services/tagged_content.rb
```diff
@@ -27,7 +27,7 @@ def search_response
       count: RummagerSearch::PAGE_SIZE_TO_GET_EVERYTHING,
```
bevanloon: Sorry to be a pain, given that I suggested this approach, but if we have for example, 2 content ids and the first one brings back >= `PAGE_SIZE_TO_GET_EVERYTHING` documents, will rummager ignore the second content id, or does it merge the two results, order them and then send us back the first x number of them?

I know we currently have `PAGE_SIZE_TO_GET_EVERYTHING` set to 1000 so pretty big, but just want to be sure we aren't going to get surprising results here.

tijmenb: > merge the two results, order them and then send us back the first x number of them

Pretty much this. Rummager creates an `OR` query for elasticsearch.

app/models/taxon.rb
gpeng: Yep. I'm wondering whether the whole splat thing is a bit messy. In the test helper I've just made it a string or an array. https://github.com/alphagov/collections/pull/348/commits/77c6592ecf9ccbeebc2e5e2dcf4dab1d62def2a3#diff-c3f5864092d1f31b743d969d943bd38cR2 Is that clearer?

app/services/tagged_content.rb
bevanloon: 👍 cool

app/models/taxon.rb
bevanloon: Not sure I have a strong opinion on which is cleaner. I just didn't know what the splat did on this side of the message sending.

gpeng:
@bevanloon I've updated `TaggedContent` to accept an array if you could just give it another quick look.

README.md
```diff
@@ -45,6 +45,11 @@ Collections serves the GOV.UK browse and topic pages.
   `child_taxons` links. None of those child taxons' links have `child_taxons`,
   in which case we display an accordion view:
   [gov.uk/education/school-governance](https://www-origin.integration.publishing.service.gov.uk/education/school-governance)
+- **Taxon with associated taxons**: a content item of type taxon that has
+  `assoicated_taxons` links. In this case the tagged content of the taxon will
```
gpeng: thanks... Fixed in 8106944

PR #352, opened 2017-06-22
Add summary description for News pages

bevanloon:
Worldwide News pages will be tagged to World Loctaions and as such
will display on World Location navigation pages. We want a specific hard
coded summary of "Updates, news and events from the UK government in
[country]" to display under these.

This commit adds a description helper to display the summary.

The only way to determine if a page is a worldwide news page is to test
the base path returned by Rummager.

### DIFF
--- a/app/helpers/taxon_helper.rb
+++ b/app/helpers/taxon_helper.rb
@@ -1,0 +1,22 @@
+module TaxonHelper
+  def taxon_description(taxon, presented_taxon)
+    return taxon.description if taxon.description.present?
+    worldwide_news_description(taxon, presented_taxon)
+  end
+
+private
+
+  # There is currently no good way to determine whether or not the taxon
+  #  is a worldwide news page from the results returned by rummager.
+  #  This is because worlwide news pages are just aggregates of publications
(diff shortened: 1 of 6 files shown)


app/helpers/taxon_helper.rb
```diff
@@ -0,0 +1,22 @@
...
+  def taxon_description(taxon, presented_taxon)
+    return taxon.description if taxon.description.present?
+    worldwide_news_description(taxon, presented_taxon)
+  end
+
+private
+
+  # There is currently no good way to determine whether or not the taxon
+  #  is a worldwide news page from the results returned by rummager.
+  #  This is because worlwide news pages are just aggregates of publications
```
rubenarakelyan: Nitpick spelling :)

tijmenb:
Hmm, I think I don't really understand what's going on and what problem this is trying to solve. Could you add some screenshots of before/after? I think we should avoid hard coding this stuff at all cost, but I'm not sure what it's trying to achieve.

bevanloon:
Closing in favour of https://github.com/alphagov/whitehall/pull/3314

PR #353, opened 2017-06-28
Set specific order for world taxonomy accordion content

davidbasalla:
For https://trello.com/c/A5XJMJzD/265-fix-ordering-issue-for-world-taxon-pages

For the new world taxon pages (eg /world/germany), we require the child taxons to be displayed in a specific order. This commit adds short-term solution. Ideally we would like to be determine the order somewhere else in the publishing platform, but given the time constraints, this might be the best solution at the moment.

We rely on the title of the taxons so that's quite brittle, but not sure if there's anything better?

### DIFF
--- a/app/presenters/taxon_presenter.rb
+++ b/app/presenters/taxon_presenter.rb
@@ -41,7 +41,7 @@ class TaxonPresenter
 
   def accordion_content
     return [] unless renders_as_accordion?
-    accordion_items = taxon.child_taxons.map { |taxon| TaxonPresenter.new(taxon) }
+    accordion_items = ordered_child_taxons.map { |taxon| TaxonPresenter.new(taxon) }
 
     general_information_title = 'General information and guidance'
 
@@ -62,5 +62,11 @@ class TaxonPresenter
     end
 
     accordion_items
+  end
+
+private
+
+  def ordered_child_taxons
+    taxon.world_related? ? WorldTaxonomySorter.call(taxon.child_taxons) : taxon.child_taxons
   end
 end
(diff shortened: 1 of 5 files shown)


app/presenters/taxon_presenter.rb
```diff
@@ -63,4 +63,10 @@ def accordion_content
 
     accordion_items
   end
+
+private
+
+  def ordered_child_taxons
+    taxon.world_related? ? WorldTaxonomySorter.call(taxon.child_taxons) : taxon.child_taxons
```
bevanloon: This is fairly minor - but I wonder if pushing the `if` into `WorldTaxonomySorter` would make this work easier to "upgrade" when we come up with a more robust solution. If `WorldTaxonomySorter` is completely responsible for deciding when to sort, it pushes the extra logic into one place which makes changing the way (or even the logic of when) we sort a bit easier for the future.

I might be tempted to actually test in `WorldTaxonomySorter` if the `base_path` starts with `/world` - I know it would duplicate code from the `Taxon` model, but it would allow the Taxon model to keep the `world_related?` private - that way, nobody else would be tempted to rely on that portion of the `Taxon` model interface which might make it better in the future if/when we want to make that more generic?

It's probably fairly academic though, so just something to discuss maybe.

davidbasalla: I like the idea of putting as much of the logic into one place. However if we do that, then in TaxonPresenter we'd always be returning the `child_taxons` via the `WorldTaxonomySorter.call` method (if I understand you correctly) which also feels a bit odd? Maybe the naming isn't 100% right here... anyway, let's discuss tomorrow

PR #354, opened 2017-06-28
Skip document validation for world content

rubenarakelyan:
This commit adds a `validate` flag to `TaggedContent` to allow taxon tagged content validation to be skipped for world content. This is because we do not restrict the types of content that can be tagged to the worldwide taxonomy.


davidbasalla:
👍

PR #355, opened 2017-06-29
Remove /government from Whitehall email subscription links

bevanloon:
Whitehall bound email subscriptions links for World Locations should be
sending a subscription link of the form `/world/<location>.atom` instead
of `/government/world/<location>.atom`.


davidbasalla:
Looks good!

PR #363, opened 2017-07-20
Fix broken email links for /world content

rubenarakelyan:
This commit fixes broken email subscription links for `/world` content:

* Links to email subscriptions and feeds inside accordions are not displayed for pages under `/world` since these links simply replicate what is shown at the top of the page

* Links to email subscriptions and feeds at the top of “leaf” pages under `/world` are replaced by the default email subscription links to allow for taxonomy-based email subscriptions - these previously linked to broken whitehall pages

Trello: https://trello.com/c/AcE14fLU/40-remove-non-subscription-email-subscriptions-link-from-world-navigation

Before:

screen shot 2017-07-20 at 09 54 52

screen shot 2017-07-20 at 09 55 02

After:

screen shot 2017-07-20 at 09 54 39

screen shot 2017-07-20 at 09 55 10

### DIFF
--- a/app/helpers/email_helper.rb
+++ b/app/helpers/email_helper.rb
@@ -1,8 +1,6 @@
 module EmailHelper
-  WHITEHALL_BOUND_BASE_PATH = "/world".freeze
-
-  def render_whitehall_email_links?(base_path)
-    base_path.starts_with?(WHITEHALL_BOUND_BASE_PATH) if base_path
+  def render_whitehall_email_links?(presented_taxon)
+    presented_taxon.world_related? && presented_taxon.renders_as_accordion?
   end
 
   def whitehall_atom_url
--- a/app/models/taxon.rb
+++ b/app/models/taxon.rb
@@ -69,6 +69,7 @@ class Taxon
 
   def can_subscribe?
     return @can_subscribe if defined?(@can_subscribe)
+    return false if world_related?
 
     true
   end
--- a/app/presenters/taxon_presenter.rb
+++ b/app/presenters/taxon_presenter.rb
@@ -14,6 +14,7 @@ class TaxonPresenter
     :child_taxons,
     :most_popular_content,
     :can_subscribe?,
+    :world_related?,
     to: :taxon
   )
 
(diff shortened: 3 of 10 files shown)


app/helpers/email_helper.rb
```diff
@@ -2,7 +2,8 @@ module EmailHelper
   WHITEHALL_BOUND_BASE_PATH = "/world".freeze
 
   def render_whitehall_email_links?(base_path)
-    base_path.starts_with?(WHITEHALL_BOUND_BASE_PATH) if base_path
+    @presented_taxon.rendering_type == TaxonPresenter::ACCORDION &&
```
gpeng: I think this would be a bit easier to understand if the two parts of the condition were broken into methods e.g. `is_an_accordion? && handled_by_whitehall?(base_path)`

rubenarakelyan: Done

app/helpers/email_helper.rb
```diff
@@ -28,4 +27,12 @@ def whitehall_base_email_sign_up_url
   def equivalent_whitehall_url
     root + request.fullpath
   end
+
+  def is_accordion?(presented_taxon)
```
davidbasalla: The `TaxonPresenter` class has the `renders_as_accordion?` method, we could use that instead?

app/helpers/email_helper.rb
```diff
@@ -28,4 +27,12 @@ def whitehall_base_email_sign_up_url
...
+    presented_taxon.rendering_type == TaxonPresenter::ACCORDION
+  end
+
+  def is_handled_by_whitehall?(presented_taxon)
+    presented_taxon.base_path.starts_with?(WHITEHALL_BOUND_BASE_PATH) if presented_taxon.base_path
```
davidbasalla: The `Taxon` class has the `world_related?` method, we could use that since it appears to do the same

app/helpers/email_helper.rb
rubenarakelyan: Done

app/helpers/email_helper.rb
rubenarakelyan: Done

davidbasalla:
Looks good to me! 👍

PR #367, opened 2017-08-17
Create a new tracking event navGridContentClicked

cbaines:
https://trello.com/c/FABDiyBp/150-create-new-event-tracking-navgridcontentclicked

A new event is added called 'navGridContentClicked' is added to the grid page with the following attributes:

- Event label: url of the content link clicked
- Event action: position of the content link clicked this should begin 1-12 and then return a L1, L2, L3 to signify leaf link
- Navigation section (dimension26): 1 or 2 (depend if there is leaf grid section or not)
- Navigation links (dimension27: total number of links on page (incl main and leaf section)
- Number of links(dimension28): number of links in section of link that was clicked (either main or leaf grid section)
- Link text (dimension29): displayed text of content link

### DIFF
--- a/test/integration/taxon_browsing_test.rb
+++ b/test/integration/taxon_browsing_test.rb
@@ -33,7 +33,6 @@ class TaxonBrowsingTest < ActionDispatch::IntegrationTest
     and_i_can_see_links_to_the_child_taxons_in_a_grid
     and_the_grid_has_tracking_attributes
     and_i_can_see_content_tagged_to_the_taxon
-    and_the_content_tagged_to_the_grandfather_taxon_has_tracking_attributes
     and_the_page_is_tracked_as_a_grid
     and_i_can_see_an_email_signup_link
   end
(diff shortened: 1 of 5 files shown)


test/presenters/taxon_presenter_test.rb
```diff
@@ -49,4 +52,150 @@
...
+    let(:content_item) { ContentItem.new(content_hash) }
+    let(:taxon) { Taxon.new(ContentItem.new(content_hash)) }
+    let(:child_taxon) { taxon.child_taxons.first }
+    let(:taxon_presenter) { TaxonPresenter.new(taxon) }
+
+    describe '#options_for_leaf_content' do
+      before :each do
+        @search_results = generate_search_results(15)
+        stub_content_for_taxon([taxon.content_id], @search_results)
+      end
```
whoojemaflip: Picky maybe, but could there be a newline between each block? It's hard to read such a dense wall of text

koetsier: sure, as long as 🤖 approves

test/presenters/taxon_presenter_test.rb
```diff
@@ -49,4 +52,150 @@
...
+        end
+        it 'contains track_label equal to the base path of the content' do
+          assert_equal(@search_results[10]['link'], subject[:track_label])
+        end
+      end
+      describe 'dimensions' do
+        subject { taxon_presenter.options_for_leaf_content(index: 10)[:track_options] }
+        describe 'dimension 28 - number of tagged content items' do
+          it 'contains the number of tagged content items' do
+            assert_equal('15', subject[:dimension28])
```
whoojemaflip: this could use some [custom matchers](https://relishapp.com/rspec/rspec-expectations/v/3-3/docs/custom-matchers/define-matcher) to really make it readable

test/presenters/taxon_presenter_test.rb
whoojemaflip: oh yeah the 🤖 loves it

test/integration/taxon_browsing_test.rb
```diff
@@ -375,13 +362,6 @@ def and_i_can_see_content_tagged_to_the_taxon
   alias_method :and_i_can_see_content_tagged_to_the_taxon_and_the_associate,
     :and_i_can_see_content_tagged_to_the_taxon
 
-  def and_the_content_tagged_to_the_grandfather_taxon_has_tracking_attributes
```
whoojemaflip: why did this need to be removed?

test/presenters/taxon_presenter_test.rb
koetsier: no he doesn't. he hates newlines! - anyway, added a few

koetsier: added as many as allowed now

test/presenters/taxon_presenter_test.rb
koetsier: Would love to, but this project uses minitest.

test/integration/taxon_browsing_test.rb
koetsier: Mainly because the new logic is more complicated and I thought the integration test was not the right place to test it exhaustively. So now instead of testing if each tag has the right attributes (the old way and would now be too complicated), the new integration test only tests one specific case. The rest of the logic is tested in the taxon_presenter_test.

test/presenters/taxon_presenter_test.rb
whoojemaflip: omg it totally does.

PR #379, opened 2017-09-06
Identify legacy navigation in Analytics

whoojemaflip:
Refactoring the AB test controller code, before implementing the new dimension to identify legacy navigation pages.

[Trello](https://trello.com/c/IPAyElBg/248-send-identifying-value-for-legacy-education-childcare-navigation-pages)

### DIFF
--- a/app/controllers/browse_controller.rb
+++ b/app/controllers/browse_controller.rb
@@ -34,11 +34,11 @@ private
   def show_html(page)
     taxon_resolver = TaxonRedirectResolver.new(
       ab_variant,
-      is_page_in_ab_test: lambda { top_level_redirect.present? },
+      page_is_in_ab_test: page_in_ab_test?,
       map_to_taxon: lambda { top_level_redirect }
     )
 
-    if taxon_resolver.page_ab_tested?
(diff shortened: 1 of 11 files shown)


app/controllers/browse_controller.rb
```diff
@@ -40,7 +40,7 @@ def show_html(page)
 
     configure_ab_response if page_in_ab_test?
 
-    if taxon_resolver.taxon_base_path
+    if taxon_resolver.redirect?
```
cbaines: From just reading this, I'm not sure why there is a conditional redirect here. Is there another name for the `redirect?` method that would make the purpose of this code clearer?

whoojemaflip: The `TaxonRedirectResolver#redirect?` method should have a different name? I'm not sure what it would be. To be fair, there is a lot of conditional logic in these controllers. It's not immediately obvious to me how to further simplify them though...

cbaines: (We discussed this in person, and still didn't come out with a new name, so maybe it doesn't exist)

PR #394, opened 2017-10-16
Create accordion component

andysellick:
- initial version of accordion as a component, will be iterated on
- GA code will be updated during iteration but for now should be exactly what was in place before
- CSS for taxon pages reorganised to prevent CSS collisions with component (taxon pages have blanket rules like '.taxon h2 {}' which cannot be overridden using our BEM approach)
- JS kept as a module but moved to 'components' dir. Possibly unnecessary, but the JS is very specific to this component. Could be moved back, we don't really have guidance on this yet

screen shot 2017-10-16 at 16 48 02

Notable changes from original accordion:

- accordion headers now get wrapped in buttons rather than links, solves some accessibility issues
- accordion header has darker hover colour
- BEM used for CSS (as per component conventions) but JS only classes (.js-something) don't, as BEM too verbose and not needed there
- inner wrapping element removed as unnecessary

https://govuk-collections-pr-394.herokuapp.com/component-guide/accordion
https://govuk-collections-pr-394.herokuapp.com/education/funding-and-finance-for-students

Relates to: https://trello.com/c/o0BCuFub/190-rewrite-the-accordion-to-meet-our-teams-needs


app/assets/stylesheets/components/_accordion.scss
```diff
@@ -0,0 +1,109 @@
+.app-c-accordion {
+  &.js-accordion-with-descriptions {
```
nickcolley: We use 'js-' prefixed classes for our JavaScript code to bind to, this is to avoid mixing presentation with behaviour

Maybe the JavaScript should add a class with a modifier?

`.app-c-accordion--with-descriptions`

nickcolley:
Just was thinking, maybe we could account for this usecase:

1. User visits a page
2. Opens an accordion
3. Shares link with accordion hash in URL
4. User 2 visits shared link
5. JavaScript fails
6. User 2 is not anchored to the correct heading

This could be bad if you're on a page like a manual with many sections, we could include the ID in the fallback headings and it should be ok?

fofr:
@nickcolley In step 6, wouldn't the hash in the URL scroll them to the right section?

app/views/components/docs/accordion.yml
```diff
@@ -0,0 +1,127 @@
...
+  The accordion has several features:
+
+  - Opening a section changes the URL so that if the URL is shared or the user uses the back button on the next page, the page will load at that part of the accordion with it expanded.
+  - If JavaScript is disabled the accordion expands fully. All of the accordion functionality (including the icons and aria attributes) are added using JavaScript.
+accessibility_criteria: |
+  The accordion must:
+
+  - indicate to users that each section can be expanded and collapsed
+  - be usable with a keyboard
+  - be fully expanded and not marked as expandable if JavaScript is disabled
```
fofr: Maybe "JavaScript is unavailable"?

Or:
```
When javascript is unavailable the accordion must:

- be fully expanded
- not be marked as expandable
```

app/views/components/docs/accordion.yml
```diff
@@ -0,0 +1,127 @@
...
+          title: 'Second accordion header',
+          panel: 'Second accordion panel'
+        },
+        {
+          title: 'Third accordion header lets see how long we can make it and what that does',
+          panel: 'Third accordion panel also containing some deliberately long content to show what happens with that'
+        }
+      ]
+  complex:
+    description: Complex markup can be passed for the accordion panels, but the styling for that content should not come from the component itself.
```
fofr: Love this 😄

app/assets/javascripts/application.js
```diff
@@ -3,5 +3,5 @@
 //= require support
 //= require browse-columns
 //= require modules/current-location
-//= require modules/accordion-with-descriptions
 //= require modules/feeds.js
+//= require components/accordion.js
```
fofr: You don't need the `.js` here

app/assets/stylesheets/components/_accordion.scss
```diff
@@ -0,0 +1,109 @@
...
+
+.app-c-accordion__icon {
+  position: absolute;
+  top: 50%;
+  right: 0;
+  height: 16px;
+  width: 16px;
+  margin-top: -8px;
+  background-image: image-url('icons-plus-minus.png'); // PNG fallback for SVG
+  background: image-url("icons-plus-minus.svg"), linear-gradient(transparent, transparent); // SVG sprite
```
fofr: What's the `linear-gradient(transparent, transparent)` bit doing here?

app/assets/stylesheets/components/_accordion.scss
```diff
@@ -0,0 +1,109 @@
...
+    right: 15px;
+  }
+
+  .section-is-open & {
+    background-position: 0 0;
+  }
+}
+
+.app-c-accordion__title {
+  @include core-24;
```
fofr: you could use `bold-24` and remove the line below.

app/assets/stylesheets/components/_accordion.scss
```diff
@@ -0,0 +1,109 @@
...
+
+// removes extra dotted outline from buttons in Firefox
+// on focus (standard yellow outline unaffected)
+@mixin firefox-remove-second-button-outline {
+  &::-moz-focus-inner {
+    border: 0;
+  }
+}
+
+.app-c-accordion__title-button {
```
fofr: You could de-duplicate some of the styles between `title-button` and `controls-button`

These are all shared:
```
@include firefox-remove-second-button-outline;
color: $link-colour;
cursor: pointer;
background: none;
border: 0;
```

app/views/components/docs/accordion.yml
fofr: I think there's some missing criteria regarding:
* open all button behaviour
* individual section expansion and collapse
* use of button rather than link (eg use of spacebar to open/close)

app/assets/stylesheets/components/_accordion.scss
andysellick: You're absolutely right. Can't believe I overlooked that. Will change it to a modifier class, good suggestion.

app/assets/stylesheets/components/_accordion.scss
andysellick: That's an excellent question. My rubbish answer is that it was there already. I wonder if it's some kind of browser specific thing. Perhaps @alecgibson might have an idea?

app/assets/stylesheets/components/_accordion.scss
andysellick: Much better. It's interesting that I didn't know that was an option - is there any kind of documentation on what mixins are available to us and when we can use them? If not, maybe something we should think about, would be invaluable for new starters.

fofr: No real documentation around this.

Here's the file:
https://github.com/alphagov/govuk_frontend_toolkit/blob/master/stylesheets/_typography.scss#L129

app/assets/stylesheets/components/_accordion.scss
```diff
@@ -0,0 +1,106 @@
...
+.app-c-accordion__button {
+  @include firefox-remove-second-button-outline;
+
+  color: $link-colour;
+  cursor: pointer;
+  background: none;
+  border: 0;
+
+  &--title {
+    @include core-24;
```
fofr: Another use for `bold-24`

andysellick:
Tested in IE11, IE10, IE9, IE8, Firefox latest, Firefox 48, Chrome latest, Chrome 53, Opera latest, Google Nexus 6, iphone 6s, Samsung Galaxy S5, ipad pro.

Tested in Voiceover on Chrome and Safari.

nickcolley:
@fofr I might be wrong but the default markup does not contain markup that'd anchor the user to the correct header?

andysellick:
@nickcolley it does, I've just tried it. It's got the ID in the markup.

nickcolley:
@andysellick my mistake, the page isn't long enough for the anchors to work, I scrunched it up and it worked, sorry. 👍

nickcolley:
https://github.com/alphagov/collections/pull/394#pullrequestreview-69621268

Can you share the pages that use this new component?

app/assets/stylesheets/components/_accordion.scss
```diff
@@ -0,0 +1,105 @@
+.app-c-accordion {
+  &.app-c-accordion--active {
```
nickcolley: Does this need to be nested to work?

app/assets/stylesheets/components/_accordion.scss
alecgibson: I think my equally rubbish answer is that it was there even when I got here.

app/assets/javascripts/components/accordion.js
```diff
@@ -25,20 +23,20 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
-      $element.addClass('js-accordion-with-descriptions');
+      $element.addClass('app-c-accordion--active');
 
       // Prevent FOUC, remove class hiding content
       $element.removeClass('js-hidden');
 
-      var $subsections = $element.find('.js-subsection');
-      var $subsectionHeaders = $element.find('.subsection-header');
-      var totalSubsections = $element.find('.subsection-content').length;
+      var $subsections = $element.find('.js-section');
```
alecgibson: Huge fan of renaming subsection -> section, but it would've been nice to have renamed the variables too.

app/views/taxons/accordion.html.erb
```diff
@@ -18,33 +19,31 @@
...
-                      section_index: index,
-                      tagged_content: taxon.tagged_content,
-                  } %>
-                </div>
-              </div>
+        <%
+          accordion_contents = []
+        %>
+        <% presented_taxon.accordion_content.each_with_index do |taxon, index| %>
+          <% contents = capture do %>
```
alecgibson: I personally find this `capture {}` approach a little inside-out. I think the way I would have expected to use this component is along the lines of:

```erb
accordion_content = [
  { title: "Title", description: "Description", arbitrary_field: "Foo", ... },
]

<%= render "components/accordion", sections: accordion_content do |accordion_section, index| %>
  <!-- Render stuff for the "panel" here, where we have access to the accordion section's hash -->
  <!-- This could even be delegated by rendering a different partial/component! -->
  <div>
    <%= accordion_section[:arbitrary_field] %>
  </div>
<% end %>
```

app/assets/javascripts/components/accordion.js
andysellick: That's a good point. Possible laziness on my part.

app/assets/stylesheets/components/_accordion.scss
36degrees: http://pauginer.com/post/36614680636/invisible-gradient-technique

36degrees: Full disclosure: I had to Google it (and it could definitely do with a comment!)

app/views/taxons/accordion.html.erb
nickcolley: I really like this, I didn't know it was possible to do this, only thought it was possible to yield a passed block.

alecgibson: I haven't actually looked into the internals again, but I think I had a branch once upon a time that did something like this (way back when we first looked into turning this into a component). I'm pretty sure it should work, though - you'd just have to have an enumerator in the partial, which calls `yield` for each accordion section.

andysellick: Alec has done the work on this now and I'm going to include it in a future PR - https://github.com/alphagov/collections/compare/accordion-yield?expand=1

PR #397, opened 2017-10-19
Add link for "Learn to drive a car: step by step" page into /browse

Davidslv:
We have added the `learn to drive a car: step by step` link into the 3rd level links of `Driving tests and learning to drive or ride` (2nd level).

This is behind the TaskListSidebar A/B test and only users of the B
variant should see this as we gradually increase the percentage.

This commit leaves a placeholder to create the HTML of the Tasklist page
itself, this is future work.

Paired with @deborahchua 
Trello: https://trello.com/c/xLxn5Ohy/214-add-a-b-test-functionality-into-collections-frontend-app

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -34,5 +34,7 @@ Rails.application.routes.draw do
     to: "services_and_information#index",
     as: :services_and_information
 
+  get '/learn-to-drive-a-car', to: 'tasklist#show'
+
   get '*taxon_base_path', to: 'taxons#show'
 end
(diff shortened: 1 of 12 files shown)


sihugh:
I'll rework this a little so that it follows https://github.com/alphagov/collections/pull/400

config/routes.rb
```diff
@@ -34,5 +34,7 @@
     to: "services_and_information#index",
     as: :services_and_information
 
+  get '/learn-to-drive-a-car', to: 'tasklist#show'
```
Davidslv: This will be a duplicate of https://github.com/alphagov/collections/pull/400/files#diff-21497849d8f00507c9c8dcaf6288b136R9

The reason is that I need it here to write the path.

PR #399, opened 2017-10-25
Remove education a/b test

maxgds:
- Remove redirects logic and config
- Remove references to variants
- Remove tests associated with a/b test


test/integration/taxon_browsing_test.rb
```diff
@@ -5,23 +5,22 @@ class TaxonBrowsingTest < ActionDispatch::IntegrationTest
...
+  #   Capybara.reset_sessions!
 
-    GovukAbTesting.configure do |config|
-      config.acceptance_test_framework = @existing_framework
-    end
-  end
+  #   GovukAbTesting.configure do |config|
+  #     config.acceptance_test_framework = @existing_framework
+  #   end
+  # end
```
vanitabarrett: Block comment left in by mistake?

maxgds: good spot

fofr:
There's some logic on the services and information page still:
https://github.com/alphagov/collections/blob/ab-test/app/controllers/services_and_information_controller.rb#L16-L18

test/integration/taxon_browsing_test.rb
maxgds: fixed

app/controllers/browse_controller.rb
```diff
@@ -6,8 +6,7 @@ def index
     setup_content_item_and_navigation_helpers(page)
 
     render :index, locals: {
-      page: page,
-      ab_variant: education_ab_test.requested_variant
+      page: page
```
fofr: In Ruby it's ok to leave in the trailing comma. This helps keep diffs a little smaller. No need to change this now though.

fofr:
I assume we no longer need the govuk_ab_testing gem too:
https://github.com/alphagov/govuk_ab_testing

Gemfile
```diff
@@ -3,7 +3,6 @@ source 'https://rubygems.org'
 ruby File.read(".ruby-version").strip
 
 gem 'gds-api-adapters', '~> 47.9'
-gem 'govuk_ab_testing', '~> 2.4'
```
fofr: You need to commit the updated Gemfile.lock file as well. This is why the Heroku deployment is failing.

fofr:
Couple of files left to remove:
https://github.com/alphagov/collections/blob/ab-test/app/helpers/navigation_ab_helper.rb
https://github.com/alphagov/collections/blob/ab-test/test/unit/navigation_ab_helper_test.rb

PR #402, opened 2017-11-01
[WIP] Add left to right option for accordion component

andysellick:
Move the plus/minus icons on accordions to the left of the element, rather than the right. This is an accessibility fix for those using Zoomtext or similar screen magnification tools - it makes it so that the accordion functionality is much more obvious.

- rename all subsection variables to section
- add left to right option for accordion component

screen shot 2017-11-07 at 09 08 38

PR: https://govuk-collections-pr-402.herokuapp.com/component-guide/accordion

Ideally would use https://github.com/alphagov/collections/compare/accordion-yield?expand=1 as a technique for calling the component, but the current documentation model doesn't support it.

### DIFF
--- a/app/assets/javascripts/modules/current-location.js
+++ b/app/assets/javascripts/modules/current-location.js
@@ -1,3 +1,4 @@
+// used by accordion component
 (function(root) {
   "use strict";
   root.GOVUK = root.GOVUK || {};
--- a/app/assets/javascripts/support.js
+++ b/app/assets/javascripts/support.js
@@ -1,3 +1,4 @@
+// used by accordion component
 if(typeof window.GOVUK === 'undefined'){ window.GOVUK = {}; }
 if(typeof window.GOVUK.support === 'undefined'){ window.GOVUK.support = {}; }
 
--- a/app/views/components/_accordion.html.erb
+++ b/app/views/components/_accordion.html.erb
@@ -1,8 +1,9 @@
 <%
   sections ||= false
+  controls_on_left ||= false
 %>
 <% if sections %>
-  <div data-module="accordion" class="app-c-accordion js-hidden">
+  <div data-module="accordion" class="app-c-accordion js-hidden <%= "app-c-accordion--controls-on-left" if controls_on_left %>">
     <% sections.each_with_index do |section, index| %>
       <%
         id = section[:id] || section[:title].downcase.tr(" ","-")
--- a/app/assets/javascripts/components/accordion.js
+++ b/app/assets/javascripts/components/accordion.js
@@ -28,28 +28,28 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
       // Prevent FOUC, remove class hiding content
       $element.removeClass('js-hidden');
 
-      var $subsections = $element.find('.js-section');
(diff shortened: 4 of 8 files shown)


nickcolley:
With a PR like this it's good to break the changes up into commits, this'll record what you're doing better and make it easier to review.

nickcolley:
Can you explain why this change is an optional behaviour? This seems to be something that should be applied to all variations?

Could we add this to the accessibility acceptance critera?

app/assets/javascripts/modules/current-location.js
```diff
@@ -1,3 +1,4 @@
+// used by accordion component
```
alecgibson: Can we rely on Static for this?

app/assets/javascripts/support.js
```diff
@@ -1,3 +1,4 @@
+// used by accordion component
```
alecgibson: Can we rely on Static for this?

app/views/components/_accordion.html.erb
```diff
@@ -1,8 +1,9 @@
 <%
   sections ||= false
+  controls_on_left ||= false
 %>
 <% if sections %>
-  <div data-module="accordion" class="app-c-accordion js-hidden">
+  <div data-module="accordion" class="app-c-accordion js-hidden <% if controls_on_left %>app-c-accordion--controls-on-left<% end %>">
```
alecgibson: (Again, I'd personally opt for `<%= "app-c-accordion--controls-on-left" if controls_on_left %>`

nickcolley:
@alecgibson we'd need to enable our documentation yml files to work with yields over at  https://github.com/alphagov/govuk_publishing_components to

alecgibson:
@nickcolley yes I know - I'm trying to have a very quick look this morning (a spikelet?), but not sure how far I'll get.

app/assets/javascripts/modules/current-location.js
andysellick: We can in the browser but not in our spec tests 👎

alecgibson: Yup. That's really sad. Hopefully will get fixed if we can move Static into a Rails engine.

andysellick:
@nickcolley have moved the 'Open all' to the left as well. As for making this the default behaviour, I wasn't sure I should go ahead with such a big change, seemed better to make this optional, try it out, and maybe convert accordions later.

nickcolley:
@andysellick 

I'd be worried no one would swap them and forget, is there any reason this cannot be applied to all versions?

I'm not sure why we'd want to support having a variation that is not accessible.

andysellick:
Going to put this on hold for the moment. Talked to Mia and the better option is to get rid of the plus/minus altogether and replace with the new Show/Hide option being prototyped in the task list. Will return to this once the design/code for that is stable.

sihugh:
I've marked this a WIP to appease the angry seal for a bit until we can get it moving again.

tijmenb:
I'm going to close this in the interest of having fewer PRs open. Please reopen if needed.

andysellick:
Okay, maybe don't delete the branch though.

tijmenb:
You can restore it :)

PR #405, opened 2017-11-03
Add learn to drive link to https://www.gov.uk/browse/driving/driving-licences

sihugh:
We've already added it to https://www.gov.uk/browse/driving/learning-to-drive but need to add it to 
the driving licences page too.


app/models/task_list_ab_test_request.rb
```diff
@@ -14,9 +14,11 @@ def initialize(request)
 
   def show_tasklist_link?(list_title, params)
```
Davidslv: params is not necessary anymore.

Davidslv:
Merge away when you ready

PR #410, opened 2017-11-07
Add Dockerfile

RyanMacG:
As part of extending out the end-to-end tests we require a Dockerfile
for running this application as part of them.


kevindew:
Looks like there's a PR for a dockerfile here: https://github.com/alphagov/collections/pull/409 which only popped up in the last day. From a quick scan I see this has capybara-webkit which I don't think we need, the app name slightly different (seems to be just "collections" rather than "collections-frontend") and this has: `ENV GOVUK_ASSET_ROOT http://assets-origin.dev.gov.uk` is that needed? if so might be best to add that as a commit to the other PR and we can get that merged?

PR #413, opened 2017-11-20
Update Jenkinsfile to avoid buildProject running twice

adrianclay:


adrianclay:
Yep - now we have a better understanding of how the build works we realised our mistake.

PR #417, opened 2017-12-07
[REVIEW ONLY] Add end civil partnership tasklist page

deborahchua:
The Modelling Services team are implementing a new tasklist "End a civil partnership: step by step".

It uses the new tasklist component that is being updated in `static` (alphagov/static#1201).

We will merge this once the component is ready.

Trello card: [Add end-a-civil-partnership task list page](https://trello.com/c/NewvyNH7/)

end-civil-partnership-tasklist

### DIFF
--- a/app/services/tasklist_content.rb
+++ b/app/services/tasklist_content.rb
@@ -1,9 +1,17 @@
 class TasklistContent
   def self.learn_to_drive_config
-    @learn_to_drive_config ||= JSON.parse(
+    @learn_to_drive_config ||= find_file("learn-to-drive-a-car")
+  end
+
+  def self.end_a_civil_partnership_config
+    @end_a_civil_partnership_config ||= find_file("end-a-civil-partnership")
+  end
+
+  def self.find_file(filename)
+    JSON.parse(
       File.read(
-        Rails.root.join("config", "tasklists", "learn-to-drive-a-car.json")
-        )
-      ).deep_symbolize_keys
+        Rails.root.join("config", "tasklists", "#{filename}.json")
+      )
+    ).deep_symbolize_keys
   end
 end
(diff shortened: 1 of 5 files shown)


app/services/tasklist_content.rb
```diff
@@ -6,4 +6,16 @@ def self.learn_to_drive_config
         )
       ).deep_symbolize_keys
   end
+
```
sihugh: could you refactor the learn-to-drive method to use the `find_file` method?  Add it as a separate commit maybe so that it's obvious what's happened?

app/views/tasklist/show.html.erb
```diff
@@ -1,4 +1,4 @@
-<% content_for :title, "Learn to drive a car: step by step" %>
+<% content_for :title, tasklist[:title] %>
 
 <% content_for :meta_tags do %>
   <meta name="description" content="Learn to drive a car in the UK - get a provisional licence, take driving lessons, prepare for your theory test, book your practical test.">
```
sihugh: We need to fix this too.  We should probably get it from the description field of the content item that we publish in [collections-publisher](https://github.com/alphagov/collections-publisher/blob/release/lib/tasks/publishing_api.rake#L52) as that's where Rummager will get it from.  There's a `ContentItem` model that might be useful to help retrieve that.

(We also need to get some better words to put in those descriptions from Jon, but that's another story)

test/integration/tasklist_page_test.rb
```diff
@@ -1,47 +1,57 @@
 require 'integration_test_helper'
 
 class TasklistPageTest < ActionDispatch::IntegrationTest
-  it "renders the learn to drive page" do
-    visit "/learn-to-drive-a-car"
+  it "renders the end a civil partnership tasklist page" do
```
sihugh: We're not getting rid of the learn to drive page, so we probably want to keep some tests around for it too.

PR #441, opened 2017-12-18
Add print styles to collections for task list

andysellick:
- add mechanism for displaying print styles
- add print css, includes only print styles for task list

Work motivated by the addition of the new task list component design, noticed that collections has no print styles and the task list needs some. Not merging until we've sorted out the components gem.

Wait for [this PR to be merged](https://github.com/alphagov/collections/pull/449) before merging.

Trello card: https://trello.com/c/ELorPDMS/393-do-all-the-task-list-gem-component-things

### DIFF
--- a/app/assets/stylesheets/print.scss
+++ b/app/assets/stylesheets/print.scss
@@ -1,0 +1,3 @@
+@import "typography";
+
+@import "components/task-list-print";
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -13,6 +13,7 @@
 
   <%= render partial: 'govuk_component/analytics_meta_tags',
     locals: { content_item: @content_item } %>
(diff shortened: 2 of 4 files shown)


nickcolley:
Is this only imported into the application when they opt into it?

andysellick:
It assumes that the task list is being provided by the components gem, so the styles are included by default, but collections now has three (and counting) task list pages included.

nickcolley:
So the application can't opt into these styles? Every application we put this gem into will get CSS it might not use?

andysellick:
Ah, I see what you mean now @nickcolley. No, this is a change to collections, to add a print stylesheet to pull in the print styles for the task list. The change is to the application itself, so in your terms collections is opting in to having these print styles. To answer the general question - no application gets the govuk_publishing_components gem styles automatically if the gem is installed - the application always has to be modified to pull in the stylesheets from the gem.

PR #485, opened 2018-01-29
Refactor tasklist_controller

sihugh:
Change the show action to cater for all hard coded step by step navigation
pages rather than having an action per page.

This gives us a head start on rendering pages from the content store and is
the first part of https://trello.com/c/3pQPJDmj/440-allow-collections-to-render-task-lists-from-the-publishing-api

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -7,9 +7,8 @@ Rails.application.routes.draw do
   mount GovukPublishingComponents::Engine, at: "/component-guide" if defined?(GovukPublishingComponents)
 
   get "/learn-to-drive-a-car", to: 'tasklist#show'
-  get "/get-a-divorce", to: 'tasklist#show_get_a_divorce'
-
-  get "/end-a-civil-partnership", to: 'tasklist#show_end_a_civil_partnership'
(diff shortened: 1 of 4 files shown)


config/routes.rb
```diff
@@ -7,9 +7,8 @@
   mount GovukPublishingComponents::Engine, at: "/component-guide" if defined?(GovukPublishingComponents)
 
   get "/learn-to-drive-a-car", to: 'tasklist#show'
-  get "/get-a-divorce", to: 'tasklist#show_get_a_divorce'
-
-  get "/end-a-civil-partnership", to: 'tasklist#show_end_a_civil_partnership'
+  get "/get-a-divorce", to: 'tasklist#show'
```
Davidslv: I think we can even do `get /:base_path, , to: 'tasklist#show, constraints: TasklistConstraints.new`

I'm just trying to think of a way of not having all those base paths written in this app... but then it will have to be inside the Constraint... so not winning much with this approach either... any other ideas?

sihugh: We also need to have published things using the updated schema before we can do funky routing in this way

PR #503, opened 2018-02-16
Upgrade to latest govuk_app_config

tijmenb:
Removes config and gems that are now taken care of by `govuk_app_config`.

https://github.com/alphagov/govuk_app_config/blob/master/CHANGELOG.md


cbaines:
I think you should remove this file as well? [1]

1: https://github.com/alphagov/collections/blob/master/config/initializers/logstasher.rb

PR #505, opened 2018-02-19
Only show email alert option if taxon is live

vanitabarrett:
Trello: https://trello.com/c/L6CEUHtK/74-only-show-email-notification-link-on-live-taxon-pages

We are planning on publishing the taxonomy so that all the taxonomy is live, but in different phases, e.g: 'alpha', 'beta' and 'live'. This will make it easier for teams who need to work and prototype with the draft taxonomy, e.g: Topic Pages team. However, we need to make sure that pages for alpha and beta don't allow users to subscribe to email alerts.

### DIFF
--- a/app/helpers/email_helper.rb
+++ b/app/helpers/email_helper.rb
@@ -11,6 +11,10 @@ module EmailHelper
     whitehall_base_email_sign_up_url + whitehall_atom_url
   end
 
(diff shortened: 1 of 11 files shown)


vanitabarrett:
@leenagupte Oops, good spot! I'll fix that now

PR #507, opened 2018-02-20
Hide non live taxon from google

fmlharrison:
Added this check so that if the taxon that is being
rendered is not live, then we hide it from being indexed
by search engines.

### DIFF
--- a/test/integration/taxon_browsing_test.rb
+++ b/test/integration/taxon_browsing_test.rb
@@ -83,6 +83,13 @@ class TaxonBrowsingTest < ActionDispatch::IntegrationTest
     then_i_cannot_see_the_blue_box_section
   end
 
+  it 'hides page from search engine when taxon is not live' do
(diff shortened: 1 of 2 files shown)


test/integration/taxon_browsing_test.rb
```diff
@@ -233,6 +247,16 @@ def then_i_can_see_the_meta_description
     )
   end
 
+  def then_page_has_meta_robots
+    content = page.find('meta[name="robots"]', visible: false)['content']
+
+    assert_equal(
+      "noindex, nofollow",
+      content,
+      "The content of the meta description should be the taxon description"
```
leenagupte: Minor thing. The content of the optional failure message doesn't match what's being tested. You might want to change to something like:

'The content of the robots meta tag should be "noindex, nofollow"'

fmlharrison: Oh nice catch!! 
This is why you don't copy and paste, kids!!

PR #508, opened 2018-02-20
Add integration test for nonlive email alert logic

vanitabarrett:
Trello: https://trello.com/c/L6CEUHtK/74-only-show-email-notification-link-on-live-taxon-pages

This adds an integration test for the logic introduced in PR #505.

We need to make sure that an email signup link is not displayed on taxon pages that do not have a phase of 'live'.

**See the [original PR](https://github.com/alphagov/collections/pull/505) for more details on why this is being added**

### DIFF
--- a/test/integration/taxon_browsing_test.rb
+++ b/test/integration/taxon_browsing_test.rb
@@ -91,6 +91,17 @@ class TaxonBrowsingTest < ActionDispatch::IntegrationTest
     and_i_cannot_see_an_email_signup_link
   end
 
+  it 'sets up a non-live taxon with no grandchildren' do
+    given_there_is_a_taxon_without_grandchildren
+    and_the_taxon_without_grandchildren_is_not_live
+    and_there_are_popular_items_for_the_taxon
+    when_i_visit_the_taxon_page
+    then_i_can_see_there_is_a_page_title
+    and_i_can_see_the_general_information_section_in_the_accordion
+    and_i_can_see_links_to_the_child_taxons_in_an_accordion
+    and_i_cannot_see_an_email_signup_link
+  end
+
 private
 
   def then_i_can_see_the_blue_box_with_its_details
@@ -177,7 +188,7 @@ private
     assert_not_nil @parent
 
     @child_taxons = student_finance['links']['child_taxons']
-    assert @child_taxons.length > 0
+    assert @child_taxons.length.positive?
 
     content_store_has_item(@base_path, student_finance)
     content_store_has_item(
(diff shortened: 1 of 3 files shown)


test/integration/taxon_browsing_test.rb
```diff
@@ -90,6 +90,13 @@ class TaxonBrowsingTest < ActionDispatch::IntegrationTest
     then_page_has_meta_robots
   end
 
+  it 'does not show an email signup link on a non-live page' do
+    given_there_is_a_taxon_with_grandchildren
+    and_the_taxon_is_not_live
+    when_i_visit_the_taxon_page
+    then_i_cannot_see_an_email_signup_link
```
fmlharrison: Would it be better to add this assertion to the test above it (as well as changing the test name) as the set up for both tests are the same?

test/integration/taxon_browsing_test.rb
```diff
@@ -396,6 +403,13 @@ def and_i_can_see_an_email_signup_link
     )
   end
 
+  def then_i_cannot_see_an_email_signup_link
+    assert page.has_no_link?(
+      'Get email alerts for this topic',
+      href: "/email-signup/?topic=#{current_path}"
+    )
+  end
+
```
fmlharrison: The email sign up link appears for in each accordion section, see: https://www.gov.uk/education/funding-and-finance-for-students#/education/financial-help-for-students-who-are-parents-or-carers

Should we also check that email sign up links are not rendered for each section of the accordion when the taxon is not live?. The opposite of what the method on line 413 is testing for.

vanitabarrett: Good spot 👍

test/integration/taxon_browsing_test.rb
```diff
@@ -91,6 +91,17 @@ class TaxonBrowsingTest < ActionDispatch::IntegrationTest
...
 
+  it 'sets up a non-live taxon with no grandchildren' do
+    given_there_is_a_taxon_without_grandchildren
+    and_the_taxon_without_grandchildren_is_not_live
+    and_there_are_popular_items_for_the_taxon
+    when_i_visit_the_taxon_page
+    then_i_can_see_there_is_a_page_title
+    and_i_can_see_the_general_information_section_in_the_accordion
+    and_i_can_see_links_to_the_child_taxons_in_an_accordion
+    and_i_cannot_see_an_email_signup_link
```
fmlharrison: I think we also need to test that the email subscription link does appear in each of the sections in the accordion (ie child taxons).... 

You could do a test like `and_all_sections_do_not_have_an_email_signup_link`, which is the opposite to this test: https://github.com/alphagov/collections/blob/e36a7752674116c853b83c45d1798b9619019396/test/integration/taxon_browsing_test.rb#L424

vanitabarrett: `and_i_cannot_see_an_email_signup_link` checks the whole page, so the test should already fail if sections in the accordion have email signup links.

fmlharrison: Oh of course!! Nice!

PR #514, opened 2018-02-23
Revert "Only show email alert option if taxon is live"

fmlharrison:
Reverts alphagov/collections#505

There was a problem that arose in staging. The email alerts are not being shown for live child taxons in the accordion layout as the child taxons on the content item don't have the 'phase' field.


fmlharrison:
I made a mistake when revert a few PRs so I had to close them and start again.

PR #515, opened 2018-02-23
Revert "Add integration test for nonlive email alert logic"

fmlharrison:
Reverts alphagov/collections#508

There was a problem that arose in staging. The email alerts are not being shown for live child taxons in the accordion layout as the child taxons on the content item don't have the 'phase' field.


fmlharrison:
I made a mistake while reverting a few connected PRs so I am closing them and trying again.

PR #516, opened 2018-02-23
Revert #505, #507, #508

kevindew:
Paired with @fmlharrison on this.

This reverts https://github.com/alphagov/collections/pull/505, https://github.com/alphagov/collections/pull/507, and https://github.com/alphagov/collections/pull/508 which had some problems when it came to deploying. The automatic github revert didn't work so we did it the old school `git revert <hash>` method.


vanitabarrett:
@tijmenb Child taxons in a content item don't have a 'phase', so the email alerts were being hidden in the accordion on pages like this: https://www.gov.uk/education/funding-and-finance-for-students

PR #522, opened 2018-02-27
Only show email alert for taxons and child taxons when phase is live

vanitabarrett:
This commit reverts #516, which in turn reverted #505, #507 and #508.

In preparation for publishing the taxonomy in phases of alpha, beta or live, we want to make sure that we only show email alert buttons for taxons that are live.

We had to revert the original commit as child taxons in a content item did not have a 'phase'. This has now been added so we can add this logic back in.

**Note: [last commit](https://github.com/alphagov/collections/pull/522/commits/493220f1f0f5bf40421dce1f0f8d9b4359019db5) introduces use of `fetch` to better handle if taxons have no phase. We have chosen to raise an error in this case, but this is up for debate and worth reviewing more closely to make sure it's been implemented in the right way.**

### DIFF
--- a/app/models/content_item.rb
+++ b/app/models/content_item.rb
@@ -35,4 +35,8 @@ class ContentItem
   def merge(to_merge)
     ContentItem.new(content_item_data.merge(to_merge))
   end
+
+  def phase
+    @content_item_data["phase"]
+  end
 end
--- a/app/models/content_item.rb
+++ b/app/models/content_item.rb
@@ -37,6 +37,8 @@ class ContentItem
   end
 
   def phase
-    @content_item_data["phase"]
+    @content_item_data.fetch("phase")
+  rescue KeyError => e
+    raise "Error searching content item: #{e}"
   end
 end
--- a/app/views/taxons/_common.html.erb
+++ b/app/views/taxons/_common.html.erb
@@ -1,5 +1,9 @@
 <% content_for :title, presented_taxon.title %>
 <% content_for :meta_tags do %>
+  <% if !presented_taxon.live_taxon? %>
+    <% # prevent search engines from indexing this page %>
+    <meta name="robots" content="noindex, nofollow">
+  <% end %>
   <meta name="description" content="<%= presented_taxon.description %>">
   <meta name="govuk:navigation-page-type" content="<%= presented_taxon.rendering_type %>" />
 <% end %>
--- a/test/integration/taxon_browsing_test.rb
+++ b/test/integration/taxon_browsing_test.rb
@@ -83,6 +83,25 @@ class TaxonBrowsingTest < ActionDispatch::IntegrationTest
     then_i_cannot_see_the_blue_box_section
   end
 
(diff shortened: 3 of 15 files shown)


app/models/content_item.rb
```diff
@@ -35,4 +35,8 @@ def to_hash
   def merge(to_merge)
     ContentItem.new(content_item_data.merge(to_merge))
   end
+
+  def phase
+    @content_item_data.fetch("phase", "alpha")
```
tijmenb: I wonder if we should crash here if `phase` doesn't exist. It would've caused us to detect the previous bug earlier.

vanitabarrett: @tijmenb Would that mean pages would error if the taxon didn't have the phase? Or doing something like this:

```
  def phase
    @content_item_data.fetch("phase")
  rescue
    nil
  end
```

tijmenb: Yes, I think that might be the preferable option. All the taxons should have the phase now, so if they haven't something is wrong and an throwing an error will alert us to it.

vanitabarrett:
@tijmenb Tweaked to not return a default value anymore - not sure how we normally handle/raise errors though? Do we log them somewhere?

test/integration/taxon_browsing_test.rb
```diff
@@ -141,7 +141,7 @@ def given_there_is_a_taxon_with_grandchildren
 
     @child_taxons =
       funding_and_finance_for_students_taxon['links']['child_taxons']
-    assert @child_taxons.length > 0
+    assert @child_taxons.length.positive?
```
leenagupte: Doesn't this already happen in the previous commit? Or were there two linting errors that happened to end up on line 144?

app/views/taxons/_common.html.erb
```diff
@@ -1,5 +1,9 @@
 <% content_for :title, presented_taxon.title %>
 <% content_for :meta_tags do %>
+  <% if !presented_taxon.live_taxon? %>
```
leenagupte: This part of the change is from the PR to hide draft taxons from search engines. I think to be clear that we're reverting this as well, this change (and it's accompanying test) should either be separated out into another commit or the commit message for this commit should be updated.

app/models/content_item.rb
```diff
@@ -37,6 +37,8 @@ def merge(to_merge)
   end
 
   def phase
-    @content_item_data["phase"]
+    @content_item_data.fetch("phase")
+  rescue KeyError => e
+    puts "Error searching content item: #{e}"
```
leenagupte: Should we raise an error here instead of writing a message? How will that affect the way it's logged?

vanitabarrett:
@tijmenb @leenagupte Have updated this PR following your comments - tidied up git history and we are now raising an error if `phase` cannot be found in the content item.

fmlharrison:
@vanitabarrett If we go with throwing an error when using `fetch` then remember to change the note at the bottom of the PR message to reflect this.

PR #533, opened 2018-03-08
Separate Worldwide pages from other taxons

leenagupte:
Preparatory work for https://trello.com/c/GJwpHOew

## Motivation

The taxon pages for regular taxons are about to go through a major overhaul. There are no plans at present to change how the `/world/*` pages are rendered. To make the new taxon pages easier to implement, and to ensure that the world pages aren't affected by the changes, the code to render world pages has been separated out.


app/controllers/world_wide_taxons_controller.rb
```diff
@@ -0,0 +1,27 @@
+class WorldWideTaxonsController < ApplicationController
+  def show
+    setup_content_item_and_navigation_helpers(taxon)
+
+    taxon_template =
```
steventux: The return values of `presented_taxon.rendering_type` are identical to the [case statement return values](https://github.com/alphagov/collections/pull/533/files#diff-e4895bed2fbf26d3e0d977f41ef5d9cfR24) aren't they?
Seems like this case statement could go away if so.

app/models/world_wide_taxon.rb
```diff
@@ -0,0 +1,86 @@
...
+    end
+  end
+
+  def merge(to_merge)
+    WorldWideTaxon.new(content_item.merge(to_merge))
+  end
+
+  def can_subscribe?
+    # Is @can_subscribe ever used on world wide pages?
+    return @can_subscribe if defined?(@can_subscribe)
```
steventux: [`can_subscribe` is read+writeable](https://github.com/alphagov/collections/pull/533/files#diff-751dc6cbbbf38590d833b91390103a9eR3) so no need to use the `@` instance var

app/presenters/world_wide_taxon_presenter.rb
```diff
@@ -0,0 +1,109 @@
...
+    }
+  end
+
+  def options_for_child_taxon(index:)
+    {
+      module: 'track-click',
+      track_category: 'navGridContentClicked',
+      track_action: (index + 1).to_s,
+      track_label: child_taxons[index].base_path,
+      track_options: { dimension26: tagged_content.any? ? '2' : '1',
```
steventux: I'd be tempted to _golf down_ these tracking dimension hashes, there's a lot of similarity in the following 2 methods.

tijmenb:
@rubenarakelyan there should not be anything to keep in sync, the world wide pages should stay as they are and the topic pages are being [redesigned completely](https://trello.com/c/DbDHkJ0S/296-finalise-desktop-content-page-designs-for-testing). I expect there to be very little overlap between the types of pages soon.

rubenarakelyan:
I was thinking more about things like models and controllers, which will mostly be doing the same thing barring a few differences - changes in one will most likely need changes in the other, unless these things are also diverging completely.

app/controllers/world_wide_taxons_controller.rb
leenagupte: @steventux This controller was copied wholesale from the existing taxons_controller. The only changes made were to the class names for `Taxon` and `TaxonPresenter`.
But yes, I think you're right about the case statement!

tijmenb: The `:grid` case statement was also removed here.

app/presenters/world_wide_taxon_presenter.rb
leenagupte: @steventux It's only the `module` that's consistent across all of the methods. They are just different enough that I'd keep them separate for clarity.

leenagupte:
@rubenarakelyan The new taxon pages are going to be quite different. The main change is that we are going to get tagged content by [content_purpose_supergroup](https://github.com/alphagov/govuk_document_types/blob/master/data/supertypes.yml#L485) rather than `document_type` and order the results from rummager based on the `content_purpose_supergroup`, e.g. guidance_and_regulation by most popular, news_and_communications by most recent,  so the Controllers, Presenters and Models could end up looking quite different.

We won't really know until the new pages are implemented. Hopefully then we'll have the chance go back and DRY-up the code where needed.

app/models/world_wide_taxon.rb
```diff
@@ -58,7 +58,7 @@ def merge(to_merge)
 
   def can_subscribe?
     # Is @can_subscribe ever used on world wide pages?
-    return @can_subscribe if defined?(@can_subscribe)
+    return can_subscribe if defined?(can_subscribe)
```
steventux: Not sure about this as it changes the result of `if defined?(...)` because the method `can_subscribe` is always defined, whereas the instance var may or may not be defined. Very hard to understand what this even does though. I've searched the usage in the master branch I don't see that instance var set anywhere...

app/presenters/world_wide_taxon_presenter.rb
leenagupte: Ok, I'll kill the commit.

PR #555, opened 2018-03-15
Implement email-link component for topic pages

maxgds:
Uses new component (bell icon) on topic pages, eg

Example pages:
Grid: https://govuk-collections-pr-555.herokuapp.com/education
Accordion: https://govuk-collections-pr-555.herokuapp.com/education/funding-and-finance-for-students
Leaf: https://govuk-collections-pr-555.herokuapp.com/education/education-of-disadvantaged-children

screen shot 2018-03-16 at 10 45 14

The old style of email link (envelope icon) should continue to appear on other pages, eg:
https://govuk-collections-pr-555.herokuapp.com/world/afghanistan

### DIFF
--- a/app/assets/stylesheets/views/_taxons.scss
+++ b/app/assets/stylesheets/views/_taxons.scss
@@ -2,6 +2,14 @@
   @include bold-27;
 }
 
+.taxon-page__email-link-wrapper {
+  background-color: $grey-4;
+  margin-top: -$gutter;
+  margin-bottom: $gutter;
+  border-bottom: 1px solid $border-colour;
+  padding: $gutter;
+}
+
 .taxon-page__section-description {
   @include core-19;
   margin-top: $gutter-one-third;
--- a/app/views/taxons/_common.html.erb
+++ b/app/views/taxons/_common.html.erb
@@ -11,6 +11,14 @@
 
 <%= render partial: 'page_header', locals: { presented_taxon: presented_taxon } %>
 
-<div class="full-page-width-wrapper">
-  <%= render partial: 'email_alerts', locals: { presented_taxon: presented_taxon } %>
-</div>
+<% if taxon_is_live?(presented_taxon) %>
+  <div class="taxon-page__email-link-wrapper">
+    <div class="full-page-width-wrapper">
+        <%= render "components/email-link", {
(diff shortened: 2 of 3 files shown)


leenagupte:
@maxgds It's very pale, but the email alert section should have a light grey background (`$grey-4`) and a thin border on the bottom.

app/views/email_signups/new.html.erb
```diff
@@ -1,5 +1,5 @@
 <% content_for :title, "Email alert subscription for #{subtopic.combined_title}" %>
-<%= render 'shared/tag_meta', tag: subtopic, description: "Subscribe to get an email each time content is published or updated in the '#{subtopic.combined_title}' topic." %>
+==<%= render 'shared/tag_meta', tag: subtopic, description: "Subscribe to get an email each time content is published or updated in the '#{subtopic.combined_title}' topic." %>==
```
leenagupte: Should the `==` be here?

maxgds: ...no :( will kill them

maxgds:
@vanitabarrett I've styled and pushed the email but I have questions/want advice:
- not sure what class to give the link wrapper
- not sure where to put the css
based on these 2 questions I was tempted to put the wrapper and styles into the component but that would probably make it less versatile, so not sure
- header has a gutter sized bottom margin, I'm adjusting to sit snugly against it by putting a negative margin on the email wrapper, but should I remove the header margin instead? (not sure if it has that flag - could add it if not, but whether to pass the flag would be based on whether the taxon is live as that's the conditional that makes the email link appear/disappear and it seems a bit convoluted)

vanitabarrett:
@maxgds Looks good! 

- I would say we can follow the pattern we've started on the taxon page, so something like: `taxon-page__email-link-wrapper` and then we can put this CSS in `taxons.scss`. 
- I agree, I think the email component should be kept separate from this full-width styling so it will still be possible to use the component in different contexts if needed.
- I think for now what you've done is fine, we can always go and tidy this up by adding a flag to the header component later on. At least for now we're not overriding any component styles.

leenagupte:
@maxgds It looks great! 🎉

maxgds:
@vanitabarrett @leenagupte Is this good to merge now?

PR #558, opened 2018-03-16
Refactor sections

vanitabarrett:
Refactor the section methods in taxon_presenter to no longer be specific to each supergroup.

We still require the content retrieval methods to be specific, e.g: guidance_and_regulation_content

### DIFF
--- a/app/models/taxon.rb
+++ b/app/models/taxon.rb
@@ -20,12 +20,13 @@ class Taxon
     @tagged_content ||= fetch_tagged_content
   end
 
-  def services_content
-    @services_content ||= fetch_most_popular_content('services')
-  end
-
-  def guidance_and_regulation_content
-    @guidance_and_regulation_content ||= fetch_most_popular_content('guidance_and_regulation')
+  def section_content(supergroup)
+    case supergroup
+    when "guidance_and_regulation"
+      guidance_and_regulation_content
+    when "services"
+      services_content
+    end
   end
 
   def self.find(base_path)
@@ -72,5 +73,13 @@ private
 
   def fetch_most_popular_content(content_purpose_supergroup = 'guidance_and_regulation')
     MostPopularContent.fetch(content_id: content_id, filter_content_purpose_supergroup: content_purpose_supergroup)
+  end
+
+  def services_content
+    @services_content ||= fetch_most_popular_content('services')
+  end
+
+  def guidance_and_regulation_content
+    @guidance_and_regulation_content ||= fetch_most_popular_content('guidance_and_regulation')
   end
 end
--- a/app/presenters/taxon_presenter.rb
+++ b/app/presenters/taxon_presenter.rb
@@ -8,8 +8,7 @@ class TaxonPresenter
     :tagged_content,
     :child_taxons,
     :live_taxon?,
-    :guidance_and_regulation_content,
-    :services_content,
+    :section_content,
     to: :taxon
   )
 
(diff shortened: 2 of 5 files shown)


app/presenters/taxon_presenter.rb
```diff
@@ -17,27 +17,43 @@ def initialize(taxon)
...
+          document_type: document.content_store_document_type.humanize
         },
       }
     end
   end
 
-  def guidance_and_regulation_section_title
-    'guidance_and_regulation'.humanize
+  def show_section?(supergroup)
+    section_content(supergroup).count.positive?
```
emmabeynon: I think this could be achieved with `section_content(supergroup).any?`

app/presenters/taxon_presenter.rb
```diff
@@ -17,27 +17,43 @@ def initialize(taxon)
...
   end
 
-  def show_guidance_section?
-    guidance_and_regulation_content.count.positive?
+  def section_content(supergroup)
```
emmabeynon: It probably makes sense to make this a private method as it's related to the internal workings of the code and not something you would want to call directly.

test/presenters/taxon_presenter_test.rb
```diff
@@ -75,13 +75,27 @@
...
+  describe 'supergroup_sections' do
+    it 'returns a list of supergroup details' do
+      section_keys = %i(show_section title documents)
+
+      taxon = mock
+      taxon.stubs(:guidance_and_regulation_content).returns([])
+      taxon_presenter = TaxonPresenter.new(taxon)
+
+      taxon_presenter.sections.each do |section|
+        assert_equal(section.keys.sort, section_keys.sort)
```
emmabeynon: I think it would be helpful to make the expectation here more explicit, in case for whatever reason `section.keys.sort` gets messed up and is returning the wrong thing.

app/presenters/taxon_presenter.rb
```diff
@@ -17,27 +17,43 @@ def initialize(taxon)
...
+    method(supergroup + "_content").call
```
emmabeynon: @steventux Can you think of a nicer way to do this?  It looks a bit odd to me, but I'm not sure of a different way to do it?

steventux: I think there's a fair bit of indirection in this approach. The `_content` methods are delegated to the taxon instance and the dynamic method call based on convention is succinct but adds to the magic. It's not immediately obvious what's being called.
If it were me I'd call the method on the taxon instance.

test/presenters/taxon_presenter_test.rb
vanitabarrett: @emmabeynon What do you mean by making the expectation more explicit?

emmabeynon: As a person who's not familiar with the code, it would be helpful to see exactly what `section.keys.sort` returns.  Just makes the test a bit easier to read and understand.

vanitabarrett: @emmabeynon Thanks, that makes sense! 👍

vanitabarrett:
@emmabeynon @steventux Thanks for your comments! I've made some changes - would appreciate a re-review when you get a chance.

app/models/taxon.rb
```diff
@@ -20,12 +20,13 @@ def tagged_content
     @tagged_content ||= fetch_tagged_content
   end
 
-  def services_content
-    @services_content ||= fetch_most_popular_content('services')
-  end
-
-  def guidance_and_regulation_content
-    @guidance_and_regulation_content ||= fetch_most_popular_content('guidance_and_regulation')
+  def section_content(supergroup)
```
steventux: Given that this section-for-supergroup logic has been moved to the `Taxon` class you could remove all specifics of the supergroup names by passing it directly to the `fetch_most_popular_content` method...


```ruby
def section_content(supergroup)
  section_contents[supergroup]
end

def section_contents
  @section_contents ||= {}
end

def fetch_most_popular_content(content_purpose_supergroup)
  @section_contents[supergroup] ||= MostPopularContent.fetch(content_id: content_id, filter_content_purpose_supergroup: content_purpose_supergroup)
end
```

vanitabarrett: @steventux I think we wanted to keep them separate for now, as other sections (e.g: 'news and communication') will not use the `fetch_most_popular_content` method as they need to be ordered differently.

steventux: sorry @vanitabarrett initial comment wasn't entirely coherent - I've updated the code snippet accordingly

steventux: Ah I see :+1: makes sense.

PR #564, opened 2018-03-19
[WIP] Add link to finder from each section

leenagupte:
Needs to be rebased against #558 

Trello: https://trello.com/c/b9mzRtCa

Add a link to filterable list page from each supergroup section.

Escapes the query string in case passing `/` as part of the taxon `base_path` causes any issues.


leenagupte:
Replaced by #566

PR #581, opened 2018-03-21
Taxon tagged organisations

fmlharrison:
This is the work I did on querying rummager to return the tagged organisations for a taxon.

 https://trello.com/c/DFPIxvtc/143-get-all-tagged-organisations

### DIFF
--- a/app/models/organisation.rb
+++ b/app/models/organisation.rb
@@ -1,0 +1,18 @@
+require 'active_model'
+
+class Organisation
+  include ActiveModel::Model
+
+  attr_accessor(
+    :title,
+    :content_id,
+    :link,
+    :slug,
+    :organisation_state,
+    :document_count
+  )
+
+  def live?
+    organisation_state == 'live'
+  end
+end
--- a/app/models/rummager_search.rb
+++ b/app/models/rummager_search.rb
@@ -32,6 +32,20 @@ class RummagerSearch
     search_result["start"]
   end
 
+  def organisations
+    @_organisations ||= search_result.dig('aggregates', 'organisations', 'options').map do |option|
+      organisation = option['value']
+      Organisation.new(
+        title: organisation['title'],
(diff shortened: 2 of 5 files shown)


app/models/organisation.rb
```diff
@@ -0,0 +1,20 @@
...
+
+  attr_accessor(
+    :title,
+    :content_id,
+    :link,
+    :slug,
+    :organisation_type,
+    :organisation_state,
+    :document_count,
+    :acronym
```
vanitabarrett: Does it make sense to only be pulling out the attributes we know we need for now? We can always expand this later if we find out we need more. I.e: I don't think we need document_count, acronym, organisation type, possibly not slug either.

app/models/rummager_search.rb
```diff
@@ -32,6 +32,22 @@ def start
     search_result["start"]
   end
 
+  def organisations
+    @_organisations ||= search_result.dig('aggregates', 'organisations', 'options').map do |option|
```
vanitabarrett: Ah, this is what I was missing! Awesome 👍

app/models/organisation.rb
fmlharrison: I was thinking about this, I do agree that is easy to add them in later. I think I will remove what we don't need. I will keep slug. 

I think we need the the document count to make sure that the order of the organisations displayed are correct - but then again this list that comes back is in order of most tagged content....

vanitabarrett: Sounds good. 

It may be better to explicitly sort rather than rely on rummager returning these in the right order, in case that somehow changes?

fmlharrison: Yeah that was what I was thinking. I think that can be done in a presenter.

app/models/rummager_search.rb
fmlharrison: 😄

PR #582, opened 2018-03-21
Add organisation section to taxon page

fmlharrison:
This is the work that I did to add a organisation section to the taxon 
page. If the taxon has any tagged organisations, they are listed using
the document list component.

As this is V1, we just want to show a list of all the tagged organisations
Showing logos and branding may come later.

https://trello.com/c/7c6dGWkJ/144-add-an-organisations-section-to-the-topic-page

### DIFF
--- a/app/models/taxon.rb
+++ b/app/models/taxon.rb
@@ -66,6 +66,10 @@ class Taxon
     phase == "live"
   end
 
+  def organisations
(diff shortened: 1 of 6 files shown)


vanitabarrett:
Maybe double check with @miaallers that this list design is ok for V1? Pretty sure it was when I last spoke to her about it. 

@miaallers example here: https://govuk-collections-pr-582.herokuapp.com/education

PR #584, opened 2018-03-22
Show metadata based on content supergroup

fmlharrison:
This is the PR for the work @emmabeynon and I did to show the correct metadata of the list items in the supergroup sections.

https://trello.com/c/VcOm55tS/145-show-metadata-based-on-content-supergroup


app/presenters/taxon_presenter.rb
```diff
@@ -42,16 +42,24 @@ def section_title(supergroup)
...
+          path: document.base_path,
         },
         metadata: {
           public_updated_at: document.public_updated_at,
+          organisations: document.organisations,
           document_type: document.content_store_document_type.humanize
         },
       }
+
+      if supergroup == "services" || document.content_store_document_type == "guide"
```
vanitabarrett: This logic has become a bit confusing. It may be worth splitting this out into a separate method. 

It also might make it easier to read if we separate the if block by supergroup/document type rather than the action we're taking. It will mean a bit of code duplication, but we already have that with repeated `if supergroup == "services"` and I think it makes it clearer what changes are being applied where.

E.g:

```
if supergroup == "services"
   .....

if document.content_store_document_type == "guide"
  ......

File: test/integration/taxon_browsing_test.rb [hunk 2]
```diff
@@ -248,10 +266,40 @@ def tagged_organisations
     ]
   end
 
+  def tagged_content_for_services
+    @tagged_content_for_services ||= generate_search_results(5, "services")
+  end
+
+  def tagged_content_for_guides
+    @tagged_content_for_guides ||= generate_search_results(5, "guides")
```
vanitabarrett: Does this work? `generate_search_results` takes a supergroup - 'guides' isn't a supergroup?

vanitabarrett: Just realised you've change the way `generate_search_results` works, but I still think it's misleading to be passing "guides" in here in the same way you're passing supergroups.

test/support/rummager_helpers.rb
```diff
@@ -89,21 +89,22 @@ def stub_organisations_for_taxon(content_id, organisations)
     )
   end
 
-  def generate_search_results(count)
+  def generate_search_results(count, supergroup="default")
     (1..count).map do |number|
-      rummager_document_for_slug("content-item-#{number}")
```
vanitabarrett: Instead of having a method for each supergroup, can we not continue to use the `rummager_document_for_slug` method, passing in the format we want? E.g:

```
rummager_document_for_slug("content-item-#{number}", format = "local_transaction")
```

Maybe I'm missing something?

test/support/rummager_helpers.rb
```diff
@@ -89,9 +89,22 @@ def stub_organisations_for_taxon(content_id, organisations)
     )
   end
 
-  def generate_search_results(count)
+  def generate_search_results(count, supergroup = "default")
     (1..count).map do |number|
-      rummager_document_for_slug("content-item-#{number}")
+      case supergroup
```
vanitabarrett: nstead of having a method for each supergroup, can we not continue to use the rummager_document_for_slug method, passing in the format we want? E.g:

rummager_document_for_slug("content-item-#{number}", format = "local_transaction")
Maybe I'm missing something?

fmlharrison: I have made a new method for the supergroup sections that returns the results that look like the results that would come back from rummager search made to get tagged content for each section.

fmlharrison:
I have added some review mark up commits you you to have a look at. I don't know if it's the ideal way to do the metadata but it works and I think is more readable.

test/presenters/taxon_presenter_test.rb
```diff
@@ -97,14 +97,15 @@
       refute taxon_presenter.show_section?("guidance_and_regulation")
     end
 
-    it 'formats guidance and regulation content for document list' do
+    it 'formats guidance and regulation content except guides for document list' do
```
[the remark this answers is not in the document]
fmlharrison: Added comment to the tests to make it clear.

PR #538, opened 2018-03-12
[Do not merge] V1 Topic Page

leenagupte:
## This feature branch is now complete - please do not add any more commits

Trello cards: 
* https://trello.com/c/I0msexm2
* https://trello.com/c/GJwpHOew

PR #586, opened 2018-03-23
V1 Topic Pages (rebased branch)

vanitabarrett:
Trello: https://trello.com/c/231ueER9/151-deploy-the-v1-topic-page

This branches off the original topic pages feature branch https://github.com/alphagov/collections/pull/538 and has been rebased against master.

screen shot 2018-03-23 at 11 31 22

# Things to Check

- Topic pages still look as expected
- Compare between this PR and the original feature branch PR to ensure we haven't lost any commits
- World and Step by Step pages still look as expected
- Some browser testing of topic pages

# Links
Link to component guide: https://govuk-collections-pr-586.herokuapp.com/component-guide
Link to topic page root: 
- https://govuk-collections-pr-586.herokuapp.com/education
- https://govuk-collections-pr-586.herokuapp.com/childcare-parenting
- https://govuk-collections-pr-586.herokuapp.com/money


app/models/rummager_search.rb
```diff
@@ -32,6 +41,20 @@ def start
     search_result["start"]
   end
 
+  def organisations
+    @_organisations ||= search_result.dig('aggregates', 'organisations', 'options').map do |option|
+      organisation = option['value']
+      Organisation.new(
```
steventux: Given the attribute names of the `Organisation` class are identical to the keys here you could probably abbreviate this to:
```ruby
attr_keys = %w(title content_id link slug organisation_state documents)
Organisation.new(organisation.slice(*attr_keys))
```
or similar.

test/models/organisation_test.rb
```diff
@@ -0,0 +1,15 @@
+require 'test_helper'
+
+describe Organisation do
+  it 'returns the state of an organisation' do
+    organisation = Organisation.new(
```
steventux: Not all of these attrs are required for the test

```ruby
assert Organisation.new(organisation_state: 'live').live?, "Expected org to be live"
```

test/services/most_popular_content_test.rb
```diff
@@ -14,7 +14,7 @@ def taxon_content_id
   end
 
   describe '#fetch' do
-    it 'returns the results from search, sorted by title' do
+    it 'returns the results from search' do
```
steventux: Given the commit message we're not testing what the amended ordering is here - not sure if this is covered elsewhere or in another commit.

test/presenters/taxon_presenter_test.rb
```diff
@@ -216,6 +216,14 @@
   end
 
   describe 'guidance_and_regulation_section' do
+    it 'checks whether guidance section should be shown' do
+      taxon = mock
```
steventux: I'd use a `before` block for this setup and a `let` block for `TaxonPresenter.new(taxon)`.

test/models/organisation_test.rb
vanitabarrett: Good point, now amended: https://github.com/alphagov/collections/pull/586/commits/e3a7756564a5bdabd0def6bb1669dab3329664ac#diff-207daaac6cfd16d598218e9fcc875105R6

app/services/most_popular_content.rb
```diff
@@ -1,33 +1,38 @@
...
   def fetch
     search_response
       .documents
-      .sort_by(&:title)
   end
 
 private
 
   def search_response
+    search_fields = %w(title
```
steventux: If this list isn't modified it could be a class constant.

app/services/most_recent_content.rb
```diff
@@ -0,0 +1,37 @@
...
+  end
+
+  def fetch
+    search_response.documents
+  end
+
+private
+
+  def search_response
+    search_fields = %w(title
```
steventux: This is identical to the array of fields above in MostPopularContent (line 22) consider making a constant available to both.

test/services/tagged_organisations_test.rb
```diff
@@ -0,0 +1,69 @@
+require 'test_helper'
+require './test/support/custom_assertions.rb'
+
+describe TaggedOrganisations do
+  describe '#fetch' do
+    it 'return tagged organisations' do
+      search_results = {
```
steventux: Seems like this could be refactored into either a helper method or `let` vars.

test/services/tagged_organisations_test.rb
```diff
@@ -0,0 +1,69 @@
...
+      Services.rummager.stubs(:search).returns(search_results)
+
+      organisations = tagged_organisations.fetch
+      assert_equal(1, organisations.count)
+      assert_equal('Department for Education', organisations.first.title)
+      assert_equal(89, organisations.first.document_count)
+    end
+
+    it 'only returns live organisations' do
+      search_results = {
```
steventux: Similarly, these tests span many lines due to the dependency setup, consider using `before` and `let` to break them up into setup and assertion.

test/support/rummager_helpers.rb
```diff
@@ -18,15 +18,22 @@ def stub_content_for_taxon(content_ids, results, filter_navigation_document_supe
       )
   end
 
-  def stub_most_popular_content_for_taxon(content_id, results, filter_navigation_document_supertype: 'guidance')
+  def stub_most_popular_content_for_taxon(content_id, results, filter_content_purpose_supergroup: 'guidance_and_regulation')
+    fields = %w(title
```
steventux: This could be a seperate method,  it may even be able to service the same need in  `stub_most_recent_content_for_taxon`

app/models/rummager_search.rb
fmlharrison: The keys are slightly different than the attr name. I could change the attr name 'document_count' to 'documents'. 
But the organisation variable does not have the document key value pair on it so the above code does not quite work

steventux:
```ruby
keys = %w(title content_id link slug organisation_state)
attrs = organisation.slice(*keys).merge('document_count' => option['documents'])
Organisation.new(attrs)
```
or similar, though perhaps this is getting overly complex, you might be better off just being explicit about it.

app/presenters/taxon_presenter.rb
```diff
@@ -1,81 +1,101 @@
...
 
-    taxon.children? ? ACCORDION : LEAF
+  def section_title(supergroup)
+    supergroup.humanize
   end
 
-  def renders_as_accordion?
-    rendering_type == ACCORDION
+  def section_document_list(supergroup)
+    section_content(supergroup).each.map do |document|
```
steventux: I think we're conflating a few different concerns here, not necessary for this PR but I'd like to see a distinct class or module handling how supergroups generate their title and content. Perhaps a bunch of subclasses or something a bit more encapsulating of the supergroup specific logic. Seems to be spread across a few different classes here.

test/services/tagged_organisations_test.rb
vanitabarrett: Done https://github.com/alphagov/collections/pull/586/files#diff-62314a7230a7329a26b5cb721a328944

test/services/tagged_organisations_test.rb
vanitabarrett: Done https://github.com/alphagov/collections/pull/586/files#diff-62314a7230a7329a26b5cb721a328944

test/models/taxon_test.rb
```diff
@@ -96,5 +79,44 @@
 
       assert_equal ["guidance_content"], @taxon.tagged_content
     end
+
+    it "requests news_and_communications content" do
+      results = %i(result_1 result_2 result_3 result_4 result_5)
```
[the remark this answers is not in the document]
vanitabarrett: Done https://github.com/alphagov/collections/blob/7a78fa79ac05867cbfa41ddb2cf801c4e28409dc/test/models/taxon_test.rb#L7

app/models/rummager_search.rb
fmlharrison: I was going to do the above then I thought it was getting over complicated to read, so I am going to leave it as is.

vanitabarrett:
@steventux When you get a chance, can you re-review? I think we addressed most of your comments, and the others will be written up in a story to revisit once this PR is merged in.

PR #589, opened 2018-03-26
Use new breadcrumbs

tijmenb:
This PR:

- makes topic pages use a custom breadcrumbs instead of using GovukNavigationHelpers, which has been retired
- makes the mainstream browse pages and old topic pages use another type of custom breadcrumb

We don't use the contextual breadcrumbs here, because they are
specifically for use on content pages.

https://trello.com/c/WJlqy3VD


leenagupte:
@tijmenb The topic pages and the list pages use the same breadcrumb pattern. 
See:
* [topic](https://www.gov.uk/education/funding-and-finance-for-students)
* [list](https://www.gov.uk/search/advanced?group=transparency&topic=%2Feducation%2Ffunding-and-finance-for-students)

I think that taxon breadcrumbs should remain a common component that both pages can/should use.

tijmenb:
@leenagupte https://github.com/alphagov/govuk_publishing_components/pull/242 adds the shared helper.

app/views/taxons/_page_header.html.erb
```diff
@@ -2,7 +2,7 @@
   <div class="full-page-width-wrapper">
     <%= render partial: 'govuk_component/breadcrumbs',
       locals: {
```
leenagupte: Shouldn't this also render `govuk_publishing_components/components/contextual_breadcrumbs`, in the same way as the `breadcrumbs` partial?

app/views/world_wide_taxons/_common.html.erb
```diff
@@ -29,7 +29,7 @@
   %>
   <%= render partial: 'govuk_component/breadcrumbs',
```
leenagupte: Shouldn't this also render `govuk_publishing_components/components/contextual_breadcrumbs`, in the same way as the `breadcrumbs` partial?

tijmenb:
After offline conversation: the mixing in of the contextual breadcrumbs was confusing and wrong. I've pulled out the relevant code from the the components gem and put that into a `Breadcrumbs` class. Mainstream and old topic pages will use that now. Note that it kind of duplicates logic from the gem, but that's not bad because they will likely diverge in the future.

PR #595, opened 2018-03-27
Some refactoring of the step by step show page

tijmenb:
This makes the controller a little bit smaller and removes some indirection.

cc @sihugh


sihugh:
Thanks for this 👍

PR #615, opened 2018-04-10
Never show non-taxon things as taxon

tijmenb:
It seems like we have been showing the step by step pages with the taxon layout recently. I think this is caused by the `DocumentTypeRoutingConstraint` not finding the content item temporarily, and thinking it's a taxon (the fallback in the routes).

screen shot 2018-04-10 at 14 04 37

By crashing we 1) get insight into the error, 2) Fastly will serve the previous (correct) page from the mirror, instead of a wrong page.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -369,7 +369,7 @@ DEPENDENCIES
   webmock
 
 RUBY VERSION
-   ruby 2.4.2p198
+   ruby 2.4.4p296
 
 BUNDLED WITH
    1.16.1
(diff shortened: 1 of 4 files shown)


kevindew:
@tijmenb Heads up - this can cause time sensitivity issues with unpublishing (which is revealed through end-to-end tests). 

If we have the situation where we have an unpublished a taxon and the router and content store aren't in sync this will return a 500. Will probably need to embrace ability to return gone and redirect here to make this resilient like we have in government frontend: https://github.com/alphagov/government-frontend/blob/master/app/controllers/content_items_controller.rb

PR #618, opened 2018-04-11
Guidance regulation boxes

vanitabarrett:
Trello: https://trello.com/c/3iq7F2VK/16-add-white-boxes-to-guidance-and-regulation-section

We want to show the most popular guidance and regulation content in highlight boxes so they are more prominent to users.

screen shot 2018-04-12 at 10 23 52

Example Pages with varying numbers of tagged guidance and regulation content

- https://govuk-collections-pr-618.herokuapp.com/education
- https://govuk-collections-pr-618.herokuapp.com/education/help-with-school-costs
- https://govuk-collections-pr-618.herokuapp.com/education/home-schooling

### DIFF
--- a/app/assets/stylesheets/components/_highlight-boxes.scss
+++ b/app/assets/stylesheets/components/_highlight-boxes.scss
@@ -8,13 +8,13 @@
   @include core-19;
   list-style-type: none;
   padding: 5px 25px $gutter-two-thirds 0;
-  width: 33%;
+  width: 100%;
   min-height: 150px;
-  min-width: 280px;
+  min-width: 250px;
   @include box-sizing(border-box);
 
-  @include media(mobile) {
-    width: 100%;
+  @include media(tablet) {
+    width: 33%;
   }
 }
 
--- a/app/assets/stylesheets/components/_highlight-boxes.scss
+++ b/app/assets/stylesheets/components/_highlight-boxes.scss
@@ -11,7 +11,7 @@
   width: 33%;
   min-height: 150px;
   min-width: 280px;
-  box-sizing: border-box;
+  @include box-sizing(border-box);
 
   @include media(mobile) {
     width: 100%;
@@ -23,8 +23,8 @@
   padding: $gutter-half $gutter-half $gutter;
   width: 100%;
   height: 100%;
-  box-sizing: border-box;
   box-shadow: 7px 7px 0 $white, 8px 8px 0 $grey-2;
+  @include box-sizing(border-box);
 }
 
 .app-c-highlight-boxes--inverse {
--- a/app/assets/stylesheets/components/_highlight-boxes.scss
+++ b/app/assets/stylesheets/components/_highlight-boxes.scss
@@ -8,19 +8,19 @@
   @include core-19;
   list-style-type: none;
   padding: 5px 25px $gutter-two-thirds 0;
-  width: 100%;
+  width: 33%;
   min-height: 150px;
-  min-width: 250px;
+  min-width: 280px;
   @include box-sizing(border-box);
 
-  @include media(tablet) {
-    width: 33%;
+  @include media(mobile) {
+    width: 100%;
   }
 }
 
 .app-c-highlight-boxes__item {
   border: 1px solid $grey-2;
-  padding: $gutter-half $gutter-half $gutter;
+  padding: $gutter-half * 1.5;
   width: 100%;
   height: 100%;
   box-shadow: 7px 7px 0 $white, 8px 8px 0 $grey-2;
--- a/app/views/taxons/sections/_news_and_communications.html.erb
+++ b/app/views/taxons/sections/_news_and_communications.html.erb
@@ -1,9 +1,1 @@
-<div class="grid-row">
-  <div class="column-two-thirds">
-    <%= render 'govuk_publishing_components/components/document_list',
-      items: section[:documents],
-      margin_top: true,
-      margin_bottom: true
-    %>
-  </div>
-</div>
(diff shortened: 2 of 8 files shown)


app/assets/stylesheets/components/_highlight-boxes.scss
```diff
@@ -13,7 +14,7 @@
   box-sizing: border-box;
```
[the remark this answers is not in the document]
vanitabarrett: I definitely never knew these existed, thank you!

app/views/taxons/sections/_news_and_communications.html.erb
```diff
@@ -0,0 +1,9 @@
+<div class="grid-row">
```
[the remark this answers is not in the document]
vanitabarrett: I think the different section partials may become more different in the future when we implement the next stage of the design, but good point - for now I'll create a partial which they all use, and then if it diverges in future we can tackle that then 👍

app/assets/stylesheets/components/_highlight-boxes.scss
```diff
@@ -13,7 +14,7 @@
...
 
   @include media(mobile) {
```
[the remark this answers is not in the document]
vanitabarrett: Noted! Changed in https://github.com/alphagov/collections/pull/618/commits/54302a9500f1a5f8f6837268a0a1f7ce43135bc0

app/assets/stylesheets/components/_highlight-boxes.scss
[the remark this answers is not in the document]
vanitabarrett: Done https://github.com/alphagov/collections/pull/618/commits/6bd41b2c94ecf44cb515aa6879fdda5a83c21bfa

vanitabarrett:
@andysellick Thanks for your comments, I think they've all been addressed now. A small bit of additional styling and logic has also since been added - would you be able to re-review?

app/assets/stylesheets/components/_highlight-boxes.scss
```diff
@@ -20,11 +20,11 @@
 
 .app-c-highlight-boxes__item {
   border: 1px solid $grey-2;
-  padding: $gutter-half $gutter-half $gutter;
+  padding: $gutter-half * 1.5;
   width: 100%;
   height: 100%;
-  box-sizing: border-box;
   box-shadow: 7px 7px 0 $white, 8px 8px 0 $grey-2;
+  @include box-sizing(border-box);
```
andysellick: General formatting point I picked up from Paul - he suggested to always put @includes at the start of a class. It's not hugely important though.

app/assets/stylesheets/components/_highlight-boxes.scss
```diff
@@ -20,11 +20,11 @@
 
 .app-c-highlight-boxes__item {
   border: 1px solid $grey-2;
-  padding: $gutter-half $gutter-half $gutter;
+  padding: $gutter-half * 1.5;
   width: 100%;
```
andysellick: Is this a block level element? If so, width: 100% should be implied, so it doesn't need to be included.

app/assets/stylesheets/components/_highlight-boxes.scss
```diff
@@ -11,7 +11,7 @@
   width: 33%;
   min-height: 150px;
   min-width: 280px;
```
andysellick: Is this line needed? It causes the boxes to wrap onto a second line below around 870px screen width, I can't tell if that's intentional or not?

screen shot 2018-04-12 at 13 12 37

app/assets/stylesheets/components/_highlight-boxes.scss
```diff
@@ -11,7 +11,7 @@
   width: 33%;
   min-height: 150px;
```
andysellick: Again, I may not be understanding this fully, but playing around with the examples this line doesn't seem to make any difference. I've just noticed the parent is using flexbox, which would explain it - I think by default all elements in a flex element have the same height i.e. the height of the tallest element. So this line is redundant, I believe.

app/assets/stylesheets/components/_highlight-boxes.scss
```diff
@@ -20,11 +20,11 @@
 
 .app-c-highlight-boxes__item {
```
andysellick: It's possible that having this child element of the LI isn't necessary with flexbox. I mention this only in passing, it's not essential to investigate.

app/assets/stylesheets/components/_highlight-boxes.scss
```diff
@@ -11,7 +11,7 @@
   width: 33%;
```
andysellick: The code seems to be outside the scope of this PR so again this is only a mention rather than necessary for investigation, but due to the irregular padding on the LI element the boxes overall don't fill the full width of the space available, see below (I've cut out some of the elements on the page so it's more obvious when compared with the 'Guidance and Regulation' title). This could be solved using flexbox's justify-content: space-between option and removing the padding altogether.

screen shot 2018-04-12 at 13 22 37

app/assets/stylesheets/components/_highlight-boxes.scss
vanitabarrett: Ah okay, I think I learnt to put them last - I'll bear that in mind 🙂

vanitabarrett:
@andysellick Thanks for the re-review, great spots 👍 

As they're more related to the component itself, not adding it to the topic page, I'll address the issues in a separate PR.  I've created one story to track all comments, as they're fairly small: https://trello.com/c/qrnKKusX/22-frontend-tweaks-to-highlight-box-component

PR #620, opened 2018-04-12
Tweak highlight box styling

vanitabarrett:
While adding the highlight boxes to the Guidance and Regulation section, [ @andysellick noticed a few minor changes](https://github.com/alphagov/collections/pull/618) that could be made to the highlight box code.

### Highlight boxes do not take full width ([link to comment](https://github.com/alphagov/collections/pull/618#discussion_r181061782))
Andy noticed that the highlight boxes don't take up the full width of the supergroup section they are placed in. On investigation, the actual component takes up the space, but the boxes themselves do not due to a `padding-right: 25px` on each box. 

I have added a negative `-25px` margin right on the parent container to counteract this. 

**Before:**
screen shot 2018-04-12 at 16 27 07

**After:**
screen shot 2018-04-12 at 16 26 46

I've chosen this over other options because of the following:

- `justify-content: space-between` - component breaks when passed less than 3 items (see screenshot below)
- Removing padding-right on last child - difficult to apply when using flex wrap, as the "last item" in each row will change on resizing.

screen shot 2018-04-12 at 15 13 58

### Other issues
The following issues have also been addressed in this PR:

- Unnecessary `width: 100%` on `.app-c-highlight-boxes__item`
- Consistency of `@include` location in component CSS

https://govuk-collections-pr-620.herokuapp.com/component-guide/highlight-boxes
https://govuk-collections-pr-620.herokuapp.com/education

### DIFF
--- a/app/assets/stylesheets/components/_highlight-boxes.scss
+++ b/app/assets/stylesheets/components/_highlight-boxes.scss
@@ -2,19 +2,24 @@
   display: flex;
   flex-wrap: wrap;
   margin-top: $gutter-half;
+
+  @include media(tablet) {
+    margin-right: -25px;
+  }
 }
 
 .app-c-highlight-boxes__item-wrapper {
   @include core-19;
   list-style-type: none;
-  padding: 5px 25px $gutter-two-thirds 0;
-  width: 33%;
+  padding: 5px 10px 25px 0;
+  width: 100%;
   min-height: 150px;
-  min-width: 280px;
+  min-width: 250px;
   @include box-sizing(border-box);
 
-  @include media(mobile) {
-    width: 100%;
+  @include media(tablet) {
+    width: 33%;
+    padding: 5px 25px 25px 0;
   }
 }
 
--- a/app/assets/stylesheets/components/_highlight-boxes.scss
+++ b/app/assets/stylesheets/components/_highlight-boxes.scss
@@ -14,7 +14,6 @@
   list-style-type: none;
   padding: 5px 10px 25px 0;
   width: 100%;
-  min-height: 150px;
   min-width: 250px;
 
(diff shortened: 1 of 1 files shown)


app/assets/stylesheets/components/_highlight-boxes.scss
```diff
@@ -2,29 +2,33 @@
...
 
 .app-c-highlight-boxes__item-wrapper {
   @include core-19;
+  @include box-sizing(border-box);
   list-style-type: none;
-  padding: 5px 25px $gutter-two-thirds 0;
-  width: 33%;
+  padding: 5px 10px 25px 0;
+  width: 100%;
   min-height: 150px;
```
andysellick: I'm still not certain a min-height is necessary, due to the way flexbox works. Unless this min height is to try to make them all look a similar height on mobile?

vanitabarrett: Sorry, didn't spot this comment on the last PR. I agree - it isn't really needed (I think it was implemented for mobile, as you say, but I don't think they need to be a similar height on mobile because of the way they're stacked). Will remove 👍

vanitabarrett:
@andysellick Good spot, I've removed the min-height. 

I'm not sure if we can merge the `app-c-highlight-boxes__item-wrapper` element with the `app-c-highlight-boxes__item`, as the outer wrapper is being used for spacing between each item element with padding (which lets us use box-sizing: border-box). If we merge the two elements, I think we'd have to use margin and we wouldn't be able to produce the same effect (unless I'm missing something!)

andysellick:
@vanitabarrett I had a quick go at it in the inspector and it looks like it could be possible, but it could be more effort than it's worth right now, so don't worry.

PR #631, opened 2018-04-23
Add featured item for News & Communication on topic pages

vanitabarrett:
Trello: https://trello.com/c/bpfOboeR/34-show-a-photo-above-the-first-news-and-communication-item

We want to highlight the most recent item in News and Communication so it stands out to users. If an item does not have an image, it will use the default whitehall road sign image - as far as I'm aware, "correspondence" document types do not have images so will always use the default.

**FYI: some images will appear as broken because they use a relative URL and the heroku app isn't on https://gov.uk. I'll double check these on integration before deploying**

**Before:**
screen shot 2018-04-24 at 10 28 17

**After:**
screen shot 2018-04-24 at 10 26 35

**Example Links:**
Component Guide: https://govuk-collections-pr-631.herokuapp.com/component-guide
With Image: https://govuk-collections-pr-631.herokuapp.com/education/pupil-wellbeing-behaviour-and-attendance
Using default whitehall image (won't work in heroku without image URL hacking): https://govuk-collections-pr-631.herokuapp.com/education/further-and-higher-education-skills-and-vocational-training
Only one news item (won't work in heroku without image URL hacking): https://govuk-collections-pr-631.herokuapp.com/education/financial-management-for-further-education-providers

### DIFF
--- a/app/assets/stylesheets/views/_taxons.scss
+++ b/app/assets/stylesheets/views/_taxons.scss
@@ -36,19 +36,40 @@
 }
 
 .taxon-page__featured-item {
+  @extend %contain-floats;
+  margin-top: $gutter-two-thirds;
+
   .taxon-page__featured-image {
-    margin-top: 20px;
-    width: 100%;
+    @include box-sizing(border-box);
+    width: (1 / 3) * 100%;
+    float: left;
+    margin-bottom: $gutter-one-third;
+    padding-right: $gutter-half;
+
+    @include media(tablet) {
+      float: none;
+      margin-bottom: $gutter-one-third;
+      padding-right: 0;
+      width: 100%;
+    }
+  }
+
+  .taxon-page__featured-text {
+    display: inline-block;
+    width: (2 / 3) * 100%;
+
+    @include media(tablet) {
+      width: 100%;
+    }
   }
 
   .taxon-page__featured-link {
-    @include bold-27;
-    display: block;
-    margin: $gutter-half 0 $gutter-one-third 0;
+    @include bold-19;
   }
 
   .taxon-page__featured-metadata-wrapper {
     @include core-16;
+    margin-top: $gutter-one-third / 2;
 
     .taxon-page__featured-metadata {
       display: inline-block;
--- a/app/assets/stylesheets/views/_taxons.scss
+++ b/app/assets/stylesheets/views/_taxons.scss
@@ -36,6 +36,11 @@
 }
 
 .taxon-page__featured-item {
+  .taxon-page__featured-image {
+    margin-top: 20px;
+    width: 100%;
+  }
+
   .taxon-page__featured-link {
     @include bold-27;
     display: block;
--- a/app/assets/stylesheets/views/_taxons.scss
+++ b/app/assets/stylesheets/views/_taxons.scss
@@ -39,6 +39,10 @@
   @extend %contain-floats;
   margin-top: $gutter-two-thirds;
 
+  &--single {
+    min-width: 330px;
+  }
+
   .taxon-page__featured-image {
     @include box-sizing(border-box);
     width: (1 / 3) * 100%;
@@ -76,6 +80,13 @@
       margin-right: $gutter-two-thirds;
     }
   }
+}
+
+.taxon-page__featured-see-more {
+  display: block;
+  padding-top: $gutter-two-thirds;
+  padding-left: $gutter-half;
+  clear: both;
 }
 
 .taxon-page__grid {
--- a/app/assets/stylesheets/views/_taxons.scss
+++ b/app/assets/stylesheets/views/_taxons.scss
@@ -35,6 +35,23 @@
   margin-top: 0;
 }
 
+.taxon-page__featured-item {
+  .taxon-page__featured-link {
+    @include bold-27;
+    display: block;
+    margin: $gutter-half 0 $gutter-one-third 0;
+  }
+
+  .taxon-page__featured-metadata-wrapper {
+    @include core-16;
+
+    .taxon-page__featured-metadata {
+      display: inline-block;
+      margin-right: $gutter-two-thirds;
+    }
+  }
+}
+
 .taxon-page__grid {
   background-color: $grey-4;
   padding-top: $gutter;
--- a/app/presenters/supergroups/news_and_communications.rb
+++ b/app/presenters/supergroups/news_and_communications.rb
@@ -9,5 +9,10 @@ module Supergroups
     def tagged_content(taxon_id)
       @content = MostRecentContent.fetch(content_id: taxon_id, filter_content_purpose_supergroup: @name)
     end
+
+
+    def promoted_content_count(*)
+      1
+    end
   end
 end
--- a/app/presenters/supergroups/news_and_communications.rb
+++ b/app/presenters/supergroups/news_and_communications.rb
@@ -36,9 +36,14 @@ module Supergroups
     end
 
     def news_item_photo(base_path)
+      default_news_image = {
+        "url": "/government/assets/placeholder.jpg",
+        "alt_text": ""
+      }.with_indifferent_access
+
       news_item = ::Services.content_store.content_item(base_path)
 
-      news_item["details"]["image"]
+      news_item["details"]["image"] || default_news_image
     end
 
     def promoted_content_count(*)
--- a/app/presenters/supergroups/news_and_communications.rb
+++ b/app/presenters/supergroups/news_and_communications.rb
@@ -10,6 +10,36 @@ module Supergroups
       @content = MostRecentContent.fetch(content_id: taxon_id, filter_content_purpose_supergroup: @name)
     end
 
+    def document_list(taxon_id)
+      tagged_content(taxon_id).each_with_index.map do |document, index|
+        data = {
+          link: {
+            text: document.title,
+            path: document.base_path
+          },
+          metadata: {
+            public_updated_at: document.public_updated_at,
+            organisations: document.organisations,
+            document_type: document.content_store_document_type.humanize
+          }
+        }
+
+        if index < promoted_content_count
+          document_image = news_item_photo(document.base_path)
+          data[:image] = {}
+          data[:image][:url] = document_image["url"]
+          data[:image][:alt] = document_image["alt_text"]
+        end
+
+        data
+      end
+    end
+
+    def news_item_photo(base_path)
+      news_item = ::Services.content_store.content_item(base_path)
+
+      news_item["details"]["image"]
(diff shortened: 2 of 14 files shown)


app/assets/stylesheets/views/_taxons.scss
```diff
@@ -35,6 +35,59 @@
...
+    @include bold-19;
+    display: block;
+    margin-bottom: $gutter-one-third;
+  }
+
+  .taxon-page__featured-metadata-wrapper {
+    @include core-16;
+
+    .taxon-page__featured-metadata {
+      display: inline-block;
```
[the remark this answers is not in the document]
vanitabarrett: This follows the pattern we currently have with metadata in the document list component, e.g:
screen shot 2018-04-24 at 12 44 21

We may address this as part of: https://trello.com/c/wMmaf5ut/59-improve-readability-of-metadata-in-consultation-boxes

@miaallers what do you think?

app/presenters/supergroups/news_and_communications.rb
```diff
@@ -9,5 +9,45 @@ def initialize
...
+        end
+
+        data
+      end
+    end
+
+    def news_item_photo(base_path)
+      default_news_image = {
+        "url": "https://assets.publishing.service.gov.uk/government/assets/placeholder-a6e35fc15d4bfcc9d0781f3888dc548b39288f0fe10a65a4b51fef603540b0c5.jpg",
+        "alt_text": ""
```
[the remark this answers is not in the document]
vanitabarrett: I've decided to remove it as the default image is this...

screen shot 2018-04-24 at 12 46 04

...which I don't think really adds to the user experience/provides them with any more context than they'd already get from the article title. At this point, the image is only really there for consistency of layout, so I think hiding the image from screen readers makes sense

app/assets/stylesheets/views/_taxons.scss
```diff
@@ -35,6 +35,59 @@
   margin-top: 0;
 }
 
+.taxon-page__featured-item {
+  margin-top: $gutter-two-thirds;
+  font-size: 0;
```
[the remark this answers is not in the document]
vanitabarrett: Good point, I think I started implementing one solution and switched halfway through, which meant I forgot about floats. 👍

app/assets/stylesheets/views/_taxons.scss
```diff
@@ -35,6 +35,59 @@
...
+
+    .taxon-page__featured-metadata {
+      display: inline-block;
+      margin-right: $gutter-two-thirds;
+    }
+  }
+}
+
+.taxon-page__featured-see-more {
+  @include core-16;
```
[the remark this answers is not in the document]
vanitabarrett: This has been addressed by using floats, so removing the `font-size: 0` that was requiring us to set `@include core-16`. The link now follows the same pattern as the other see more links on the page.

app/views/taxons/sections/_news_and_communications.html.erb
```diff
@@ -1 +1,44 @@
...
+        <% end %>
+      </div>
+    </div>
+
+  <% if section[:documents].empty? %>
+      <%= link_to(
+        section[:see_more_link][:text],
+        section[:see_more_link][:url], class: "taxon-page__featured-see-more"
+      )%>
+    </div>
```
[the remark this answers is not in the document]
vanitabarrett: I was hesitant about doing it, but at the time decided to put it above readability of code. I've implemented a different solution which now uses 2 if statements rather than 1 if/elsif. The code is longer, but I think it makes it easier to read?

app/views/taxons/sections/_news_and_communications.html.erb
```diff
@@ -1 +1,32 @@
-<%= render partial: 'document_list_with_grid', locals: { documents: section[:documents], see_more_link: section[:see_more_link] } %>
+
+<% featured_item = section[:documents].shift(section[:promoted_content_count]).first %>
```
leenagupte: Do you need the `.first` here?

app/presenters/supergroups/news_and_communications.rb
```diff
@@ -36,9 +36,14 @@ def document_list(taxon_id)
     end
 
     def news_item_photo(base_path)
+      default_news_image = {
```
leenagupte: Is this stored anywhere? Can the default image be retrieved from anywhere rather than being hard-coded here?

vanitabarrett:
@andysellick Thanks for your comments - really useful! FYI the see more link is already tested in [TaxonBrowsingTest](https://github.com/alphagov/collections/blob/master/test/integration/taxon_browsing_test.rb#L171). I think I've addressed most of your comments if you want to re-review

app/views/taxons/sections/_news_and_communications.html.erb
```diff
@@ -18,18 +22,25 @@
...
     </div>
+    <% if no_more_news_items %>
+      <%= link_to(
+        section[:see_more_link][:text],
+        section[:see_more_link][:url], class: "taxon-page__featured-see-more"
+      ) %>
+    <% end %>
   </div>
 
+  <% if !no_more_news_items %>
```
leenagupte: I would remove the double-negative, and instead change this to an `else` statement.

app/views/taxons/sections/_news_and_communications.html.erb
```diff
@@ -1,4 +1,3 @@
-
 <%
```
leenagupte: I'd fix this up to the commit that introduced it.

app/views/taxons/sections/_news_and_communications.html.erb
vanitabarrett: I think so - `section[:documents].shift(section[:promoted_content_count])` returns the first item in an array of hashes, so we need to get the first item from that array.

app/presenters/supergroups/news_and_communications.rb
vanitabarrett: Have adjusted to use relative URL: '/government/assets/placeholder.jpg'

app/views/taxons/sections/_news_and_communications.html.erb
vanitabarrett: Good point, I've kept the if/if structure the same but renamed the variable to `documents`. So it now checks for `documents` or `!documents`, which I think is clearer?

vanitabarrett:
@leenagupte @andysellick Thanks both for your comments. I've hopefully addressed them all now - if you have some time to re-review that would be great 😃

app/assets/stylesheets/views/_taxons.scss
```diff
@@ -35,6 +35,59 @@
...
+    @include media(tablet) {
+      float: none;
+      margin-bottom: $gutter-one-third;
+      padding-right: 0;
+      width: 100%;
+    }
+  }
+
+  .taxon-page__featured-text {
+    display: inline-block;
```
andysellick: Interestingly you can remove all of the CSS for this element and it still behaves (almost) correctly, because the image next to it is floated left. Unfortunately that doesn't work if at any point the text is taller than the image, because it wraps onto the next line. I'd advise for consistency to also float this element, even though you'd have to add another line to turn off float above tablet.

app/views/taxons/sections/_news_and_communications.html.erb
```diff
@@ -1 +1,44 @@
-<%= render partial: 'document_list_with_grid', locals: { documents: section[:documents] } %>
+<%
+  featured_item = section[:documents].shift(section[:promoted_content_count]).first
+  featured_item_layout_class = "taxon-page__featured-item--single" if section[:documents].empty?
+%>
+
+<div class="grid-row">
+  <div class="taxon-page__featured-item column-one-third <%= featured_item_layout_class %>">
+    <img class="taxon-page__featured-image" src="<%= featured_item[:image][:url] %>" alt="<%= featured_item[:image][:alt] %>">
```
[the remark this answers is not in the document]
andysellick: It's a tricky problem. One thing you can do is set the tabindex of the image to -1 so at least keyboard users don't have to deal with it.

Another approach is the one I'm considering for the image/link component we discussed yesterday, where I'm going to face the exact same problem. The BBC news homepage has an interesting approach for their main story, where the image has no link, but the story title link next to it has a before pseudo that's positioned absolute 100% width 100% height to the containing div, so any part of that div is clickable. Any text in the same container is given position relative and a higher z-index than the pseudo element so it sits above it and still seems to behave normally i.e. can be selected etc.

Again though, you probably don't have to solve this problem, you could just wait for me to build that component 😃

app/assets/stylesheets/views/_taxons.scss
```diff
@@ -35,6 +35,59 @@
...
+
+    .taxon-page__featured-metadata {
+      display: inline-block;
+      margin-right: $gutter-two-thirds;
+    }
+  }
+}
+
+.taxon-page__featured-see-more {
+  display: block;
```
andysellick: Suggest taking display block off this element and moving the margin to the bottom of the element above. Display block on links can be confusing for users, because if they click near the link they might hit it without intending to. Also because this seems to be the same (in appearance) link as the one that appears when there's a list of stories, and that isn't block.

app/views/taxons/sections/_news_and_communications.html.erb
```diff
@@ -1 +1,45 @@
...
+            <time class="taxon-page__featured-metadata" datetime="<%= metadata_value.iso8601 %>">
+              <%= l(metadata_value, format: '%e %B %Y') %>
+            </time>
+          <% else %>
+            <p class="taxon-page__featured-metadata"><%= metadata_value %></p>
+          <% end %>
+        <% end %>
+      </div>
+    </div>
+    <% if !documents %>
```
andysellick: I could be wrong about this, but is this logic overly complicated? Am I right in thinking that this 'see more' link should always appear? Then you wouldn't need to have it twice in the code, just take it out of the if statements and put it at the bottom.

(thinking out loud here) If you did do that, it would mean that the see more link would appear like this:

screen shot 2018-04-26 at 09 34 16

...which obviously wouldn't be good. However, you might be able to move the logic to the div class instead: `<div class="<%= "column-two-thirds" if documents %>">`

Sorry, just wondering if you'd thought about this.

vanitabarrett: Nice! You're right in thinking the see more link needs to appear and that I put the logic in place purely due to placement of that link, but... I didn't think of moving the logic to the `div`!

I've managed to get that working with `<div class="<%= documents ? "column-two-thirds" : "taxon-page__featured-see-more" %>">`. Great spot, thanks! 👍💯

PR #650, opened 2018-05-08
[WIP] Create organisations index page

MahmudH:
New route created pointing to new organisations index page. This page displays stubbed organisations. 

https://trello.com/c/F3Hx0P7T/28-create-organisations-index-page


app/controllers/organisations_controller.rb
```diff
@@ -0,0 +1,59 @@
+class OrganisationsController < ApplicationController
+  attr_reader :content_item
```
Davidslv: What's the reason to add the `attr_reader` here?

app/views/organisations/index.html.erb
```diff
@@ -0,0 +1,34 @@
+<% content_for :title, "Departments, agencies and public bodies - GOV.UK" %>
+
+<h1>Departments, agencies and public bodies - GOV.UK</h1>
+<br>
+<h2>Ministerial departments</h2>
+<%= @content.select { |item| item[:department_type] == "Ministerial" }.count %>
```
Davidslv: This logic shouldn't belong in the view

app/views/organisations/index.html.erb
```diff
@@ -0,0 +1,34 @@
+<% content_for :title, "Departments, agencies and public bodies - GOV.UK" %>
+
+<h1>Departments, agencies and public bodies - GOV.UK</h1>
+<br>
```
vanitabarrett: Is this extra spacing needed? If yes, can we do this with CSS instead? (we should avoid using HTML tags for styling)

app/controllers/organisations_controller.rb
MahmudH: This was to display the content item from the content store on the page for stubbing purposes but I believe this can be removed

app/views/organisations/index.html.erb
```diff
@@ -0,0 +1,34 @@
...
+<h2>Ministerial departments</h2>
+<%= @content.select { |item| item[:department_type] == "Ministerial" }.count %>
+<ul>
+  <% @content.each do |item| %>
+    <% if item[:department_type] == "Ministerial" %>
```
Davidslv: it seems you need two methods to filter out "Ministerial" and "Non Ministerial".
There's a few things here that are problematic: 

- There's a lot of repetition in this view
- You are cycling through the whole `@content` twice which adds complexity
- There's a lot of logic in this view that should belong somewhere else, for example: a view presenter, or a ContentItemHelper with the logic seems a more appropriate place to send this.

app/views/organisations/index.html.erb
```diff
@@ -0,0 +1,34 @@
...
+    <li class="app-c-taxon-list__item"><%= link_to item[:title], item[:base_path] %></li>
```
vanitabarrett: You're using a component class here instead of using the component itself. We should never reuse component classes outside of a component. There's some docs here on how to call the taxon-list component: collections.dev.gov.uk/component-guide/taxon-list (when you're running Collections locally) 🙂

For reference, it might be something like: 

```
<%= render 'components/taxon-list', {
          items: @content
      } %>
```

app/controllers/organisations_controller.rb
Davidslv: That's not a good idea to add code for testing purposes.
I believe you can remove it and it will still work

app/views/organisations/index.html.erb
```diff
@@ -0,0 +1,34 @@
+<% content_for :title, "Departments, agencies and public bodies - GOV.UK" %>
+
+<h1>Departments, agencies and public bodies - GOV.UK</h1>
```
vanitabarrett: We might want to put hardcoded strings like this into a locale file so that, in the future, we can add translations into other languages.

app/views/organisations/index.html.erb
```diff
@@ -0,0 +1,34 @@
...
+<br>
+<h2>Ministerial departments</h2>
+<%= @content.select { |item| item[:department_type] == "Ministerial" }.count %>
+<ul>
+  <% @content.each do |item| %>
+    <% if item[:department_type] == "Ministerial" %>
+    <li class="app-c-taxon-list__item"><%= link_to item[:title], item[:base_path] %></li>
+    <% end %>
+    <% if item[:agencies].length > 1 && item[:department_type] == "Ministerial" %>
```
vanitabarrett: If would suggest moving some of this logic out of the view and into a presenter to make things a bit tidier

app/controllers/organisations_controller.rb
MahmudH: Yes, I will remove it

MahmudH:
Closing as will be superseded by another PR

PR #651, opened 2018-05-09
Topic list add branding

andysellick:
Add colour branding and a 'small' option to the topic list component.

Component guide: https://govuk-collections-pr-651.herokuapp.com/component-guide/topic-list

Trello card: https://trello.com/c/SvSNQlsa/50-modify-component-list-of-links-topic-list


app/views/components/_topic-list.html.erb
```diff
@@ -1,20 +1,22 @@
 <%
   items ||= []
   see_more_link ||= false
+  size ||= false
+  component_size = "app-c-topic-list--small" if size == 'small'
```
vanitabarrett: Would passing in `small: true` make more sense here? Otherwise people might make typos and/or expect other sizes to be able to be passed in.

andysellick: Ha. I came up with both options but thought if I used small: true someone would say why not do a general one for size? But I guess we don't build things that we don't need yet, so I'll take a look.

On the fixup, I think I tried but it didn't work. I'll give it another attempt - I might be thinking of another PR.

andysellick: @vanitabarrett I can't fixup that commit - I must have changed something that means the commits can't be reordered. Have addressed the other point though.

app/assets/stylesheets/components/_topic-list.scss
```diff
@@ -4,10 +4,20 @@
   margin-bottom: $gutter-one-third;
 }
 
-.app-c-topic-list__link {
+.app-c-topic-list__link--no-underline {
   text-decoration: none;
```
chris-gds, line 8: Out of scope but - can we remove this? 
I feel like [these](https://www.gov.uk/topic/environmental-management) should have an underline as they're links

PR #652, opened 2018-05-09
Add tracking to highlight boxes within supergroup sections

vanitabarrett:
Trello: https://trello.com/c/ezNswqdS/73-add-tracking-to-highlight-boxes-on-topic-pages

This PR covers expanding the highlight box component to include data tracking. It *does not* cover using these tracking abilities on topic pages. This will now be addressed in a separate PR.

Component Guide: https://govuk-collections-pr-652.herokuapp.com/component-guide/highlight-boxes

### DIFF
--- a/app/helpers/highlight_boxes_helper.rb
+++ b/app/helpers/highlight_boxes_helper.rb
@@ -1,0 +1,9 @@
+module HighlightBoxesHelper
+  def data_tracking?(items)
+    items.each do |item|
+      return true if item[:link][:data_attributes]
+    end
+
+    false
+  end
+end
--- a/app/views/components/_highlight-boxes.html.erb
+++ b/app/views/components/_highlight-boxes.html.erb
@@ -4,16 +4,17 @@
   inverse_class = "app-c-highlight-boxes--inverse" if inverse
 %>
 <% if items.any? %>
-  <ol class="app-c-highlight-boxes">
+  <ol class="app-c-highlight-boxes" <%= "data-module=track-click" if data_tracking?(local_assigns[:items]) %>>
     <% items.each do |content_item| %>
       <li class="app-c-highlight-boxes__item-wrapper">
         <div class="app-c-highlight-boxes__item <%= inverse_class %>">
-          <a
-            class="app-c-highlight-boxes__title <%= "app-c-highlight-boxes__title--featured" if content_item[:link][:featured] %>"
-            href="<%= content_item[:link].fetch(:path) %>"
-          >
-            <%= content_item[:link].fetch(:text)  %>
-          </a>
+          <%= link_to(
+            content_item[:link].fetch(:text),
+            content_item[:link].fetch(:path),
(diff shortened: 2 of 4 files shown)


app/views/components/_highlight-boxes.html.erb
```diff
@@ -4,16 +4,17 @@
...
-            class="app-c-highlight-boxes__title <%= "app-c-highlight-boxes__title--featured" if content_item[:link][:featured] %>"
-            href="<%= content_item[:link].fetch(:path) %>"
-          >
-            <%= content_item[:link].fetch(:text)  %>
-          </a>
+          <%= link_to(
+            content_item[:link].fetch(:text),
+            content_item[:link].fetch(:path),
+            class: "app-c-highlight-boxes__title #{"app-c-highlight-boxes__title--featured" if content_item[:link][:featured]}",
+            data: (highlight_box_tracking_attributes(local_assigns[:data_track_category], content_item, index))
```
leenagupte: This function feels like something that should be called in the presenter rather than the view.

app/views/taxons/sections/_guidance_and_regulation.html.erb
```diff
@@ -1,5 +1,6 @@
 <%= render 'components/highlight-boxes',
-  items: section[:documents].shift(section[:promoted_content_count])
+  items: section[:documents].shift(section[:promoted_content_count]),
+  data_track_category: "guidanceAndRegulationHighlightBoxClicked"
```
leenagupte: This value feels like something that should be set in the supergroup presenter. It could also probably be created programmatically.

app/views/components/_highlight-boxes.html.erb
andysellick: Is this approach flexible enough? What if I want to use this component in a different context, and pass additional tracking attributes, like dimensions? Is that possible?

vanitabarrett: I think we're going to change the approach to be more flexible (to match the document list component). 👍

vanitabarrett:
@leenagupte @andysellick This is now ready for re-review. Please note the scope of the PR has changed - it now only covers adding the tracking functionality to the component, not making use of it on topic pages. This will be covered in a separate PR.

app/helpers/highlight_boxes_helper.rb
```diff
@@ -0,0 +1,9 @@
+module HighlightBoxesHelper
+  def data_tracking?(items)
+    items.each do |item|
+      return true if item[:link][:data_attributes]
+    end
+
+    false
```
andysellick: Do you need false here? Presumably if there's no data attributes this method doesn't return anything.

leenagupte: @andysellick If the `each` block doesn't evaluate to anything, it returns the whole `items` array back to the call. This doesn't happen with `map`, because `map` creates a brand new array and `each` doesn't.

PR #654, opened 2018-05-14
Add tracking to topic pages

vanitabarrett:
Trello: https://trello.com/c/xYWhPOMg/74-add-tracking-to-topic-pages

Adds tracking to all highlight boxes, document lists and featured items on topic pages.

##  Highlight Boxes
Each link should have the following attributes:

- __data-track-category="servicesHighlightBoxClicked"__ (where services is the name of the supergroup section)
- __data-track-action="[position within section]"__ (starting at 1)
- __data-track-label="[anchor value]"__ (the page the user will go to once the anchor is clicked)

Output from GA Debugger:
screen shot 2018-05-14 at 10 27 20

## Document List
Each link should have the following attributes:

- __data-track-category="servicesDocumentListClicked"__ (where services is the name of the supergroup section)
- __data-track-action="[position within section]"__ (starting at 1)
- __data-track-label="[anchor value]"__ (the page the user will go to once the anchor is clicked)

Output from GA Debugger:
screen shot 2018-05-14 at 10 27 35

## News Featured Item
Same as document list, but with track category "newsAndCommunicationsFeaturedItemClicked"

Output from GA Debugger:
screen shot 2018-05-14 at 10 27 52

## Organisations
All organisations should be treated as one list, regardless of having logos or not, and regardless of whether hidden under a Show More or not.

Output from GA Debugger:
screen shot 2018-05-14 at 10 28 12

Component Guide: https://govuk-collections-pr-654.herokuapp.com/component-guide
Example Page: https://govuk-collections-pr-654.herokuapp.com/education

### DIFF
--- a/app/presenters/supergroups/supergroup.rb
+++ b/app/presenters/supergroups/supergroup.rb
@@ -59,10 +59,6 @@ module Supergroups
       []
     end
 
-    def promoted_content_count(*)
-      3
-    end
-
     def data_module_label
       name.camelize(:lower)
     end
@@ -75,6 +71,10 @@ module Supergroups
         track_action: index + 1,
         track_label: base_path
       }
+    end
+
+    def promoted_content_count(*)
+      3
     end
   end
 end
--- a/app/presenters/supergroups/supergroup.rb
+++ b/app/presenters/supergroups/supergroup.rb
@@ -37,7 +37,9 @@ module Supergroups
     end
 
     def document_list(taxon_id)
-      format_document_data(tagged_content(taxon_id))
+      items = tagged_content(taxon_id).drop(promoted_content_count)
+
+      format_document_data(items)
     end
 
     def promoted_content(*)
--- a/app/presenters/supergroups/supergroup.rb
+++ b/app/presenters/supergroups/supergroup.rb
@@ -37,11 +37,12 @@ module Supergroups
     end
 
     def document_list(taxon_id)
-      tagged_content(taxon_id).each.map do |document|
+      tagged_content(taxon_id).each_with_index.map do |document, index|
         data = {
           link: {
             text: document.title,
-            path: document.base_path
+            path: document.base_path,
+            data_attributes: data_attributes(document.base_path, index)
           },
           metadata: {
             public_updated_at: document.public_updated_at,
@@ -60,6 +61,16 @@ module Supergroups
 
     def data_module_label
       name.camelize(:lower)
+    end
+
+  private
+
+    def data_attributes(base_path, index)
+      {
+        track_category: data_module_label + "DocumentListClicked",
+        track_action: index + 1,
+        track_label: base_path
+      }
     end
   end
 end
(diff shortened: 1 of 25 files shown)


app/views/taxons/_organisation_logos_and_list.html.erb
```diff
@@ -12,7 +12,7 @@
 </div>
 
 <% if organisations_without_logos.any? %>
-  <div>
+  <div <%= "data-module=track-click" if track_click?(organisations_without_logos) %>>
```
andysellick: I wonder if this could be simplified - instead of checking each bit of the page and adding the data module attribute, could it be applied just once to a higher level element? On the education page example there's 71 instances of it.

app/views/taxons/_document_list_with_grid.html.erb
```diff
@@ -1,6 +1,6 @@
 <% if documents.any? %>
   <div class="grid-row">
-    <div class="column-two-thirds">
+    <div class="column-two-thirds" <%= "data-module=track-click" if track_click?(documents) %>>
```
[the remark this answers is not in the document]
vanitabarrett: Yes, we've added a card for it, but we didn't do the change within this piece of work as it will also mean updating all the other places where the document list is used https://trello.com/c/fDWv3Kle/117-add-track-click-wrapper-to-document-list-component

app/views/taxons/_organisation_logos_and_list.html.erb
vanitabarrett: I don't think that's linked to this line of code specifically? I think it's because we wanted the organisation logo component in static to be self-contained, e.g: contain the `data-module=track-click` itself so that it can be easily used by other people who need to add tracking without having to add that data-module themselves (like the problem we have with the document list).

However, this does mean that every time we use the logo, it comes with its own track-click... Hence the 71 track clicks.

I wasn't really happy with that solution but couldn't think of any alternative at the time. I guess we could have an argument passed to the component saying whether we're providing our own data-module wrapper or not?

app/presenters/supergroups/supergroup.rb
```diff
@@ -37,11 +37,40 @@ def tagged_content(_taxon_id)
...
+
+    def data_attributes(base_path, index)
+      {
+        track_category: data_module_label + "DocumentListClicked",
+        track_action: index,
+        track_label: base_path
+      }
+    end
+
+    def promoted_content_count(*)
```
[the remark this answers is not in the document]
vanitabarrett: It's overridden here with params https://github.com/alphagov/collections/pull/654/files/b0efc59bb633d0ccc62ed0297398ae60e385133d#diff-2d219904a10337ff4d4eb11663fb2db6R46

app/views/taxons/_organisation_logos_and_list.html.erb
vanitabarrett: Added a card for this: https://trello.com/c/VSi9cy7j/132-improve-track-click-functionality-for-organisation-logo-component

PR #656, opened 2018-05-14
Create index organisations page

MahmudH:
- New organisations index page created. 
- Data on page is displayed from the content store.

https://trello.com/c/F3Hx0P7T/28-create-organisations-index-page


app/views/organisations/index.html.erb
```diff
@@ -0,0 +1,19 @@
+<% content_for :title, "Departments, agencies and public bodies - GOV.UK" %>
+
+<h1>Departments, agencies and public bodies - GOV.UK</h1>
+<% @content_item["details"].each do |organisation_id, organisation| %>
```
sihugh: I think we might want to split this out so that we don't automatically iterate through the content item's details.  If someone adds something to the details of the content item, then this page will try and render it, which could be interesting!  

Also, we want to be overt about which order we display things in, which might not be the same order that things are sent from the content store.  Splitting the innards into a partial, but making separate calls for each type that we want to show, in the order that we want to show them should reduce our risk of doing something inadvertent.

app/models/content_store_organisation.rb
```diff
@@ -0,0 +1,61 @@
...
+  def organisations_list(department_type)
+    @content_item.content_item_data["details"][department_type]
+  end
+
+  def organisation_department_count(organisation)
+    organisation["works_with"].flatten.count
+  end
+
+  def get_name_of_organisation(organisation_id)
+    org_name = {
```
sihugh: We should move this declaration to a class level constant so that we don't instantiate a new hash each time we use this method.  We could think about a more descriptive name for it too.

app/models/content_store_organisation.rb
```diff
@@ -0,0 +1,99 @@
...
+
+  def organisations_list(department_type)
+    details[department_type]
+  end
+
+  def organisation_department_count(organisation)
+    organisation["works_with"].values.flatten.count
+  end
+
+  def get_name_of_organisation(organisation_type)
```
sihugh: You might want to check for things that aren't used anywhere any more - I think we've got a few bits to tidy up.

app/models/content_store_organisation.rb
```diff
@@ -0,0 +1,99 @@
...
+        "ordered_non_ministerial_departments" => "Non ministerial departments",
+        "ordered_agencies_and_other_public_bodies" => "Agencies and other public bodies",
+        "ordered_high_profile_groups" => "High profile groups",
+        "ordered_public_corporations" => "Public corporations",
+        "ordered_devolved_administrations" => "Devolved administrations"
+    }
+    org_name[organisation_type]
+  end
+
+  def get_name_of_department_category(category_type)
```
sihugh: Using `get_` is not particularly rubyish.  It reads better (in the view for example) if we use something like
```
@organisations.category_name(category_type)
```
We should also move the hash outside of this method and make it a constant - otherwise we instantiate it every time we call the method.

app/views/organisations/_organisations.html.erb
```diff
@@ -0,0 +1,35 @@
...
+                         category_type: category_type,
+                         departments: departments %>
+            <% end %>
+          </div>
+        </div>
+      <% end %>
+    </div>
+    <hr>
+  </div>
+</div>
```
sihugh: The red symbol here means that the file hasn't been finished with a new line which it should.  Most editors have a setting to ensure that this happens.  I've been got by this - see: https://github.com/alphagov/frontend/pull/907 ;-)

test/integration/content_store_organisations_test.rb
```diff
@@ -0,0 +1,72 @@
...
+
+  it "is possible to see a list of organisations" do
+    visit "/government/organisations"
+    assert page.has_content?(@content_item[:details][:ordered_ministerial_departments][0][:title])
+  end
+
+  it "displays Ministerial department to be one" do
+    visit "/government/organisations"
+    assert page.has_selector?(".department-count", text: "1")
+  end
```
sihugh: We should probably add a couple more tests now.  The "works_with" count seems important, so we should validate that.  Maybe also that the right "works_with" sections are shown.  Also, we could verify that things are shown in the right order, given that we've put a lot of work into it?

app/models/content_store_organisation.rb
MahmudH: 👍

app/models/content_store_organisation.rb
MahmudH: 👍

sihugh:
This is a good start and makes it easier for us to create the full page when we're ready to do so.

PR #658, opened 2018-05-16
Use breadcrumbs component from gem

tijmenb:
This has now moved from static to the gem:

https://trello.com/c/XLuzv0pB

### DIFF
--- a/app/assets/javascripts/browse-columns.js
+++ b/app/assets/javascripts/browse-columns.js
@@ -12,7 +12,7 @@
     this.$root = this.$el.find('#root');
     this.$section = this.$el.find('#section');
     this.$subsection = this.$el.find('#subsection');
-    this.$breadcrumbs = $('.govuk-breadcrumbs');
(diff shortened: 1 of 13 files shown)


app/assets/javascripts/browse-columns.js
```diff
@@ -337,7 +337,10 @@
       }
     },
     updateBreadcrumbs: function(state) {
-      var $newBreadcrumbs = $(state.sectionData.breadcrumbs);
+      // Calling `.last()` is necessary because the breadcrumbs component has 2
```
thomasleese: Will this still work if Slimmer moves the `<script>` to the bottom of the page?

tijmenb: Yeah, `state.sectionData` comes directly from the [json endpoint](https://github.com/alphagov/collections/blob/73ad7c021500b31881fbb0a6a51fba5607c303bb/app/controllers/browse_controller.rb#L22-L27) so slimmer doesn't interfere.

PR #667, opened 2018-05-31
Present data on organisation show view

MahmudH:
Render organisation content store data on show view
- There are two organisation show page templates.
- The view that is rendered depends on the govuk status given by the content store.
- The govuk status of exempt displays a link to a separate website, whereas the
status of live does not.

https://trello.com/c/k74yKHtG/107-present-data-on-organisations-show-view

### DIFF
--- a/app/helpers/organisation_helper.rb
+++ b/app/helpers/organisation_helper.rb
@@ -1,5 +1,16 @@
 module OrganisationHelper
-  def child_organisations_count(organisation)
-    organisation["works_with"].values.flatten.count
+  def child_organisation_count(organisation)
+    child_orgs = organisation.content_item.content_item_data["links"]["ordered_child_organisations"]
+    child_orgs.present? ? child_orgs.count : 0
+  end
+
+  def show_organisation(organisation)
+    if organisation.content_item.content_item_data["details"]["organisation_govuk_status"]["status"] == "live"
+      render partial: 'show_organisation',
+             locals: { organisation: @organisation }
+    else
+      render partial: 'separate_website',
+             locals: { organisation: @organisation }
+    end
   end
 end
--- a/app/models/organisation.rb
+++ b/app/models/organisation.rb
@@ -18,8 +18,60 @@ class Organisation
     @content_item.content_item_data["title"]
   end
 
+  def logo
+    details["logo"]
+  end
+
+  def base_path
+    @content_item.content_item_data["base_path"]
+  end
+
+  def ordered_featured_links
+    details["ordered_featured_links"]
+  end
+
+  def body
+    details["body"]
+  end
+
+  def ordered_featured_documents
+    details["ordered_featured_documents"]
+  end
+
+  def child_organisation_count
+    links["ordered_child_organisations"].count
+  end
+
+  def social_media_links
+    details["social_media_links"]
+  end
+
+  def organisation_type
+    details["organisation_type"].tr("_", " ")
+  end
+
+  def ordered_ministers
+    details["ordered_ministers"]
+  end
+
+  def board_members
+    details["ordered_board_members"]
+  end
+
+  def ordered_contacts
+    links["ordered_contacts"]
+  end
+
+  def ordered_corporate_information_pages
+    details["ordered_corporate_information_pages"]
+  end
+
   def ordered_parent_organisations
     links["ordered_parent_organisations"]
+  end
+
+  def ordered_featured_policies
+    links["ordered_featured_policies"]
   end
 
 private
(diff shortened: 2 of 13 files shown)


app/helpers/organisation_helper.rb
```diff
@@ -2,4 +2,20 @@ module OrganisationHelper
   def child_organisations_count(organisation)
     organisation["works_with"].values.flatten.count
   end
+
+  def child_organisation_count(organisation)
+    if organisation.content_item.content_item_data["links"]["ordered_child_organisations"].count != 0
```
Davidslv: you can use `.zero?` instead for example:

```ruby
unless organisation.content_item.content_item_data["links"]["ordered_child_organisations"].count.zero?
```

app/models/organisation.rb
```diff
@@ -18,8 +18,62 @@ def title
...
+    links["ordered_contacts"]
+  end
+
+  def ordered_corporate_information_pages
+    details["ordered_corporate_information_pages"]
+  end
+
   def ordered_parent_organisations
-    links["ordered_parent_organisations"]
+    if links["ordered_parent_organisations"].present?
```
Davidslv: you can leave the if statement as the result of this method will be nil when the key is not present.
Which is the same as doing `links["ordered_parent_organisations"]`

example:

```ruby
irb(main):001:0> p = {}
=> {}
irb(main):002:0> p["a"]
=> nil
```

app/helpers/organisation_helper.rb
MahmudH: I've had to do organisation.content_item.content_item_data["links"]["ordered_child_organisations"].presence as some organisations do not have this field in the content store to begin with so it is nil therefore doing a count throws an error

app/models/organisation.rb
MahmudH: i have removed the statement out of here and put it in the view instead as some orgs don't have a parent organisation

app/helpers/organisation_helper.rb
```diff
@@ -2,4 +2,32 @@ module OrganisationHelper
   def child_organisations_count(organisation)
     organisation["works_with"].values.flatten.count
   end
+
+  def child_organisation_count(organisation)
+    if organisation.content_item.content_item_data["links"]["ordered_child_organisations"].presence
```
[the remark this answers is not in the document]
MahmudH: 👍

app/helpers/organisation_helper.rb
```diff
@@ -2,4 +2,31 @@ module OrganisationHelper
...
+      render partial: 'show_organisation',
+             locals: { organisation: @organisation }
+    else
+      render partial: 'separate_website',
+             locals: { organisation: @organisation }
+    end
+  end
+
+  def ordered_featured_policies(organisation)
+    featured_policies = []
```
sihugh: This feels like it could be a bit simpler.  It seems to be just making a new array from another array, or default to an empty array if it doesn't exist.
It also seems very similar to a lot of things in the `organisation` model, so it feels like it might belong there? 
Perhaps:
```
def ordered_featured_policies(organisation)
  organisation.content_item.content_item_data["links"]["ordered_featured_policies"].presence || []
end
```
might work?  Check that you've got a test that covers it before changing anything.

app/helpers/organisation_helper.rb
```diff
@@ -2,4 +2,31 @@ module OrganisationHelper
   def child_organisations_count(organisation)
     organisation["works_with"].values.flatten.count
   end
+
+  def child_organisation_count(organisation)
```
sihugh: We've got this method, and a `child_organisation_count` method in the `organisation` model.  Do we need both of them?

app/helpers/organisation_helper.rb
```diff
@@ -2,4 +2,31 @@ module OrganisationHelper
   def child_organisations_count(organisation)
```
sihugh: I know this isn't part of this PR, but it has a very similar name to other `count` methods.  Maybe consider whether changing the name of it would help understand which one you should use?  What's the difference between `works_with` and `ordered_child_organisations`?

app/helpers/organisation_helper.rb
MahmudH: I have changed it and added tests for empty or present featured policies in the content store and it's passing 👍

app/helpers/organisation_helper.rb
MahmudH: The method in organisation model is removed now, instead organisation_helper handles this 👍

PR #675, opened 2018-06-06
Add components to organisation page header

andysellick:
Add components into the header section of the new organisation pages.

- add organisation logo
- add translation nav
- add featured links
- add no10 banner
- add a presenter for the header section
- add tests for each page type to ensure the right bits are being shown on them

There's some commented out code in this PR that isn't currently being used but is being kept for reference. Normally we should remove this but as these pages aren't live even if deployed it's handy to keep it for reference for the moment - commented sections will be removed in a following PR.

Some example pages:

- [with translation nav](https://govuk-collections-pr-675.herokuapp.com/government/organisations/office-of-the-secretary-of-state-for-wales)
- [with large featured links](https://govuk-collections-pr-675.herokuapp.com/government/organisations/charity-commission)
- [with normal featured links](https://govuk-collections-pr-675.herokuapp.com/government/organisations/attorney-generals-office)
- [with no10 banner](https://govuk-collections-pr-675.herokuapp.com/government/organisations/prime-ministers-office-10-downing-street)

Trello card: https://trello.com/c/v7Izogff/137-add-organisation-header-to-organisation-page


vanitabarrett:
The large featured links are smaller than they previously were in Whitehall. This may not be a problem, but may be worth noting in the Bugs and Issues board so we can keep track of these differences.

**Current:**
screen shot 2018-06-08 at 15 29 49

**New:**
screen shot 2018-06-08 at 15 29 45

vanitabarrett:
Also one of your commit messages says "- TO BE CHANGED: comment out all the code below as it's going to change but need it for reference" 🤔(I know this is WIP, but worth noting case you just forgot to remove it)

app/presenters/organisations/header_presenter.rb
```diff
@@ -0,0 +1,88 @@
+module Organisations
+  class HeaderPresenter
+    attr_reader :org
+
+    def initialize(organisation)
+      @org = organisation
```
vanitabarrett: Nitpick: `@org` can just be `@organisation`.

app/presenters/organisations/header_presenter.rb
```diff
@@ -0,0 +1,88 @@
...
+      end
+    end
+
+    def is_no_10?
+      true if base == "/government/organisations/prime-ministers-office-10-downing-street"
+    end
+
+  private
+
+    def base
```
vanitabarrett: To be clear this should probably be `base_path` too

app/views/organisations/_separate_website.html.erb
```diff
@@ -1,25 +1,26 @@
...
+      organisation: @header.organisation_logo
+    } %>
+  </div>
 </div>
 
-<div class="organisation-body">
-  <%= organisation.body %>
-</div>
+<%= render "govuk_publishing_components/components/notice", {
+  title: "#{organisation.title} has a <a href='/'>separate website</a>".html_safe
```
vanitabarrett: I wonder if this can be translated.

app/views/organisations/_show_organisation.html.erb
```diff
@@ -1,9 +1,33 @@
...
+  <h1 class="visually-hidden"><%= @header.org.title %></h1>
+  <div class="grid-row">
+    <div class="<%= @header.logo_wrapper_class %>">
+      <%= render "govuk_publishing_components/components/organisation_logo", {
+        organisation: @header.organisation_logo
+      } %>
+    </div>
+
+    <div class="<%= @header.link_wrapper_class %>">
+      <%= render "govuk_publishing_components/components/translation-nav", @header.translation_links %>
```
vanitabarrett: On mobile, the no-margin-top means that there's not enough spacing between this and the logo:

screen shot 2018-06-08 at 16 33 01

vanitabarrett: Same with links and organisation logo: 

screen shot 2018-06-08 at 16 37 04

app/views/organisations/_show_organisation.html.erb
```diff
@@ -1,9 +1,33 @@
...
+      <%= render "components/topic-list", @header.ordered_featured_links %>
```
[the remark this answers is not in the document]
andysellick: Good spot, thanks, hadn't even noticed. This is why we need a mechanism for comparing the old and new pages! I've added this but still need to figure out how to do the acronym for orgs like DEFRA.

app/assets/stylesheets/views/_organisations.scss
```diff
@@ -20,3 +20,12 @@
     }
   }
 }
+
+.no10-banner {
+  @include bold-19;
+  padding-top: 140px;
+  background-image: image-url('prime-ministers-office-header.jpg');
+  background-repeat: no-repeat;
+  background-position: 50% 0;
```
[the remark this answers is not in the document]
andysellick: Yes, it keeps the 'no10' badge in view in the middle if the screen is narrow.

If we had time I'd be tempted to do this properly i.e. with a mobile image and resizing etc. but this is a straight copy from the existing site to save time (also it means it behaves the same as the current one, and I think fewer design changes is better).

app/views/organisations/_show_organisation.html.erb
andysellick: Good spot. Annoyingly I'm going to have to put in some glue code for many of the page sections since the margin bottom of the components is inconsistent - the logo has no margin at all. Hopefully I can just do a generic class and extend it, will add.

app/views/organisations/_separate_website.html.erb
andysellick: Probably. I think I'll leave this for now as I started work on this section before realising it was out of scope for this PR.

app/presenters/organisations/header_presenter.rb
andysellick: I'm trying to keep it terse. I was originally using organisation and the code was beginning to look pretty dense.

app/assets/stylesheets/views/_organisations.scss
```diff
@@ -20,3 +20,20 @@
     }
   }
 }
+
+.margin-bottom {
```
vanitabarrett: This is a really vague class name. I know we don't tend to do BEM in our views, but we should probably at least scope it to be inside `.organisations` or rename it to `organisations__margin-bottom`. It feels like this is being used more as a mixin too - perhaps that would make it neater?

app/presenters/organisations/header_presenter.rb
```diff
@@ -0,0 +1,118 @@
...
+
+      component_data
+    end
+
+    def ordered_featured_links
+      links = []
+
+      if has_services_and_information_link?
+        see_more_link = {
+          text: "All #{org.title} services and information", # need to use acronym where available
```
vanitabarrett: It's not clear this is something that still needs to be implemented. Can we add ` TODO: ` to make it clearer/reference a Trello card?

app/presenters/organisations/header_presenter.rb
```diff
@@ -0,0 +1,118 @@
...
+      true if org.organisation_type == "executive_office"
+    end
+
+  private
+
+    def slug
+      org.base_path.gsub("/government/organisations/", "")
+    end
+
+    def small_header_links?
```
vanitabarrett: I feel like this method may be mis-named. I know it's being used to decide what layout we use for the header links, but it actually just returns whether something is news or not. Technically it could be used somewhere else for a different purpose later on. So it may be worth renaming it to make this clearer?

app/presenters/organisations/header_presenter.rb
```diff
@@ -0,0 +1,118 @@
...
+      true if org.organisation_featuring_priority == "news"
+    end
+
+    def has_services_and_information_link?
```
vanitabarrett: Can this method and `organisations_with_services_and_information_link` be rolled into one?

app/views/organisations/_show_organisation.html.erb
```diff
@@ -1,9 +1,33 @@
...
-<%= render 'featured_links',
-           featured_links: organisation.ordered_featured_links %>
+    <div class="<%= @header.link_wrapper_class %> margin-bottom">
+      <%= render "govuk_publishing_components/components/translation-nav", @header.translation_links %>
+      <%= render "components/topic-list", @header.ordered_featured_links %>
+    </div>
+  </div>
+<% end %>
+
+<% if false %>
```
vanitabarrett: I would still be tempted to remove this and the code beneath as we can always refer back to the Git history for this, but up to you 🙂

app/views/taxons/_organisation_logos_and_list.html.erb
```diff
@@ -2,7 +2,7 @@
   <ol>
     <% organisations_with_logos.each do |promoted_organisation| %>
       <li class="taxon-page__organisation">
-        <%= render 'govuk_component/organisation_logo', {
+        <%= render 'govuk_publishing_components/components/organisation_logo', {
```
vanitabarrett: This should perhaps be part of the first commit?

test/integration/content_store_organisation_test.rb
```diff
@@ -161,6 +151,6 @@ class OrganisationTest < ActionDispatch::IntegrationTest
     }
     content_store_has_item('/government/organisations/prime-ministers-office-10-downing-street', content_item_with_empty_featured_policies)
     visit "/government/organisations/prime-ministers-office-10-downing-street"
-    assert page.has_css?(".organisation-policies")
+    #assert page.has_css?(".organisation-policies")
```
vanitabarrett: I would probably comment out the whole test so it's more obvious that we're not running this. Or marking it as to skip so we don't run it at all and it flags up as 'pending' in the test run? 

May also be worth noting this in https://trello.com/c/Qw7Y6CKE/150-add-our-policies-list-to-what-we-do-section so the person who works on that card definitely knows this test is available to them.

app/views/organisations/_separate_website.html.erb
vanitabarrett: It feels a bit weird to commit/merge half-finished work - can we cherrypick this to a new in progress branch and remove it from this one?

app/presenters/organisations/header_presenter.rb
andysellick: I'll rename it later if it gets used for something more generic, but for now I wanted a method that clearly explained what it's for - organisations have too many things in them already that don't make sense unless you know lots about organisations.

app/views/organisations/_show_organisation.html.erb
andysellick: I intend to address this in the very next card, the news section.

app/views/taxons/_organisation_logos_and_list.html.erb
andysellick: Not sure I agree, not sure I have time to address anyway.

test/integration/content_store_organisation_test.rb
andysellick: I'll take it out completely - it's something we should cover in future work.

app/views/taxons/_organisation_logos_and_list.html.erb
vanitabarrett: This change is related to the taxons view rather than organisation pages, which is why I don't think it belongs in this commit. Happy to help if you decide you want to split this out.

andysellick: Done.

PR #676, opened 2018-06-06
Add 'works with' text to organisation list

vanitabarrett:
Trello: https://trello.com/c/DUFbZjjM/138-add-works-with-info-to-organisation-list-page

Adds "Works with 1 public body" or "Works with [x] agencies and public bodies" to organisations that have the appropriate data. This follows the same logic as is currently in Whitehall, i.e:

- Use 'public body' if organisation only works with one other
- Use 'agencies and public bodies' if organisation works with more than one

## With one organisation:
screen shot 2018-06-06 at 17 32 08

## With multiple organisations:
screen shot 2018-06-06 at 17 32 21

Review app: https://govuk-collections-pr-676.herokuapp.com/government/organisations


vanitabarrett:
@andysellick Note: I'll address moving things out of the document list component in a separate PR I think, when I pick up the view toggle card.

app/presenters/organisations_presenter.rb
```diff
@@ -1,4 +1,6 @@
 class OrganisationsPresenter
```
andysellick: You might want to move this file to `app/presenters/organisations/index_presenter.rb` to fit with the changes I'm making. Having all the presenters in an organisations directory means we don't have to keep putting 'organisations' at the start of all the filenames.

test/presenters/organisations_presenter_test.rb
```diff
@@ -120,7 +125,8 @@
           link: {
             text: "The Charity Commission",
             path: "/government/organisations/charity-commission",
-            context: nil
+            context: nil,
+            description: nil
```
andysellick: Do we need a test where both context and description aren't nil?

vanitabarrett: I was going to add a test, but I just realised I'm deleting/modifying all this code in the next PR anyway so it's probably ok to leave as is until then

app/presenters/organisations_presenter.rb
vanitabarrett: Good point. Done 👍

PR #680, opened 2018-06-08
Add organisations "Works With" view all toggle

vanitabarrett:
Implements the View All toggle that currently exists on the Whitehall /government/organisations page.

screen shot 2018-06-08 at 10 55 43
screen shot 2018-06-08 at 13 39 08

Note: The first commit for this PR is a cleanup commit to stop using the document list component on the organisation list page. Using this component was putting limitations in place when it came to building the View All toggle and, as it's only a simple list, we decided it would be better to stop using the component and add the styling into `organisations.scss` itself.

Review App: https://govuk-collections-pr-680.herokuapp.com/government/organisations

### DIFF
--- a/app/assets/stylesheets/views/_organisations.scss
+++ b/app/assets/stylesheets/views/_organisations.scss
@@ -1,24 +1,33 @@
 .organisations {
-  .organisations__section {
+  .organisations-list__section {
     margin-bottom: $gutter * 2;
 
     .department-count {
       @include bold-80;
     }
 
-    .organisations__logo-wrapper {
+    .organisations-list__item {
       list-style-type: none;
-      padding-bottom: $gutter-two-thirds;
       margin-bottom: $gutter-two-thirds;
       border-bottom: 1px solid $grey-2;
 
-      .organisation__works-with {
+      .organisation-list__item-title {
+        @include core-19;
+        margin-right: $gutter-one-third;
+      }
+
+      .organisation-list__item-context {
+        color: $grey-1;
+      }
+
+      .organisation-list__item-works-with {
         @include core-16;
-        margin-top: $gutter-half;
+        margin-top: $gutter-one-third;
+        margin-bottom: $gutter-one-third;
       }
     }
 
-    .organisations__list-without-number {
+    .organisations-list__without-number {
       @include media(tablet) {
         float: right;
       }
--- a/app/assets/stylesheets/views/_organisations.scss
+++ b/app/assets/stylesheets/views/_organisations.scss
@@ -1,36 +1,66 @@
 .organisations {
   .organisations-list__section {
     margin-bottom: $gutter * 2;
+  }
 
-    .department-count {
-      @include bold-80;
+  .department-count {
+    @include bold-80;
+  }
+
+  .organisations-list__item {
+    list-style-type: none;
+    padding-bottom: $gutter-half;
+    margin-bottom: $gutter-half;
+    border-bottom: 1px solid $grey-2;
+  }
+
+  .organisation-list__item-title {
+    @include core-19;
+    margin-right: $gutter-one-third;
+  }
+
+  .organisation-list__item-context {
+    color: $grey-1;
+    white-space: nowrap;
+  }
+
+  .organisation-list__item-works-with {
+    @include core-16;
+    display: block;
+    margin-top: $gutter-half;
+    margin-right: $gutter-two-thirds;
+
+    @include media(tablet) {
+      display: inline-block;
     }
+  }
 
-    .organisations-list__item {
-      list-style-type: none;
-      margin-bottom: $gutter-two-thirds;
-      border-bottom: 1px solid $grey-2;
+  .organisation-list__works-with-toggle {
+    @include core-16;
+  }
 
-      .organisation-list__item-title {
-        @include core-19;
-        margin-right: $gutter-one-third;
-      }
+  .organisation-list__works-with {
+    background-color: $panel-colour;
+    margin-top: $gutter-one-third;
+    padding: $gutter-one-third;
+  }
 
-      .organisation-list__item-context {
-        color: $grey-1;
-      }
+  .organisation-list__works-with-title {
+    font-size: 16px;
+    font-weight: bold;
+  }
 
-      .organisation-list__item-works-with {
-        @include core-16;
-        margin-top: $gutter-one-third;
-        margin-bottom: $gutter-one-third;
-      }
-    }
+  .organisation-list__works-with-list {
+    margin-bottom: $gutter-half;
+  }
 
-    .organisations-list__without-number {
-      @include media(tablet) {
-        float: right;
-      }
+  .organisation-list__works-with-list-item {
+    padding-bottom: 5px;
+  }
+
+  .organisations-list__without-number {
+    @include media(tablet) {
+      float: right;
     }
   }
 }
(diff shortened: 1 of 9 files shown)


app/views/organisations/_organisations_list.html.erb
```diff
@@ -1,45 +1,56 @@
...
+              <%= render "govuk_component/organisation_logo", {
+                organisation: {
+                  name: organisation["logo"]["formatted_title"],
+                  url: organisation["href"],
+                  brand: organisation["brand"],
+                  crest: organisation["logo"]["crest"]
+                }
+              } %>
+            <% else %>
+              <div>
```
andysellick: This div isn't necessary?

app/assets/stylesheets/views/_organisations.scss
```diff
@@ -1,24 +1,50 @@
...
-  .organisations__section {
+  .organisations-list__section {
     margin-bottom: $gutter * 2;
 
     .department-count {
       @include bold-80;
     }
 
-    .organisations__logo-wrapper {
+    .organisations-list__item {
```
andysellick: Since you're using BEM, we probably don't need to nest all this SASS, would make it a bit more readable.

app/presenters/organisations/index_presenter.rb
```diff
@@ -40,24 +37,21 @@ def works_with(organisation)
...
-              description: works_with(organisation)
-            }
-          }
+    def link_for_non_ministerial_organisation(organisation)
+      link_title_class = "organisation-list__item-title"
 
-          data
-        end
+      if works_with_statement(organisation)
+        tag.h3(link_to(organisation["title"], organisation["href"]), class: link_title_class)
```
andysellick: This is causing some inconsistency in the output of headings:

screen shot 2018-06-08 at 14 10 56

'Food standards Agency' is a heading but 'Crown Prosecution Service' isn't. That's a bit confusing.

app/assets/stylesheets/views/_organisations.scss
```diff
@@ -1,24 +1,50 @@
...
-      padding-bottom: $gutter-two-thirds;
-      margin-bottom: $gutter-two-thirds;
+      padding-bottom: $gutter-half;
+      margin-bottom: $gutter-half;
       border-bottom: 1px solid $grey-2;
 
-      .organisation__works-with {
+      .organisation-list__item-title {
+        @include core-19;
+        display: inline-block;
```
andysellick: Where this is a heading tag, this display attribute is necessary. Where it's a link, it isn't.

app/assets/stylesheets/views/_organisations.scss
```diff
@@ -1,24 +1,50 @@
...
+        margin-top: $gutter-one-third;
+        margin-right: $gutter-two-thirds;
+      }
+
+      .organisation-list__works-with {
+        background-color: $panel-colour;
+        margin-top: $gutter-one-third;
+        padding: $gutter-one-third;
+
```
andysellick: Might be worth adding a little bit of spacing between the list items themselves, I'm a bit worried about the links being too close together, particularly on mobile.

app/assets/stylesheets/views/_organisations.scss
```diff
@@ -1,24 +1,50 @@
...
+        .organisation-list__works-with-title {
```
andysellick: On mobile this element appears smaller than the text it's a heading for (14px compared to 16px). Also would suggest a little spacing beneath it, it looks a little squashed.

screen shot 2018-06-08 at 14 17 00

app/assets/stylesheets/views/_organisations.scss
andysellick: Maybe outside the scope of this PR, but something we might want to address is that the clickable area of the logo extends all the way to the right. I've accidentally clicked on it a few times trying to open the 'view all' link. Maybe if it was inline-block? Would be a change to the component, I think.

screen shot 2018-06-08 at 14 18 44

app/presenters/organisations/index_presenter.rb
```diff
@@ -40,24 +37,21 @@ def works_with(organisation)
       end
     end
 
-  private
+    def executive_office?(organisation_type)
+      organisation_type.eql?(:number_10)
```
andysellick: It's possible that some of the code here is going to duplicate some of the code for the org pages themselves, I wonder if there's a good way around that or if a little duplication is to be expected?

app/views/organisations/_expanded_works_with_organisations.html.erb
```diff
@@ -0,0 +1,17 @@
+<a href="#" data-controls="toggle_<%= current_organisation.parameterize %>" data-expanded="false" data-toggled-text="<%= t('view_less') %>">
+  <%= t('view_all') %>
+</a>
+
+<div class="organisation-list__works-with js-hidden" id="toggle_<%= current_organisation.parameterize %>">
+  <% works_with_organisations.each do |type, work_with_organisations| %>
+    <h4 class="organisation-list__works-with-title"><%=  t('organisations.type.' + type) %></h4>
+    <ul class="organisation-list__works-with-list">
+      <% work_with_organisations.each do |work_with_organisation| %>
+      <li>
```
andysellick: Indent?

app/views/organisations/_organisations_list.html.erb
```diff
@@ -1,45 +1,56 @@
...
+          heading_level: 2
+        } %>
+        <div class="department-count">
+          <p class="visuallyhidden">There are <%= organisations.count %> <%= organisation_type.to_s.humanize %></p>
+          <span aria-hidden="true"><%= organisations.count %></span>
         </div>
+      <% end %>
+    </div>
+
+    <div class="column-two-thirds <%= " organisations-list__without-number" if @presented_organisations.executive_office?(organisation_type) %>">
```
andysellick: Not sure you need the space in `" organisations-list__without-number"`

app/views/organisations/_expanded_works_with_organisations.html.erb
```diff
@@ -0,0 +1,17 @@
+<a href="#" data-controls="toggle_<%= current_organisation.parameterize %>" data-expanded="false" data-toggled-text="<%= t('view_less') %>">
```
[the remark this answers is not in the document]
vanitabarrett: Good catch

app/presenters/organisations/index_presenter.rb
vanitabarrett: This was intentional, but perhaps this is more confusing than I intended. 

When this was in the document list component, anything with a description (i.e: "works with X public bodies") was a heading, anything else was just a link because it wasn't acting as a heading to any further content. I agree this might mean people navigating by headings miss content though/get spammed with too much content. Not sure of the best solution? Just make them all links rather than headings?

app/presenters/organisations/index_presenter.rb
vanitabarrett: I think for now we should leave this where it is and address any potential duplication on the org pages as it arises

app/presenters/organisations/index_presenter.rb
andysellick: I think so, yes. Most of them don't even count as headings, in that there's no content beneath them. Good to be consistent.

app/views/organisations/_expanded_works_with_organisations.html.erb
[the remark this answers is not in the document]
andysellick: I'd have gone with just putting it onto the next line by default for mobile, in order to avoid an inconsistent layout, as shown:

screen shot 2018-06-11 at 11 45 24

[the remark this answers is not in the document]
vanitabarrett: @andysellick Good point, I'll change that

PR #690, opened 2018-06-13
Add latest news and what we do section to new org pages

andysellick:
This PR adds the 'latest news', 'what we do' and 'our policies' sections to the new organisation pages.

screen shot 2018-06-14 at 15 13 57

The code is currently not returning data for the 'latest' section from the correct source, but this will be addressed in a later PR.

URLs to test with:
- [org page with all three sections](https://govuk-collections-pr-690.herokuapp.com/government/organisations/office-of-the-secretary-of-state-for-wales)
- [org page with just the first two](https://govuk-collections-pr-690.herokuapp.com/government/organisations/attorney-generals-office)

Trello cards:
- https://trello.com/c/BdGlYIwl/175-add-latest-news-and-communication-section-to-organisation-page
- https://trello.com/c/vlrxzeFp/174-add-what-we-do-section-to-organisation-page
- https://trello.com/c/Qw7Y6CKE/177-add-our-policies-list-to-what-we-do-section

### DIFF
--- a/app/presenters/organisations/show_presenter.rb
+++ b/app/presenters/organisations/show_presenter.rb
@@ -11,8 +11,12 @@ module Organisations
       (prefix + org.title)
     end
 
-    def latest
-      stories = []
+    def has_latest_documents?
+      org.ordered_featured_documents.length.positive?
+    end
+
+    def latest_documents
+      documents = []
 
       org.ordered_featured_documents.each do |story|
         metadata = {}
@@ -22,7 +26,7 @@ module Organisations
           metadata[:public_updated_at] = Date.parse(story["public_updated_at"])
         end
 
-        stories << {
+        documents << {
           link: {
             text: story["title"],
             path: story["href"]
@@ -32,15 +36,15 @@ module Organisations
       end
 
       {
-        items: stories,
+        items: documents,
         brand: org.brand
       }
     end
 
     def subscription_links
       {
-        email_signup_link: "TODO",
-        feed_link_box_value: "TODO",
+        email_signup_link: "/government/email-signup/new?email_signup[feed]=#{org.web_url}.atom",
+        feed_link_box_value: "#{org.web_url}.atom",
         brand: org.brand
       }
     end
--- a/app/presenters/organisations/show_presenter.rb
+++ b/app/presenters/organisations/show_presenter.rb
@@ -1,0 +1,59 @@
+module Organisations
+  class ShowPresenter
+    attr_reader :org
+
+    def initialize(organisation)
+      @org = organisation
+    end
+
+    def prefixed_title
+      prefix = needs_definite_article?(org.title) ? "the " : ""
+      (prefix + org.title)
+    end
+
+    def latest
+      stories = []
+
+      org.ordered_featured_documents.each do |story|
+        metadata = {}
+        metadata[:document_type] = story["document_type"]
+
+        if story["public_updated_at"]
+          metadata[:public_updated_at] = Date.parse(story["public_updated_at"])
+        end
+
+        stories << {
+          link: {
+            text: story["title"],
+            path: story["href"]
+          },
+          metadata: metadata
+        }
+      end
+
+      {
+        items: stories,
+        brand: org.brand
+      }
+    end
+
+    def subscription_links
+      {
+        email_signup_link: "TODO",
+        feed_link_box_value: "TODO",
+        brand: org.brand
+      }
+    end
+
+  private
+
+    def needs_definite_article?(phrase)
+      exceptions = [/civil service resourcing/, /^hm/, /ordnance survey/]
+      !has_definite_article?(phrase) && !(exceptions.any? { |e| e =~ phrase.downcase })
+    end
+
+    def has_definite_article?(phrase)
+      phrase.downcase.strip[0..2] == 'the'
+    end
+  end
+end
--- a/app/presenters/organisations/show_presenter.rb
+++ b/app/presenters/organisations/show_presenter.rb
@@ -1,24 +1,22 @@
 module Organisations
   class ShowPresenter
-    attr_reader :org
-
     def initialize(organisation)
       @org = organisation
     end
 
     def prefixed_title
-      prefix = needs_definite_article?(org.title) ? "the " : ""
-      (prefix + org.title)
+      prefix = needs_definite_article?(@org.title) ? "the " : ""
+      (prefix + @org.title)
     end
 
     def has_latest_documents?
-      org.ordered_featured_documents.length.positive?
+      @org.ordered_featured_documents.length.positive?
     end
 
     def latest_documents
       documents = []
 
-      org.ordered_featured_documents.each do |story|
+      @org.ordered_featured_documents.each do |story|
         metadata = {}
         metadata[:document_type] = story["document_type"]
 
@@ -37,15 +35,44 @@ module Organisations
 
       {
         items: documents,
-        brand: org.brand
+        brand: @org.brand
       }
     end
 
     def subscription_links
       {
-        email_signup_link: "/government/email-signup/new?email_signup[feed]=#{org.web_url}.atom",
-        feed_link_box_value: "#{org.web_url}.atom",
-        brand: org.brand
+        email_signup_link: "/government/email-signup/new?email_signup[feed]=#{@org.web_url}.atom",
+        feed_link_box_value: "#{@org.web_url}.atom",
+        brand: @org.brand
+      }
+    end
+
+    def has_featured_policies?
+      true if @org.ordered_featured_policies && @org.ordered_featured_policies.length.positive?
+    end
+
+    def featured_policies
+      policies = []
+
+      @org.ordered_featured_policies.each do |policy|
+        policies << {
+          link: {
+            text: policy["title"],
+            path: policy["base_path"]
+          }
+        }
+      end
+
+      policies << {
+        link: {
+          text: "See all our policies",
+          path: "/government/policies?organisations[]=#{@org.slug}"
+        }
+      }
+
+      {
+        items: policies,
+        brand: @org.brand
       }
     end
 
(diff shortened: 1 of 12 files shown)


app/presenters/organisations/show_presenter.rb
```diff
@@ -0,0 +1,59 @@
...
+    def initialize(organisation)
+      @org = organisation
+    end
+
+    def prefixed_title
+      prefix = needs_definite_article?(org.title) ? "the " : ""
+      (prefix + org.title)
+    end
+
+    def latest
```
vanitabarrett: Would it make sense to have a 'News' or 'Documents' presenter, which this could then live in along with the featured news stuff. As we're splitting things into different presenters according to use, I'm not sure it makes sense for this to live here.

vanitabarrett: If you do choose to keep it here though, it should have a name like `latest_documents` so it's clear what this actually does.

app/presenters/organisations/show_presenter.rb
```diff
@@ -0,0 +1,59 @@
...
+      stories = []
```
vanitabarrett: Not sure this is a good variable name when the section displays the content it's meant to (i.e: not just news stories). Maybe `documents`. Same for here: https://github.com/alphagov/collections/pull/690/files#diff-ce57b95cdb3de98ec1f5711ba7dc9f24R17

On a related note, might be worth updating the Trello and PR to reflect that this section is not related to supergroups any more.

app/presenters/organisations/show_presenter.rb
```diff
@@ -0,0 +1,59 @@
...
+
+      org.ordered_featured_documents.each do |story|
```
vanitabarrett: I think this could be done more simply using `.map`. We have an example of formatting data for use with the document_list component in another place in Collections: https://github.com/alphagov/collections/blob/master/app/presenters/supergroups/supergroup.rb#L67

Would be something like:

```
      stories = org.ordered_featured_documents.each.map do |story|
        data = {
          link: {
            text: story["title"],
            path: story["href"]
          },
          metadata: {
            document_type: story["document_type"]
            public_updated_at: (Date.parse(story["public_updated_at"]) if story["public_updated_at"]),
          }
        }

        data
      end
```

app/views/organisations/_show_organisation.html.erb
```diff
@@ -49,16 +49,34 @@
   <% end %>
 </div>
 
+<% if @organisation.ordered_featured_documents.length %>
```
vanitabarrett: Should this check `latest` rather than ordered_featured_documents? We'll definitely need to make that change when we get the correct document data to show.

vanitabarrett: Also I'm not sure this works. It would need to be something like:

`if @organisation.ordered_featured_documents.length > 0`
or
`if @organisation.ordered_featured_documents.any?`

At the moment, even if length is 0 the section will still display.

app/views/organisations/_show_organisation.html.erb
```diff
@@ -49,16 +49,34 @@
...
+  <%= render "govuk_publishing_components/components/heading", {
```
vanitabarrett: This will also need to go in a grid column-two-thirds.

app/views/organisations/_show_organisation.html.erb
andysellick: Fixed.

app/views/organisations/_show_organisation.html.erb
andysellick: Fixed.

app/presenters/organisations/what_we_do_presenter.rb
```diff
@@ -0,0 +1,43 @@
...
+      end
+
+      {
+        stacked: true,
+        brand: org.brand,
+        links: links
+      }
+    end
+
+    def what_we_do_title
```
vanitabarrett: Think this can just be called 'title', as when we reference it in  the view it's already prefixed with: `@what_we_do`

test/integration/organisation_test.rb
```diff
@@ -178,10 +221,32 @@ class OrganisationHeaderTest < ActionDispatch::IntegrationTest
       }
     }
 
+    @content_item_blank = {
```
vanitabarrett: Not sure this belongs in this commit, especially as other tests are added separately.

app/presenters/organisations/show_presenter.rb
```diff
@@ -0,0 +1,59 @@
+module Organisations
+  class ShowPresenter
```
vanitabarrett: There is a lot of logic in this presenter, I think it's worth writing some tests for.

app/views/organisations/_show_organisation.html.erb
```diff
@@ -79,12 +79,16 @@
   </div>
 
   <div class="grid-row">
-    <div class="column-two-thirds">
+    <div class="column-two-thirds brand--<%= @organisation.brand %>">
```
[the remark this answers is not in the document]
andysellick: It won't do it for some reason.

app/views/organisations/_show_organisation.html.erb
```diff
@@ -46,7 +46,7 @@
     <div class="grid-row">
       <div class="column-two-thirds">
         <%= render "govuk_publishing_components/components/heading", {
-          text: "Latest from #{@show.prefixed_title}",
+          text: t('organisations.latest_from') + @show.prefixed_title,
```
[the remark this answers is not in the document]
andysellick: Ah, cool. Thanks.

test/integration/organisation_test.rb
```diff
@@ -303,13 +312,35 @@ class OrganisationHeaderTest < ActionDispatch::IntegrationTest
...
   it "shows a large news item only on news organisations" do
     visit "/government/organisations/attorney-generals-office"
     assert page.has_css?(".gem-c-image-card.gem-c-image-card--large")
 
     visit "/government/organisations/charity-commission"
+    assert page.has_css?(".gem-c-image-card")
     refute page.has_css?(".gem-c-image-card.gem-c-image-card--large")
+
+    visit "/government/organisations/civil-service-resourcing"
+    refute page.has_css?(".gem-c-image-card")
```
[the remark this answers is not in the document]
andysellick: Good point.

app/presenters/organisations/what_we_do_presenter.rb
andysellick: Actually this is now duplicate code as it's been put into the much more generic show_presenter, so I'll remove it.

test/integration/organisation_test.rb
andysellick: Which commit is this in?

test/integration/organisation_test.rb
```diff
@@ -311,6 +311,7 @@ class OrganisationTest < ActionDispatch::IntegrationTest
 
   it "shows the no10 banner element only on no10's page" do
     visit "/government/organisations/prime-ministers-office-10-downing-street"
+    puts(page.html)
```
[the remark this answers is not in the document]
andysellick: D'oh

app/presenters/organisations/show_presenter.rb
```diff
@@ -0,0 +1,59 @@
...
+    attr_reader :org
```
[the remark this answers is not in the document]
andysellick: Yes, I think so. It gets called in prefixed_title - `prefix = needs_definite_article?(org.title) ? "the " : ""`

app/views/organisations/_show_organisation.html.erb
```diff
@@ -23,32 +23,104 @@
...
-      <% news.each do |news_item| %>
-        <div class="column-one-third">
-          <%= render "govuk_publishing_components/components/image_card", news_item %>
-        </div>
+  <div class="grid-row">
+    <div class="column-two-thirds brand--<%= @organisation.brand %>">
+      <%= render "govuk_publishing_components/components/govspeak", {
+        } do %>
+          <p><%= @organisation.body.html_safe %></p>
+          <p>TODO: add the rest</p>
```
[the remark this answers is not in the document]
andysellick: Unsure, but I'll take it out and add a note to the card.

app/presenters/organisations/show_presenter.rb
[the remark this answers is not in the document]
vanitabarrett: You should be able to replace usages of org with @org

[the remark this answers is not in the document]
andysellick: I know, but there's loads of them in the code and it makes it slightly cleaner.

test/presenters/organisations/show_presenter_test.rb
```diff
@@ -0,0 +1,48 @@
...
+  it 'adds a prefix to a title when it should' do
+    content_item = ContentItem.new({ title: "Attorney General's Office" }.with_indifferent_access)
+    organisation = Organisation.new(content_item)
+    @show_presenter = Organisations::ShowPresenter.new(organisation)
+
+    expected = "the Attorney General's Office"
+    assert_equal expected, @show_presenter.prefixed_title
+  end
+
+  it 'does not add a prefix to a title when it should not' do
```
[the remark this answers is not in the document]
andysellick: Added.

app/presenters/organisations/show_presenter.rb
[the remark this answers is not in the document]
vanitabarrett: It's a bit misleading though because it creates a getter method which could be called outside the class, which we don't need. Someone unfamiliar with the code may think this is an indication of it being needed elsewhere.

You could make it private, e.g:

```
private
    attr_reader :org

public
    #rest of presenter code goes here


```

app/presenters/organisations/show_presenter.rb
andysellick: Added. It'll need expanding and modifying over time, but it's started.

PR #692, opened 2018-06-13
Hide toggle button when javascript disabled

vanitabarrett:
When Javascript is disabled, we show the "Works With" organisation info by default.
This makes the 'view all' toggle redundant. We can therefore hide unless JS is enabled.

**Before (with JS disabled):**
screen shot 2018-06-13 at 13 22 50

**After (with JS disabled:**
screen shot 2018-06-13 at 13 23 16


vanitabarrett:
@andysellick Good point, I'll change that now

PR #693, opened 2018-06-13
Add people and role pages

rubenarakelyan:
This commit adds pages for people and roles. The views are barebones since they will be filled in with components.

Trello: https://trello.com/c/CeLJAdEq/182-add-people-and-role-pages-to-collections


rubenarakelyan:
Yeah, I mean they are closely related to organisations but either works. Is there a preference?

PR #694, opened 2018-06-13
Cleanup organisation test files

vanitabarrett:
We have two test files which do the same thing: content_store_organisation_test.rb and organisation_test.rb.

We have been updating organisation_test.rb, so this commit takes any tests from content_store_organisation_test.rb which we still want and moves them into organisation_test.rb.


test/integration/organisation_test.rb
```diff
@@ -184,6 +184,18 @@ class OrganisationHeaderTest < ActionDispatch::IntegrationTest
     content_store_has_item("/government/organisations/office-of-the-secretary-of-state-for-wales", @content_item_wales_office)
   end
 
+  it "sets the page title" do
+    visit "/government/organisations/prime-ministers-office-10-downing-street"
+    assert page.has_title?("Prime Minister's Office, 10 Downing Street - GOV.UK")
+
+    visit "/government/organisations/attorney-generals-office"
+    print page.html
```
tijmenb: I think this line can be removed?

vanitabarrett: @tijmenb removed!

PR #707, opened 2018-06-19
Add FOI section to organisation pages

andysellick:
Adds the FOI section. It's not in a 2/3rds grid column yet since it's below the contacts section, but this will be addressed in a future PR.

There was a problem with some of the org pages where the data used by this code wasn't getting into the content item, this problem is being addressed and tracked by [this trello card](https://trello.com/c/a4N9yTgl/14-some-orgs-dont-have-updated-content-items-yet).

Some example pages:

- [page with just the general FOI information](https://govuk-collections-pr-707.herokuapp.com/government/organisations/attorney-generals-office)
- [page with simple FOI contact information](https://govuk-collections-pr-707.herokuapp.com/government/organisations/charity-commission)
- [page with more complex FOI information](https://govuk-collections-pr-707.herokuapp.com/government/organisations/department-for-environment-food-rural-affairs)
- [page with FOI information and a description](https://govuk-collections-pr-707.herokuapp.com/government/organisations/office-of-the-secretary-of-state-for-wales)
- [organisation exempt from FOI](https://govuk-collections-pr-707.herokuapp.com/government/organisations/security-vetting-appeals-panel) (see card for other examples)
- [no10 has no FOI information at all](https://govuk-collections-pr-707.herokuapp.com/government/organisations/prime-ministers-office-10-downing-street)

screen shot 2018-06-19 at 10 46 24

Trello card: https://trello.com/c/gD2hhE0K/193-foi-request-section


app/models/organisation.rb
```diff
@@ -110,6 +114,18 @@ def social_media_links
...
 
+  def foi_exempt?
+    details["foi_exempt"]
+  end
+
+  def foi_contacts
+    links["ordered_foi_contacts"]
+  end
+
+  def ordered_contacts
```
vanitabarrett: This means we can delete the `ordered_contacts` method at the bottom of this file.

app/presenters/organisations/show_presenter.rb
```diff
@@ -148,6 +178,19 @@ def search_results_to_documents(search_results)
       }
     end
 
+    def contact_line(line)
+      return line + "<br/>" if line && line.length.positive?
+      ""
```
vanitabarrett: Do we need to return `""` at all?

app/presenters/organisations/show_presenter.rb
```diff
@@ -120,6 +122,34 @@ def featured_policies
...
     end
 
+    def has_foi_contacts?
+      true if @org.foi_contacts && @org.foi_contacts.length.positive?
+    end
+
+    def foi_contacts
+      @org.foi_contacts.map do |foi_contact|
+        data = {
+          post_addresses: foi_contact["details"]["post_addresses"].map { |post|
```
vanitabarrett: You're mixing use of `map do...end` and `map {}` here - we seem to prefer `do...end` in general, so probably makes it more readable if we always do that.

app/presenters/organisations/show_presenter.rb
```diff
@@ -148,6 +178,19 @@ def search_results_to_documents(search_results)
       }
     end
 
+    def contact_line(line)
+      return line + "<br/>" if line && line.length.positive?
+      ""
+    end
+
+    def make_email_link(email)
+      "<a href=\"#{email}\" class=\"brand__color\">#{email}</a>"
```
vanitabarrett: Can we use `link_to` here instead?

app/presenters/organisations/show_presenter.rb
```diff
@@ -1,5 +1,7 @@
 module Organisations
   class ShowPresenter
+    include ActionView::Helpers
```
vanitabarrett: Is this so you can use `content_tag`? You can pull in the bit that is specifically needed, e.g: `include ActionView::Helpers::TagHelper`

vanitabarrett:
I noticed [this page](https://www.gov.uk/government/organisations/department-for-environment-food-rural-affairs) on live also indicates that the contact details can be used for Environmental Information Regulations (EIR) requests - do we think we still need that?

screen shot 2018-06-19 at 13 16 49

vanitabarrett:
Also the link in the last paragraph for [Cabinet Office](https://govuk-collections-pr-707.herokuapp.com/government/organisations/cabinet-office) isn't actually linking anywhere, it's just rendering as plain text.

app/views/organisations/_show_organisation.html.erb
```diff
@@ -130,6 +130,10 @@
   <% end %>
 <% end %>
 
+<% unless @organisation.is_promotional_org? %>
+  <%= render partial: 'foi' %>
```
vanitabarrett: I think calling the partial `freedom_of_information` (or similar) may be clearer - wouldn't want to assume someone knows what FOI means.

vanitabarrett:
This also looks a bit more confusing now than it did previously on live:

**Live:**
screen shot 2018-06-19 at 13 35 08

**New:**
screen shot 2018-06-19 at 13 35 15

app/presenters/organisations/show_presenter.rb
andysellick: Yes, because we're appending the return value onto an existing string. If the method returns `nil` it breaks. I could add more logic to get round this but I think this is the easiest approach.

app/presenters/organisations/show_presenter.rb
andysellick: I tried `do...end` but it didn't work, I think because it's inside an attribute declaration? Unless I'm misunderstanding. Can you show what you mean?

app/presenters/organisations/show_presenter.rb
andysellick: No. Actually, I just realised, we should be using `mail_to` otherwise these links don't work! Whoops.

app/presenters/organisations/show_presenter.rb
vanitabarrett: The following works for me (I think):

```
data = {
    title: foi_title(foi_contact["details"]["title"]),
    post_addresses: foi_contact["details"]["post_addresses"].map do |post|
        data = ""
        .....etc
    end,
    email_addresses: foi_contact["details"]["email_addresses"].map do |email|
        make_email_link(email["email"])
    end,
    description: foi_description(foi_contact["details"]["description"])
}
```

app/presenters/organisations/show_presenter.rb
```diff
@@ -1,5 +1,9 @@
 module Organisations
   class ShowPresenter
+    include ActionView::Helpers::TagHelper
+    include ActionView::Helpers::UrlHelper
+    include ActionView::Helpers::TranslationHelper
```
vanitabarrett: Maybe I'm missing something - what is the TranslationHelper needed for?

andysellick: It's for link 94 `t('organisations.foi.freedom_of_information_requests')`

andysellick:
@vanitabarrett have fixed the link not rendering as a link issue.

app/presenters/organisations/show_presenter.rb
vanitabarrett: If you don't want to pull in another helper you can use `I18n.t('...')` without needing that, but works in the same way 👍

PR #711, opened 2018-06-19
Cherrypick add orgs document lists

vanitabarrett:


app/views/organisations/_latest_documents_by_type.html.erb
```diff
@@ -0,0 +1,25 @@
...
+  </div>
+
+  <% @documents.latest_documents_by_type.in_groups_of(2, false) do |documents_group| %>
+    <div class="grid-row">
+      <% documents_group.each do |documents| %>
+        <div class="column-half">
+          <%= render "govuk_publishing_components/components/heading", {
+            text: documents[:title],
+            padding: true,
+            font_size: 24
```
andysellick: This should be a H3 as it's under the 'Documents' H2. However I note that the component is using a H3 for each document, we may have to modify the component to use a H4 (or preferably no heading tag at all, doesn't really seem necessary).

vanitabarrett: @andysellick Good spot, I've made this a h3 now! I'll create a card in the backlog for the document list component changes separately - I don't think it should block this.

andysellick: Agreed.

PR #715, opened 2018-06-20
Use partials for organisations

andysellick:
Clean up the new organisations pages code...

- use partials for each section
- use SECTION tags for sections
- wrap 'what we do' section in if logic
- standardise spacing classes on sections


andysellick:
@sihugh two things could be added to this PR...

- the logic changes we discussed for the org's status live/exempt/?
- we could get rid of the `_show.html.erb` template as all it does is then call `_show_organisation.html.erb` (but I can't see how to do it)

sihugh:
We're looking at the statuses soon.
We've got [this card](https://trello.com/c/OnoGN1Cw/206-tech-debt-tidy-up-the-organisation-page-views) to look at tidying up the views

PR #721, opened 2018-06-21
Add corporate info

vanitabarrett:
Adds the corporate information page links to the right hand side of the page. As the contact information section has not been done yet, the links line up with FOI.

## Please read before reviewing
This PR is ready to review in terms of the visual appearance of the section, but some of the actual link content is different to what's on live:

- About Us and Social Media Use links are missing from Promotional Orgs
- Missing secondary corporate information links (paragraph at the bottom of the section)
- All orgs now show links to Transparency Data and Corporate Reports, regardless of whether they have published any.

These are being addressed by https://github.com/alphagov/whitehall/pull/4147, but this hasn't been merged yet.

## Design Changes
Design changes (using topic list instead of bulleted list) has been noted here: https://trello.com/c/tkhnPF8N/21-corporate-information-page-links-shown-without-bullets

**Before:**
screen shot 2018-06-21 at 16 33 39

**After:**
screen shot 2018-06-21 at 16 33 44

- No 10: https://govuk-collections-pr-721.herokuapp.com/government/organisations/prime-ministers-office-10-downing-street
- Ministerial Org: https://govuk-collections-pr-721.herokuapp.com/government/organisations/department-for-exiting-the-european-union
- Organisation with 0 published transparency data docs: https://govuk-collections-pr-721.herokuapp.com/government/organisations/building-regulations-advisory-committee
- Organisation with custom jobs URL: https://govuk-collections-pr-721.herokuapp.com/government/organisations/competition-and-markets-authority

### DIFF
--- a/app/assets/stylesheets/views/_organisations.scss
+++ b/app/assets/stylesheets/views/_organisations.scss
@@ -110,6 +110,17 @@
   border-top-style: solid;
 }
 
+.organisation__section-border-top {
+  border-top: 5px solid $black;
+  padding-top: $gutter-half;
+}
+
+.organisation__corporate-information .organisation__float-section {
+  @include media(tablet) {
+    float: right;
+  }
+}
+
 .organisations__margin-bottom {
   margin-bottom: $gutter-half;
 
--- a/app/presenters/organisations/show_presenter.rb
+++ b/app/presenters/organisations/show_presenter.rb
@@ -101,6 +101,30 @@ module Organisations
       }
     end
 
+    def corporate_information
+      corporate_information_links = @org.ordered_corporate_information.map do |link|
+        {
+          text: link["title"],
+          path: link["href"],
+        }
+      end
+
+      job_links = separate_job_links(corporate_information_links)
+
+      {
+        corporate_information_links: {
+          items: corporate_information_links - job_links,
+          brand: @org.brand,
+          margin_bottom: true
+        },
+        job_links: {
+          items: job_links,
+          brand: @org.brand,
+          margin_bottom: true
+        }
+      }
+    end
+
   private
 
     def contact_line(line)
@@ -156,6 +180,18 @@ module Organisations
 
     def has_definite_article?(phrase)
       phrase.downcase.strip[0..2] == 'the'
+    end
+
+    def separate_job_links(corporate_information_links)
+      job_links = []
+
+      corporate_information_links.each do |link|
+        if link[:path].end_with?("/recruitment", "/procurement") || link[:text].eql?("Jobs")
+          job_links << link
+        end
+      end
+
+      job_links
     end
   end
 end
--- a/app/views/organisations/_corporate_information.html.erb
+++ b/app/views/organisations/_corporate_information.html.erb
@@ -1,11 +1,22 @@
-<div class="organisation-corporate-info">
-  <%= render "govuk_publishing_components/components/heading", {
-      text: "Corporate Information",
-      heading_level: 4
-  } %>
-  <% corporate_info.each do |info| %>
-    <p>
-      <%= link_to info["title"], info["href"] %>
-    </p>
+<div class="column-one-third">
+  <% if @show.corporate_information[:corporate_information_links][:items].any? %>
+    <%= render "govuk_publishing_components/components/heading", {
+      text: t('organisations.corporate_information'),
+      margin_bottom: 2,
+      padding: true,
+      border_top: 5,
+      brand: @organisation.brand,
+    } %>
+    <%= render "components/topic-list", @show.corporate_information[:corporate_information_links] %>
+  <% end %>
+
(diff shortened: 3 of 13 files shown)


app/presenters/organisations/show_presenter.rb
```diff
@@ -139,5 +163,17 @@ def needs_definite_article?(phrase)
     def has_definite_article?(phrase)
       phrase.downcase.strip[0..2] == 'the'
     end
+
+    def separate_job_links(corporate_information_links)
+      job_links = []
+
+      corporate_information_links.each do |link|
+        if link[:path].end_with?("/recruitment", "/procurement") || link[:text].eql?("Jobs")
```
andysellick: I can see why this has been done but it seems a bit precarious to separate the links like this. Not a blocking issue, but the publishing interface should let the publisher decide which list a link belongs to.

app/views/organisations/_corporate_information.html.erb
```diff
@@ -1,11 +1,22 @@
...
-  <%= render "govuk_publishing_components/components/heading", {
-      text: "Corporate Information",
-      heading_level: 4
-  } %>
-  <% corporate_info.each do |info| %>
-    <p>
-      <%= link_to info["title"], info["href"] %>
-    </p>
+<div class="column-one-third">
+  <% if @show.corporate_information[:corporate_information_links][:items].any? %>
```
andysellick: Should this IF statement encompass the wrapping column-one-third div?

test/support/organisation_helper.rb
```diff
@@ -359,4 +359,23 @@ def organisation_with_foi
...
+      details: {
+        brand: "attorney-generals-office",
+        ordered_corporate_information_pages: [
+          {
+            title: "Corporate Information page",
+            href: "/corporate-info"
+          },
+          {
+            title: "Jobs",
+            href: "/jobs"
```
andysellick: Do we need to add a few more links here to more thoroughly test the logic for separating job links?

app/views/organisations/_corporate_information.html.erb
```diff
@@ -1,16 +1,20 @@
...
+    <% unless @organisation.is_promotional_org? %>
+      <%= render "govuk_publishing_components/components/heading", {
+        text: t('organisations.corporate_information'),
+        margin_bottom: 2,
+        padding: true,
+        border_top: 5,
+        brand: @organisation.brand
+      } %>
+    <% end %>
+    <div class="<%= "organisation__section-border-top" if @organisation.is_promotional_org? %>">
```
andysellick: Should we be using the branding classes (used elsewhere) here? I note the additional padding so maybe we still need a class to add that as well.

app/assets/stylesheets/views/_organisations.scss
```diff
@@ -110,6 +110,17 @@
   border-top-style: solid;
 }
 
+.organisation__section-border-top {
+  border-top: 5px solid $black;
+  padding-top: $gutter-half;
+}
+
+.organisation__corporate-information .organisation__float-section {
```
andysellick: Could this be more specific, like `.organisation__float-right`? Also does it need to nest below the corporate information class like this? Seems unnecessarily specific.

Frustrating that there isn't just a utility class to float right.

app/views/organisations/_show_organisation.html.erb
```diff
@@ -57,6 +57,8 @@
     <% end %>
   </div>
   <% if @organisation.ordered_corporate_information && @organisation.ordered_corporate_information.any? %>
-    <%= render partial: 'corporate_information' %>
+    <div class="organisation__corporate-information">
```
andysellick: What does this class do? It doesn't seem to be declared anywhere?

app/assets/stylesheets/views/_organisations.scss
vanitabarrett: I had to make it more specific so it overrode the `.grid-row .column-one-third` styling which floats it left (which makes it appear under the content when there's no FOI/Contact Info/High Profile Groups, rather than sticking to the right-hand side).

app/views/organisations/_show_organisation.html.erb
vanitabarrett: Simply to help with the overriding problem mentioned above.

app/views/organisations/_corporate_information.html.erb
vanitabarrett: I don't think so - the check in `show_organisation` checks if there's any corporate information before rendering this partial, so it won't render if there's nothing. In here, we just want to check whether we're rendering corporate information links or job links, but at this point we know there's at least one.

app/views/organisations/_show_organisation.html.erb
andysellick: In that case I'd suggest nesting it under the more general `.organisations` class?

PR #722, opened 2018-06-21
Add foi links

andysellick:
Adds the optional 'contact form' link to the foi section for the new organisations pages.

Examples:
- [with contact link](https://govuk-collections-pr-722.herokuapp.com/government/organisations/ministry-of-defence)
- [without](https://govuk-collections-pr-722.herokuapp.com/government/organisations/hm-courts-and-tribunals-service)
- [without contact link and without address](https://govuk-collections-pr-722.herokuapp.com/government/organisations/charity-commission)

screen shot 2018-06-21 at 15 29 32

This includes a slight change from the original to have some more descriptive text for the link in question. Original shown here:

screen shot 2018-06-21 at 15 30 07

Trello card: https://trello.com/c/gD2hhE0K/193-foi-request-section

### DIFF
--- a/app/presenters/organisations/show_presenter.rb
+++ b/app/presenters/organisations/show_presenter.rb
@@ -69,15 +69,13 @@ module Organisations
         {
           title: foi_title(foi_contact["details"]["title"]),
           post_addresses: foi_contact["details"]["post_addresses"].map do |post|
-            data = ""
-            data << contact_line(post["title"])
-            data << contact_line(post["street_address"].gsub("\r\n", "<br/>"))
-            data << contact_line(post["locality"])
-            data << contact_line(post["postal_code"])
-            data << contact_line(post["world_location"])
+            foi_address(post)
           end,
           email_addresses: foi_contact["details"]["email_addresses"].map do |email|
             make_email_link(email["email"])
+          end,
+          links: foi_contact["details"]["contact_form_links"].map do |link|
+            make_link(link)
           end,
           description: foi_description(foi_contact["details"]["description"])
         }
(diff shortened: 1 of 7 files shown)


app/presenters/organisations/show_presenter.rb
```diff
@@ -110,6 +108,16 @@ def contact_line(line)
       ""
     end
 
+    def make_link(link)
+      return link_to(make_link_text(link["description"]), link["link"]) if link["link"].length.positive?
+      nil
+    end
+
+    def make_link_text(text)
+      return I18n.t('organisations.foi.contact_form') unless text.length.positive?
```
vanitabarrett: Can you do `text.empty?`. Not sure how the link descriptions work - are they enforced or could it be `nil`?

vanitabarrett: Also I think this could be if you wanted to avoid having the additional line that returns `text`:

```
text.length.positive? ? text : I18n.t('organisations.foi.contact_form')
```

app/presenters/organisations/show_presenter.rb
```diff
@@ -119,8 +127,20 @@ def foi_title(title)
...
-      content_tag(:p, contact.gsub("\r\n", "<br/>").html_safe) if contact
+    def foi_address(post)
+      data = ""
+      data << contact_line(post["title"])
+      data << contact_line(post["street_address"].gsub("\r\n", "<br/>"))
+      data << contact_line(post["locality"])
+      data << contact_line(post["postal_code"])
+      data << contact_line(post["world_location"])
+      return data if data.length.positive?
+      nil
```
vanitabarrett: I would have thought this method would return nil automatically, without this? You could then probably remove the need for `return`

app/presenters/organisations/show_presenter.rb
```diff
@@ -119,8 +127,20 @@ def foi_title(title)
...
+    end
+
+    def foi_description(description)
+      return content_tag(:p, description.gsub("\r\n", "<br/>").html_safe) if description.length.positive?
+      nil
```
vanitabarrett: See other comment about returning nil

andysellick: Weird. That's what I thought too, but it kept returning an empty P tag. Works now though. Oh well.

PR #727, opened 2018-06-22
Add contacts section

vanitabarrett:
Trello: https://trello.com/c/s2ir5SNd/194-add-contact-section-to-organisation-page

screen shot 2018-06-22 at 16 38 33

Examples:

- With lots of contacts: https://govuk-collections-pr-727.herokuapp.com/government/organisations/department-for-international-trade
- Promotional Org: https://govuk-collections-pr-727.herokuapp.com/government/organisations/civil-service
- Long email address: https://govuk-collections-pr-727.herokuapp.com/government/organisations/cabinet-office
- No contacts: https://govuk-collections-pr-727.herokuapp.com/government/organisations/independent-family-returns-panel


app/presenters/organisations/contacts_presenter.rb
```diff
@@ -0,0 +1,105 @@
...
+    def has_contacts?
+      @org.ordered_contacts && @org.ordered_contacts.length.positive?
+    end
+
+    def contacts
+      format_contacts(@org.ordered_contacts)
+    end
+
+    def has_foi_contacts?
+      true if @org.foi_contacts && @org.foi_contacts.length.positive?
```
andysellick: Minor: do we need `true if` here? (and yes, this is code you've moved so it's not on you, but I just noticed it in comparison to the `has_contacts?` method above).

app/assets/stylesheets/views/_organisations.scss
```diff
@@ -146,3 +146,12 @@
 .organisation__address {
   font-style: normal;
 }
+
+.organisation__border-top {
```
andysellick: Not sure if this'll help or not, but the heading component can have a border-top of 1px if needed. Colour might be an issue though.

app/views/organisations/_contacts.html.erb
```diff
@@ -0,0 +1,52 @@
...
+      <% if contact[:post_addresses].any? %>
+        <address class="column-half organisations__margin-bottom organisation__address">
+          <% contact[:post_addresses].each do |post| %>
+            <%= post.html_safe %>
+          <% end %>
+        </address>
+      <% end %>
+
+      <% if contact[:email_addresses].any? || contact[:links].any? || contact[:phone_numbers].any?  %>
+        <div class="<%= contact[:post_addresses].any? ? "column-half" : "column-two-thirds" %>">
```
andysellick: Just wondering about the `column-two-thirds` here. I see what it does, is it worth the effort or could we use the full width in this circumstance?

app/presenters/organisations/contacts_presenter.rb
vanitabarrett: Good spot, probably not. I'll try removing 👍

app/assets/stylesheets/views/_organisations.scss
vanitabarrett: Yeah, I noticed that, but the colour can only be passed via a brand, so it wouldn't work. Perhaps we can extend that component in the future, but think it works for now.

PR #728, opened 2018-06-22
Add promotional features

andysellick:
Add promotional features to organisation pages. They look like this:

screen shot 2018-06-25 at 09 44 09

They should appear on the following pages:

- [Prime Ministers Office](https://govuk-collections-pr-728.herokuapp.com/government/organisations/prime-ministers-office-10-downing-street) - [Live version](https://www.gov.uk/government/organisations/prime-ministers-office-10-downing-street)
- [Civil Service](https://govuk-collections-pr-728.herokuapp.com/government/organisations/civil-service) - [Live version](https://www.gov.uk/government/organisations/civil-service)

Promotional features are confusingly complex.

- Each feature can contain up to three items, which is why the PM's page has two items each with their own title and then 3 items below with one title. The first two are features that contain one item ('Transparency' and 'Meet the Prime Minister'), the second is a feature that contains 3 items ('History and tour')
- If a feature has one item it should appear as a one third grid column with the item inside full width
- If a feature has two items it should appear as a two thirds grid column with the items inside each half width (there don't appear to be any live examples at the moment of this)
- If a feature has three items it should appear as a full width grid column with the items inside each one third grid columns

The GOV.UK grid doesn't support this kind of complexity so I've had to write some custom grid classes using flexbox: 
- necessary as floats can't cope with the varying height of these items (see the Civil Service page) and attempting a min-height fix isn't ideal)
- only works for IE10 but nothing older. I've not attempted a fallback because our IE 8-10 traffic is incredibly low

Trello card: https://trello.com/c/UayWcPCE/192-add-no10-specific-sections


vanitabarrett:
Line breaks have disappeared:

(New vs Old)
screen shot 2018-06-25 at 10 34 08

app/presenters/organisations/documents_presenter.rb
```diff
@@ -174,5 +204,14 @@ def latest_statistics
...
       search_results_to_documents(@latest_statistics)
     end
+
+    def promotions_child_column_class(number_of_items)
+      return "column-half" if number_of_items == 2
+      "column-one-third" if number_of_items == 3
+    end
+
+    def promotional_feature_link(link)
+      link if link.length.positive?
```
vanitabarrett: This will error if `item["href"]` is `nil` - not sure if that's ever a possibility but good to handle just in case. You can do `link if link && link.length.positive?`

app/views/organisations/_promotional_features.html.erb
```diff
@@ -0,0 +1,22 @@
+<section class="organisations__margin-bottom organisations__promotions">
+  <% @documents.promotional_features.each do |feature| %>
+    <div class="<%= feature[:parent_column_class] %>">
+      <%= render "govuk_publishing_components/components/heading", {
+        text: feature[:title],
+        heading_level: 2,
+        padding: true,
+        border_top: 5,
+        brand: @organisation.brand,
+        font_size: 19
```
vanitabarrett: These appear smaller than the other section headings on the page - is this intended?

app/views/organisations/_show_organisation.html.erb
```diff
@@ -16,6 +16,10 @@
   <%= render partial: 'what_we_do' %>
 <% end %>
 
+<% if @organisation.is_promotional_org? %>
+  <%= render partial: 'promotional_features' %>
```
vanitabarrett: This means this will render even if the promotional org does not have any promotional features - is it worth doing a check here or in the promotional_features partial?

app/views/organisations/_promotional_features.html.erb
andysellick: It is, although maybe it shouldn't be. I noticed that a lot of the headings on the Civil Service page were too long and wrapping onto the next line unnecessarily, making the layout look wonky.

andysellick: Have changed - we shouldn't be developing to fix content.

PR #744, opened 2018-06-26
Change translation nav link order

andysellick:
Change order of language list to match existing pages.

Looks like this:

screen shot 2018-06-26 at 13 30 52

Example link: https://govuk-collections-pr-744.herokuapp.com/government/organisations/office-of-the-secretary-of-state-for-wales

Trello card: https://trello.com/c/DB72AxQM/67-cymraeg-english-toggle-is-inverted

### DIFF
--- a/app/presenters/organisations/header_presenter.rb
+++ b/app/presenters/organisations/header_presenter.rb
@@ -76,7 +76,7 @@ module Organisations
         {
           brand: org.brand,
           no_margin_top: true,
-          translations: links
+          translations: links.sort_by { |t| t[:locale] == I18n.default_locale.to_s ? '' : t[:locale] }
         }
       end
     end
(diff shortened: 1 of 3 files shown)


app/presenters/organisations/header_presenter.rb
```diff
@@ -76,7 +76,7 @@ def translation_links
         {
           brand: org.brand,
           no_margin_top: true,
-          translations: links
+          translations: links.reverse
```
vanitabarrett: This assumes the order of translations in the content item - not sure if we can rely on that. Also think it makes sense to be consistent with how we sort translations in other places, e.g: in government-frontend we sort by locale.

`translations.sort_by { |t| t["locale"] == I18n.default_locale.to_s ? '' : t["locale"] }`

test/presenters/organisations/header_presenter_test.rb
```diff
@@ -0,0 +1,32 @@
+require 'test_helper'
+
+describe Organisations::HeaderPresenter do
+  #include RummagerHelpers
```
vanitabarrett: This can be deleted.

andysellick: D'oh!

PR #764, opened 2018-06-27
[DO NOT MERGE] Add UK aid logo to DFID's page

andysellick:
Adds the UK aid logo section to a specific org page.

screen shot 2018-06-27 at 16 28 53

This PR exists only if we really need it. In order to add this to the org page in question some very specific logic has been written and this is not the way that GOV.UK should be built - if there is a requirement for something on an org page it should be an option available to all orgs to configure as they wish, not hard coded into a template like this.

This PR will not be merged.

Trello card: https://trello.com/c/1W0ySgAs/57-uk-aid-component-missing

Example: https://govuk-collections-pr-764.herokuapp.com/government/organisations/department-for-international-development


app/assets/stylesheets/views/_organisation.scss
```diff
@@ -88,3 +88,16 @@
   background-position: 50% 0;
   text-align: center;
 }
+
+.organisation__uk-aid {
+  @include core-19;
+  padding: $gutter-half;
+  padding-left: 120px;
```
vanitabarrett: Could merge the padding into one line

app/views/organisations/_what_we_do.html.erb
```diff
@@ -20,6 +20,10 @@
             </a>
           </p>
       <% end %>
+
+      <% if @organisation.is_dfid? %>
+        <%= render partial: 'uk_aid' %>
```
vanitabarrett: I'm not sure it's worth having a partial for this as the actual code for rendering the UK Aid bit is so small and we're already in a partial separate from the main show view here?

andysellick: Maybe not. I was trying to keep it separate from the rest of the code.

sihugh:
We haven't been asked to add this, so I'll close it.

PR #790, opened 2018-07-17
Refactor OrganisationFeed system to allow extending

tijmenb:
We're going to migrate https://www.gov.uk/government/feed to this application. The organisation feeds were added in https://github.com/alphagov/collections/pull/682, and have identical functionality. 

This PR refactors the existing code to make it reusable for other types of feeds. See commits for details.

The actual feed will be migrated in a follow up PR.

https://trello.com/c/wGkUiiGq


app/controllers/feeds_controller.rb
```diff
@@ -1,23 +1,23 @@
...
   enable_request_formats organisation: :atom
 
-  def organisation
-    set_access_control_allow_origin_header if request.format.atom?
+  before_action do
+    # Allows ajax requests as per https://govuk.zendesk.com/agent/tickets/1935680
+    if request.format.atom?
+      response.headers["Access-Control-Allow-Origin"] = "*"
+    end
+  end
```
kevindew: Not for this PR but this feels like something we should move into nginx

app/services/feeds/feed_entry_presenter.rb
```diff
@@ -0,0 +1,41 @@
+module Feeds
+  class FeedEntryPresenter
```
kevindew: I think this adds some app confusion having this class here within services when there is already a presenters directory for the other ones in this app

tijmenb: Do you think it's better to put them back into their directories without module, or do you see another option?

PR #792, opened 2018-07-17
Serve /goverment/feed

tijmenb:
This is an Atom feed currently served by Whitehall. It is generated using the same method as the organisations feed. When we [migrated that feed][1] we identified two functional changes that will also apply here:

- The `id` of the feed will change from `tag:www.gov.uk,2005:DocumentCollection/231443` to a URL.
- We can no longer provide long form feed content as Rummager doesn't let us obtain that. We fall back to providing a title and a summary.

In addition, we will be serving more content in the feed, not just things published by Whitehall (AAIB reports for example). We've not heard any complaints about that, according to @sihugh. 

[1]: https://github.com/alphagov/collections/pull/682

https://trello.com/c/wGkUiiGq

### DIFF
--- a/app/controllers/feeds_controller.rb
+++ b/app/controllers/feeds_controller.rb
@@ -1,5 +1,5 @@
 class FeedsController < ApplicationController
-  enable_request_formats organisation: :atom
+  enable_request_formats organisation: :atom, all: :atom
 
   before_action do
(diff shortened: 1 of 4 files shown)


sihugh:
In addition, I tried before an after tests on the org feeds with Feedly and Vienna (a desktop RSS feed reader) and the change in ID was transparent to both.  So hopefully people won't notice anything different

PR #796, opened 2018-07-19
[DO NOT MERGE] Update test for breadcrumb component

andysellick:
This updates collections ahead of a change to the breadcrumb component in [this PR](https://github.com/alphagov/govuk_publishing_components/pull/435). 

Do not merge this PR until a new version of the components gem is created and added to this PR (until then tests will fail anyway).

### DIFF
--- a/test/integration/organisation_test.rb
+++ b/test/integration/organisation_test.rb
@@ -504,13 +504,13 @@ class OrganisationTest < ActionDispatch::IntegrationTest
 
   it "displays breadcrumbs" do
     visit "/government/organisations/prime-ministers-office-10-downing-street"
-    assert page.has_css?(".gem-c-breadcrumbs .gem-c-breadcrumbs--item", text: "Prime Minister's Office, 10 Downing Street")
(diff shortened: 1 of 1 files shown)


test/integration/organisation_test.rb
```diff
@@ -504,13 +504,13 @@ class OrganisationTest < ActionDispatch::IntegrationTest
 
   it "displays breadcrumbs" do
     visit "/government/organisations/prime-ministers-office-10-downing-street"
-    assert page.has_css?(".gem-c-breadcrumbs .gem-c-breadcrumbs--item", text: "Prime Minister's Office, 10 Downing Street")
+    assert page.has_css?(".gem-c-breadcrumbs .gem-c-breadcrumbs__item", text: "Prime Minister's Office, 10 Downing Street")
```
tijmenb: Maybe we can one-up this and remove the class name altogether, and test just test that the string appears in the `gem-c-breadcrumbs` element? That way we would have prevented this PR.

andysellick: Fair point, done.

PR #811, opened 2018-07-30
Fix non-typographic apostrophe

quis:
> “Smart quotes,” the correct quotation marks and apostrophes, are curly or sloped. "Dumb quotes," or straight quotes, are a vestigial constraint from typewriters when using one key for two different marks helped save space on a keyboard

– http://smartquotesforsmartpeople.com

;)

### DIFF
--- a/app/views/organisations/index.html.erb
+++ b/app/views/organisations/index.html.erb
@@ -13,7 +13,7 @@
 </div>
 
 <form class="filter-organisations-list__form" data-filter="form">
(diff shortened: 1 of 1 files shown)


vanitabarrett:
Closing this as I've created a new PR which fixes the test which is failing here (I couldn't work out how to do it in this PR, sorry!) 

New one here: https://github.com/alphagov/collections/pull/967

PR #833, opened 2018-08-14
Use taxonomy list component on topic pages

vanitabarrett:
Trello: https://trello.com/c/PFbuMOGm/100-spike-adapt-highlight-box-component-to-work-in-a-list-of-links

This commit replaces using the highlight box and document list component separately for one list, instead replacing it with the [taxonomy_list component](https://govuk-publishing-components.herokuapp.com/component-guide/taxonomy_list) which will render both wrapped in one list. This is an accessibility improvement, so the items are read as belonging to one list.

Visual changes:

- Document lists are now full width instead of 2/3 width. This wasn't intended, we will revisit this.

## Before (Services)
screen shot 2018-08-14 at 16 58 12

## After (Services)
screen shot 2018-08-14 at 16 58 34

Example: https://govuk-collections-pr-833.herokuapp.com/education


sihugh:
Looks like we could extract a common partial at some point too 👍

PR #854, opened 2018-09-10
Fix schema.org breadcrumbs on org pages

sihugh:
## Before
screen shot 2018-09-10 at 15 54 31

## After
screen shot 2018-09-10 at 15 54 14

The [breadcrumb component uses the `:is_current_page` property](https://govuk-publishing-components.herokuapp.com/component-guide/breadcrumbs/highlight_current_page) to set extra attributes on the current page (such as `aria-current`) and helps determine the url to set on any links.

We haven't been setting these on org pages which means that the schema implementation is invalid on pages which include the current page in the breadcrumb.

An example page is:

- https://www.gov.uk/government/organisations/department-for-work-pensions

Which fails schema validation on the breadcrumb because the `@id` is mandatory:

- https://search.google.com/structured-data/testing-tool/u/0/#url=https%3A%2F%2Fwww.gov.uk%2Fgovernment%2Forganisations%2Fdepartment-for-work-pensions

Setting the `is_current_page` property:

- makes the link clickable (it links to the #content section on the same page)
- sets the `aria-current` property to the current page which helps with accessibility
- enables click tracking on the link, so we can see whether our breadcrumbs are
confusing.
- and also fixes the schema.org thing which was the basis for this ticket anyway

https://trello.com/c/GOPIPoAm/159-error-in-schema-breadcrumblist-for-certain-pages

### DIFF
--- a/app/views/organisations/_breadcrumb.html.erb
+++ b/app/views/organisations/_breadcrumb.html.erb
@@ -7,10 +7,11 @@
       },
       {
         title: "Organisations",
-        url: "/government/organisations/"
+        url: "/government/organisations"
       },
       {
-        title: @header.org.title
+        title: @header.org.title,
+        is_current_page: true
       }
     ]
   } %>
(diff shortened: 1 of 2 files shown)


sihugh:
Good plan, thanks.
  
I've kept the multiple tests as they're testing slightly different pages I think.  I've extended the first one to ensure that the appropriate other links are actually there too, and tweaked the remaining tests to be a bit more specific.

PR #865, opened 2018-09-18
Use heading component on topic pages

vanitabarrett:
This shouldn't change how the title of each section looks to the user.

## Before:
screen shot 2018-09-18 at 10 34 24

## After:
screen shot 2018-09-18 at 10 34 03

Review App: https://govuk-collections-pr-865.herokuapp.com/government/europe

### DIFF
--- a/app/views/taxons/show.html.erb
+++ b/app/views/taxons/show.html.erb
@@ -15,7 +15,11 @@
     <div class="taxon-page__section-group">
       <div class="grid-row">
         <div class="column-two-thirds">
-          <h2 class="taxon-page__section-heading"><%= section[:title] %></h2>
+          <%= render "govuk_publishing_components/components/heading", {
+            text: section[:title],
+            heading_level: 2,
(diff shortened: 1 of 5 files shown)


app/views/taxons/show.html.erb
```diff
@@ -15,7 +15,10 @@
     <div class="taxon-page__section-group">
       <div class="grid-row">
         <div class="column-two-thirds">
-          <h2 class="taxon-page__section-heading"><%= section[:title] %></h2>
```
andrewgarner: I wondered if we could also remove the CSS definition for `.taxon-page__section-heading` but it's also being used in `app/views/taxons/_organisations.html.erb`. Perhaps we could update that heading too and then remove the CSS?

vanitabarrett: Good spot, thanks!

vanitabarrett: @andrewgarner done 🙂

PR #874, opened 2018-09-21
Remove our policies from org pages

1pretz1:
Remove `Our policies` from all organisation pages, this is due to policies being retired.

Trello:
https://trello.com/c/W0A9H0Js/211-remove-our-policies-from-org-pages

## Before:
screen shot 2018-09-21 at 13 49 10

## After:
screen shot 2018-09-21 at 13 51 37


test/presenters/organisations/show_presenter_test.rb
```diff
@@ -80,48 +80,6 @@
     assert_equal expected, @show_presenter.parent_organisations
   end
 
-  it 'formats policies correctly' do
```
vanitabarrett: I think this means we can also remove the example this test uses: https://github.com/alphagov/collections/blob/master/test/support/organisation_helper.rb#L237

1pretz1: Yep, now looking there is a few helpers that need to be removed too. Good spot!

1pretz1: Removed! 👍

PR #875, opened 2018-09-24
Get news image url from Rummager

oscarwyatt:
News and communications on taxon pages have an image, the url was gathered from Content Store, we can do now this from the Rummager request. This removes the unnecessary extra call, speeding up page load times.

Rummager doesn't return the alt text but by using an image_card instead of an image, then we don't need alt text as the link text overrides it for accessibility purposes (the whole image is part of the link).

Trello: https://trello.com/c/9Ckuzwca/178-taxon-pages-get-the-news-imageurl-from-rummager

### DIFF
--- a/app/views/taxons/sections/_news_and_communications.html.erb
+++ b/app/views/taxons/sections/_news_and_communications.html.erb
@@ -5,36 +5,28 @@
 
 <div class="grid-row">
   <div class="taxon-page__featured-item column-one-third <%= featured_item_layout_class %>" <%= "data-module=track-click" if track_click?(section[:promoted_content]) %>>
-    <img class="taxon-page__featured-image" src="<%= featured_item[:image][:url] %>" alt="<%= featured_item[:image][:alt] %>">
-    <div class="taxon-page__featured-text">
-      <%= link_to(
-        featured_item[:link].fetch(:text),
-        featured_item[:link].fetch(:path),
-        class: "taxon-page__featured-link",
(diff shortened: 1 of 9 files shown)


vanitabarrett:
@oscarwyatt I'm not sure what's happening here, but the featured article at /education is showing the default image, even though the article has its own image? 

https://govuk-collections-pr-875.herokuapp.com/education

oscarwyatt:
Interesting, I'll look into it

app/views/taxons/sections/_news_and_communications.html.erb
```diff
@@ -5,7 +5,10 @@
 
 <div class="grid-row">
   <div class="taxon-page__featured-item column-one-third <%= featured_item_layout_class %>" <%= "data-module=track-click" if track_click?(section[:promoted_content]) %>>
-    <img class="taxon-page__featured-image" src="<%= featured_item[:image][:url] %>" alt="<%= featured_item[:image][:alt] %>">
+    <%= render "govuk_publishing_components/components/image_card", {
+        href: featured_item[:link].fetch(:path),
+        image_src: featured_item[:image][:url]
+    } %>
     <div class="taxon-page__featured-text">
```
vanitabarrett: If we render the heading and metadata with the component, we can then remove this and any associated `taxon-page__featured` CSS

sihugh:
I think we need to add an environment variable to heroku apps.  I've done it for this review app, but will have to look into doing it around the houses.

PR #901, opened 2018-10-11
 Translate content purpose into document types

sihugh:
Iterating which document types belong to which supertype is very slow if it means republishing content and getting it all reindexed by rummager.  We are circumventing this by ignoring the `content_purpose_supergroup` properties of content items and rummager results, and instead translating them prior to performing queries.

We use the [govuk_document_types](https://github.com/alphagov/govuk_document_types) gem to help us with this. In the tests which exercise this path where we use the gem, we've stubbed the response to ensure that we reduce the brittleness of the tests as much as possible.

So that we can retain confidence that the gem returns reasonable responses for each of the groups (and that the groups still exist in the gem), we've added a few tests to assert how we expect the gem to behave.

These changes have already been done in [finder frontend](alphagov/finder-frontend#619), and this PR is to ensure that collections is in line with those changes.

https://trello.com/c/Q88UJzp5/177-update-collections-to-use-the-govukdocumenttypes-gem

### DIFF
--- a/test/presenters/supergroups/guidance_and_regulation_test.rb
+++ b/test/presenters/supergroups/guidance_and_regulation_test.rb
@@ -102,4 +102,12 @@ describe Supergroups::GuidanceAndRegulation do
       assert_equal expected, guidance_and_regulation_supergroup.promoted_content(taxon_id)
     end
   end
+
+  describe "document types" do
+    it "returns appropriate things" do
+      document_types = GovukDocumentTypes.supergroup_document_types("guidance_and_regulation")
+
+      assert_includes document_types, "guidance"
+    end
+  end
 end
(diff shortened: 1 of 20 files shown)


test/presenters/supergroups/guidance_and_regulation_test.rb
```diff
@@ -102,4 +102,12 @@
       assert_equal expected, guidance_and_regulation_supergroup.promoted_content(taxon_id)
     end
   end
+
+  describe "document types" do
+    it "returns appropriate things" do
+      document_types = GovukDocumentTypes.supergroup_document_types("guidance_and_regulation")
+
+      assert document_types.include? "guidance"
```
andrewgarner: How about we use the `assert_includes` assertion?

```rb
assert_includes document_types, "guidance"
```

At the moment our failure message will be "Expected false to be truthy.", with a more specific assertion our error message would be "Expected 'document_types' to include 'guidance'".

sihugh: Good call, will update, thanks

PR #905, opened 2018-10-22
Add Topic Pages multivariate test

oscarwyatt:
Trello card: https://trello.com/c/tBIXwhGy/35-remove-blue-white-content-boxes-on-topic-pages

This adds the TopicPagesTest multivariate test and configures it for variants A-E. At the moment, each view is configured to point to A (the control). This will be changed in the future as we develop each variant.


app/controllers/concerns/taxon_pages_testable.rb
```diff
@@ -0,0 +1,35 @@
...
+    @taxon_page_test ||= GovukAbTesting::AbTest.new(
+      "TopicPagesTest",
+      dimension: CUSTOM_DIMENSION,
+      allowed_variants: %w(A B C D E),
+      control_variant: "A"
+    )
+  end
+
+  def set_test_response_header
+    taxon_page_variant.configure_response(response)
```
vanitabarrett: Do we only want to do this `if is_testable_taxon_page?` ?

oscarwyatt: As far as I'm aware they're all part of the A/B so we don't need this(?)

sistephens:
Just to keep everyone on their toes... this is now not going to be a variant 😬

We made a call in planning to remove blue/white boxes without running it as an A/B test.

So we should deploy this and it will become the control in our other A/B tests.

app/controllers/concerns/taxon_pages_testable.rb
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+private
+
+  def taxon_page_test
+    @taxon_page_test ||= GovukAbTesting::AbTest.new(
+      "TopicPagesTest",
+      dimension: CUSTOM_DIMENSION,
+      allowed_variants: %w(A B C D E),
+      control_variant: "A"
```
sihugh: I didn't know about this option :-)

vanitabarrett:
@sihugh Removing of the boxes is no longer a variant - we're just going to ship it as part of the A variant (I think the PR will need a rebase once this is merged though)

PR #909, opened 2018-10-26
[DO NOT MERGE] Add in-page sub-navigation

karlbaker02:
This PR adds an in-page sub-navigation menu to the left-hand side of the page, in a one-third column. All page content is placed in a two-thirds column to the right, which now includes sub-topics as a list rather than in a grid.

I've raised the PR as it is to get a Heroku app built for team inspection. It is tagged with `DO NOT MERGE` for the following reasons:

- The image card component used in the `_news_and_communications` partial currently only displays an image with the title and description to the right-hand side if `large: true` is specified as part of the initialization. Large image cards have the image set to use two-thirds of the available space, but the images available do not fill this. Additionally, the design in the Trello card specifies a 50-50 split between image and text.

- All of the content in the `_news_and_communications` partial is slightly wider than the rest of the content in the two-thirds column.

- The link in the sub-nav which is active (if applicable) should have an underline applied to them.

### DIFF
--- a/app/controllers/taxons_controller.rb
+++ b/app/controllers/taxons_controller.rb
@@ -4,7 +4,7 @@ class TaxonsController < ApplicationController
   def show
     setup_content_item_and_navigation_helpers(taxon)
 
-    render :show, locals: {
+    render "show_#{taxon_page_variant.variant_name.downcase}", locals: {
       presented_taxon: presented_taxon,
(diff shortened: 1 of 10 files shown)


vanitabarrett:
Closing as this is superseded by https://github.com/alphagov/collections/compare/add-in-page-nav

PR #915, opened 2018-11-01
Add in page nav

MahmudH:
This change adds an in page navigation to topic pages. This change is part of an A/B Test, which will be present in Variant C.

Trello card: https://trello.com/c/5hFNt9yn

**Before**
screen shot 2018-11-02 at 10 07 03

This PR adds a in page nav to the left. This will allow users to navigate through content in each supergroup linked to the topic.

**After**
screen shot 2018-11-02 at 09 54 07

### DIFF
--- a/app/views/taxons/_organisations.html.erb
+++ b/app/views/taxons/_organisations.html.erb
@@ -5,7 +5,8 @@
         <%= render "govuk_publishing_components/components/heading", {
             text: t('taxons.organisations'),
             heading_level: 2,
-            margin_bottom: 2
+            margin_bottom: 2,
+            id: "organisations"
           } %>
 
 
--- a/app/views/taxons/_organisations.html.erb
+++ b/app/views/taxons/_organisations.html.erb
@@ -1,18 +1,17 @@
 <% if presented_organisations.show_organisations? %>
-  <div class="taxon-page__section-group taxon-page__section-group--organisation" data-module="toggle">
+  <div id="organisations" class="taxon-page__section-group taxon-page__section-group--organisation" data-module="toggle">
     <div class="grid-row">
       <div class="column-two-thirds">
         <%= render "govuk_publishing_components/components/heading", {
-            text: t('taxons.organisations'),
-            heading_level: 2,
-            margin_bottom: 2,
-            id: "organisations"
+              text: t('taxons.organisations'),
+              heading_level: 2,
+              margin_bottom: 2
           } %>
 
 
         <%= render partial: 'organisation_logos_and_list', locals: {
-          organisations_with_logos: presented_organisations.promoted_organisation_list[:promoted_with_logos],
-          organisations_without_logos: presented_organisations.promoted_organisation_list[:promoted_without_logos]
+              organisations_with_logos: presented_organisations.promoted_organisation_list[:promoted_with_logos],
+              organisations_without_logos: presented_organisations.promoted_organisation_list[:promoted_without_logos]
         } %>
 
         <% if presented_organisations.show_more_organisations? %>
@@ -25,8 +24,8 @@
 
           <div id="more-organisations" class="js-hidden">
             <%= render partial: 'organisation_logos_and_list', locals: {
-              organisations_with_logos: presented_organisations.show_more_organisation_list[:organisations_with_logos],
-              organisations_without_logos: presented_organisations.show_more_organisation_list[:organisations_without_logos]
+                  organisations_with_logos: presented_organisations.show_more_organisation_list[:organisations_with_logos],
+                  organisations_without_logos: presented_organisations.show_more_organisation_list[:organisations_without_logos]
             } %>
           </div>
         <% end %>
--- a/app/views/taxons/sections/_news_and_communications.html.erb
+++ b/app/views/taxons/sections/_news_and_communications.html.erb
@@ -3,9 +3,9 @@
   featured_item_layout_class = "taxon-page__featured-item--single" if section[:documents].empty?
 %>
 
-<div class="grid-row">
-  <div class="taxon-page__featured-item column-one-third <%= featured_item_layout_class %>" <%= "data-module=track-click" if track_click?(section[:promoted_content]) %>>
+  <div class="taxon-page__featured-item <%= featured_item_layout_class %>" <%= "data-module=track-click" if track_click?(section[:promoted_content]) %>>
     <%= render "govuk_publishing_components/components/image_card", {
+        large: true,
         href: featured_item[:link].fetch(:path),
         image_src: featured_item[:image][:url],
         heading_text: featured_item[:link].fetch(:text),
@@ -17,7 +17,7 @@
     } %>
   </div>
   <% if section[:documents].any? %>
-    <div class="column-two-thirds" <%= "data-module=track-click" if track_click?(section[:documents]) %>>
+    <div <%= "data-module=track-click" if track_click?(section[:documents]) %>>
       <%= render 'govuk_publishing_components/components/document_list',
         items: section[:documents],
         margin_top: true,
@@ -29,4 +29,3 @@
     <%= render partial: "taxons/sections/see_more_link", locals: { section: section } %>
 
   </div>
-</div>
--- a/app/views/taxons/show_a.html.erb
+++ b/app/views/taxons/show_a.html.erb
@@ -2,10 +2,10 @@
 <% content_for :is_taxon_with_subtopics, presented_taxon.show_subtopic_grid? %>
 <%=
   render(
-      partial: 'common',
-      locals: {
-          presented_taxon: presented_taxon
-      }
+    partial: 'common',
+    locals: {
+      presented_taxon: presented_taxon
+    }
   )
 %>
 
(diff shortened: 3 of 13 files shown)


app/views/taxons/_organisations.html.erb
```diff
@@ -5,7 +5,8 @@
         <%= render "govuk_publishing_components/components/heading", {
             text: t('taxons.organisations'),
             heading_level: 2,
-            margin_bottom: 2
+            margin_bottom: 2,
+            id: "organisations"
```
vanitabarrett: I would suggest moving the id of each section to the `taxon-page__section-group` so the heading isn't so close to the top of the screen and the blue border at the top of each section is visible on the screen when the user clicks a link and jumps down to it.

## As is now:
screen shot 2018-11-02 at 11 22 13

## Suggested:
screen shot 2018-11-02 at 11 21 58

app/views/taxons/show_a.html.erb
```diff
@@ -2,32 +2,32 @@
 <% content_for :is_taxon_with_subtopics, presented_taxon.show_subtopic_grid? %>
 <%=
   render(
-    partial: 'common',
-    locals: {
-      presented_taxon: presented_taxon
-    }
+      partial: 'common',
```
vanitabarrett: I think this (and a few lines underneath) are indented one tab too far.

app/views/taxons/show_a.html.erb
```diff
@@ -2,32 +2,32 @@
...
+          <div class="column-two-thirds">
+            <%= render "govuk_publishing_components/components/heading", {
+                text: t(section[:id], scope: :content_purpose_supergroup, default: section[:title]),
+                heading_level: 2,
+                margin_bottom: 2
+            } %>
+          </div>
         </div>
+        <%= render(
+                partial: section[:partial_template],
```
vanitabarrett: Same comment here about indenting, I think it's more easily readable if it's only indented one tab.

app/views/taxons/show_a.html.erb
```diff
@@ -38,25 +38,25 @@
   <nav role="navigation" class="taxon-page__grid">
     <div class="full-page-width-wrapper">
       <%= render "govuk_publishing_components/components/heading", {
-        text: t('taxons.sub_topics'),
-        heading_level: 2,
-        font_size: 19,
-        margin_bottom: 2
+          text: t('taxons.sub_topics'),
```
vanitabarrett: Same comment about indenting

app/views/taxons/show_a.html.erb
```diff
@@ -38,25 +38,25 @@
...
       <ol class="taxon-page__grid-wrapper">
         <% presented_taxon.child_taxons.each_with_index do |child_taxon, index| %>
           <li class="taxon-page__grid-item">
             <h3 class="taxon-page__grid-heading">
               <%= link_to(
-                child_taxon.title,
-                child_taxon.base_path,
-                data: presented_taxon.options_for_child_taxon(index: index)
-              )%>
+                      child_taxon.title,
```
vanitabarrett: Indenting

app/views/taxons/show_a.html.erb
```diff
@@ -38,25 +38,25 @@
...
+                  )%>
             </h3>
             <p class="taxon-page__grid-item-description"><%= child_taxon.description %></p>
           </li>
         <% end %>
       </ol>
     </div>
   </nav>
-<% end %>
+<% end %>
```
vanitabarrett: You're missing a newline at the end of this file. You can add this in manually or [configure your IDE](https://stackoverflow.com/questions/12415147/ensure-newline-at-eof-in-rubymine) to add these automatically.

app/views/taxons/show_a.html.erb
vanitabarrett: Indenting comments will also probably apply to `show_c` as it's been copied from `show_a`

app/views/taxons/show_c.html.erb
```diff
@@ -1,5 +1,69 @@
...
 %>
+
+<div class="full-page-width-wrapper">
+  <div class="grid-row">
+    <div class="column-one-third">
+      <%= render "govuk_publishing_components/components/contents_list", {
+          hide_title: true,
+          contents: [
+              {
+                  text: "On this page:",
```
vanitabarrett: We should probably add this to `en.yml` so it can be translated in future.

app/views/taxons/show_c.html.erb
```diff
@@ -1,5 +1,69 @@
...
+                    text: t(section[:id], scope: :content_purpose_supergroup, default: section[:title]),
+                    heading_level: 2,
+                    margin_bottom: 2,
+                    id: section[:id]
+                } %>
+              </div>
+            </div>
+            <%= render(
+                    partial: section[:partial_template],
+                    locals: {section: section}
```
vanitabarrett: We normally try and keep a space between the `{` and the content, e.g: `{ section: section }`.

app/views/taxons/show_c.html.erb
```diff
@@ -1,5 +1,69 @@
...
+                )
+            %>
+          </div>
+        <% end %>
+      <% end %>
+
+      <%= render partial: 'organisations', locals: {presented_organisations: presented_taxon.organisations_section} %>
```
vanitabarrett: Same comment here about spacing, should be `{ presented_organisations.....etc }`

test/integration/taxon_browsing_test.rb
```diff
@@ -20,6 +21,7 @@ class TaxonBrowsingTest < ActionDispatch::IntegrationTest
     and_i_can_see_the_research_and_statistics_section
     and_i_can_see_the_organisations_section
     and_i_can_see_the_sub_topics_grid
+    and_i_can_see_the_in_page_nav
```
vanitabarrett: I would add this in a new test, e.g: `it 'renders an in-page nav in Variant C`. Otherwise it becomes difficult to see what has been added for normal taxon page changes and what has been added for variants.

test/integration/taxon_browsing_test.rb
```diff
@@ -283,6 +285,20 @@ def and_i_can_see_the_sub_topics_grid
     end
   end
 
+  def and_i_can_see_the_in_page_nav
```
vanitabarrett: I don't think this test actually tests the presence of the in-page nav, it seems to check the new design of the sub-topic list. It makes sense to still keep this as something we want to check, but you might want to write an additional separate check for seeing the in-page nav. Things we may want to check there:

- Presence of the contents list component
- Presence of links in the contents list component, e.g: links to the sub-topic list

vanitabarrett:
Also just spotted this:

screen shot 2018-11-02 at 11 40 15

If you look closely, the feedback bar at the bottom of the page has a grey background surrounding it. We added this so previously so it matched the grid background. As we don't have the grid in Variant C, we probably no longer want that background there either.

I'd also add some spacing between the sub-topic list/in-page nav and the feedback bar - they seem quite close.

app/views/taxons/sections/_news_and_communications.html.erb
```diff
@@ -3,9 +3,9 @@
   featured_item_layout_class = "taxon-page__featured-item--single" if section[:documents].empty?
 %>
 
-<div class="grid-row">
-  <div class="taxon-page__featured-item column-one-third <%= featured_item_layout_class %>" <%= "data-module=track-click" if track_click?(section[:promoted_content]) %>>
+  <div class="taxon-page__featured-item <%= featured_item_layout_class %>" <%= "data-module=track-click" if track_click?(section[:promoted_content]) %>>
     <%= render "govuk_publishing_components/components/image_card", {
+        large: true,
```
vanitabarrett: Once [the fix](https://github.com/alphagov/collections/pull/918) is merged, you should be able to rebase to fix the issue with the news image card layout.

test/integration/taxon_browsing_test.rb
MahmudH: 👍

test/integration/taxon_browsing_test.rb
MahmudH: 👍

app/views/taxons/show_c.html.erb
MahmudH: 👍

app/views/taxons/show_c.html.erb
MahmudH: 👍

app/views/taxons/show_c.html.erb
MahmudH: 👍

app/views/taxons/show_a.html.erb
MahmudH: 👍

app/views/taxons/show_a.html.erb
MahmudH: 👍

app/views/taxons/_organisations.html.erb
MahmudH: 👍

PR #956, opened 2018-12-05
Don't show alpha taxons

cbaines:
The phase indicates that these are not ready for people to use, so
don't display them.


app/controllers/taxons_controller.rb
```diff
@@ -2,6 +2,8 @@ class TaxonsController < ApplicationController
   include TaxonPagesTestable
 
   def show
+    return error_404 if taxon.alpha_taxon?
```
andrewgarner: How about moving this into `Taxon` so that `find` only returns live taxons?

I say this because `Taxon.find` returns alpha and live taxons, but `taxon.child_taxons` only returns live taxons so the handling of the taxon phase isn't consistent.

cbaines: Sure, I've got `find` to ignore alpha taxons now.

cbaines: Sure, I've changed this now, so that `Taxon.find` raises an exception.

PR #952, opened 2018-12-03
Add Campaign page

vanitabarrett:
Adds campaign page in preparation for EU Exit. This will be populated by taxons with Brexit content.

screen shot 2018-12-10 at 13 41 57

PR #960, opened 2018-12-10
[DO NOT MERGE] Add tracking to campaign page

vanitabarrett:
Requires https://github.com/alphagov/collections/pull/952 to be merged in first


app/presenters/brexit_taxon_presenter.rb
```diff
@@ -17,4 +18,12 @@ def description
   def finder_link
     "https://finder-frontend-pr-706.herokuapp.com/prepare-individual-uk-leaving-eu?topic=#{taxon.base_path}"
   end
+
+  def data_attributes
+    {
+      "track-category": "navGridContentClicked",
+      "track-action": @index,
```
bilbof: Do we usually set a number as "track-action"? I wonder if instead of passing an index we could pass the id of the taxon or something like that, to simplify this class.

app/presenters/brexit_taxons_presenter.rb
```diff
@@ -15,6 +15,6 @@ def call
   def featured_taxons
     @featured_taxons ||= FEATURED_TAXONS
       .map { |base_path| Taxon.find(base_path) }
-      .map { |taxon| BrexitTaxonPresenter.new(taxon) }
+      .each.with_index(1).map { |taxon, index| BrexitTaxonPresenter.new(taxon, index) }
```
bilbof: Did you mean to offset the index by 1? If not, you could use `.each_with_index`. https://stackoverflow.com/a/20258160

app/presenters/brexit_taxon_presenter.rb
vanitabarrett: The track-action is the link's position within other similar links. E.g: if there are 6 taxons, when you click on the first link it should have a track-action of 1, whereas if you click on the last taxon link, it should have a track-action of 6 (which is also why I've offset the index by 1).

vanitabarrett:
Replacing with https://github.com/alphagov/collections/pull/962 because this was proving to be difficult to rebase

PR #962, opened 2018-12-10
Add tracking to campaign page

vanitabarrett:
Adds tracking to the various elements in the new campaign page.

### DIFF
--- a/app/presenters/brexit_taxon_presenter.rb
+++ b/app/presenters/brexit_taxon_presenter.rb
@@ -6,8 +6,9 @@ class BrexitTaxonPresenter
     to: :taxon
   )
 
-  def initialize(taxon)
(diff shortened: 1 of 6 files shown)


vanitabarrett:
@bilbof Do you mind reviewing again? I had to create a new branch because the old one got very difficult to rebase.

app/presenters/brexit_taxon_presenter.rb
```diff
@@ -6,8 +6,9 @@ class BrexitTaxonPresenter
     to: :taxon
   )
 
-  def initialize(taxon)
+  def initialize(taxon, index = false)
     @taxon = taxon
+    @index = index if index
```
bilbof: You could use `index = nil` in the params and remove the conditional on line 11, so `@index = index`

PR #967, opened 2018-12-11
Fix non-typographic apostrophe

vanitabarrett:
Original change by @quis. I just added the test fix.

> “Smart quotes,” the correct quotation marks and apostrophes, are curly or sloped. "Dumb quotes," or straight quotes, are a vestigial constraint from typewriters when using one key for two different marks helped save space on a keyboard

– http://smartquotesforsmartpeople.com

;)

### DIFF
--- a/app/views/organisations/index.html.erb
+++ b/app/views/organisations/index.html.erb
@@ -13,7 +13,7 @@
 </div>
 
 <form class="filter-organisations-list__form" data-filter="form">
-  <label class="filter-organisations-list__label" for="filter-organisations-list">What's the latest<br />from</label>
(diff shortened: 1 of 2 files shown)


sihugh:
Does govspeak render smart quotes or straight quotes?

vanitabarrett:
@sihugh 

Smart...? This is from gov.uk/student-finance:

## Example
screen shot 2018-12-11 at 13 27 23




Although outside govspeak, they always seem to be straight, e.g:

## Examples
screen shot 2018-12-11 at 13 28 03
screen shot 2018-12-11 at 13 28 09

sihugh:
Cool, I was just worried that if govspeak only rendered the straight quotes, but we were putting curly ones in static content then it would look weird.

PR #972, opened 2018-12-11
Refactor citizen landing page to improve performance

karlbaker02:
This PR refactors the citizen landing page to remove the need to create new `Topic` instances, instead using `ContentItem` data. For `other_taxons` in `BrexitTaxonsPresenter`, the `ContentItem` of `/` can be used directly to create `ContentItem`s and remove the need for any additional calls to the Content Store.

### DIFF
--- a/app/presenters/brexit_taxons_presenter.rb
+++ b/app/presenters/brexit_taxons_presenter.rb
@@ -14,9 +14,13 @@ class BrexitTaxonsPresenter
 
   def featured_taxons
     @featured_taxons ||= FEATURED_TAXONS
-      .map { |base_path| ContentItem.find!(base_path) }
-      .each_with_index
-      .map { |content_item, index| BrexitTaxonPresenter.new(content_item, index) }
+      .to_enum
+      .with_index(1)
+      .map do |base_path, index|
+        content_item = ContentItem.find!(base_path)
+
+        BrexitTaxonPresenter.new(content_item, index)
+      end
   end
 
   def other_taxons
--- a/app/presenters/brexit_taxons_presenter.rb
+++ b/app/presenters/brexit_taxons_presenter.rb
@@ -10,9 +10,9 @@ class BrexitTaxonsPresenter
 
   def featured_taxons
     @featured_taxons ||= FEATURED_TAXONS
-      .map { |base_path| Taxon.find(base_path) }
-      .map
-      .with_index(1) { |taxon, index| BrexitTaxonPresenter.new(taxon, index) }
+      .map { |base_path| ContentItem.find!(base_path) }
+      .each_with_index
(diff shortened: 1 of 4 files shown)


app/presenters/brexit_taxons_presenter.rb
```diff
@@ -10,9 +10,9 @@ class BrexitTaxonsPresenter
 
   def featured_taxons
     @featured_taxons ||= FEATURED_TAXONS
-      .map { |base_path| Taxon.find(base_path) }
-      .map
-      .with_index(1) { |taxon, index| BrexitTaxonPresenter.new(taxon, index) }
+      .map { |base_path| ContentItem.find!(base_path) }
+      .each_with_index
```
bilbof: I believe @vanitabarrett used `with_index(1)` rather than `each_with_index` because the index needs to start at 1 rather than zero.

app/presenters/brexit_taxons_presenter.rb
```diff
@@ -21,8 +21,7 @@ def other_taxons
       .reject { |content_item| FEATURED_TAXONS.include?(content_item.base_path) }
       .reject { |content_item| content_item.base_path == "/government/all" }
       .select { |content_item| search_response_including_brexit(content_item.content_id, document_types).total.positive? }
-      .map { |content_item| Taxon.find(content_item.base_path) }
-      .map { |taxon| BrexitTaxonPresenter.new(taxon) }
+      .map { |content_item| BrexitTaxonPresenter.new(content_item) }
```
bilbof: Yeah nice work :)

app/presenters/brexit_taxons_presenter.rb
```diff
@@ -8,21 +8,28 @@ class BrexitTaxonsPresenter
...
+  ).freeze
+
   def featured_taxons
     @featured_taxons ||= FEATURED_TAXONS
-      .map { |base_path| Taxon.find(base_path) }
-      .map
-      .with_index(1) { |taxon, index| BrexitTaxonPresenter.new(taxon, index) }
+      .to_enum
+      .with_index(1)
+      .map do |base_path, index|
```
bilbof: Perhaps you can do `FEATURED_TAXONS.map.with_index(1) do ...` rather than using `to_enum`

karlbaker02: You still need the `map` after `with_index(1)` as otherwise `@featured_taxons` isn't going to have the correct content (an array of presenters).

As far as I'm aware, `to_enum` and `map` (without any block) are equivalent, and I think it makes the intention of the code more obvious when used as we have above, rather than `map` implicitly returning an enumerable.

Happy to discuss this further.

PR #984, opened 2018-12-20
[DO NOT MERGE] Replace organisation announcements with news and communications

andrewgarner:
Changes the "Announcements" list to a "News and communications" list on organisation pages and links to the new news and communications finder.

See https://trello.com/c/fZM6JhkU


config/locales/cy.yml
```diff
@@ -13,18 +13,26 @@ cy:
     corporate_information: "Gwybodaeth gorfforaethol"
     document_types:
       documents: "Dogfennau"
-      announcements:
+      news_and_communications:
         title: "Cyhoeddiadau"
-        see_all: "Gweld ein holl cyhoeddiadau"
+        see_all:
+          text: "Gweld ein holl cyhoeddiadau"
```
karlbaker02: If we want to be consistent, this should instead be a translation to say _see all news and communications_

andrewgarner: We'll need to source a translation for this, something I think we're best doing in a separate PR.

karlbaker02: Should we put in the request now so that it has chance to happen possibly over Christmas?

andrewgarner: 👍 

I'm not sure what the process is though?

sihugh:
I've rebased and added the organisation breadcrumb

PR #1008, opened 2019-01-21
Replace groupings on org pages with supergroups

bilbof:
https://trello.com/c/loZpQ8we/160-replace-lists-on-org-pages-with-6-supergroup-lists

### What's this?

This replaces organisation groupings with supergroups. It links to new finders from the bottom of each list.

Before:

screen shot 2019-01-23 at 13 15 53

View live: https://www.gov.uk/government/organisations/department-for-environment-food-rural-affairs

After _(There are more supergroups than in the screenshot)_:

screen shot 2019-01-23 at 13 15 39

View live: https://govuk-collections-pr-1008.herokuapp.com/government/organisations/department-for-environment-food-rural-affairs

### Also included:

Some refactoring! I extracted some supergroup logic from Organisations::DocumentPresenter into a new Supergroups presenter. I also moved some logic to Services for fetching supergroups and their data. This involved creating a few new classes. I tried not to touch things unrelated to my change.

See commit messages for more details.

Still to come: Welsh language translations.


app/controllers/organisations_controller.rb
```diff
@@ -16,6 +16,7 @@ def show
     @show = Organisations::ShowPresenter.new(@organisation)
     @header = Organisations::HeaderPresenter.new(@organisation)
     @documents = Organisations::DocumentsPresenter.new(@organisation)
+    @supergroups = Organisations::SupergroupsPresenter.new(@organisation)
```
bilbof: This is a good entry point for a review. Essentially, supergroups like 'announcements' used to be a part of `@documents` but have been moved to `@supergroups`. Having two presenters is intended to make things simpler.

config/locales/cy.yml
```diff
@@ -34,6 +34,36 @@ cy:
         see_all:
           text: "Gweld ein holl hystadegau"
           path: "/government/statistics?departments[]=%{organisation}"
+      news_and_communications:
+        title: "News and communications"
+        see_all:
+          text: "See all news and communications"
+          path: "/news-and-communications?organisations[]=%{organisation}"
```
bilbof: Still waiting on translations and correct links for the translation files.

app/controllers/organisations_controller.rb
bilbof: Example page: https://govuk-collections-pr-1008.herokuapp.com/government/organisations/department-for-environment-food-rural-affairs

Versus production: https://www.gov.uk/government/organisations/department-for-environment-food-rural-affairs

app/lib/supergroups_api/supergroup_collection.rb
```diff
@@ -0,0 +1,29 @@
+module SupergroupsApi
+  class SupergroupCollection
+    SUPERGROUP_TYPES = %w(
```
leenagupte: Is it worth reading the supergroups from here: https://github.com/alphagov/govuk_document_types/blob/master/data/supergroups.yml in case they change?

app/controllers/organisations_controller.rb
leenagupte: @bilbof I'd add these comments to the commit message so they don't get lost.

app/lib/supergroups_api/supergroup_collection.rb
bilbof: @leenagupte I thought about it, but it [says here](https://github.com/alphagov/govuk_document_types#govuk-document-types):

> This gem is only to be used in publishing-api and rummager. **Don't use it in your project.**

leenagupte: I think @sihugh was thinking about ways to make the supergroups my dynamic. I'm not sure how far he got though.

bilbof: That would be great. The stern warning on govuk_document_types scared me off!

config/locales/cy.yml
leenagupte: You don't need to add these placeholder translations. If an entry doesn't exist in the locale, then the entry for "en" is used as the default.

app/lib/supergroups_api/supergroup.rb
```diff
@@ -0,0 +1,37 @@
+module SupergroupsApi
```
leenagupte: I'd think about renaming this module/class. SupergroupsApi doesn't tell what it does. It's doing a rummager search so perhaps the module or class name should be "Search"?

app/lib/supergroups_api/supergroup_collection.rb
```diff
@@ -0,0 +1,29 @@
+module SupergroupsApi
+  class SupergroupCollection
```
leenagupte: There are a lot of classes with "Supergroup" in the name. What is a "SupergroupCollection"?

leenagupte:
@bilbof There's a lot going on in this one commit that make the changes (and the reasons for them) hard to follow. Perhaps you could try splitting them out into to atomic commits, so we can easily compare the code change to it's related tests?

app/lib/supergroups_api/supergroup_collection.rb
bilbof: SupergroupCollection is a collection of Supergroup instances. Do you think it will be more understandable as `Search::Supergroups`?

leenagupte: And a Supergroup is the result of a search by supergroup and organisation?

leenagupte: There's already a Supergroup module that's used to power the Taxon pages: https://github.com/alphagov/collections/blob/master/app/presenters/supergroups/guidance_and_regulation.rb

leenagupte: The different Rummager searches seem to live here: https://github.com/alphagov/collections/tree/master/app/services, so you might want to consider following the same pattern.

bilbof: Yeah that is a good idea.

app/presenters/organisations/supergroups_presenter.rb
```diff
@@ -0,0 +1,49 @@
...
+      @org = organisation
+    end
+
+    def has_groups?
+      supergroups_collection.has_groups?
+    end
+
+    def latest_documents_by_supergroup
+      supergroups_collection.groups.map { |supergroup|
+        next unless supergroup.has_documents?
```
koetsier: bit nicer to use ```select``` then map

app/services/search/supergroups.rb
```diff
@@ -0,0 +1,29 @@
+module Search
+  class Supergroups
+    SUPERGROUP_TYPES = %w(
```
koetsier: Maybe we should use the document_types gem here:
```
    def supergroup_types
      @supergroup_types ||= GovukDocumentTypes::SUPERGROUPS.dig('content_purpose_supergroup', 'items').map { |item| item['id'] }
    end
```

what do you think?

bilbof: In principle that's a good idea, but I worry if we add a new supergroup then this will automatically start appearing on org pages.

leenagupte: @bilbof It should, but only if there are any documents of those types tagged to the organisation. The bonus is we have less places to maintain the list of supergroups.

bilbof: Awesome, lets change it :)

bilbof:
Product team has approved this, so it can now be merged in.

sihugh:
We still need the Welsh translations I think?  Check out the DWP page in the review app and view the translated page....

bilbof:
Good spot @sihugh, I added the translations: https://govuk-collections-pr-1008.herokuapp.com/government/organisations/department-for-work-pensions.cy

PR #1016, opened 2019-01-25
Get consultation closing dates from rummager

sihugh:
Topic pages currently look up consultation closing date information from content store, which means up to 5 extra api requests are required to render a page.  By bundling this information in the initial rummager request we can skip all of this.

This should speed up page rendering (particularly when content store is slow), as well as making the render process more robust as we need to make fewer api requests.

Consultation closing dates are shown in the "Policy papers and consultations" sections on pages such as https://www.gov.uk/education and https://gov.uk/transport/road-infrastructure.  

This was added to Whitehall in https://github.com/alphagov/whitehall/pull/4611 (and all consultations have been reindexed) so all the information should be present in rummager.

Example pages (should look identical to current live versions): 
- https://govuk-collections-pr-1016.herokuapp.com/education
- https://govuk-collections-pr-1016.herokuapp.com/transport/road-infrastructure

I deployed this to staging for 10 minutes or so (16:30ish to 16:40ish) and the percentiles for page duration show an immediate improvement:

screenshot 2019-01-25 16 50 02

### DIFF
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -8,6 +8,7 @@ class Document
     :description,
     :base_path,
     :public_updated_at,
+    :end_date,
     :change_note,
     :format,
     :content_store_document_type,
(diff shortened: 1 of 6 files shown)


sihugh:
Thanks!

PR #1035, opened 2019-02-13
Enable in-page navigation on topic pages

koetsier:
This is variant C of the Q3 A/B test:

Trello: https://trello.com/c/Touii2Tz/305-ship-variant-c-from-topic-pages-a-b-test

image

image

### DIFF
--- a/app/views/taxons/show.html.erb
+++ b/app/views/taxons/show.html.erb
@@ -1,67 +1,65 @@
 <% content_for :page_class, "taxon-page taxon-page--grid" %>
-<% content_for :is_taxon_with_subtopics, presented_taxon.show_subtopic_grid? %>
 <% content_for :brexit_taxon, presented_taxon.brexit? %>
 <%=
   render(
-    partial: 'common',
-    locals: {
-      presented_taxon: presented_taxon
-    }
+      partial: 'common',
+      locals: {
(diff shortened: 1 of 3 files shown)


app/views/taxons/show.html.erb
```diff
@@ -1,67 +1,65 @@
...
-          </li>
-        <% end %>
-      </ol>
+        </div>
+      <% end %>
     </div>
-  </nav>
-<% end %>
+  </div>
+</div>
```
oscarwyatt: No newline, tut tut

test/integration/taxon_browsing_test.rb
```diff
@@ -43,6 +42,23 @@ class TaxonBrowsingTest < ActionDispatch::IntegrationTest
     then_i_can_see_navigation_to_brexit_pages
   end
 
+  it 'renders an in-page nav in variant C' do
```
oscarwyatt: Super pedantic I know but it might be confusing to future people that is says something about variant c so I'd remove that

koetsier: fixed

app/views/taxons/show.html.erb
koetsier: fixed

PR #1042, opened 2019-02-21
Hide the govuk-accordion controls

alex-ju:
After the new accordion (based on `govuk-frontend`) was introduced in the `govuk_publishing_components` gem it started to clash (they are using the same `data-module` value for initialisation) with the accordion component defined in this repository.

Long term fix is to replace this component with the one in the in `govuk_publishing_components`, however the local component contains a lot of logic including tracking with Google Analytics that we need to make sure are preserved before doing the switch. Meanwhile the...

Short term fix is to hide the controls exposed by the script coming from `govuk_publishing_components` and leave the local controls do their job.

### Before
screenshot-www gov uk-2019 02 21-17-32-23

### After
screenshot-www gov uk-2019 02 21-17-32-39

### DIFF
--- a/app/assets/stylesheets/components/_accordion.scss
+++ b/app/assets/stylesheets/components/_accordion.scss
@@ -103,3 +103,8 @@
 .app-c-accordion__controls {
   @extend %contain-floats;
 }
+
+
+.govuk-accordion__controls {
+  display: none;
+}


nickcolley:
I think the correct answer is that both GOV.UK Frontend and you should be using namespaces, it'd be a breaking change for us but I'll raise an issue, and see if there's anything we can do to move it forwards...

alex-ju:
@andysellick @nickcolley cool, namespace will be. wasn't aware of the precedent. will update now.

alex-ju:
Closed in favour of #1043

PR #1052, opened 2019-03-14
Add custom link to prepare-eu-brexit page

sihugh:
The first commit has the bulk of the changes and is easier to see the diff. 
The second two commits are moves/renames and are slightly clunky to allow git to follow along.
The last two address a couple of accessibility issues.

This allows us to add links to the prepare-eu-exit page that are to arbitrary locations on GOV.UK

The link we've added is to a new mainstream piece of content that user research has validated as a better solution for most people's needs.

As part of this, we're removing the transport and going-staying-abroad links because these are the things that are most closely reflected in this new publication.

I've moved the citizen readiness presenters together, but I've done this in two clunky steps to preserve the git history better.  The move and rename seems to be too much for git to cope with in one step.  (Github seems to struggle with this across the PR too, but the individual commits seem OK)

I also added a few tests for the actual page which we were somehow missing before 🤔 .

page affected is: https://govuk-collections-pr-1052.herokuapp.com/prepare-eu-exit

https://trello.com/c/Xaxcsyiv/131-change-citizen-landing-page-to-interim-solution


app/presenters/citizen_readiness/link_presenter.rb
```diff
@@ -0,0 +1,30 @@
+module CitizenReadiness
+  class LinkPresenter
+    include Tracking
+
+    attr_reader :base_path
+
+    def initialize(base_path, index)
+      @base_path = base_path
+      @index = index #used in link tracking
```
bilbof: Might be worth adding the same comment to `BrexitTaxonPresenter`, as it's quite useful here and would serve the same purpose there.

app/presenters/brexit_taxons_presenter.rb
```diff
@@ -1,28 +1,30 @@
...
   ).freeze
 
-  def featured_taxons
-    @featured_taxons ||= FEATURED_TAXONS.map.with_index do |base_path, index|
-      featured_taxon = level_one_taxons.detect { |taxon| taxon.base_path == base_path }
-      BrexitTaxonPresenter.new(featured_taxon, index)
-    end
+  def featured_links
+    custom_links +
+      featured_taxons(custom_links.length)
```
bilbof: Is there a way that you could instantiate all of the presenters here? It would be a bit more code I suppose, but it would make the index stuff a bit clearer.

app/presenters/citizen_readiness/link_presenter.rb
```diff
@@ -0,0 +1,30 @@
...
+    end
+
+    def link
+      base_path
+    end
+
+  private
+
+    def slug
+      link.delete("/")
```
bilbof: This won't support multiple `/` in a base_path. Is that an issue?

app/views/brexit_campaign/_featured_taxons.html.erb
```diff
@@ -1,12 +1,12 @@
-<% @taxons.in_groups_of(local_assigns.fetch(:display_full_width, true) ? 3 : 2, false).each_with_index do |taxons_group, row_index| %>
+<% @featured_links.in_groups_of(local_assigns.fetch(:display_full_width, true) ? 3 : 2, false).each_with_index do |links_group, row_index| %>
   <div class="grid-row" data-module="track-click">
-    <% taxons_group.each do |taxon| %>
+    <% links_group.each do |link| %>
       <div class="<%= local_assigns.fetch(:display_full_width, true) ? 'column-one-third' : 'column-half' %>">
         <div class="campaign__taxon <%= "campaign__taxon--variable-border" if row_index.eql?(0) %> ">
           <h3 class="campaign__taxon-title">
-            <%= link_to(taxon.title, taxon.finder_link, data: taxon.data_attributes, "aria-label": "#{t('campaign.see_more_in_taxon', topic: taxon.title.downcase)}") %>
+            <%= link_to(link.title, link.link, data: link.featured_data_attributes, "aria-label": "#{t('campaign.see_more_in_taxon', topic: link.title.downcase)}") %>
```
bilbof: If you have the index here, from the `each_with_index` above, could you provide this index value to `link.featured_data_attributes` and save yourself some work in the presenters? I suppose it's nice to keep the view lean, but since we already have the indexes to hand it would be simpler.

app/presenters/citizen_readiness/link_presenter.rb
sihugh: It's just to drop the leading `/` so should be fine for now I think?

app/presenters/citizen_readiness/link_presenter.rb
sihugh: 👍  done

app/presenters/citizen_readiness/link_presenter.rb
bilbof: Cool :)

app/views/brexit_campaign/_featured_taxons.html.erb
sihugh: Nice option.  I'll have a quick go at that, thanks

PR #1074, opened 2019-04-03
Cache common usages of search for a short period

sihugh:
These code paths are commonly used, and the parameters are generally fixed which makes the calls to rummager fairly cacheable (as opposed to finders which have lots of combinations).  We don't mind if the results are a little stale as we'll otherwise fall back to the mirrors on a 5xx error which can be up to 24 hours old.

Memcache is already set up on frontend machines

### DIFF
--- a/app/controllers/organisations_api_controller.rb
+++ b/app/controllers/organisations_api_controller.rb
@@ -28,7 +28,7 @@ private
   RESULTS_PER_PAGE = 20
 
   def presented_organisations(start:)
-    organisations = get_organisations(count: RESULTS_PER_PAGE, start: start)
+    organisations = organisations_from_search(count: RESULTS_PER_PAGE, start: start)
     OrganisationsApiPresenter.new(
       organisations["results"],
       current_page: current_page,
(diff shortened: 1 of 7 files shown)


app/controllers/organisations_api_controller.rb
```diff
@@ -53,22 +53,30 @@ def presented_organisation(slug:)
...
-    Services.rummager.search(
-      filter_format: "organisation",
-      order: "title",
-      count: count,
-      start: start
-    )
+  def organisations_from_search(count:, start:)
+    cache_key = "api/organisations/#{count}/#{start}"
+
+    @organisations_from_search ||= Rails.cache.fetch(cache_key, expires_in: 10.minutes) do
```
barrucadu: Rather than adding caching everywhere which calls `Services.rummager.search`, could the caching be done there?  Or maybe a new `cached_search` function.  It could take the cache duration as an optional argument, defaulting to 5 minutes, and another optional argument to disable caching if needed.

This works for now, but if any new searches are added people will need to remember to add the explicit caching.

sihugh: Yeah, I was doing this as a short term thing - but might do the rework now as you've moved collections over to search-api already 👍

PR #1095, opened 2019-04-24
Sort documents in the "services" and "guidance & regulation" supergroups by popularity descending

barrucadu:
[Trello card](https://trello.com/c/wTJRIpnh/672-change-sort-order-of-2-supergroup-lists-on-org-pages)

### DIFF
--- a/app/services/search/supergroup.rb
+++ b/app/services/search/supergroup.rb
@@ -1,10 +1,13 @@
 module Search
   class Supergroup
-    attr_reader :content_purpose_supergroup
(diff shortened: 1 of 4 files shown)


app/services/search/supergroup.rb
```diff
@@ -29,7 +32,7 @@ def search_rummager(additional_params)
     def default_rummager_params
       {
         count: 2,
-        order: "-public_timestamp",
+        order: @sort_order,
```
sihugh: we can omit the `@` on `sort_order` as we've used `attr_reader` above

PR #1145, opened 2019-06-14
Change 'See more...' links on topic pages to point to supergroup finders

bilbof:
As part of our work to retire topic page finders (/search/advanced), we need to change the 'See more...' links on topic pages to point to appropriate supergroup finders.

'See more...' links should point to the right supergroup finder, filtered by the taxon.

Should be merged in after https://github.com/alphagov/finder-frontend/pull/1188 is deployed.

https://trello.com/c/qhxNedVR/766-change-see-more-links-on-topic-pages-to-point-to-supergroup-finders-s

### DIFF
--- a/app/presenters/supergroups/supergroup.rb
+++ b/app/presenters/supergroups/supergroup.rb
@@ -10,15 +10,10 @@ module Supergroups
       name.humanize
     end
 
-    def finder_link(base_path)
-      query_string = {
-        topic: base_path,
-        group: name
-      }.to_query
-
+    def finder_link(base_path, taxon_id)
       {
         text: see_all_link_text,
-        url: "/search/advanced?#{query_string}",
+        url: finder_url(base_path, taxon_id),
         data: see_all_link_data_attributes(base_path, name)
       }
     end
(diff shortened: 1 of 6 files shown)


app/presenters/supergroups/supergroup.rb
```diff
@@ -47,6 +42,15 @@ def data_module_label
 
   private
 
+    def finder_path
+      "/search/#{name.dasherize}"
+    end
+
+    def finder_url(base_path)
+      query_string = { parent: base_path }.to_query
```
barrucadu: This should set the `level_one_taxon` and `level_two_taxon` parameters - though I'm not actually sure `level_two_taxon` is required, eg https://www-origin.publishing.service.gov.uk/search/news-and-communications?level_one_taxon=17fb491a-15f3-4c71-8e25-f645ff8b3ed3 works and isn't a level one taxon

barrucadu: This should actually set the new `topic` parameter @koetsier is adding: https://trello.com/c/rjrtLu4N/821-allow-topic-parameter-in-supergroup-finder

bilbof: Thanks, changed.

PR #1188, opened 2019-07-30
Update supergroup finder link

hannako:
Update the 'See all research and statistics' link on the org pages from the old whitehall finder to the new supergroup finder (government/statistics --> search/research-and-statistics)

It's not necessary to pass in `content_store_document_type=published_statistics` or `order=updated-newest` because they are defaults. 

https://trello.com/c/RQVHQstP/673-change-research-statistics-link-on-pages-to-point-to-the-new-finder

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -32,7 +32,7 @@ en:
         title: "Research and statistics"
         see_all:
           text: "See all research and statistics"
-          path: "/government/statistics?departments[]=%{organisation}&parent=%{organisation}"
+          path: "/search/research-and-statistics?organisations[]=%{organisation}&parent=%{organisation}"
       policy_and_engagement:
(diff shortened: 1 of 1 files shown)


config/locales/en.yml
```diff
@@ -32,7 +32,7 @@ en:
         title: "Research and statistics"
         see_all:
           text: "See all research and statistics"
-          path: "/government/statistics?departments[]=%{organisation}&parent=%{organisation}"
+          path: "/search/research-and-statistics?organisations[]=%{organisation}"
```
bilbof: Would be good to add `parent`, too:     

```
path: "/search/policy-papers-and-consultations?organisations[]=%{organisation}&parent=%{organisation}"
```

hannako: Ok will do.

PR #1189, opened 2019-07-30
Update cookie serialiser

hannako:
Update cookie serializer from `:marshal` to `:json`. Collections does not read or write any cookies so it's ok to switch to `:json` without transitioning to `:hybrid` first

### DIFF
--- a/config/brakeman.ignore
+++ b/config/brakeman.ignore
@@ -10,7 +10,29 @@
       "line": 17,
       "link": "https://brakemanscanner.org/docs/warning_types/cross_site_scripting",
       "code": "Organisation.find!(request.path).body",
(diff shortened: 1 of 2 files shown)


hannako:
@barrucadu Is this the right approach for the brakeman errors around cookie serialization? Again. 
(https://github.com/alphagov/specialist-publisher/pull/1471)

barrucadu:
You should also change the cookie serialisation method to `:hybrid`, which will let us switch to `:json` eventually.

cbaines:
> You should also change the cookie serialisation method to `:hybrid`, which will let us switch to `:json` eventually.

Does collections actually set Rails cookies? If it doesn't you could just change the mode to :json, as there's no need to transition cookies.

hannako:
Thanks Chris. I don't think it does set any Rails cookies, or read any either. I'll switch to `:json`

PR #1205, opened 2019-08-09
[DO NOT MERGE] Preliminary countdown element

andysellick:
Prototyping a countdown for the Brexit page, design is unfinalised.

Counts down to midnight on the 31st of October, after which it will disappear. Brexit technically occurs at 11pm on the 31st but Ruby's date calculation stuff does some rounding around midnight. Since this is complicated to achieve precisely and the behaviour at this time is undecided (see below) I've left this as is for now and we can address nearer the time when decisions are made.

It's currently undecided what it should say on the 31st, as "0 days to go" is a little strange.

Tested in a few old and mobile browsers and looks okay.

Screen Shot 2019-08-12 at 17 51 20

### DIFF
--- a/test/integration/brexit_landing_page_test.rb
+++ b/test/integration/brexit_landing_page_test.rb
@@ -4,7 +4,7 @@ require_relative '../support/brexit_landing_page_steps'
 class BrexitLandingPageTest < ActionDispatch::IntegrationTest
   include BrexitLandingPageSteps
 
-  it 'renders the brexit page' do
+  it "renders the brexit page" do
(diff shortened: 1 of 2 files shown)


vanitabarrett:
This PR has become quite out of date with master and is tricky to rebase as we've since updated Collections to remove govuk_frontend_toolkit so there are a lot of conflicts. I've copied over the relevant changes from this PR into a new branch which is up to date with master, in case this work is needed (talks still ongoing).

The new branch is here: https://github.com/alphagov/collections/compare/countdown-component?expand=1

PR #1294, opened 2019-09-20
Add AB test setup for Brexit landing page

vanitabarrett:
This PR sets up two variants of the landing page, A and you guessed it... B. These variants are currently the same but B will be updated with the new landing page designs in a later PR.

Further information in following commits: ->

Co-authored: @vanitabarrett @1pretz1 

Trello:
https://trello.com/c/Az3SiQxJ/265-add-a-b-test-logic-to-the-landing-page

Also thanks to @kevindew for bashing his head against a problem due for my (@1pretz1) gross negligence of the rules of html! 😆😭

### DIFF
--- a/app/controllers/brexit_landing_page_controller.rb
+++ b/app/controllers/brexit_landing_page_controller.rb
@@ -1,4 +1,9 @@
 class BrexitLandingPageController < ApplicationController
+  # 68 was chosen to uniquely identify BrexitLandingPageTest in GA
+  CUSTOM_DIMENSION = 68
+
+  helper_method :page_variant
+
   def show
     setup_content_item_and_navigation_helpers(taxon)
 
@@ -7,9 +12,24 @@ class BrexitLandingPageController < ApplicationController
       presentable_section_items: presentable_section_items,
       show_dynamic_list: show_dynamic_list?,
     }
+
+    page_variant.configure_response(response)
   end
 
 private
+
+  def page_variant
+    @page_variant ||= begin
+                        ab_test = GovukAbTesting::AbTest.new(
+                          "BrexitLandingPageTest",
+                          dimension: CUSTOM_DIMENSION,
+                          allowed_variants: %w(A B),
+                          control_variant: "A",
+                        )
+
+                        ab_test.requested_variant(request.headers)
+                      end
+  end
 
   def taxon
     Taxon.find(request.path)
@@ -22,8 +42,8 @@ private
   def presentable_section_items
     presented_taxon.supergroup_sections.select { |section| section[:show_section] }.map do |section|
       {
-          href: "##{section[:id]}",
-          text: t(section[:id], scope: :content_purpose_supergroup, default: section[:title]),
+        href: "##{section[:id]}",
+        text: t(section[:id], scope: :content_purpose_supergroup, default: section[:title]),
       }
     end
   end
--- a/app/views/brexit_landing_page/show.html.erb
+++ b/app/views/brexit_landing_page/show.html.erb
@@ -1,14 +1,14 @@
 <% content_for :page_class, "taxon-page taxon-page--grid" %>
-<% content_for :extra_headers do %>
-  <link rel="preload" as="image" href="<%= asset_path 'chevron-extend-hover.svg' %>">
+<% content_for :meta_tags do %>
+  <%= page_variant.analytics_meta_tag.html_safe %>
 <% end %>
 
 <%=
   render(
       partial: 'page_header',
       locals: {
-          presented_taxon: presented_taxon,
-          show_dynamic_list: show_dynamic_list
+        presented_taxon: presented_taxon,
+        show_dynamic_list: show_dynamic_list
       }
   )
 %>
--- a/app/views/email_signups/new.html.erb
+++ b/app/views/email_signups/new.html.erb
@@ -2,7 +2,7 @@
 <%= render 'shared/tag_meta', tag: subtopic, description: "Subscribe to get an email each time content is published or updated in the '#{subtopic.combined_title}' topic." %>
 <% content_for :page_class, "email-signup-new" %>
 
-<% content_for :meta_section do %>
+<% content_for :meta_tags do %>
   <meta name="govuk:section" content="<%= meta_section %>">
 <% end %>
 
--- a/app/views/latest_changes/index.html.erb
+++ b/app/views/latest_changes/index.html.erb
@@ -6,7 +6,7 @@
       } %>
 <% end %>
 
-<% content_for :meta_section do %>
+<% content_for :meta_tags do %>
   <meta name="govuk:section" content="<%= meta_section %>">
   <meta name="robots" content="noindex">
 <% end %>
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -7,7 +7,6 @@
   <%= csrf_meta_tags %>
   <%= yield :meta_tags %>
   <%= yield :meta_section %>
-  <%= yield :extra_headers %>
   <%= render 'govuk_publishing_components/components/meta_tags', content_item: @content_item %>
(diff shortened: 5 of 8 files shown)


app/controllers/concerns/brexit_landing_page_testable.rb
```diff
@@ -0,0 +1,27 @@
+module BrexitLandingPageTestable
+  CUSTOM_DIMENSION = 68
```
kevindew: Does this have to be a number? if so perhaps we could have a comment above it about why the analytics team have given it this number or at least a clue about it. There's a little bit of a `MEANING_OF_LIFE = 42` vibe to it - that I appreciate is not confusion you've introduced but we could try alleviate for others.

app/views/brexit_landing_page/show_A.html.erb
```diff
@@ -1,6 +1,6 @@
 <% content_for :page_class, "taxon-page taxon-page--grid" %>
 <% content_for :extra_headers do %>
-  <link rel="preload" as="image" href="<%= asset_path 'chevron-extend-hover.svg' %>">
+  <% @page_variant.analytics_meta_tag.html_safe %>
```
kevindew: Does this need to be a `<%= `, and if I'm right on that should a test be failing?

app/views/brexit_landing_page/show_B.html.erb
```diff
@@ -0,0 +1,112 @@
...
+  <%= @page_variant.analytics_meta_tag.html_safe %>
+<% end %>
+
+<%=
+  render(
+      partial: 'page_header',
+      locals: {
+          presented_taxon: presented_taxon,
+          show_dynamic_list: show_dynamic_list
+      }
```
kevindew: we seem to have double indentation here

app/controllers/brexit_landing_page_controller.rb
```diff
@@ -1,16 +1,28 @@
...
-    render locals: {
-      presented_taxon: presented_taxon,
-      presentable_section_items: presentable_section_items,
-      show_dynamic_list: show_dynamic_list?,
-    }
+    if page_variant.variant?("A")
+      render "show_A", view_locales
+    else
+      render "show_B", view_locales
+    end
```
kevindew: It's unconventional to use capital letters in the view names, I'd suggest lowercasing them.

Also it might be overkill having the view_locales in a method when it'd be a bit of a simpler ready if they were set before the conditional (as you have to go off to mentally break off to read the method to understand the data)

You may even want to go along the lines of:

```
view = page_variant.variant?("A") ? "show_a" : "show_b"

render view, locals: { ... }
```

Also will this default to A if CDN sets nothing? I'm guessing so based on the control group part.

app/views/brexit_landing_page/show_B.html.erb
```diff
@@ -0,0 +1,112 @@
...
+    <%= link_to(t('brexit_landing_page.email'), "/email-signup/?topic=#{presented_taxon.base_path}", {
+      class: "govuk-body govuk-link",
+      data: {
+        module: "track-click",
+        "track-category": "emailAlertLinkClicked",
+        "track-action": presented_taxon.base_path
+      }
+    }) %>
+  </div>
+</div>
```
kevindew: Is this view basically the same as variant A at the moment? If so I wonder if we want to just using the same view until divergence is sufficiently significant that it warrants complete new views. Otherwise in the meantime we have to maintain two pages of near identical HTML.

app/controllers/brexit_landing_page_controller.rb
```diff
@@ -1,16 +1,28 @@
 class BrexitLandingPageController < ApplicationController
+  include BrexitLandingPageTestable
```
kevindew: I'd suggest we actually move the code from this include into the controller. I'm not sure if we'd intend to use it in any other controllers and I think we could reduce the amount of code in doing so.

We could probably have:
```
helper_method :page_variant
after_action { page_variant.configure_response(response) }

# ...

def page_variant
  @page_variant ||= begin
    test = GovukAbTesting::AbTest.new(
             "BrexitLandingPageTest",
             dimension: CUSTOM_DIMENSION,
             allowed_variants: %w(A B),
             control_variant: "A",
           )
    test.requested_variant(request.headers)
  end
end
```

which might make it feel a bit simpler.

1pretz1: Sure, I've made a few small changes to your suggestion but have relocated and cut down the code... success!

app/controllers/brexit_landing_page_controller.rb
1pretz1: Sure, I like your suggestion! You're correct, the default will indeed by variant 'A' if no header is set from the CDN.

app/views/brexit_landing_page/show_A.html.erb
1pretz1: Yup it does my bad.

`extra_headers` is actually a confusing name for what this content_for block is doing as it's actually only setting a meta tag and not a header here. I've renamed this to `meta_section`.

No test should be failing as there wasn't any checking the meta tag, the only test present was one checking the request header coming in. I've since amended the tests to check for the response headers and that the meta tag is being set.

app/views/brexit_landing_page/show_B.html.erb
1pretz1: Amended this in both shows.

app/views/brexit_landing_page/show_B.html.erb
1pretz1: It's identical at present, I thought for the testing and logic side of things it would be better to have two fully functioning views but I suppose I can remove this one and point both renders to `show_a` for now until `show_b` is implemented.

app/views/brexit_landing_page/show.html.erb
```diff
@@ -1,14 +1,14 @@
 <% content_for :page_class, "taxon-page taxon-page--grid" %>
-<% content_for :extra_headers do %>
-  <link rel="preload" as="image" href="<%= asset_path 'chevron-extend-hover.svg' %>">
+<% content_for :meta_section do %>
```
kevindew: perhaps this belongs in meta_tags, I don't really know why we need meta_tags and meta_section since they seem to serve similar purpose although the only thing we seem to have in meta_section is `<meta name="govuk:section" content="<%= meta_section %>">`

app/controllers/brexit_landing_page_controller.rb
```diff
@@ -7,10 +10,26 @@ def show
...
+                          "BrexitLandingPageTest",
+                          dimension: CUSTOM_DIMENSION,
+                          allowed_variants: %w(A B),
+                          control_variant: "A",
+                        )
+
+                        ab_test.requested_variant(request.headers)
+                      end
+  end
+  helper_method(:page_variant)
```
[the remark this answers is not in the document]
1pretz1: Sure

app/views/brexit_landing_page/show.html.erb
1pretz1: I've altered all content_for's using meta_section to now use meta_tags, as you said there's no clear reason and they were originally all housed under one - 'extra_headers'. See here https://github.com/alphagov/collections/commit/031d6615893076cffcd44cc18420287ce99d45f0

kevindew: Great thanks Pete.

I noticed there's a couple of files where meta tags are defined twice:
https://github.com/alphagov/collections/blob/1e53550626aeb5e653e693a35643a69c2b1efc27/app/views/second_level_browse_page/show.html.erb#L3-L10

and

https://github.com/alphagov/collections/blob/1e53550626aeb5e653e693a35643a69c2b1efc27/app/views/subtopics/show.html.erb#L3-L21

1pretz1: Now there is one per file! 🎉

PR #1317, opened 2019-10-08
Howto images

sihugh:
*Best reviewed by commit*
- tiny fix for deprecation message
- extract part of view into partial
- extract part of model into another model
- actual change

Adds images to the HowTo schema for each step in a step by step page such as https://www.gov.uk/learn-to-drive-a-car

Currently we use the same placeholder image for each of the steps.  We think this is stopping us from being presented as our preferred rich result which looks like this:

Screenshot 2019-10-10 08 10 17

In this PR, we're adding images for steps numbered 1 - 12, and others for the `and` and `or` type steps.  If our needs are different to these then we'll still use the old placeholder image.

Testing the updated schema in the [rich results testing tool](https://search.google.com/test/rich-results?view=search-preview&id=ngMOgbeS64psO37ynx2iBw) shows a valid snippet: 

Screenshot 2019-10-10 08 12 07

The new images aren't displayed, though this isn't unexpected.  From [the docs](https://support.google.com/webmasters/answer/6398748#__RICH_CARD_BLUE_LINKS): 

> If you are testing rich card markup, you can see your actual image in the preview if we confirm, or verify, that the image URL is in our index.  However, the preview displays a placeholder image if our systems cannot verify the image URL. Even so, when your rich card markup reaches our systems, the image you provide should be shown correctly in the actual Google Search results.


app/views/step_nav/_structured_data.html.erb
```diff
@@ -0,0 +1,8 @@
+<% step_image_urls = Schemas::HowTo.how_to_images.map { |id, image|
```
hannako: Is it OK to have this logic in the view? Would it be better to make a step_nav helper and put the logic there instead? Or are helpers a bit uncool because they're included in every view?

sihugh: I think it's OK.  It'd be nice to remove this stuff from the views altogether and generate the image_urls somewhere else ([like this for example](https://stackoverflow.com/a/31681098/170864)) as you say.  We've had problems in the past of the helpers not generating the hashed asset URLs properly though, so have resorted to this while we try to figure it out.

sihugh: I'm going to give it another crack though, thanks for the prod

sihugh: Should be better now

PR #1324, opened 2019-10-15
Rescue 403 errors

leenagupte:
Trello: https://trello.com/c/DAqDo4oW
Follows on from: [RFC 113: Expanding draft access for unauthenticated users to allow multi-page fact checks](https://github.com/alphagov/govuk-rfcs/blob/master/rfc-113-expanding-draft-access-for-unauthenticated-users.md)

## What's changed?
Adds error handling to handle 403 errors thrown by content-store

## Why?
Users should be redirect to a signon page if they are trying to access a content item on the draft stack and they are either not signed in, do not have a valid auth_bypass token, of the content item is restricted in some other way. They are instead being shown a "Problem has occurred" page because 403s are not being handled correctly.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -7,6 +7,7 @@ class ApplicationController < ActionController::Base
   before_action :restrict_request_formats
 
   rescue_from GdsApi::ContentStore::ItemNotFound, with: :error_404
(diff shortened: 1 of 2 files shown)


ChrisBAshton:
Tested on integration 👍 works great.

PR #1327, opened 2019-10-18
Add Welsh translation for Brexit page

emmabeynon:
For https://trello.com/c/3DUTcMYK/104-build-welsh-language-landing-page

This adds a Welsh translation of '/brexit'.  We have amended the route and controller to handle a locale so that the correct translation is rendered.

While some of the page copy can be added to a translation file, there are certain sections of the page that are not well set up to handle translations.  The Brexit Landing Page Presenter has been amended to work with a locale so that the correct translations can be rendered for buckets (note a follow up piece of tidying-up work will be needed to reorganise translation files for the landing page, including the buckets translation files).  The supergroup sections proved quite tricky to work with as they don't use translation files at all.  As a quick solution I've added a method to each supergroup presenter which returns the Welsh translation of the supergroup title so that it can be displayed on the Welsh translation page.


app/presenters/brexit_landing_page_presenter.rb
```diff
@@ -31,6 +31,10 @@ def supergroup_sections
     end
   end
 
+  def email_path
+    base_path.gsub(/(\.\w{2})$/, "")
```
benthorner: Minor: we could do `.split('.').first` here.

app/presenters/supergroups/supergroup.rb
```diff
@@ -40,6 +40,10 @@ def data_module_label
       name.camelize(:lower)
     end
 
+    def cy_title
+      raise NotImplementedError
```
benthorner: I wonder if we could avoid this by changing [the title method](https://github.com/alphagov/collections/blob/7c3229158ade1c847c273326d82eed2c682c4e48/app/presenters/supergroups/supergroup.rb#L10) to use locales. All it really does is change things like 'research_and_statistics' into 'Research and statistics'. This would reduce the total changes and complexity, not that it's a major burden or anything. It would also be reasonable to avoid [the additional test](https://github.com/alphagov/collections/pull/1327/files#diff-48963a9d9fa925c5dbdcfeb1323aeb10R106).

app/presenters/brexit_landing_page_presenter.rb
```diff
@@ -27,10 +31,30 @@ def supergroup_sections
...
+  end
+
+  def translation_links
+    links = []
+    taxon.translations.each do |link|
```
koetsier: could you use ```map``` here?

app/controllers/brexit_landing_page_controller.rb
```diff
@@ -31,4 +33,12 @@ def presentable_section_items
   def show_dynamic_list?
     true
   end
+
+  def set_locale
+    I18n.locale = locale
```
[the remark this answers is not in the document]
emmabeynon: Nice, I didn't know about this.  This makes it a bit tidier for sure 👍

app/presenters/brexit_landing_page_presenter.rb
emmabeynon: Done 👍

app/presenters/brexit_landing_page_presenter.rb
emmabeynon: Done 👍

app/presenters/supergroups/supergroup.rb
emmabeynon: I've added these as translation now.

emmabeynon:
@benthorner Thanks Ben, your comments should be addressed now.

app/services/supergroup_sections/brexit_sections.rb
```diff
@@ -10,13 +10,17 @@ def sections
...
+          title: I18n.t("brexit_sections.#{supergroup.name}.title"),
         }
       end
     end
 
   private
 
     attr_reader :taxon_id, :base_path
+
+    def cy_locale?
```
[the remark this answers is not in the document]
emmabeynon: Good catch!

PR #1358, opened 2019-11-11
Update person template header to include roles

MuriloDalRi:
Update the header of a person template to include roles so it matches the design of the page we have in Whitehall.

[Trello](https://trello.com/c/iNSK7NVU/1506-3-update-person-template-header-to-include-roles)

### DIFF
--- a/app/models/person.rb
+++ b/app/models/person.rb
@@ -22,8 +22,13 @@ class Person
     @content_item.content_item_data["title"]
   end
 
-  def roles
-    links["ordered_current_appointments"].map { |ra| ra["title"] }.to_sentence
+  def current_roles
+    links.fetch("ordered_current_appointments", [])
+      .map { |appointment| appointment["links"]["role"].first }
+  end
+
+  def current_roles_title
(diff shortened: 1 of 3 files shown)


app/models/person.rb
```diff
@@ -22,8 +22,19 @@ def title
...
+    return unless appointments
+
+    if appointments.include?(%w[links])
+      roles = appointments.map { |ra| ra["links"]["role"] }.map do |role|
+        role.map { |r| r["title"] }.join(",")
+      end
+
+      roles.to_sentence
+    else
+      appointments.map { |ra| ra["title"] }.map { |t| t.split(" - ").last }.to_sentence
```
thomasleese: This should never be the case - is there an example content item that is missing the links?

MuriloDalRi: https://www.gov.uk/api/content/government/people/aarti-thakor

thomasleese: Thanks. I'll investigate why this is the case. In the mean time you don't need to worry about supporting that kind of content item, it's incorrect.

It might also be worth adding a test with an example correct content item.

PR #1360, opened 2019-11-13
Update person template to include previous appointments

MahmudH:
This PR allows previous roles on people pages to be displayed:
Screen Shot 2019-11-13 at 13 25 25

This data is retrieved from the content store.

Trello: 
https://trello.com/c/cI4gRrdt/1508-3-update-person-template-to-include-previous-roles

### DIFF
--- a/app/models/person.rb
+++ b/app/models/person.rb
@@ -38,6 +38,20 @@ class Person
     details["body"]
   end
 
+  def previous_appointments
+    ordered_previous_appointments.map do |role|
+      {
+        link: {
+          text: role["title"],
+          path: role["base_path"],
+        },
+        metadata: {
+          appointment_duration: "#{role['start_year']} to #{role['end_year']}",
+        },
+      }
+    end
+  end
+
 private
 
   def current_roles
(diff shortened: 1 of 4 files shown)


app/models/person.rb
```diff
@@ -52,4 +66,18 @@ def links
   def details
     @content_item.content_item_data["details"]
   end
+
+  def ordered_previous_appointments
+    previous_roles = []
+    links["ordered_previous_appointments"].each do |previous_appointment|
+      if previous_appointment["links"]["role"].present?
+        previous_appointment["links"]["role"].each do |role|
```
thomasleese: There [will actually only ever be one `role` here](https://github.com/alphagov/whitehall/blob/9291b03f5e986b894c35f5f8ffa6141216d0a506/app/presenters/publishing_api/role_appointment_presenter.rb#L28-L37) so I wouldn't worry about iterating over each one and just use `role = previous_appointment["links"]["role"].first`.

MahmudH: Thanks for pointing that out. I'll update this :)

app/models/person.rb
```diff
@@ -52,4 +66,15 @@ def links
   def details
     @content_item.content_item_data["details"]
   end
+
+  def ordered_previous_appointments
+    previous_roles = []
+    links["ordered_previous_appointments"].each do |previous_appointment|
```
thomasleese: You could use a `.map` here to avoid needing the `.each` and then `.push`.

For example:

```ruby
links["ordered_previous_appointments"].each do |previous_appointment|
  role = previous_appointment["links"]["role"].first
  role["start_year"] = Time.parse(previous_appointment["details"]["started_on"]).strftime("%Y")
  role["end_year"] = Time.parse(previous_appointment["details"]["ended_on"]).strftime("%Y")
  role
end
```

PR #1363, opened 2019-11-15
Various improvements to person page

thomasleese:
This includes various improvements to the people page, notably adding margins to the bottom of headings and supporting people without any ordered previous appointments.

See commits for individual details.

Screenshot 2019-11-15 at 14 27 40

### DIFF
--- a/app/views/people/show.html.erb
+++ b/app/views/people/show.html.erb
@@ -14,11 +14,11 @@
           text: "Biography",
           href: "#biography"
         },
-        {
+        person.has_previous_roles? ? {
           text: "Previous roles",
           href: "#previous-roles"
-        },
-      ]
+        } : nil,
+      ].compact
     } %>
   </div>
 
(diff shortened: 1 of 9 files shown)


app/views/people/show.html.erb
```diff
@@ -16,24 +13,16 @@
         {
           text: "Biography",
           href: "#biography"
-        }
-      ]
+        },
+        person.has_previous_roles? ? {
```
injms: The contents heading and links shouldn't be part of the the image card component. It means that 'Contents' isn't a heading and on smaller screens it sits next to the image rather than below it.

Before:
Screenshot 2019-11-15 at 14 57 19

After:
Screenshot 2019-11-15 at 14 57 09

thomasleese: Oh interesting! Is there an alternative component available or do we construct that from ordinary HTML?

injms: Hmm 🤔 the [contents list](https://components.publishing.service.gov.uk/component-guide/contents_list) component would be the best - there's a visual difference with the heading not being bolded, but it's a heading element rather than a `div`.

thomasleese: Thanks, I don't think there's an issue with the heading not being bolded as long as we're following the publishing components. I've pushed up a commit to change this!

PR #1365, opened 2019-11-19
Fix bottom padding for Biography section on the person page

callumknights:
Done to match whitehall.

Old collections spacing shown below:

image

Vs the current spacing on whitehall:

image

### DIFF
--- a/app/views/people/_biography.html.erb
+++ b/app/views/people/_biography.html.erb
@@ -1,10 +1,11 @@
-<%= render "govuk_publishing_components/components/heading", {
-  text: "Biography",
-  id: "biography",
-  margin_bottom: 2,
-} %>
+<section id="biography" class="govuk-!-padding-bottom-9">
+  <%= render "govuk_publishing_components/components/heading", {
+    text: "Biography",
+    margin_bottom: 2,
+  } %>
 
-<%= render "govuk_publishing_components/components/govspeak", {
-} do %>
-  <%= person.biography.html_safe %>
-<% end %>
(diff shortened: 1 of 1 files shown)


app/views/people/_biography.html.erb
```diff
@@ -1,10 +1,12 @@
-<%= render "govuk_publishing_components/components/heading", {
-  text: "Biography",
-  id: "biography",
-  margin_bottom: 2,
-} %>
+<section id="biography" class="govuk-!-padding-bottom-9">
+  <%= render "govuk_publishing_components/components/heading", {
+    text: "Biography",
+    id: "biography",
```
thomasleese:
```suggestion
```

I don't think we need this if the `section` has an `id` now.

app/views/people/_biography.html.erb
```diff
@@ -1,10 +1,12 @@
...
 
-<%= render "govuk_publishing_components/components/govspeak", {
-} do %>
-  <%= person.biography.html_safe %>
-<% end %>
+  <%= render "govuk_publishing_components/components/govspeak", {
+  } do %>
+    <%= person.biography.html_safe %>
+  <% end %>
+</section>
```
cbaines: The trailing newline is missing.

callumknights: 👍

app/views/people/_biography.html.erb
callumknights: 👍

PR #1367, opened 2019-11-19
Include announcements

MahmudH:
This PR updates the people template to include announcements, which are retrieved from search api. These results are presented by the document list component, which includes the title, public_timestamp and content_store_document_type.

At the bottom of the announcements, a link has been added to view all announcements, this is because the number of announcements displayed is 10 for each person. In order to view all announcements, the link redirects the user to the news and communications finder, filtered by person (the person they are viewing).

Screen Shot 2019-11-19 at 14 18 36

This PR also adds a link to announcements to sidebar.
Screen Shot 2019-11-19 at 14 51 36

Trello: https://trello.com/c/YSTcpIwP/1509-5-update-person-template-to-include-announcements

### DIFF
--- a/app/models/person.rb
+++ b/app/models/person.rb
@@ -79,6 +79,14 @@ class Person
     "/search/news-and-communications?people=#{slug}"
   end
 
+  def email_signup
+    "/email-signup?link=/government/people/#{slug}"
+  end
+
+  def subscribe_to_feed
+    "https://www.gov.uk/government/people/#{slug}.atom"
+  end
+
 private
 
   def slug
--- a/app/models/person.rb
+++ b/app/models/person.rb
@@ -56,7 +56,40 @@ class Person
     details["body"]
   end
 
+  def announcement_items
+    announcements.map do |announcenment|
+      {
+        link: {
+          text: announcenment["title"],
+          path: announcenment["link"],
+        },
+        metadata: {
+          public_timestamp: Date.parse(announcenment["public_timestamp"]).strftime("%d %B %Y"),
+          content_store_document_type: announcenment["content_store_document_type"].humanize,
+        },
+      }
+    end
+  end
+
+  def has_announcements?
+    announcements.present?
+  end
+
 private
+
+  def announcements
+    @announcements ||= Services.cached_search(
+      count: 10,
+      order: "-public_timestamp",
+      filter_people: slug,
+      reject_content_purpose_supergroup: "other",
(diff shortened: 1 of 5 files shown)


app/views/people/show.html.erb
```diff
@@ -19,14 +19,20 @@
...
     } %>
   </div>
 
   <div class="govuk-grid-column-two-thirds">
     <%= render partial: "biography", locals: { person: person } %>
 
-    <%= render partial: "previous_roles", locals: { person: person } %>
+  <%= render partial: "previous_roles", locals: { person: person } %>
+
+  <%= render partial: "announcements", locals: { person: person } %>
```
cbaines: The indentation is slightly off here.

app/models/person.rb
```diff
@@ -56,8 +56,45 @@ def biography
     details["body"]
   end
 
+  def announcement_items
+    announcements.map do |annoucenment|
```
thomasleese:
```suggestion
    announcements.map do |announcement|
```

:)

app/models/person.rb
```diff
@@ -56,8 +56,45 @@ def biography
...
+        },
+      }
+    end
+  end
+
+  def has_announcements?
+    announcements.present?
+  end
+
+  def redirect_to_news_and_communications
```
thomasleese:
```suggestion
  def link_to_news_and_communications
```

Just because this method isn't actually performing any redirect.

app/models/person.rb
```diff
@@ -56,8 +56,45 @@ def biography
...
+  end
+
+  def announcements
+    @announcements ||= Services.cached_search(
+      count: 10,
+      order: "-public_timestamp",
+      filter_people: slug,
+      reject_content_purpose_supergroup: "other",
+      fields: %w[title link content_store_document_type public_timestamp],
+      )["results"]
```
thomasleese:
```suggestion
    )["results"]
```

app/models/person.rb
MahmudH: Makes sense so I've updated it

app/models/person.rb
MahmudH: well spotted!

app/views/people/show.html.erb
MahmudH: Good spot! I'll update now

app/models/person.rb
MahmudH: 👍

PR #1369, opened 2019-11-19
Update person template to include current roles

callumknights:
https://trello.com/c/2P3IR10o/1507-3-update-person-template-to-include-current-roles

Done as part of the whitehall migration work.

Shown below is how the page now looks rendered on using collections.

image

### DIFF
--- a/app/models/person.rb
+++ b/app/models/person.rb
@@ -102,6 +102,11 @@ class Person
     current_role_appointments.any?
   end
 
+  def current_roles
+    current_role_appointments
+        .map { |appointment| appointment["links"]["role"].first }
+  end
+
 private
 
   def slug
@@ -131,11 +136,6 @@ private
 
   def previous_role_appointments
     @previous_role_appointments ||= role_appointments(current: false)
-  end
-
-  def current_roles
-    current_role_appointments
-      .map { |appointment| appointment["links"]["role"].first }
   end
 
   def previous_roles
--- a/app/views/people/_current_roles.html.erb
+++ b/app/views/people/_current_roles.html.erb
@@ -1,0 +1,22 @@
+<section id="current-roles">
+  <% person.current_roles.each do |role| %>
+    <section class="role_appointment role govuk-!-padding-bottom-8">
+    <%= render "govuk_publishing_components/components/heading", {
+        text: person.current_roles_title,
+        id: person.current_roles_title.parameterize,
+        margin_bottom: 2,
+    } %>
+    <%= render "govuk_publishing_components/components/govspeak", {
+    } do %>
+      <%= role["details"]["body"].html_safe %>
+    <% end  %>
+
+    <% if role["document_type"] == "ministerial_role" %>
+      <p class="govuk-body-s"><a href=<%= role["base_path"] %>>More about this role</a></p>
+    <% end %>
+
+    <% parent_organisations = role["links"]["ordered_parent_organisations"].map { |parent| link_to(parent["title"], parent["base_path"]) }%>
+    <%= parent_organisations.to_sentence.html_safe %>
+    </section>
+  <% end %>
+</section>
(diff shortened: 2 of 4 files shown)


cbaines:
Brakeman seems unhappy:
```
== Overview ==

Controllers: 18
Models: 21
Templates: 80
Errors: 0
Security Warnings: 1
Ignored Warnings: 3

== Warning Types ==

Cross-Site Scripting: 1

== Warnings ==

Confidence: High
Category: Cross-Site Scripting
Check: CrossSiteScripting
Message: Unescaped model attribute
Code: (Unresolved Model).new["links"]["role"][0]["details"]["body"]
File: app/views/people/_current_roles.html.erb
Line: 12
```

app/views/people/_current_roles.html.erb
```diff
@@ -0,0 +1,24 @@
...
+    } do %>
+      <%= role["details"]["body"].html_safe %>
+    <% end  %>
+
+    <% if role["document_type"] == "ministerial_role" %>
+      <p class="govuk-body-s"><a href=<%= role["base_path"] %>>More about this role</a></p>
+    <% end %>
+
+    <% parent_organisations = [] %>
+    <% role["links"]["ordered_parent_organisations"].map { | parent | parent_organisations << link_to(parent["title"], parent["base_path"]) }%>
```
cbaines: I'd change `| parent |` to `|parent|` as I think that's neater.

cbaines: and more consistent.

app/views/people/show.html.erb
```diff
@@ -27,6 +27,9 @@
   <div class="govuk-grid-column-two-thirds">
     <%= render partial: "biography", locals: { person: person } %>
 
+    <%= render partial: "current_roles", locals: { person: person} %>
+
     <%= render partial: "previous_roles", locals: { person: person } %>
   </div>
 </div>
+
```
cbaines: You're adding a empty line at the end of this file, which seems a bit redundant.

cbaines:
Could you include a screenshot in the Pull Request description.

Also, I think the Whitehall rendered Person page includes an entry for the current roles in the Contents. I don't know if you want to add that here, or in a different Pull Request, but it still needs doing.

app/views/people/_current_roles.html.erb
thomasleese:
```suggestion
    <% parent_organisations = role["links"]["ordered_parent_organisations"].map { | parent | link_to(parent["title"], parent["base_path"]) }%>
```

Because you're using `map` you don't need to add to an array in each iteration as map does that for you.

app/models/person.rb
```diff
@@ -56,6 +56,10 @@ def biography
     details["body"]
   end
 
+  def ordered_current_appointments
```
thomasleese: I've [had to change the way the links are presented again](https://github.com/alphagov/collections/pull/1371#issuecomment-556978645), unfortunately!

When you rebase, you'll see there's a method you can use instead of this one though called `current_role_appointments`.

app/views/people/_current_roles.html.erb
callumknights: Thanks Tom 👍

app/models/person.rb
```diff
@@ -102,6 +102,15 @@ def currently_in_a_role?
     current_role_appointments.any?
   end
 
+  def current_role_appointments
```
thomasleese: I don’t think this needs to be public now that we’re not using it in the view.

app/views/people/show.html.erb
```diff
@@ -18,6 +18,10 @@
           text: "Biography",
           href: "#biography"
         },
+        person.currently_in_a_role? ? {
+            text: "Current Roles",
```
thomasleese: I think the current page in Whitehall shows this as “Roles”.

PR #1372, opened 2019-11-21
Add translations to person pages

MuriloDalRi:
Add switch language buttons to the top of the person template.

Screenshot 2019-11-21 at 15 54 51

https://trello.com/c/bZoNuOvw/1556-2-add-switch-language-links-to-people-pages


config/locales/en/shared.yml
```diff
@@ -3,5 +3,64 @@ en:
...
+      th: ไทย
+      tk: Türkmençe
+      tr: Türkçe
+      uk: Українська
+      ur: اردو
+      uz: O'zbekcha
+      vi: Tiếng Việt
+      zh: 中文
+      zh-hk: 中文
+      zh-tw: 中文
```
cbaines: The trailing newline is missing here.

app/models/person.rb
```diff
@@ -56,6 +56,10 @@ def biography
     details["body"]
   end
 
+  def translations
+    available_translations.each { |t| t[:text] = language_name(t[:locale]) }
```
thomasleese: I found this a bit confusing to read, it might be better to use `each_with_object`.

```suggestion
    available_translations.each_with_object do |translation, memo|
      memo[:locale] = translation[:locale]
      memo[:base_path] = translation[:base_path
      memo[:text] = language_name(translation[:locale])
    end
```

I prefer this because it's explicit as well, we can be sure we'll have the right fields in the dictionary.

thomasleese: Also, it would be good to support the `:active` key as well as per https://components.publishing.service.gov.uk/component-guide/translation-nav so the current locale isn't clickable.

MuriloDalRi: I thought it would be more "ruby" to keep it short but your suggestion does make it clearer! 👍 

I'll look into the getting the `active` thing working

app/models/person.rb
```diff
@@ -56,7 +56,17 @@ def biography
...
 
-private
+  def translations
+    available_translations.map do |t|
+      t[:locale] = t[:locale]
+      t[:base_path] = t[:base_path]
+      t[:text] = language_name(t[:locale])
+      t[:active] = true if locale == t[:locale]
+      t
+    end
```
thomasleese: I think it's better to be explicit with the hash here, something like this:

```ruby
available_translations.map do |translation|
  translation
    .slice(:locale, :base_path)
    .merge(
      text: language_name(translation[:locale]),
      active: locale == translation[:locale],
    )
end
```

The lines `t[:locale] = t[:locale]`, and `t[:base_path] = t[:base_path]` look a bit odd! :smile:

MuriloDalRi: I like that

PR #1374, opened 2019-11-21
Add organisations to role pages

MahmudH:
The organisations data is retrieved from the content store and displayed into a link, as there can be more than one organisation present, the to_sentence ruby function is used for readability.

Screenshot:
Screen Shot 2019-11-21 at 16 46 13

Trello: https://trello.com/c/IyNL6Q2F/1557-2-add-mini-organisations-to-role-pages

### DIFF
--- a/app/models/role.rb
+++ b/app/models/role.rb
@@ -26,6 +26,10 @@ class Role
     details["body"]
   end
 
+  def organisations
+    links["ordered_parent_organisations"]
+  end
+
 private
 
   def links
--- a/app/views/roles/_organisations.html.erb
+++ b/app/views/roles/_organisations.html.erb
@@ -1,0 +1,6 @@
+<% if role.organisations %>
+  <div class="organisations-list">
+    Organisations:
+    <%= role.organisations.map { |organisation| link_to(organisation['title'], organisation['base_path']) }.to_sentence.html_safe %>
+  </div>
+<% end %>
--- a/test/models/role_test.rb
+++ b/test/models/role_test.rb
@@ -1,0 +1,39 @@
+require "test_helper"
+
+describe Role do
+  setup do
+    @api_data = {
+      "base_path" => "/government/ministers/prime-minister",
+      "links" => {
+        "ordered_parent_organisations" => [
+          {
+            "base_path" => "/government/organisations/cabinet-office",
+            "title" => "Cabinet Office",
+          },
+          {
+            "base_path" => "/government/organisations/prime-ministers-office-10-downing-street",
+            "title" => "Prime Minister's Office, 10 Downing Street",
+          },
+        ],
+      },
+    }
+    @content_item = ContentItem.new(@api_data)
+    @role = Role.new(@content_item)
+  end
+
+  describe "organisations" do
+    it "should have organisations title and base_path" do
+      expected = [
+        {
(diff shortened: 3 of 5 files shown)


app/models/role.rb
```diff
@@ -26,6 +26,14 @@ def responsibilities
     details["body"]
   end
 
+  def organisations
+    links["organisations"]
```
```suggestion
    links["ordered_parent_organisations"]
```

`organisations` is actually an older link type which is deprecated now, instead we should use `ordered_parent_organisations` because it makes sure the organisations will appear in the right order.

app/models/role.rb
```diff
@@ -26,6 +26,14 @@ def responsibilities
...
+  end
+
+  def organisations_present?
```
```suggestion
  def has_organisations?
```

test/models/role_test.rb
```diff
@@ -0,0 +1,39 @@
+require "test_helper"
+
+describe Role do
```
thomasleese: Nice tests!

app/views/roles/_organisations.html.erb
```diff
@@ -0,0 +1,9 @@
+<% if role.organisations_present? %>
+    <div class="organisations-list">
+      <div class="organisations-list">
+        Organisations:
+        <%= role.organisations.map{ |organisation| link_to(organisation['title'], organisation['base_path']) }.to_sentence.html_safe %>
```
thomasleese:
```suggestion
        <%= role.organisations.map { |organisation| link_to(organisation['title'], organisation['base_path']) }.to_sentence.html_safe %>
```

There should be a space after `map` :)

app/views/roles/_organisations.html.erb
```diff
@@ -0,0 +1,9 @@
+<% if role.organisations_present? %>
+    <div class="organisations-list">
+      <div class="organisations-list">
```
thomasleese: It looks like these are the same divs. Do we use the `organisations-list` class anywhere as well?

MahmudH: Oh you're right! No we don't use it anywhere else. There are others similar to it for the Organisation page but not the same styling. Do you think it would be better to call it `roles-organisations-list`?

thomasleese: I was thinking if we don't use it anywhere we could just leave it as a `<div>` without a class.

MahmudH: Ok 👍

app/views/roles/_organisations.html.erb
MahmudH: 👍

app/models/role.rb
MahmudH: 👍

app/models/role.rb
MahmudH: Thanks, I'll update this 👍

app/models/role.rb
```diff
@@ -26,6 +26,14 @@ def responsibilities
     details["body"]
   end
 
+  def organisations
+    links["ordered_parent_organisations"]
+  end
+
+  def has_organisations?
+    organisations.present?
+  end
```
AlanGabbianelli: Do we need this method? Can't we use `<% if role.organisations %>` instead?

PR #1377, opened 2019-11-25
Add mini current role holder to role pages

AlanGabbianelli:
The current role holder data is retrieved from the content store and
displayed into a link.

[Trello card](https://trello.com/c/jRxT5RZc/1558-1-add-mini-current-role-holder-to-role-pages)

### DIFF
--- a/app/models/role.rb
+++ b/app/models/role.rb
@@ -15,11 +15,11 @@ class Role
   end
 
   def base_path
-    @content_item.content_item_data["base_path"]
+    content_item_data["base_path"]
   end
 
   def title
-    @content_item.content_item_data["title"]
+    content_item_data["title"]
   end
 
   def responsibilities
@@ -32,11 +32,15 @@ class Role
 
 private
 
+  def content_item_data
+    content_item.content_item_data
+  end
+
   def links
-    @content_item.content_item_data["links"]
+    content_item_data["links"]
   end
 
   def details
-    @content_item.content_item_data["details"]
+    content_item_data["details"]
   end
 end
--- a/app/models/role.rb
+++ b/app/models/role.rb
@@ -30,6 +30,14 @@ class Role
     links["ordered_parent_organisations"]
   end
 
+  def current_holder
+    links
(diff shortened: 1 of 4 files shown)


app/models/role.rb
```diff
@@ -30,13 +30,25 @@ def organisations
     links["ordered_parent_organisations"]
   end
 
+  def current_holder
+    links["role_appointments"]
+      .select { |ra| ra.dig("details", "current") == true }
```
```suggestion
      .select { |ra| ra.dig("details", "current") }
```

Since it's already a boolean, you won't need this!

app/models/role.rb
```diff
@@ -30,13 +30,25 @@ def organisations
...
+      .first
```
thomasleese: You can combine the calls to `.select({ ... }).first` [into a single call to `.find`](https://ruby-doc.org/core-2.6.5/Enumerable.html#method-i-find) (avoids unnecessarily performing the `select` block to the later elements in the array).

app/models/role.rb
```diff
@@ -30,13 +30,25 @@ def organisations
...
+      .dig("links", "person")
+      .first
+  end
+
 private
 
+  def content_item_data
```
thomasleese: Nice 👍

app/models/role.rb
```diff
@@ -30,13 +30,25 @@ def organisations
     links["ordered_parent_organisations"]
   end
 
+  def current_holder
+    links["role_appointments"]
```
```suggestion
    links.fetch("role_appointments", [])
```

Although unlikely, it's possible that a ministerial role could be created and not yet assigned any people.

app/models/role.rb
AlanGabbianelli: Of course! 😆

app/models/role.rb
AlanGabbianelli: Very true! 👍

app/models/role.rb
AlanGabbianelli: I thought that wasn't possible. 🤔 I'll change it.

PR #1380, opened 2019-11-26
Add current role holder and biography

MahmudH:
This PR adds current role, biography and a link to the person, which is retrieved from the content store. 

Screenshot:
Screen Shot 2019-11-26 at 11 15 18

Trello: https://trello.com/c/02C3OKnk/1559-2-add-current-role-holder-to-role-pages

### DIFF
--- a/app/models/role.rb
+++ b/app/models/role.rb
@@ -38,6 +38,14 @@ class Role
       .first
   end
 
+  def current_holder_biography
+    current_holder["details"]["body"]
+  end
+
+  def link_to_person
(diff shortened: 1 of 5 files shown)


app/models/role.rb
```diff
@@ -38,6 +38,14 @@ def current_holder
       .first
   end
 
+  def biography
```
```suggestion
  def current_holder_biography
```

It might be better to call it this just because a role doesn't have a biography on its own.

MahmudH: Ok sounds good to me 👍

PR #1389, opened 2019-11-29
Allow scripts to be rendered Right to Left

edwardkerry:
If a language should read Right to Left, this PR allows the shared
`config/locales/en/shared.yml` to be updated with a direction.

When rendering, required divs can use the application helper method
`page_text_direction` to set a `dir` attribute on an html element which will apply
relevant text direction to the contents..

This follows similar patterns to those in government-frontend and whitehall
https://github.com/alphagov/government-frontend/blob/71927708e97e4d7375ca719cfa861f74e3d7c024/app/helpers/application_helper.rb#L2

This enables the functionality, but individual languages will need to be identified and added to the configuration

Before
Screenshot 2019-11-29 at 13 52 47

After
Screenshot 2019-11-29 at 13 52 31

[trello](https://trello.com/c/pkVccAv2/1586-3-render-people-pages-right-to-left-for-approriate-languages)

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -1,4 +1,8 @@
 module ApplicationHelper
+  def page_text_direction
+    I18n.t("shared.language_direction.#{@content_item['locale']}", default: "ltr")
+  end
+
   def browsing_in_top_level_page?(top_level_page)
     request.path.starts_with?(top_level_page.base_path)
   end
--- a/app/views/people/show.html.erb
+++ b/app/views/people/show.html.erb
@@ -34,7 +34,7 @@
     } %>
   </div>
 
-  <div class="govuk-grid-column-two-thirds">
+  <div class="govuk-grid-column-two-thirds" dir="<%=page_text_direction%>">
     <%= render partial: "biography", locals: { person: person } %>
 
     <%= render partial: "current_roles", locals: { person: person } if person.currently_in_a_role? %>
--- a/config/locales/en/shared.yml
+++ b/config/locales/en/shared.yml
@@ -64,3 +64,5 @@ en:
       zh: 中文
       zh-hk: 中文
       zh-tw: 中文
+    language_direction:
+      ar: rtl
--- a/config/locales/en/shared.yml
+++ b/config/locales/en/shared.yml
@@ -66,3 +66,8 @@ en:
       zh-tw: 中文
     language_direction:
       ar: rtl
+      fa: rtl
+      ur: rtl
+      he: rtl
+      dr: rtl
+      ps: rtl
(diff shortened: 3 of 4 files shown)


config/locales/en/shared.yml
```diff
@@ -64,3 +64,5 @@ en:
       zh: 中文
       zh-hk: 中文
       zh-tw: 中文
+    language_direction:
+      ar: rtl
```
MuriloDalRi: There are other languages written right to left such as Hebrew, Persian and Urdu. I don't know if we have any translations in those languages but it would be nice to add support

MuriloDalRi: Let's not even get into top-to-bottom writing!

app/helpers/application_helper.rb
```diff
@@ -1,4 +1,12 @@
 module ApplicationHelper
+  def page_text_direction
+    I18n.t("shared.language_direction.#{@content_item["locale"]}", default: "ltr")
+  end
+
+  def text_direction_wrapper_class
```
injms: Though a class is useful, this could also be a [`dir` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/dir) on the wrapping element:

```html
<p dir="ltr">This is left to right</p>
```

If the CSS doesn't load, then the browser can still display correctly.

If you wanted to remove the class completely, you can still style the element to make it go right to left:

```css
[dir="rtf"] {
  direction: rtl;
  text-align: start;
}
```

config/locales/en/shared.yml
edwardkerry: Yes, I'm not sure where to get a definitive list from, but I agree. Perhaps these can be added in a later PR when I have found the details

app/helpers/application_helper.rb
edwardkerry: Thanks @injms I had just noticed that and pushed up a simplified commit using the `dir` attribute :+1:

config/locales/en/shared.yml
MuriloDalRi: Sounds sensible. I found this page: https://www.w3.org/International/questions/qa-scripts.en

app/helpers/application_helper.rb
injms: Super!

config/locales/en/shared.yml
edwardkerry: i've added some more harvested from whitehall and government-frontend

app/views/people/show.html.erb
```diff
@@ -34,7 +34,7 @@
     } %>
   </div>
 
-  <div class="govuk-grid-column-two-thirds">
+  <div class="govuk-grid-column-two-thirds" dir="<%=page_text_direction%>">
```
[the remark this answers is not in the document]
injms: Rather than moving it up the DOM it's better to add it to the specific areas that need it, even if that means adding the `dir` attribute to a larger number of elements.

PR #1395, opened 2019-12-02
Small fixes to people page templates

thomasleese:
This tidies up some of the code in the templates for people pages and fixes a few small issues. See individual commits for more details.

### DIFF
--- a/app/views/people/_current_roles.html.erb
+++ b/app/views/people/_current_roles.html.erb
@@ -17,10 +17,12 @@
           <p class="govuk-body-s"><a href=<%= role["base_path"] %>>More about this role</a></p>
         <% end %>
 
-        <%= role["links"].fetch("ordered_parent_organisations", [])
-          .map { |parent| link_to(parent["title"], parent["base_path"]) }
-          .to_sentence
-          .html_safe %>
+        <p class="govuk-body">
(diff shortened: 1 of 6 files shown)


app/views/people/_current_roles.html.erb
```diff
@@ -1,22 +1,27 @@
...
+        <% if role["document_type"] == "ministerial_role" %>
+          <p class="govuk-body-s"><a href=<%= role["base_path"] %>>More about this role</a></p>
+        <% end %>
 
-    <% parent_organisations = role["links"]["ordered_parent_organisations"].map { |parent| link_to(parent["title"], parent["base_path"]) }%>
-    <%= parent_organisations.to_sentence.html_safe %>
-    </section>
-  <% end %>
-</section>
+        <%= role["links"].fetch("ordered_parent_organisations", [])
```
vanitabarrett: This may need the `govuk-body` class on it (or it's parent section) so that it gets our usual font styling and matches other text on the page.

It'll also need `govuk-link` to get the GOVUK Frontend hover, visited and focus styles.

thomasleese: Thanks! I've pushed up a commit which fixes this.

PR #1399, opened 2019-12-03
Render translations with right to left scripts

edwardkerry:
This PR allows any role pages with a translation in a right-to-left script to be rendered accurately.

It also adjusts the right-to-left rendering for people, adding the necessary `dir` attribute to individual elements rather than a shared parent element. 

For accessibility the `lang` attribute is used, which required making a previously private `locale` method public.

### DIFF
--- a/app/views/people/_announcements.html.erb
+++ b/app/views/people/_announcements.html.erb
@@ -1,5 +1,5 @@
 <% if person.has_announcements? %>
-  <section id="announcements">
+  <section id="announcements" dir="<%= page_text_direction %>" lang="<%= person.locale %>">
     <%= render "govuk_publishing_components/components/heading", {
         text: "Announcements",
         id: "announcements",
--- a/app/views/people/_biography.html.erb
+++ b/app/views/people/_biography.html.erb
@@ -1,4 +1,4 @@
-<section id="biography" class="govuk-!-padding-bottom-9">
+<section id="biography" class="govuk-!-padding-bottom-9" dir="<%= page_text_direction %>" lang="<%= person.locale %>">
   <%= render "govuk_publishing_components/components/heading", {
     text: "Biography",
     margin_bottom: 2,
--- a/app/views/people/_current_roles.html.erb
+++ b/app/views/people/_current_roles.html.erb
@@ -1,5 +1,5 @@
 <% if person.currently_in_a_role? %>
-  <section id="current-roles">
+  <section id="current-roles" dir="<%= page_text_direction %>" lang="<%= person.locale %>">
     <% person.current_roles.each do |role| %>
       <section class="role_appointment role govuk-!-padding-bottom-8">
         <%= render "govuk_publishing_components/components/heading", {
--- a/app/views/people/_previous_roles.html.erb
+++ b/app/views/people/_previous_roles.html.erb
@@ -1,5 +1,5 @@
 <% if person.has_previous_roles? %>
-  <section class="previous-roles govuk-!-padding-bottom-9">
+  <section class="previous-roles govuk-!-padding-bottom-9" dir="<%= page_text_direction %>" lang="<%= person.locale %>">
     <%= render "govuk_publishing_components/components/heading", {
       text: "Previous roles in government",
       id: "previous-roles",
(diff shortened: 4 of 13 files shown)


app/views/roles/_organisations.html.erb
```diff
@@ -1,5 +1,5 @@
 <% if role.organisations %>
-  <div class="organisations-list">
+  <div class="organisations-list" dir="<%= page_text_direction %>" lang="<%= role.locale %>">
```
injms: This shouldn't be set `rtl`, since it contains a left-to-right language ('Organisations').

app/views/roles/show.html.erb
```diff
@@ -7,7 +7,7 @@
 <%= render partial: "current_holder", locals: { role: role } %>
 
 <div class="govuk-grid-row">
-  <div class="govuk-grid-column-two-thirds">
+  <div class="govuk-grid-column-two-thirds" dir="<%= page_text_direction %>" lang="<%= role.locale %>">
```
injms: Same here - this shouldn't be `rtl` as it contains a left-to-right language and the lang is always added regardless of the default language.

app/views/roles/_header.html.erb
```diff
@@ -1,4 +1,4 @@
-<header class="govuk-grid-row">
+<header class="govuk-grid-row" dir="<%= page_text_direction %>" lang="<%= role.locale %>">
```
injms: This shouldn't be rtl since it contains a left to right language - in this case, 'Ministerial role'.

app/views/people/_announcements.html.erb
```diff
@@ -1,5 +1,5 @@
 <% if person.has_announcements? %>
-  <section id="announcements">
+  <section id="announcements" dir="<%= page_text_direction %>" lang="<%= person.locale %>">
```
injms: Since this contains a left to right language (in this instance, 'Announcements') that isn't translated it shouldn't be displayed right-to-left.

It's also putting out lang regardless of the default language, which seems a bit messy - on frontend there's a helper that adds the lang attribute only when the language is different to the default language (https://github.com/alphagov/frontend/blob/master/app/helpers/locale_helper.rb) which would prevent attributes being added when they're not needed.

app/views/people/_biography.html.erb
```diff
@@ -1,4 +1,4 @@
-<section id="biography" class="govuk-!-padding-bottom-9">
+<section id="biography" class="govuk-!-padding-bottom-9" dir="<%= page_text_direction %>" lang="<%= person.locale %>">
```
injms: Same here - this shouldn't have right to left since it contains text that won't be translated.

app/views/people/_current_roles.html.erb
```diff
@@ -1,5 +1,5 @@
 <% if person.currently_in_a_role? %>
-  <section id="current-roles">
+  <section id="current-roles" dir="<%= page_text_direction %>" lang="<%= person.locale %>">
```
injms: This shouldn't be set as right to left, since it could contain 'More about this role' in English.

app/views/people/_previous_roles.html.erb
```diff
@@ -1,5 +1,5 @@
 <% if person.has_previous_roles? %>
-  <section class="previous-roles govuk-!-padding-bottom-9">
+  <section class="previous-roles govuk-!-padding-bottom-9" dir="<%= page_text_direction %>" lang="<%= person.locale %>">
```
injms: This shouldn't be rtl, as it contains untranslated content in a left-to-right language.

app/views/roles/_current_holder.html.erb
```diff
@@ -1,5 +1,5 @@
 <% if role.current_holder %>
-  <div>
+  <div dir="<%= page_text_direction %>" lang="<%= role.locale %>">
```
injms: This shouldn't be rtl, as it contains a left to right language.

chao-xian:
Hi @edwardkerry - thanks for your work on this migration. The changes @injms suggests are to maintain this template is WCAG 2.1 compliant. I understand that Ian's going to be providing further help on this. It would be great to keep this template accessible.

edwardkerry:
Thanks @chao-xian and @injms, would be good to sit with you to make sure these changes are correct and sensible. These templates are still being rendered by Whitehall-frontend so there is time to iterate and improve these before we go live.

PR #1402, opened 2019-12-04
Add contents to roles

callumknights:
Add Contents list to the Roles page in Collections

https://trello.com/c/cp4b3mue/1589-1-add-contents-list-to-role-pages

Part of the Whitehall migration work

This is how the page now looks with the Contents list showing

image

I have chosen to leave out "Announcements", "Previous Roles" and "Policies" as those features have not yet been implemented in Collections, and adding them to the Contents list actually requires doing a fair portion of the work to add the feature as a whole. It makes sense to me to leave adding these items for now and do them when their corresponding features are migrated into Collections.

### DIFF
--- a/app/models/role.rb
+++ b/app/models/role.rb
@@ -30,6 +30,10 @@ class Role
     links["ordered_parent_organisations"]
   end
 
+  def currently_occupied?
+    current_holder.present?
+  end
+
   def current_holder
     links
       .fetch("role_appointments", [])
--- a/app/views/roles/show.html.erb
+++ b/app/views/roles/show.html.erb
@@ -7,6 +7,20 @@
 <%= render partial: "current_holder", locals: { role: role } %>
 
 <div class="govuk-grid-row">
+  <div class="govuk-grid-column-one-third">
+    <%= render "govuk_publishing_components/components/contents_list", {
+      contents: [
+        {
+          text: "Responsibilities",
+          href: "#responsibiities"
+        },
+        role.currently_occupied? ? {
+          text: "Current Role Holder",
+          href: "#current-role-holder"
+        } : nil
+      ].compact
+    } %>
+  </div>
   <div class="govuk-grid-column-two-thirds" dir="<%= page_text_direction %>" lang="<%= role.locale %>">
     <%= render "govuk_publishing_components/components/heading", {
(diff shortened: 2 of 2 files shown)


app/views/roles/show.html.erb
```diff
@@ -7,7 +7,22 @@
...
+    <%= render "govuk_publishing_components/components/contents_list", {
+      contents: [
+        {
+          text: "Responsibilities",
+          href: "#responsibiities"
+        },
+        role.currently_occupied? ? {
+          text: "Current Role Holder",
+          href: "#current-role-holder"
+        }: nil
```
AlanGabbianelli:
```suggestion
        } : nil
```

app/views/roles/show.html.erb
```diff
@@ -7,7 +7,22 @@
...
+      ].compact
+    } %>
+  </div>
   <div class="govuk-grid-column-two-thirds" dir="<%= page_text_direction %>" lang="<%= role.locale %>">
+
```
bevanloon: did you mean to leave this line break here?

bevanloon:
Have we captured the work to add "Announcements", "Previous Roles" and "Policies" somewhere?

app/models/role.rb
```diff
@@ -30,6 +30,12 @@ def organisations
     links["ordered_parent_organisations"]
   end
 
+  def currently_occupied?
```
thomasleese: This method is very similar to the one below is, `current_holder`. To avoid duplicate code, you could write this method as:

```ruby
def currently_occupied?
  current_holder.present?
end
```

But then I wonder whether this method has much benefit, and in your view, you could just do:

```
role.current_holder ? { ... } : nil
```

callumknights:
We have a card for announcements, which Mahmoud has done and raised a PR for https://github.com/alphagov/collections/pull/1406. I don't believe there are currently any cards for Previous Roles and Policies, either in To Do or Planning.

app/models/role.rb
callumknights: Okay definitely an oversight on my part, it kind of doesn't have much actual benefit, but using a methods like that keeps it consistent with similar views across Collections, which I personally think is a good thing. You may disagree though

PR #1406, opened 2019-12-06
Add announcements to role pages

MahmudH:
Announcements are retrieved from search api. These results are presented by the document list component, which includes the title, public_timestamp and content_store_document_type.

Links are also added for email alerts, atom feed subscriptions and to view all announcements.

This PR also refactors replication of code between person and role by adding an announcements presenter.

Trello: https://trello.com/c/Z6SmadpI/1588-2-add-announcements-to-role-pages

Screenshot:
Screen Shot 2019-12-06 at 11 33 50

Similar to PR: https://github.com/alphagov/collections/pull/1367


app/views/roles/_announcements.html.erb
```diff
@@ -0,0 +1,23 @@
+<% if role.announcements %>
```
thomasleese: This view and the one for people are very similar, could we merge the two somehow? They'd only need to take a `announcements` object and then could just `announcements.items` and `announcements.links`. I don't think there's anything specific to the person or the role.

PR #1414, opened 2019-12-13
Display previous holders on role pages

benjamineskola:
Adds a `previous_holders` method to the role, and then alters the view
to show these.

https://trello.com/c/Rk6ChPtp/1629-2-add-previous-holders-list-to-role-pages

### DIFF
--- a/app/models/role.rb
+++ b/app/models/role.rb
@@ -35,15 +35,27 @@ class Role
   end
 
   def current_holder
-    links
-      .fetch("role_appointments", [])
-      .find { |ra| ra.dig("details", "current") }
+    @current_holder ||= role_holders(current: true)
+      .first
       &.dig("links", "person")
       &.first
   end
 
   def current_holder_biography
     current_holder["details"]["body"]
+  end
+
+  def past_holders
+    @past_holders ||= role_holders(current: false)
+      .reverse
+      .map do |rh|
+        rh.dig("links", "person")
+          &.first
+          &.tap do |hash|
+            hash["details"]["start_year"] = Time.zone.parse(rh["details"]["started_on"]).year
+            hash["details"]["end_year"] = Time.zone.parse(rh["details"]["ended_on"]).year
+          end
+      end
   end
 
   def link_to_person
@@ -93,5 +105,11 @@ private
 
   def slug
     link_to_person.split("/").last
+  end
+
+  def role_holders(current:)
+    links.fetch("role_appointments", [])
+      .find_all { |rh| rh["details"]["current"] == current }
+      .sort_by { |rh| rh["details"]["started_on"] }
   end
 end
--- a/app/views/roles/_past_role_holders.html.erb
+++ b/app/views/roles/_past_role_holders.html.erb
@@ -1,0 +1,17 @@
+<% unless role.past_holders.empty? %>
+  <section id="past-role-holders" class="govuk-!-padding-bottom-9">
+    <%= render "govuk_publishing_components/components/heading", {
+      text: "Previous holders of this role",
+    } %>
+
+    <%= render "components/taxon-list", {
+      items: role.past_holders.map do |rh|
+        {
+          text: rh['title'],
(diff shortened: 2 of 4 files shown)


app/views/roles/_past_role_holders.html.erb
```diff
@@ -0,0 +1,17 @@
+<% unless role.past_holders.empty? %>
+  <section id="past-role-holders" class="govuk-!-padding-bottom-9">
+    <%= render "govuk_publishing_components/components/heading", {
+      text: "Past role holders:",
+    } %>
+
+    <%= render "components/taxon-list", {
+      items: role.past_holders.map { |rh|
```
thomasleese:
```suggestion
      items: role.past_holders.map do |rh|
        ...
      end
```

I think for longer blocks, it's better to use a `do/end`, as it's more readable.

app/models/role.rb
```diff
@@ -46,6 +46,18 @@ def current_holder_biography
     current_holder["details"]["body"]
   end
 
+  def past_holders
```
thomasleese: Since we now have methods for fetching the previous and current role appointments in this class (`current_holder` and `past_holders`) it might be better to generalise it a bit like I did for people: https://github.com/alphagov/collections/blob/4be621d6d5d4a8358e21cea78269590a8636b691/app/models/person.rb#L93-L106

I also wonder (maybe not for this PR) but whether it might make sense to have some sort of mixin class like `HasRoleAppointments` which we can `include` in both this class and the person class. But on the other hand this might be confusing as the logic is then somewhere else. 🤷‍♂

app/models/role.rb
```diff
@@ -46,6 +45,22 @@ def current_holder_biography
...
+    @past_holders ||= role_holders(current: false)
+      .reverse
+      .map do |rh|
+        rh["details"]["start_year"] = Time.zone.parse(rh["details"]["started_on"]).year
+        rh["details"]["end_year"] = Time.zone.parse(rh["details"]["ended_on"]).year if rh["details"]["ended_on"]
+        rh
+      end
+  end
+
+  def role_holders(current:)
```
thomasleese: I think this method can be `private`.

app/views/roles/_past_role_holders.html.erb
```diff
@@ -0,0 +1,17 @@
+<% unless role.past_holders.empty? %>
+  <section id="past-role-holders" class="govuk-!-padding-bottom-9">
+    <%= render "govuk_publishing_components/components/heading", {
+      text: "Past role holders:",
```
thomasleese:
```suggestion
      text: "Previous holders of this role",
```

This is what it says in Whitehall: https://www.gov.uk/government/ministers/parliamentary-under-secretary-of-state#previous-holders-of-this-role

benjamineskola:
I'm a little unhappy that `current_holder` returns a Person but `past_holders` returns a list of RoleAppointments; but since the view needs the RoleAppointment dates, it doesn't exactly make sense to make these match up.

I suppose I could add the dates to the Person hashes instead, but I'm not sure if it's worth it just for the sake of this.

app/models/role.rb
```diff
@@ -46,6 +45,21 @@ def current_holder_biography
...
+      .reverse
+      .map do |rh|
+        rh.dig("links", "person")
+          &.first
+          &.deep_merge(
+            "details" => {
+            "start_year" => Time.zone.parse(rh["details"]["started_on"]).year,
+            "end_year" => Time.zone.parse(rh["details"]["ended_on"]).year,
+            },
+          )
```
thomasleese: An alternative to this (perhaps a more Ruby solution):

```ruby
&.tap do |hash|
  hash["details"]["start_year"] = Time.zone.parse(rh["details"]["started_on"]).year
  hash["details"]["end_year"] = Time.zone.parse(rh["details"]["ended_on"]).year
end
```

`tap` takes a block which executes against the object, and then it returns the object.

PR #1421, opened 2019-12-17
Set the direction of right to left text 

edwardkerry:
This ensures that any right to left language translations tag components with the `dir` attribute for accessibility, and are assigned the necessary class to make use of govuk-publishing-components stylings. These are only set for specified RTL languages.

For all non-default locales, it also assigns a `lang` attribute.

[Trello Card](https://trello.com/c/G77SR2Nx/1616-improve-accessiblity-for-right-to-left-translations)

### DIFF
--- a/app/views/people/_current_roles.html.erb
+++ b/app/views/people/_current_roles.html.erb
@@ -1,5 +1,5 @@
 <% if person.currently_in_a_role? %>
-  <section id="current-roles" class="<%= direction_rtl_class %>" <%= dir_attribute %> <%= lang_attribute %>>
+  <section id="current-roles" <%= direction_rtl_class(prefix: true) %> <%= dir_attribute %> <%= lang_attribute %>>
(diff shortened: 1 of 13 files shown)


app/views/people/_current_roles.html.erb
```diff
@@ -1,5 +1,5 @@
 <% if person.currently_in_a_role? %>
-  <section id="current-roles" dir="<%= page_text_direction %>" lang="<%= person.locale %>">
+  <section id="current-roles" class="<%= direction_rtl_class %>" <%= dir_attribute %> <%= lang_attribute %>>
```
nickcolley: Will this output `class=""` if `page_text_direction` set to `ltr`?

edwardkerry: ah, it will. good spot.
We could do something like:
`<%= "class=#{direction_rtl_class}" if direction_rtl_class %>` but that is not particularly tidy

PR #1442, opened 2020-01-14
Update colours

maxgds:
1) Turns off legacy colours and addresses an error this generated.
2) Sets the focus state for selected items on browse pages 

### Before
Non selected item in focus:
Screenshot 2020-01-14 at 09 50 28
Selected item in focus
Screenshot 2020-01-14 at 09 50 38

### After
Non selected item in focus:
Screenshot 2020-01-14 at 09 49 44
Selected item in focus:
Screenshot 2020-01-14 at 09 49 52

3) Fixes classes on some areas of the organisations page that were previously overlooked

http://www.gov.uk/government/organisations
Adds appropriate classes to these parts of the page:
Screenshot 2020-01-14 at 09 03 23
Screenshot 2020-01-14 at 09 16 46

4) Fixes classes on some overlooked areas of individual organisation pages

http://www.gov.uk/government/organisations/innovate-uk

Screenshot 2020-01-14 at 10 13 16

Screenshot 2020-01-14 at 10 11 59

### DIFF
--- a/app/assets/stylesheets/views/_browse.scss
+++ b/app/assets/stylesheets/views/_browse.scss
@@ -177,16 +177,22 @@
       }
 
       ul .active a {
-        background: $govuk-link-colour;
+        background: $govuk-brand-colour;
         color: govuk-colour("white");
-
-        &:hover {
-          background: $govuk-link-colour;
-        }
 
         p {
           color: govuk-colour("white");
         }
+
+        &:focus {
+          background: $govuk-focus-colour;
+          color: $govuk-focus-text-colour;
+
+          p {
+          color: $govuk-focus-text-colour;
+          }
+        }
+
       }
     }
 
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -1,5 +1,6 @@
 $govuk-compatibility-govuktemplate: true;
 $govuk-compatibility-govukelements: true;
+$govuk-use-legacy-palette: false;
 
 // gem components
 @import 'govuk_publishing_components/all_components';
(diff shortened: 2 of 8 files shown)


app/assets/stylesheets/views/_browse.scss
```diff
@@ -177,16 +177,22 @@
       }
 
       ul .active a {
-        background: $govuk-link-colour;
+        background: $govuk-brand-colour;
```
injms: Is this because the brand colour is used for backgrounds so will be tested for contrast, whilst the link colour might not be? Could be worth a comment to make sure that it's not changed in the future by someone well-meaning.

injms:
Only a couple of minor aesthetic quibbles that I've been able to see; aside from these, I think it's good to go.

On Commonwealth Scholarship Commission in the UK `/government/organisations/commonwealth-scholarship-commission-in-the-uk` the super-long social media name and line-height mean some gaps have appeared:

Screenshot 2020-01-14 at 13 36 39

---

More line-height oddness in the header when links go over multiple lines, as seen on Innovate UK `/government/organisations/innovate-uk` and the Cabinet Office `/government/organisations/cabinet-office` pages:

Screenshot 2020-01-14 at 13 42 25

---

If a link stretches over two lines in the news and communications (or any of the other sections in this part of the page), then creates this block effect - which isn't consistent with how other links behave. 

Screenshot 2020-01-14 at 13 42 44

Changing the link to `display: inline` fixed this in the inspector, but of course actually changing it is much more difficult!

Screenshot 2020-01-14 at 13 54 14

app/assets/stylesheets/views/_browse.scss
maxgds: No, it's because both colours translate to the same thing but semantically saying "I am highlighting this item because the user selected it previously" doesn't seem to be a case for using a link colour - :active and .active combined with a variable called link-colour was overly confusing.

maxgds:
> /government/organisations/commonwealth-scholarship-commission-in-the-uk

Hmm... this comes from the gem and the line height is set to be the same as the icon height, so I'm not sure how to address it as it's clearly to make the text sit nicely next to the image, but it probably wasn't written anticipating the text wrapping. My instinct here is to leave it as is - but either way, it shouldn't block this being merged as it's in a different codebase.

PR #1451, opened 2020-01-17
Transition css tweaks

huwd:
## What
Designers flagged a few tweaks they wanted following from yesterdays update to the Brexit landing page.

1. Alignment of header with translation options.
Screenshot 2020-01-17 12 34 51

2. A bit more breathing space between actions and their subheadings.
Screenshot 2020-01-17 12 34 44

## Changes

**Breathing space**
### Before
Screenshot_2020-01-17 Brexit(2)

## After
Screenshot_2020-01-17 Brexit(1)

**Header**
Now aligned,

In order to do this I needed more granular control over CSS, so moved this out of using a govuk_publishing_component and into being a normal dumb h1 with some govuk css applied.

The direction of travel from yesterdays change seems to be to prefer using css design system styling over use of publishing components. This makes sense to me as it will allow us to tweak and fiddle a bit more in response to frequent changes.

Screenshot 2020-01-17 at 14 50 27

**Tidy up**
The initial implementation continued some tweaks the the CSS. @andysellick pointed out that some of that functionality should be covered by the govuk spacing system.

This PR now removes a whole welt of code that causes divergence from standard GOV.UK layouts.
The result keeps all horizontal lines (light grey and darker black ones) aligned, but also uses the grid system to set spacing internal to rows.

The result allows us to remove a bunch of custom sass for this app.

Changes also mean a couple tweaks to tests that were looking for specific classes.

### DIFF
--- a/app/assets/stylesheets/views/_brexit-landing-page.scss
+++ b/app/assets/stylesheets/views/_brexit-landing-page.scss
@@ -34,6 +34,13 @@ $red: #E61E32;
   }
 }
 
+.landing-page__page_title {
+  @include govuk-font($size: 48, $weight: bold);
+  color: govuk-colour("white");
+  margin-top: govuk-spacing(7);
+  margin-bottom: govuk-spacing(4);
+}
+
 .landing-page__translation {
   @include govuk-media-query($from: tablet) {
     float: right;
--- a/app/assets/stylesheets/views/_brexit-landing-page.scss
+++ b/app/assets/stylesheets/views/_brexit-landing-page.scss
@@ -85,7 +85,7 @@ $red: #E61E32;
 }
 
 .landing-page__section-content {
-  padding-left: 0;
+  padding-left: govuk-spacing(3);
 
   @include govuk-media-query($from: tablet) {
     padding-right: 0;
--- a/app/assets/stylesheets/views/_brexit-landing-page.scss
+++ b/app/assets/stylesheets/views/_brexit-landing-page.scss
@@ -14,6 +14,7 @@ $red: #E61E32;
 .landing-page__header {
   padding-bottom: govuk-spacing(6);
   padding-top: govuk-spacing(3);
+  margin-bottom: govuk-spacing(6);
   background-color: $govuk-brand-colour;
   color: govuk-colour("white");
 }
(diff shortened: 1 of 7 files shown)


andysellick:
Happy to review, but can you add a link to the description to where this is/what it looked like previously? Also is there a trello card for this for reference?

app/assets/stylesheets/views/_brexit-landing-page.scss
```diff
@@ -78,7 +85,7 @@ $red: #E61E32;
 }
 
 .landing-page__section-content {
-  padding-left: 0;
+  padding-left: govuk-spacing(3);
```
andysellick: I don't think this is needed - this class is applied to a govuk column, which already has a padding left of 15px.

app/views/brexit_landing_page/_page_header.html.erb
```diff
@@ -28,11 +28,9 @@
...
       </div>
       <div class="govuk-grid-column-two-thirds">
-        <%= render "govuk_publishing_components/components/title", {
-          title: I18n.t("brexit_landing_page.page_header"),
-          inverse: true,
-          margin_bottom: 5
-        } %>
+        <h1 class="landing-page__page_header">
+          <%= I18n.t("brexit_landing_page.page_header") %>
+        </h1>
```
andysellick: I wish the component model had been able to accommodate this requirement. Sorry.

huwd: I started by trying to add `margin-top` with a value in the same way it takes a `margin-bottom`. Though i can see the rational for not giving us a ton of css leavers and buttons. 

I wonder how much of a weird use case is this?

The language toggle comes with a margin-top of `45px` our h1 comes with a `40px` margin-top.

app/assets/stylesheets/views/_brexit-landing-page.scss
```diff
@@ -34,6 +35,13 @@ $red: #E61E32;
   }
 }
 
+.landing-page__page_header {
```
andysellick: This should be `.landing-page__page-header` - but actually would it be possible to change it to something else? I was briefly confused by the previous class in this PR `.landing-page__header`. Maybe `.landing-page__title`?

app/assets/stylesheets/views/_brexit-landing-page.scss
```diff
@@ -48,40 +56,23 @@ $red: #E61E32;
...
-  padding-left: 0;
-}
-
-.landing-page__section-content {
-  padding-left: 0;
-
-  @include govuk-media-query($from: tablet) {
-    padding-right: 0;
+  margin-bottom: govuk-spacing(6);
+  &:before {
```
andysellick: Nitpick: I'd put a newline in before the `before`.

vanitabarrett:
Looks good 👍 

I spotted this spacing looks a bit large on mobile - 

Screenshot 2020-01-21 at 10 27 05

app/assets/stylesheets/views/_brexit-landing-page.scss
huwd: Sounds good, added

app/assets/stylesheets/views/_brexit-landing-page.scss
huwd: @andysellick discussed with @vanitabarrett we opted for `.landing-page__page-title` you think that would have been clearer or still cause the pause for thought?

PR #1456, opened 2020-01-23
Add schema.org People structured data

sihugh:
Also adds a basic spec to make sure people pages render. In doing that, the spec failed on several potentially `nil` bits (based on the content item generated by `GovukSchemas::RandomExample.for_schema`) , so I've added some defence around those.


sihugh:
@thomasleese do say if you disagree with the nil checks and defaults I've added.

PR #1464, opened 2020-01-27
Landing page spacing

vanitabarrett:
Trello: https://trello.com/c/v7YlL73F/413-tweak-landing-page-spacing-on-smaller-devices

## What
Add responsive spacing to landing page.

## Why
Some areas of the landing page were using `govuk-spacing` instead of `govuk-responsive-spacing`. As a result, some spacing on the page looked too large when on smaller screens.

## Section heading (desktop before vs after)
Screenshot 2020-01-27 at 11 21 06
Screenshot 2020-01-27 at 11 20 30

## Section heading (mobile before vs after)
Screenshot 2020-01-27 at 11 21 15
Screenshot 2020-01-27 at 11 20 35

## Bucket heading (mobile before vs after)
Screenshot 2020-01-27 at 11 52 24

Screenshot 2020-01-27 at 11 52 13

### DIFF
--- a/app/assets/stylesheets/views/_brexit-landing-page.scss
+++ b/app/assets/stylesheets/views/_brexit-landing-page.scss
@@ -58,7 +58,7 @@ $red: #E61E32;
 
 .landing-page__section {
   border-top: 2px solid govuk-colour("black");
-  padding-top: govuk-spacing(7);
+  @include govuk-responsive-padding(6, "top");
 }
 
 .landing-page__guidance_subheader {
--- a/app/assets/stylesheets/views/_brexit-landing-page.scss
+++ b/app/assets/stylesheets/views/_brexit-landing-page.scss
@@ -73,7 +73,9 @@ $red: #E61E32;
     border-top: 1px solid govuk-colour("mid-grey", $legacy: "grey-2");
     content: "";
     display: block;
-    margin: 0 govuk-spacing(3) govuk-spacing(6);
+    margin-right: govuk-spacing(3);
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/views/_brexit-landing-page.scss
```diff
@@ -73,7 +73,8 @@ $red: #E61E32;
     border-top: 1px solid govuk-colour("mid-grey", $legacy: "grey-2");
     content: "";
     display: block;
-    margin: 0 govuk-spacing(3) govuk-spacing(6);
+    margin: 0 govuk-spacing(3);
+    @include govuk-responsive-margin(6, "bottom");
```
huwd: Interesting, so what does the compile to, something like:

```
margin: 0 15px;
margin-bottom: 30px
@media only screen and (min-width: $tablet) {
  margin-bottom: 20px
}
```

Line 76 already would already apply a margin-bottom 15px right? so is that just overridden by the mixin at small screen sizes or does it also give a desktop overwrite?

vanitabarrett: Yes, the `margin-bottom` will override  any margin applied by `margin`. Line 76 actually means the margin-bottom will be set to 0 and the left value will be set to 15px.

I guess a clearer way of writing it might be to split each one out separately, e.g:

```
margin-right: 15px
margin-left: 15px
@include govuk-responsive-margin(6, "bottom");
```

I'll amend to make it clearer 👍

PR #1465, opened 2020-01-27
Add time based intl helper and test intl

huwd:
Trello: https://trello.com/c/feQSo8wO/412-spike-release-cache-11pm-launch

## Background
We will need to deploy new content to the landing page at a fixed time to meet upcoming brexit deadlines.

Deployment pipelines vary and are unreliable, so in order to get a time-specific deploy we don't want tor rely on them.

Instead this adds flow control to use alongside an intl migration with a timegate.
It also allows for an override to check that all is functioning well, based on a shared secret in ENV.

## Specific changes
- Added fake test integration intl translations to test
- Set a test time of 13:45, at which we can clear a cache and see if it rolls over in <1min
- Standardised a few differences between where we were using `I18t.t()` and `t()`

### DIFF
--- a/app/helpers/transition_period_helper.rb
+++ b/app/helpers/transition_period_helper.rb
@@ -1,0 +1,28 @@
+module TransitionPeriodHelper
+  SWITCHOVER_TIME = Time.zone.parse("2020-01-31 23:00:00").in_time_zone
+
+  def time_based_intl
+    if before_switchover_and_not_overridden?
+      "brexit_landing_page"
+    else
+      "transition_landing_page"
+    end
+  end
+
+private
+
+  def before_switchover_and_not_overridden?
+    before_switchover? && !override?
+  end
+
+  def before_switchover?
+    Time.zone.now < SWITCHOVER_TIME
+  end
+
+  def override?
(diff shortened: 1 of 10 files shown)


app/helpers/transition_period_helper.rb
```diff
@@ -0,0 +1,12 @@
+module TransitionPeriodHelper
+  SWITCHOVER_TIME = Time.zone.parse('2020-01-01 17:00:00').in_time_zone
```
vanitabarrett: Date needs updating if we want to test this?

vanitabarrett:
I'd be hesitant to approve and merge before we've tested this. Can we test it by deploying a branch to integration/staging instead?

vanitabarrett:
This approach looks good for pure content changes, but I'm not sure how it would work for view changes, e.g: moving components around, turning elements of the page on/off, adjusting colours. If we test this and it works for content changes, we need to think about these things too to see if it's viable for Friday.

app/helpers/transition_period_helper.rb
huwd: sorry ran `git push` not `git push --force`

PR #1475, opened 2020-02-03
Remove time dependent deployment from transition landing page

huwd:
Trello: https://trello.com/c/vkNzXwcO/444-tidy-up-behind-ourselves-timed-deployment-edition

For the 31st of January (Brexit night) we had to deploy a version of the page that held both the before and after content, with a time based switch to flick between the two.

This allowed us to do a to-the-minute update.

With that moment past, we can now remove all that complexity!

**Also**
1. [Fixes one part](https://github.com/alphagov/collections/pull/1475/commits/37ad4446cdaec8508766313776cb1812c0e712d7) where we were still reading from the old Brexit translations (the content was the same so we didn't spot it)
2. [Changes the mailto subject lines](https://github.com/alphagov/collections/pull/1475/commits/833212ddde13054570d1698c07508d4635cd02d5) (for email + linked in) to exist in the translations file with everything else.
  - [Mailto Link in English](mailto://?body=https://www.gov.uk/transition/&subject=Get ready for 2021 - GOV.UK)
  - [Mailto link in Welsh](mailto://?body=https://www.gov.uk/transition.cy/&subject=Get ready for 2021 - GOV.UK)
  - [LinkedIn link in English](http://www.linkedin.com/shareArticle?url=https%3A%2F%2Fwww.gov.uk%2Ftransition&title=Get%20ready%20for%202021%20-%20GOV.UK)
  - [LinkedIn link in Welsh](http://www.linkedin.com/shareArticle?url=https%3A%2F%2Fwww.gov.uk%2Ftransition.cy&title=Get%20ready%20for%202021%20-%20GOV.UK)

**NB**: We have no welsh translation for the subject line currently. Will request this, and add to a later PR. There is no change from current functionality on live.
Would be especially good if someone could check Linkedin works, as I don't have an account!


vanitabarrett:
Seems good, but looks like the build is failing due to tests

huwd:
yeah sorry just spotted that, fixing now!

vanitabarrett:
If we're removing reference to the dynamic list, we should also remove it here: https://github.com/alphagov/collections/pull/1475/commits/8df4752f2798374605da6ad66c16bdc1054e9939#diff-605aa9fbcbf62bbb1853aa386ed7ad86R7

It's also scattered elsewhere in the code, probably worth removing as I think we've removed all the related view code that actually shows the chevron/link, so it doesn't look any different for true or false.

app/presenters/brexit_landing_page_presenter.rb
```diff
@@ -50,22 +50,8 @@ def translation_links
...
-    end
-  end
-
-  def before_switchover?
-    Time.zone.now < SWITCHOVER_TIME
-  end
-
   def fetch_buckets
-    buckets = I18n.t("#{time_based_intl}.campaign_buckets")
+    buckets = I18n.t("transition_landing_page.campaign_buckets")
```
laurentqro: somehow not in the scope of this PR, but should we have `t` instead of `I18n.t` for consistency?

laurentqro: or is `t` available only as a view helper?

huwd: Yeah, i checked this - so if you remove it here in the presenter it errors!!
However there appears to be some kind of override in helpers that must do something like:
`t = I18n.t`, which is a bit annyoing.

Would kind of like to throw it out to the community and ask what a good standard should be then write a rubocop rule for it?

`t()` is concise, but seems a bit orphan `I18n.t()` always seems clear what it is and where it comes from... so i might stay towards the more verbose?

laurentqro: Ok, so it looks like `t` is an ActionView helper, so we can't use `t` outside of views.

https://apidock.com/rails/v5.2.3/ActionView/Helpers/TranslationHelper/translate

PR #1484, opened 2020-02-05
Update Organisation breadcrumb to show parent

RobBond-GDS:
Adjust the breadcrumb to show the parent organisation link as the last item instead of the current page the anchor links to itself.

Removed "is_current_page" attribute from the organisation breadcrumb as it is inconsistent with mobile breadcrumb styling

https://trello.com/c/i1LS8Acr/24-fix-breadcrumbs-for-organisation-pages

### DIFF
--- a/test/support/organisation_helper.rb
+++ b/test/support/organisation_helper.rb
@@ -701,11 +701,7 @@ module OrganisationHelpers
     visit content["base_path"]
   end
 
-  def i_see_the_current_page_breadcrumb
-    assert page.has_css?(".gem-c-breadcrumbs", text: @title)
-  end
-
-  def and_the_courts_breadcrumb
+  def i_see_the_courts_breadcrumb
     assert page.has_css?(".gem-c-breadcrumbs", text: "Courts and Tribunals")
   end
 
--- a/app/presenters/organisations/header_presenter.rb
+++ b/app/presenters/organisations/header_presenter.rb
@@ -13,10 +13,6 @@ module Organisations
           url: "/",
         },
         index_page_breadcrumb,
-        {
-          title: @org.title,
-          is_current_page: true,
-        },
       ]
     end
 
--- a/test/integration/courts_pages_test.rb
+++ b/test/integration/courts_pages_test.rb
@@ -5,8 +5,7 @@ class CourtsPagesTest < ActionDispatch::IntegrationTest
 
   it "renders a courts page correctly" do
     when_i_visit_a_courts_page
-    i_see_the_current_page_breadcrumb
-    and_the_courts_breadcrumb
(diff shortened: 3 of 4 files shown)


vanitabarrett:
Change looks good! Can you update the PR title and description (it would be good to reflect the removal of the current page link as a whole, not just the is_current_page attribute). Then you can squash the 2 commits 🙂

vanitabarrett:
Really picky, but can we update the commit message to be shorter so it doesn't get cut off?

Something like: "don't show current page in breadcrumbs on org pages"?

test/support/organisation_helper.rb
```diff
@@ -701,12 +701,8 @@ def setup_and_visit_page(content)
...
   end
 
-  def i_see_the_current_page_breadcrumb
-    assert page.has_css?(".gem-c-breadcrumbs", text: @title)
-  end
-
-  def and_the_courts_breadcrumb
-    assert page.has_css?(".gem-c-breadcrumbs", text: "Courts and Tribunals")
+  def i_see_the_courts_breadcrumb
+    assert page.has_css?(".gem-c-breadcrumbs")
```
vanitabarrett: This needs to actually test what the test name suggests, i.e: look for the courts breadcrumb text.

RobBond-GDS: not sure what to do with this as the breadcrumb would only display "Home > Organisation" 

what do suggest?

test/support/organisation_helper.rb
```diff
@@ -702,11 +702,11 @@ def setup_and_visit_page(content)
   end
 
   def i_see_the_current_page_breadcrumb
-    assert page.has_css?(".gem-c-breadcrumbs", text: @title)
+    assert page.has_css?(".gem-c-breadcrumbs")
   end
 
   def and_the_courts_breadcrumb
-    assert page.has_css?(".gem-c-breadcrumbs", text: "Courts and Tribunals")
```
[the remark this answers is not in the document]
RobBond-GDS: not sure what to do with this as the breadcrumb would only display "Home > Organisation" 

what do suggest?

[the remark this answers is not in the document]
vanitabarrett: I think for HM Courts and Tribunal pages, the breadcrumb will now show Home > Courts and Tribunals. Like we do for the [organisation tests](https://github.com/alphagov/collections/pull/1484/files#diff-653b30bd42942791bb593adf881a10a5R531) we can test for the presence of the Courts and Tribunals link

test/support/organisation_helper.rb
vanitabarrett: Replied on your previous comment.

PR #1495, opened 2020-02-12
Remove references to dynamic lists on Brexit landing page

laurentqro:
This PR removes a lot of old tech debt referencing `dynamic lists`.

https://trello.com/c/KVmOh3Uk/474-remove-dynamic-lists-tech-debt-from-collections

### DIFF
--- a/test/support/transition_landing_page_steps.rb
+++ b/test/support/transition_landing_page_steps.rb
@@ -6,32 +6,15 @@ module TransitionLandingPageSteps
   include GdsApi::TestHelpers::ContentItemHelpers
   include RummagerHelpers
 
+  CONTENT_ID = "d6c2de5d-ef90-45d1-82d4-5f2438369eea".freeze
+  TRANSITION_TAXON_PATH = "/transition".freeze
+
   def given_there_is_a_transition_taxon
-    stub_content_store_has_item(transition_taxon_path, content_item)
-  end
-
-  def transition_taxon_path
-    "/transition"
-  end
-
-  def content_id
-    "d6c2de5d-ef90-45d1-82d4-5f2438369eea"
-  end
-
-  def content_item
-    GovukSchemas::RandomExample.for_schema(frontend_schema: "taxon") do |item|
-      item.merge(
-        "base_path" => transition_taxon_path,
-        "content_id" => content_id,
-        "title" => "Transition",
-        "phase" => "live",
-        "links" => {},
-      )
-    end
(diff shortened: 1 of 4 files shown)


test/support/transition_landing_page_steps.rb
```diff
@@ -6,32 +6,15 @@ module TransitionLandingPageSteps
...
   include RummagerHelpers
 
-  def given_there_is_a_transition_taxon
-    stub_content_store_has_item(transition_taxon_path, content_item)
-  end
-
-  def transition_taxon_path
-    "/transition"
-  end
+  CONTENT_ID = "d6c2de5d-ef90-45d1-82d4-5f2438369eea".freeze
```
vanitabarrett: What is this the content id of - the taxon? Can we make it clear?

huwd: This is the `/transition` taxon (formerly of the parish of `/brexit`)
[Matches here](https://github.com/alphagov/search-api/blob/1f91a38234c2f8c88b43854a3d039d9c26e2a4bf/lib/govuk_index/presenters/common_fields_presenter.rb#L11)

huwd: But yeah i agree, could leave these as a comment? Or a link? What's most likely to be updated if we have to change the name again?

vanitabarrett: Can we call it `TRANSITION_TAXON_CONTENT_ID`, to make that clearer? Like we call the path `TRANSITION_TAXON_PATH` (below)?

test/support/transition_landing_page_steps.rb
```diff
@@ -6,38 +6,15 @@ module TransitionLandingPageSteps
...
-    stub_content_store_has_item(brexit_taxon_path, content_item)
-  end
-
-  def brexit_taxon_path
-    "/brexit"
-  end
+  CONTENT_ID = "d6c2de5d-ef90-45d1-82d4-5f2438369eea".freeze
+  TRANSITION_TAXON_PATH = "/transition".freeze
 
-  def content_id
```
huwd: Or perhaps we resurrect a method like this called `transition_taxon_id` and just shove it in there?

test/support/transition_landing_page_steps.rb
laurentqro: Good idea, I like `TRANSITION_TAXON_CONTENT_ID`. 👏

PR #1516, opened 2020-02-27
Add a news story to the Transition Landing Page

koetsier:
And fixed a small typo

Trello: https://trello.com/c/abg1wivz/551-update-the-transition-page-to-include-the-uks-negotiating-mandate

### DIFF
--- a/config/locales/en/transition_landing_page.yml
+++ b/config/locales/en/transition_landing_page.yml
@@ -15,6 +15,13 @@ en:
     comms:
       links:
       - link:
+          text: "Our approach to the Future Relationship with the EU"
+          path: "/government/publications/our-approach-to-the-future-relationship-with-the-eu"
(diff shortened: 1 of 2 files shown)


config/locales/en/transition_landing_page.yml
```diff
@@ -14,6 +14,13 @@ en:
     comms_header: "News"
     comms:
       links:
+      - link:
+          text: "Our approach to the Future Relationship with the EU"
+          path: "/government/publications/our-approach-to-the-future-relationship-with-the-eu"
+          description: "The UK’s approach to negotiations with the European Union."
+          metadata:
```
vanitabarrett: This needs to be less indented - metadata should be on the same level as "link" (see below for examples).

koetsier: Good catch

PR #1523, opened 2020-02-28
[Do Not Merge] Add eCommerce tracking to taxon browse

huwd:
Trello: https://trello.com/c/9UR6LMz0/140-add-ecommerce-tracking-to-taxon-pages

## Dependent on
- Fix to the large payload ecommerce bug
- Fix to double event listners firing in ecommerce
##What
We should add Google Analytics Ecommerce tracking to the navigational links across the taxon pages

**List level atributes**

- ` data-analytics-ecommerce`
- ` data-ecommerce-start-index="1"`
- ` data-list-title="Taxon Browse"`

with the level reflecting how far the user is into the navigation

Also if [the generalisation mentioned here](https://trello.com/c/BTt4TmUG/143-%F0%9F%97%A1-generalise-statics-ecommerce-module) has not been completed the list will require the following empty attribute to work: ` data-search-query`

**Row level attributes**
-  `ecommerce_row: true`
-  `ecommerce_subheading: "All categories"`

if [this PR has been merged](https://github.com/alphagov/static/pull/2068) we can also have an optional `data-ecommerce-subheading` to describe which groups a row item falls into.

##Why
So we can understand the % of users who interact with each element. 

This will inform the taxon pages redesign and provide a benchmark against which to judge the re-designed header's performance.

### DIFF
--- a/app/presenters/taxon_organisations_presenter.rb
+++ b/app/presenters/taxon_organisations_presenter.rb
@@ -41,12 +41,11 @@ private
   def organisation_list_with_logos
     organisations_with_logos.map.with_index do |org, index|
       {
-
           name: org.logo_formatted_title,
           url: org.link,
           brand: org.brand,
           crest: org.crest,
-          data_attributes: data_attributes(org.link, org.title, index + 1),
+          data_attributes: data_attributes(org.link, org.title, index + 1, "Organisations"),
       }
     end
   end
@@ -61,16 +60,17 @@ private
         link: {
           text: organisation.title,
           path: organisation.link,
-          data_attributes: data_attributes(organisation.link, organisation.title, tracking_number + index),
+          data_attributes: data_attributes(organisation.link, organisation.title, tracking_number + index, "Organisations"),
         },
       }
     end
   end
 
-  def data_attributes(base_path, link_text, index)
+  def data_attributes(base_path, link_text, index, secton_title)
     {
       ecommerce_row: true,
       ecommerce_path:  base_path,
+      ecommerce_subheading: secton_title,
       track_category: "organisationsDocumentListClicked",
       track_action: index,
       track_label: base_path,
--- a/app/presenters/taxon_organisations_presenter.rb
+++ b/app/presenters/taxon_organisations_presenter.rb
@@ -69,6 +69,8 @@ private
 
   def data_attributes(base_path, link_text, index)
     {
+      ecommerce_row: true,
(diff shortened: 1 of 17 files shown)


huwd:
Ok so there's a bug with this one.
As it is, this PR work 💯 on every section _except organisations_.

Theoretically speaking, to make organisations appear I just need to add a `data-analyitics-ecommerce` [attribute on this line](https://github.com/alphagov/collections/pull/1523/files#diff-99c34807ca55ceefe2e8b4eb97bd41f1R2)...

... If i do so, then all eCommerce tracking mysteriously stops working on the page.
I can't work out why this would be...

I've swapped out all the attributes on this element, I've made them match other sections, I've made them different, swapped each out in turn... no joy.

My best guess is that the 'other organisations' sections contains a lot of items... when they render the page has over 100+ products...

I tried removing tracking from the others leaving only organisations, but still nothing.
I think i could do with a second pair of eyes to help think this through.

app/views/taxons/_organisations.html.erb
```diff
@@ -1,5 +1,5 @@
 <% if presented_organisations.show_organisations? %>
-  <div id="organisations" class="taxon-page__section-group taxon-page__section-group--organisation" data-module="toggle">
+  <div id="organisations" class="taxon-page__section-group taxon-page__section-group--organisation" data-ecommerce-start-index="1" data-list-title="Taxonomy browse" data-search-query="bloop" data-module="toggle">
```
huwd: If we want to not have organisation tracking, I should remove this set of attributes

app/presenters/taxon_organisations_presenter.rb
```diff
@@ -72,6 +72,8 @@ def data_attributes(base_path, link_text, index)
       track_category: "organisationsDocumentListClicked",
       track_action: index,
       track_label: base_path,
+      ecommerce_row: index,
```
huwd: I should remove this if we want to go live without organisations

huwd:
NB: analytics feature tests were removed with the revert.
The review here brought the testing strategy into question.
The attention at the point of pause on this feature was to return and rewrite them ensuring data attributes were present for a complete journey.

Need to also find a way to test both curated and a-z paths!

benthorner:
@huwd this PR is very old. Do we still need it?

kevindew:
Closing since this is stale, can always be re-opened if needed later.

PR #1537, opened 2020-03-09
[Do Not Merge] Add eCommerce tracking to mainstream browse

huwd:
Trello: https://trello.com/c/wyHXRVll/138-add-ecommerce-tracking-to-mainstream-browse

## Dependent on
- Fix to the large payload ecommerce bug
- Fix to double event listners firing in ecommerce

##What
We should add Google Analytics Ecommerce tracking to the navigational links across the mainstream browse pages

**List level atributes**

- ` data-analytics-ecommerce`
- ` data-ecommerce-start-index="1"`
- ` data-list-title="Mainstream Browse - Level {x}"`

with the level reflecting how far the user is into the navigation

Also if [the generalisation mentioned here](https://trello.com/c/BTt4TmUG/143-%F0%9F%97%A1-generalise-statics-ecommerce-module) has not been completed the list will require the following empty attribute to work: ` data-search-query`

**Row level attributes**
-  `ecommerce_row: true`
-  `ecommerce_subheading: "All categories"`

if [this PR has been merged](https://github.com/alphagov/static/pull/2068) we can also have an optional `data-ecommerce-subheading` to describe which groups a row item falls into.

##Why
So we can understand the % of users who interact with each element. 

This will inform the mainstream redesign and provide a benchmark against which to judge the re-designed header's performance.

### DIFF
--- a/app/views/second_level_browse_page/_links.html.erb
+++ b/app/views/second_level_browse_page/_links.html.erb
@@ -1,3 +1,4 @@
+<% ecommerce_page_row_count = 1 %>
 <div class="pane-inner <%= page.lists.curated? ? 'curated-list' : 'a-to-z' %>">
   <h2 tabindex="-1" class="js-heading"><%= page.title %></h2>
 
@@ -8,13 +9,19 @@
       <p class="sort-order"><%= hairspace list.title %></p>
     <% end %>
 
-    <ul>
+    <ul data-analytics-ecommerce
+        data-search-query
+        data-ecommerce-start-index=<%= "#{section_index + 1}" %>
+        data-list-title="Mainstream browse - Column 3">
       <% list.contents.each_with_index do |list_item, index| %>
         <li>
           <%= link_to(
             list_item.title,
             list_item.base_path,
             data: {
+              ecommerce_row: true,
+              ecommerce_path: list_item.base_path,
+              ecommerce_subheading: page.lists.curated? ? list.title : 'A to Z',
               track_category: 'thirdLevelBrowseLinkClicked',
               track_action: "#{section_index + 1}.#{index + 1}",
               track_label: list_item.base_path,
@@ -25,6 +32,7 @@
             },
             class: "govuk-link",
           ) %>
+          <% ecommerce_page_row_count += 1 %>
         </li>
       <% end %>
     </ul>
@@ -33,13 +41,19 @@
   <% if page.related_topics.any? %>
     <div class="detailed-guidance">
       <h2>Detailed guidance</h2>
-      <ul>
+      <ul data-analytics-ecommerce
+          data-search-query
+          data-ecommerce-start-index=<%= "#{ecommerce_page_row_count}" %>
+          data-list-title="Mainstream browse - Column 3">
         <% page.related_topics.each_with_index do |related_topic, index| %>
           <li>
             <%= link_to(
               related_topic.title,
               related_topic.base_path,
               data: {
+                ecommerce_row: true,
+                ecommerce_path: related_topic.base_path,
+                ecommerce_subheading: "Detailed guidance",
                 track_category: 'thirdLevelBrowseLinkClicked',
                 track_action: "#{page.lists.count + 1}.#{index + 1}",
                 track_label: related_topic.base_path,
--- a/app/views/second_level_browse_page/_second_level_browse_pages.html.erb
+++ b/app/views/second_level_browse_page/_second_level_browse_pages.html.erb
@@ -1,4 +1,8 @@
-<div class="pane-inner <%= curated_order ? "curated" : "alphabetical" %>">
+<div class="pane-inner <%= curated_order ? "curated" : "alphabetical" %>"
+     data-analytics-ecommerce
+     data-search-query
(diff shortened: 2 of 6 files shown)


app/views/second_level_browse_page/_links.html.erb
```diff
@@ -35,13 +41,19 @@
   <% if page.related_topics.any? %>
     <div class="detailed-guidance">
       <h2>Detailed guidance</h2>
-      <ul>
+      <ul data-analytics-ecommerce
+          data-search-query
+          data-ecommerce-start-index=<%= "#{ecommerce_page_row_count}" %>
```
theseanything:
```suggestion
          data-ecommerce-start-index=<%= "#{page.lists.sum(&:size)}" %>
```
and remove the `ecommerce_page_row_count`?

app/views/second_level_browse_page/_second_level_browse_pages.html.erb
```diff
@@ -9,8 +13,9 @@
         <%= link_to(
           browse_page.base_path,
           data: {
-            ecommerce_row: "#{index + 1}",
+            ecommerce_row: true,
             ecommerce_path: browse_page.base_path,
+            ecommerce_subheading: "#{title} - #{ curated_order ? 'Curated' : 'A to Z'}",
```
theseanything:
```suggestion
            ecommerce_subheading: "#{title} - #{curated_order ? 'Curated' : 'A to Z'}",
```

features/step_definitions/analytics_steps.rb
```diff
@@ -1,20 +1,19 @@
...
-  assert page.has_selector?('div#root[data-list-title="Mainstream browse"]')
-  assert page.has_selector?('div#root[data-search-query="all categories"]')
+  assert page.has_selector?('div#root[data-list-title="Mainstream browse - Column 1"]')
 
-  assert_equal page.find("div#root li:nth-child(1) a")['data-ecommerce-row'], '1'
+  assert page.find("div#root li:nth-child(1) a")['data-ecommerce-row']
   assert_equal page.find("div#root li:nth-child(1) a")['data-ecommerce-path'], '/browse/benefits'
 
-  assert_equal page.find("div#root li:nth-child(2) a")['data-ecommerce-row'], '2'
+  assert page.find("div#root li:nth-child(2) a")['data-ecommerce-row']
```
theseanything: For consistency could this be?
```suggestion
  assert page.has_selector?("div#root li:nth-child(2) a['data-ecommerce-row']")
```

features/step_definitions/viewing_browse_steps.rb
```diff
@@ -78,11 +78,11 @@
 end
 
 Then(/^the A to Z label should be present$/) do
-  assert page.has_content?("A\u200Ato\u200AZ")
+  assert page.has_text?("A\u200Ato\u200AZ")
 end
 
 Then(/^the A to Z label should not be present$/) do
-  assert page.has_no_content?("A\u200Ato\u200AZ")
+  assert page.has_no_text?("A\u200Ato\u200AZ")
```
theseanything: Why is this not just spaces instead of '\u200A' (https://codepoints.net/U+200A?lang=en)

features/step_definitions/analytics_steps.rb
huwd: Great shout, amended

app/views/second_level_browse_page/_second_level_browse_pages.html.erb
huwd: great spot, amended

features/step_definitions/viewing_browse_steps.rb
huwd: We're using a hairspace, i think for design reasons

app/views/second_level_browse_page/_links.html.erb
huwd: unfortunately that won't work, because it's not the sum of the lists, there's these list.contents bits too.

So i could do something like `page.lists.sum{|list| list.contents.sum(&:size)}`... (<- i think that might work?)
But i figure putting a variable within the <li> that I'm trying to count would be stupider but perhaps easier to grok if you were browisng the code? 

What'd ya reckon?

huwd:
rebased off master, should resolve conflicts

app/views/second_level_browse_page/_links.html.erb
theseanything: Hmm, yeah not sure. Either way it seems like logic that should be placed in a view. Would a helper function with a descriptive function name be more ideal?

huwd:
NB: Analyitics feature tests were removed with the revert.
The review here brought the testing strategy into question.
The attention at the point of pause on this feature was to return and rewrite them ensuring data attributes were present for a complete journey.

Need to also find a way to test both curated and a-z paths!

benthorner:

benthorner:
Removed CD warning.

@huwd this PR is very old. Do we still need it?

kevindew:
Closing since this is stale, can always be re-opened if needed later.

PR #1550, opened 2020-03-18
Accordion sections

sihugh:
Initial stab at using accordion sections.

This also changes the yaml structure to allow subsections within the sections, as per the design.

https://govuk-collec-accordion-cavqnth.herokuapp.com/coronavirus

### DIFF
--- a/app/views/coronavirus_landing_page/_accordion_sections.html.erb
+++ b/app/views/coronavirus_landing_page/_accordion_sections.html.erb
@@ -1,14 +1,18 @@
   <% accordion_contents = [] %>
   <% t("coronavirus_landing_page.sections").each do | section | %>
-    <% contents = capture do %>
+    <% content = capture do %>
       <%= render( partial: 'section', locals: section ) %>
     <% end %>
     <%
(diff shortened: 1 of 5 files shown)


app/views/coronavirus_landing_page/_accordion_sections.html.erb
```diff
@@ -0,0 +1,14 @@
...
+    <% end %>
+    <%
+      accordion_contents << {
+        title: section[:title],
+        id: section[:title].underscore,
+        panel: contents
+      }
+    %>
+  <% end %>
+  <%= render 'components/accordion', sections: accordion_contents %>
```
andysellick: You probably want the accordion component from the gem?

sihugh: Yeah, I wasn't sure, given that it says "experimental".  I used the same thing as the world location news pages do.

sihugh: I've pushed that as a new commit

PR #1552, opened 2020-03-19
NHS component for landing page

DilwoarH:
Adds a header notice component for the landing page.

### Mobile
Screenshot 2020-03-19 at 11 28 59

### Desktop
Screenshot 2020-03-19 at 11 29 07

## Trello
https://trello.com/c/MeU64Pv0/39-nhs-component-for-landing-page


DilwoarH:
> Looks great! Do we need to add the PHE branding too, or has that come off again?

This has not been decided yet. Merging for now, will update if the designs change.

app/views/components/_header-notice.html.erb
```diff
@@ -0,0 +1,39 @@
+<%
+  title ||= false
+  title_logo_url ||= false
+  heading ||= false
+  list ||= []
+  call_to_action ||= false
+  branding_classes ||= false
```
andysellick: We shouldn't be passing arbitrary classes to a component like this as it encourages restyling of restyling of components by applications, which breaks isolation. Instead there should be an option to enable a particular look for the component e.g. `nhs_style: true`.

DilwoarH: Ah yes - thanks for pointing this out. I’ll update this.

PR #1553, opened 2020-03-19
Update branding to be flags rather than classes

DilwoarH:
Changed based on the following comment on PR:
https://github.com/alphagov/collections/pull/1552#discussion_r394993977

### DIFF
--- a/app/views/components/_header-notice.html.erb
+++ b/app/views/components/_header-notice.html.erb
@@ -4,7 +4,8 @@
   heading ||= false
   list ||= []
   call_to_action ||= false
-  branding_classes ||= false
+  nhs_branding ||= false
+  branding_classes = 'app-c-header-notice__branding--nhs' if nhs_branding
 %>
 <% if title.present? || title_logo_url.present? %>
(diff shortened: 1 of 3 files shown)


app/views/components/_header-notice.html.erb
```diff
@@ -4,7 +4,11 @@
   heading ||= false
   list ||= []
   call_to_action ||= false
-  branding_classes ||= false
+
+  # branding
```
andysellick: I'm not sure the comment is necessary here?

app/views/components/_header-notice.html.erb
```diff
@@ -4,7 +4,11 @@
...
+  nhs_branding ||= false
+  branding_classes = ''
+  branding_classes += 'app-c-header-notice__branding--nhs' if nhs_branding
```
andysellick: I can see you're planning ahead, but I think for now it's enough to do something simpler, like:

`branding_classes = 'app-c-header-notice__branding--nhs' if nhs_branding`

and remove line 10.

PR #1560, opened 2020-03-20
Design updates for CV-19

DilwoarH:
Trello tickets:

- https://trello.com/c/lk56cLWI/54-reduce-the-margin-below-the-nhs-box-to-30px-on-mobile
- https://trello.com/c/9HlvAiNB/51-nhs-box-link-needs-to-be-19px
- https://trello.com/c/kU55ud5e/55-remove-subheading-data-on-reported-cases-of-coronavirus-around-the-uk

### DIFF
--- a/app/assets/stylesheets/components/_header-notice.scss
+++ b/app/assets/stylesheets/components/_header-notice.scss
@@ -38,5 +38,5 @@ $nhs_brand_color: #005EB8;
 }
 
 .app-c-header-notice__call-to-action-link {
-  @include govuk-font(24, $weight: bold);
+  @include govuk-font(19, $weight: bold);
 }
--- a/app/assets/stylesheets/components/_header-notice.scss
+++ b/app/assets/stylesheets/components/_header-notice.scss
@@ -1,7 +1,7 @@
 $nhs_brand_color: #005EB8;
 
 .app-c-header-notice {
-  margin-bottom: govuk-spacing(8);
+  @include govuk-responsive-margin(8, "bottom");
 }
 
 .app-c-header-notice__branding--nhs {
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/components/_header-notice.scss
```diff
@@ -21,11 +21,11 @@ $nhs_brand_color: #005EB8;
 
 .app-c-header-notice__title {
   color: govuk-colour("white");
-  margin: 0;
```
andysellick: This change doesn't do anything - except increase the sass compilation time by a tiny amount. I'd suggest keeping it as 0, it's more readable anyway.

app/assets/stylesheets/components/_header-notice.scss
```diff
@@ -21,11 +21,11 @@ $nhs_brand_color: #005EB8;
...
+  margin: govuk-spacing(0);
 }
 
 .app-c-header-notice__title-logo {
-  max-width: 90px;
```
andysellick: Personally I'd avoid the spacing scale for anything where width is concerned, I think the intent of it is only for margin and padding. Just because a number matches doesn't mean we should use the spacing scale all the time.

app/assets/stylesheets/components/_header-notice.scss
DilwoarH: Good point

app/assets/stylesheets/components/_header-notice.scss
andysellick: I assume you meant to change this back and reduce it?

DilwoarH: I kind of disagree :) but I've updated it as I sort of agree with the width argument

DilwoarH: > I assume you meant to change this back and reduce it?

yeah I reduced it to 60px

PR #1576, opened 2020-03-23
Rework landing page frontend

andysellick:
Restructuring frontend code for coronavirus landing page. Uses new component changes for some of the new stuff instead of CSS in the app.

- Uses https://github.com/alphagov/govuk_publishing_components/pull/1396
- Uses https://github.com/alphagov/govuk_publishing_components/pull/1397


app/views/coronavirus_landing_page/_page_header.html.erb
```diff
@@ -30,25 +30,48 @@
...
-      heading_level: 2
+      inverse: true
     } %>
     <div class="govuk-grid-row">
       <div class="govuk-grid-column-one-third">
-        <h3 class="gem-c-heading  gem-c-heading--font-size-19 gem-c-heading--padding gem-c-heading--border-top-2">
-          <a href="https://www.gov.uk/government/publications/closure-of-educational-settings-information-for-parents-and-carers/closure-of-educational-settings-information-for-parents-and-carers">
+        <% link = capture do %>
+          <a href="https://www.gov.uk/government/publications/closure-of-educational-settings-information-for-parents-and-carers/closure-of-educational-settings-information-for-parents-and-carers" class="govuk-link covid__page-header-link">
             Schools and childcare are only open to children of key workers
```
DilwoarH: Should we not have all of this in the content yaml file:

https://github.com/alphagov/collections/blob/master/config/locales/en/coronavirus_landing_page.yml

andysellick: Yes, good point, will have a look.

sihugh: Leave it for now, I'll pick it up after the latest update

PR #1580, opened 2020-03-23
Add accordion analytics

andysellick:
Adds analytics to the accordion on the coronavirus landing page.

It does this by wrapping the accordion in a div with the tracking module JS attached and using the accordion component option to attach data attributes to each accordion section for tracking. There's an additional new module to provide toggling of the state of the accordion, so that the GA event includes whether the accordion was opened or closed.

Sadly you can't attach more than one module to a single element, so there's a bit of nesting here.

Trello card: https://trello.com/c/fE13lx7e/42-review-tracking-implementation-on-the-landing-page

### DIFF
--- a/spec/javascripts/modules/toggle-attribute_spec.js
+++ b/spec/javascripts/modules/toggle-attribute_spec.js
@@ -1,0 +1,31 @@
+describe('A toggle attribute module', function () {
+  "use strict"
+
+  var $element
+  var toggle
+  var clickon
+  var html = '\
+    <div id="element" data-module="toggle-attribute">\
+      <div id="clickon" data-toggle-attribute="data-state" data-when-closed-text="closed" data-when-open-text="open" data-state="closed">\
+      </div>\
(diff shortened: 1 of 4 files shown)


spec/javascripts/modules/toggle-attribute_spec.js
```diff
@@ -0,0 +1,31 @@
...
+  afterEach(function () {
+    $(document).off()
+  })
+
+  it("sets the state to open when clicked and back again", function () {
+    expect(clickon).toHaveAttr("data-state", "closed")
+    clickon.click()
+    expect(clickon).toHaveAttr("data-state", "open")
+    clickon.click()
+    expect(clickon).toHaveAttr("data-state", "closed")
```
DilwoarH: Should you have a test for analytics being trigger like this:
https://github.com/alphagov/collections/blob/master/spec/javascripts/components/accordion_spec.js#L94

PR #1586, opened 2020-03-23
Adds stay at home sections and guidance sections

DilwoarH:


config/locales/en/coronavirus_landing_page.yml
```diff
@@ -3,6 +3,17 @@ en:
     meta_title: "Coronavirus (COVID-19): what you need to do"
     meta_description: "Find out about the government response to coronavirus (COVID-19) and what you need to do."
     page_header: "Coronavirus (COVID-19): what you need to do"
+    stay_at_home:
+      pretext: "Stay at home:"
```
hannako: This doesn't match the content doc. Maybe need to check with ben. Content doc has:
`Stay at home. Protect the NHS. Save lives.`
If it's meant to match the design then need to lose the `:`

app/views/coronavirus_landing_page/_page_header.html.erb
```diff
@@ -28,6 +28,32 @@
...
+        </p>
+        <ul class="govuk-list govuk-list--bullet govuk-font-colour-white">
+          <% t("coronavirus_landing_page.stay_at_home.list").each do | item | %>
+            <li><%= item %></li>
+          <% end %>
+        </ul>
+      </div>
+    </div>
+
+    <div class="govuk-grid-row covid__page-header-guidance-section">
```
andysellick: Does this need to be in a whole new grid row?

app/views/coronavirus_landing_page/_page_header.html.erb
```diff
@@ -28,6 +28,32 @@
...
+      <div class="govuk-grid-column-two-thirds">
+        <p class="govuk-body govuk-font-colour-white govuk-!-margin-bottom-1">
+          <%= t("coronavirus_landing_page.guidance.pretext") %>
+        </p>
+        <p class="govuk-body govuk-font-colour-white">
```
andysellick: `govuk-font-colour-white` isn't needed on this para, as the only content is a link that already has a white font colour applied.

config/locales/en/coronavirus_landing_page.yml
hannako: have asked Ben

hannako: Ok ignore - sam confirmed it's just mean to be Stay at home, so just need to lost the `:`

app/views/coronavirus_landing_page/_page_header.html.erb
```diff
@@ -28,6 +28,32 @@
       </div>
     </div>
 
+    <div class="govuk-grid-row">
+      <div class="govuk-grid-column-two-thirds">
+        <p class="govuk-body govuk-!-font-weight-bold govuk-!-margin-bottom-1 govuk-font-colour-white">
```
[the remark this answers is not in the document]
DilwoarH: Yeah - I thought this would be quicker to get out rather than creating all of the CSS.

Thanks for raising

config/locales/en/coronavirus_landing_page.yml
DilwoarH: Done

config/locales/en/coronavirus_landing_page.yml
```diff
@@ -3,6 +3,17 @@ en:
     meta_title: "Coronavirus (COVID-19): what you need to do"
     meta_description: "Find out about the government response to coronavirus (COVID-19) and what you need to do."
     page_header: "Coronavirus (COVID-19): what you need to do"
+    stay_at_home:
+      pretext: "Stay at home"
```
andysellick: Are quote marks necessary in this yml? Seems like it works elsewhere without. Another thing to look at tidying later...

DilwoarH: It was required before as we had `"Stay at home:"` but not anymore

sihugh:
Merge away 👍

PR #1593, opened 2020-03-24
Fix css debt

DilwoarH:
https://trello.com/c/Sr6cZpfb/72-code-tidy-up-from-monday-night-rapid-deploys

### DIFF
--- a/app/views/coronavirus_landing_page/_page_header.html.erb
+++ b/app/views/coronavirus_landing_page/_page_header.html.erb
@@ -57,7 +57,7 @@
     <%= image_tag("corona-landing-page-logos.png", class: "covid__logos", alt: "Stay at home. Protect the NHS. Save lives.") %>
 
     <%= render "govuk_publishing_components/components/heading", {
-      text: "Announcements",
+      text: t("coronavirus_landing_page.announcements_label"),
       padding: true,
       font_size: 19,
       inverse: true
--- a/app/views/coronavirus_landing_page/_page_header.html.erb
+++ b/app/views/coronavirus_landing_page/_page_header.html.erb
@@ -30,27 +30,28 @@
 
     <div class="govuk-grid-row">
       <div class="govuk-grid-column-two-thirds">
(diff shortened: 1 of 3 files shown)


app/views/coronavirus_landing_page/_page_header.html.erb
```diff
@@ -43,10 +46,10 @@
...
       <div class="govuk-grid-column-two-thirds">
-        <p class="govuk-body govuk-font-colour-white govuk-!-margin-bottom-1">
+        <p class="govuk-body covid__page-header-body-pretext">
           <%= t("coronavirus_landing_page.guidance.pretext") %>
         </p>
-        <p class="govuk-body govuk-font-colour-white">
+        <p class="govuk-body">
           <a class="govuk-link covid__page-header-link" href="<%= t("coronavirus_landing_page.guidance.link.href") %>">
             <%= t("coronavirus_landing_page.guidance.link.text") %>
           </p>
```
andysellick: I think this trailing `</p>` is unnecessary - should be an `</a>`?

app/views/coronavirus_landing_page/_page_header.html.erb
```diff
@@ -43,10 +46,10 @@
 
     <div class="govuk-grid-row covid__page-header-guidance-section">
```
andysellick: I think you could drop the `govuk-grid-row` from this element and fold it into the two thirds column that precedes it?

app/views/coronavirus_landing_page/_page_header.html.erb
sihugh: This could be why the tests are all failing - just checking in a branch

sihugh: It was - that fix is merged if you rebase now

app/views/coronavirus_landing_page/_page_header.html.erb
DilwoarH: done

app/views/coronavirus_landing_page/_page_header.html.erb
DilwoarH: omg! #fail

PR #1600, opened 2020-03-25
Automatically hide the global banner after viewing coronavirus landing page

vanitabarrett:
## What
When a user lands on the coronavirus landing page, set the global_bar_seen cookie to `999` (manually dismissed) so the main portion of the banner is hidden (see https://github.com/alphagov/static/pull/2109 for the relevant changes to the banner)

## Why
The banner is being changed to include the key information that's also on the coronavirus landing page. We assume that if someone has viewed the landing page, they don't need to see the same information repeated in the banner.

**Note: this doesn't work if a user hasn't accepted cookies before landing on the page. This requires a change to `govuk_publishing_components` - I'll raise a PR for this shortly.**

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -7,4 +7,5 @@
 //= require modules/feeds.js
 //= require modules/toggle-attribute
 //= require components/accordion
+//= require modules/coronavirus-landing-page
 //= require govuk_publishing_components/all_components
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/application.js
```diff
@@ -7,4 +7,5 @@
 //= require modules/feeds.js
 //= require modules/toggle-attribute
 //= require components/accordion
+//= require modules/coronavirus_landing_page
```
alex-ju: minor, we tend to use dash-separated filenames for JavaScript, including Collections

PR #1605, opened 2020-03-26
Now makes width 100% of all sections up to 950px

DilwoarH:
This makes it easier for users to view on tablet devices

https://trello.com/c/GTwS7EgD/92-investigate-landing-page-breakpoints

### DIFF
--- a/app/assets/stylesheets/views/_covid.scss
+++ b/app/assets/stylesheets/views/_covid.scss
@@ -1,5 +1,14 @@
 $covid-green: #7eff8e;
 
+.covid__page {
+  .govuk-grid-column-one-third,
+  .govuk-grid-column-two-thirds {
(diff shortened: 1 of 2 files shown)


andysellick:
I'm not convinced by this change - I agree it looks a bit squashed at the extremes but this alternative leaves the NHS box looking too wide on tablet. Have we thought about alternatives? For example, we could reduce the font size and the padding of the NHS box on tablet, e.g.

Screenshot 2020-03-26 at 11 32 16

DilwoarH:
@andysellick as per our conversation on slack Mia and Sam think this approach is better for now. Let's review this again if the layout or components change.

DilwoarH:
> Okay. I think there are better ways to solve this problem but we'll go with this for now.

I'm happy to do a better alternative once we have something better.

PR #1615, opened 2020-03-27
Slight redesign to covid landing page

andysellick:
- make social distancing link more prominent
- move images to the right
- amend design of announcements section
- wrap all in a grey, not black, box


app/assets/stylesheets/views/_covid.scss
```diff
@@ -10,43 +11,73 @@ $covid-green: #7eff8e;
...
   &:link,
   &:visited {
     color: govuk-colour("white");
   }
   &:focus {
     color: $govuk-text-colour;
   }
+}
+
+.covid__highlight-link {
```
vanitabarrett: The green SVG on yellow focus is a bit painful - perhaps the focus state should only be on the text? 

Screenshot 2020-03-27 at 18 28 00

andysellick: Ah yes, I'd forgotten about that. Okay, good point, will take a look.

andysellick: @vanitabarrett have fixed.

vanitabarrett: Looks awesome! 👏

andysellick:
@DilwoarH yeah, I think hard coded numbers are the inevitable by-product of high pressure changes like this. We can look at alternatives next week if it's a problem.

PR #1621, opened 2020-03-30
Fix green arrow and NHS svg

andysellick:
Text for the green arrow link was changed so no longer wraps onto two lines by default. Applying flexbox to fix vertical layout. Before:

Screenshot 2020-03-30 at 09 16 22

After:

Screenshot 2020-03-30 at 09 16 50

At a smaller screen size:

Screenshot 2020-03-30 at 09 17 01

At an extremely small screen size:

Screenshot 2020-03-30 at 09 17 26

Also fixes a problem with the NHS SVG logo on Internet explorer. There's a weird IE thing where the canvas size doesn't play well with scaling the SVG, but adding a height (apparently even to the containing element) crops it.

Before:

Screenshot 2020-03-30 at 09 15 22

After:

Screenshot 2020-03-30 at 09 15 59

Trello card: https://trello.com/c/aUQMJcEV/101-re-align-the-green-arrow-icon-to-the-featured-link-in-the-header

### DIFF
--- a/app/assets/stylesheets/views/_covid.scss
+++ b/app/assets/stylesheets/views/_covid.scss
@@ -58,15 +58,19 @@ $covid-grey: #272828;
 }
 
 .covid__highlight-link {
-  box-sizing: border-box;
   display: inline-block;
+  display: -ms-flexbox;
+  display: flex;
+  -ms-flex-align: center;
+  align-items: center;
+  box-sizing: border-box;
   padding-left: 50px;
   min-height: 45px;
   max-width: 460px;
   background-image: image-url("covid-arrow.png"); // PNG fallback for SVG
-  background: image-url("covid-arrow.svg");
+  background-image: image-url("covid-arrow.svg");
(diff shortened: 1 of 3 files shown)


app/assets/stylesheets/views/_covid.scss
```diff
@@ -75,8 +79,12 @@ $covid-grey: #272828;
     margin-bottom: govuk-spacing(6);
   }
 
-  @include govuk-media-query($from: tablet) {
-    background-position: 0 3px;
+  // IE10/11 specific - vertical align in flexbox is ignored
+  // unless an explicit height is set
+  @include govuk-media-query($from: desktop) {
+    @media screen and (min-width:0\0) {
```
DilwoarH: `0\0`? - can this not just be 0?

andysellick: No - this is an IE10/11 specific hack. See the commit message for details.

PR #1622, opened 2020-03-30
Changes covid SVG to work with IE11

DilwoarH:
# Before
Screenshot 2020-03-30 at 10 08 02

# After
Screenshot 2020-03-30 at 10 54 57


DilwoarH:
> Good stuff. Do we know what was the problem with the original SVG? Would be good to know/document.

This is what Mia said: 
> they were a combination of grouped elements and 'flattened' elements. The flattened elements were the ones that displayed. So I flattened everything into one element

PR #1627, opened 2020-03-31
Add initial FE for livestream video

andysellick:
Trello card: https://trello.com/c/fYzEgmTP/116-build-the-ability-to-feature-a-youtube-livestream-on-the-landing-page

## With cookie consent
Screenshot 2020-03-31 at 15 56 48

## Without cookie consent
Screenshot 2020-03-31 at 15 37 28


app/views/coronavirus_landing_page/_page_header.html.erb
```diff
@@ -95,6 +95,16 @@
           </div>
         <% end %>
       </div>
+
```
hannako: Are we not doing the following from the trello card?
```
Off - Time not announced:
The next live press conference will be shown here
View past coronavirus press conferences on YouTube
```

DilwoarH: I think we'll have to do that as an iteration

hannako: ok

PR #1628, opened 2020-03-31
Expect new live stream fields from collections publisher

hannako:
Adds a new `live_stream` method to the presenter. 

Need to add the new field to the content item before we can merge this pr.

### DIFF
--- a/app/presenters/coronavirus_landing_page_presenter.rb
+++ b/app/presenters/coronavirus_landing_page_presenter.rb
@@ -1,5 +1,5 @@
 class CoronavirusLandingPagePresenter
-  COMPONENTS = %w(stay_at_home guidance announcements_label announcements nhs_banner sections topic_section country_section notifications).freeze
+  COMPONENTS = %w(live_stream stay_at_home guidance announcements_label announcements nhs_banner sections topic_section country_section notifications).freeze
 
   def initialize(content_item)
     COMPONENTS.each do |component|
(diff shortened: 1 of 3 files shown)


app/presenters/coronavirus_landing_page_presenter.rb
```diff
@@ -8,4 +8,39 @@ def initialize(content_item)
       end
     end
   end
+
+  def parsed_stream_date
+    if live_stream_date.present?
+      ints = live_stream_date.split("/").map(&:to_i)
+      begin
+        DateTime.new(ints[2], ints[1], ints[0]).strftime("%d %B %Y").sub(/^0/, "")
```
DilwoarH: Hey @hannako - if we're just displaying the date, for now, do we need to parse it?

hannako: The current instruction on the yaml file is to enter the dates as `31/03/2020` and time as `17:00` so we would need to parse them if they're to be displayed in the style that's in the design eg `31 March 2020`.

But we could require that the dates are inputted exactly as they'll be displayed instead, eg enter `1 March 2020` into the yaml file. But it depends if we're using the date to decide whether or not to display certain fields. I wasn't sure what your logic is. We don't have any validations on the date that's inputed in the yaml, or sent to publishing api by collections publisher.

hannako:
Closing PR for today as for current requirements we can just depend on the `show_video` flag rather than the date field

PR #1631, opened 2020-04-01
Add check for campaign cookie

hannako:
Check if a user has agreed to campaign cookies and pass that state through to the views. This can be used to make https://github.com/alphagov/collections/pull/new/toggle_video_placeholder work.

### DIFF
--- a/app/controllers/coronavirus_landing_page_controller.rb
+++ b/app/controllers/coronavirus_landing_page_controller.rb
@@ -5,7 +5,7 @@ class CoronavirusLandingPageController < ApplicationController
   def show
     @content_item = content_item.to_hash
     breadcrumbs = [{ title: "Home", url: "/", is_page_parent: true }]
-    render "show", locals: { breadcrumbs: breadcrumbs, details: presenter }
+    render "show", locals: { breadcrumbs: breadcrumbs, details: presenter, campaign_cookie: campaign_cookie }
   end
 
 private
@@ -16,5 +16,17 @@ private
 
   def presenter
     CoronavirusLandingPagePresenter.new(content_item.to_hash)
+  end
+
+  def cookie_policy
+    if cookies.has_key?("cookies_policy")
+      JSON.parse(cookies["cookies_policy"])
+    else
+      {}
+    end
+  end
+
+  def campaign_cookie
+    cookie_policy["campaigns"]
   end
 end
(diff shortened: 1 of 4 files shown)


app/controllers/coronavirus_landing_page_controller.rb
```diff
@@ -17,4 +17,16 @@ def content_item
...
+  def cookie_policy
+    if cookies.has_key?("cookies_policy")
+      JSON.parse(cookies["cookies_policy"])
+    else
+      {}
+    end
+  end
+
+  def campaign_cookie
+    cookie_policy["campaigns"]
```
DilwoarH, line 30: Won't this fail if the above returns {}

hannako: it will return nil which is ok

hannako: In the `_live_stream` partial you'd have 
```
<% if campaign_cookie == true %>
 show video
<% else %>
show placeholder
<% end %>

File: app/controllers/coronavirus_landing_page_controller.rb [hunk 2]
```diff
@@ -17,4 +17,16 @@ def content_item
   def presenter
     CoronavirusLandingPagePresenter.new(content_item.to_hash)
   end
+
+  def cookie_policy
+    if cookies.has_key?("cookies_policy")
```
sihugh, line 22: What's `cookies` here?

hannako: it's this:
```
pry(#<CoronavirusLandingPageController>)> cookies
=> #<ActionDispatch::Cookies::CookieJar:0x00007f7f41518138
 @committed=false,
 @cookies=
  {"cookies_policy"=>"{\"essential\":true,\"settings\":true,\"usage\":true,\"campaigns\":true}",
   "cookies_preferences_set"=>"true",
   "global_bar_seen"=>"{\"count\":999,\"version\":6}",
   "_ga"=>"GA1.1.1469247533.1585502073",
   "_gid"=>"GA1.1.1336703964.1585502073"},
 @delete_cookies={},
 @request=
  #<ActionDispatch::Request:0x00007f7f416dcac8
   @env= ...
```

sihugh: We can't generally do cookie parsing/behaviour in ruby code because most things are served from Fastly which serves cached html/js/css.  Feels like we might need to do this client side with CSS?

hannako: ah yes, thanks for explaining.

PR #1639, opened 2020-04-02
Extract partials from coronavirus views

sihugh:
This should make working with subsections of the page easier.


app/views/coronavirus_landing_page/_page_header.html.erb
```diff
@@ -1,17 +1,3 @@
-<% content_for :is_full_width_header, true %>
```
sihugh, line 1: I extracted this from _page_header altogether and included it at the top of show.html.erb

app/views/coronavirus_landing_page/_topic_section.html.erb
```diff
@@ -1,22 +1,3 @@
-<section class="covid__topic-wrapper">
```
sihugh, line 1: this is now back in show.html.erb as well

PR #1640, opened 2020-04-02
Add beginning of business page

sihugh:
The skeleton of the business page is based on the coronavirus landing page.

This will be iterated tomorrow.

Note that the `stay_at_home` key in the content item is `header_section` in the new business content item and there are differences between them which may not be totally reflected in the fixture yet

We'll have an alignment session after this is built.


sihugh:
We can update the fixture with a copy paste from the content item when published

PR #1651, opened 2020-04-03
Update breadcrumbs

vanitabarrett:
Trello: https://trello.com/c/UnaD3yvm/134-fix-breadcrumbs-to-look-like-the-design-topic-breadcrumbs

## What

- Removes the coronavirus page from the business breadcrumbs
- Links to the coronavirus page from title context
- Switches hardcoded title for component
- Makes the whole business header section grey rather than black.

## Before
Screenshot 2020-04-03 at 14 09 05

## After
Screenshot 2020-04-03 at 14 08 49

### DIFF
--- a/app/assets/stylesheets/views/_covid.scss
+++ b/app/assets/stylesheets/views/_covid.scss
@@ -27,6 +27,14 @@ $covid-grey: #272828;
   color: govuk-colour("white");
 }
 
+.covid__page-header--business {
+  background-color: $covid-grey;
+}
+
+.covid__page-header--hub {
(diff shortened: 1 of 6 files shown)


DilwoarH:
On the design - the COVID 19 subheading is a link. Did we want to bring that across?

vanitabarrett:
@DilwoarH I'm not sure what you mean, that should be a link in this PR?

vanitabarrett:
@DilwoarH That will require a change in the components gem - I would prefer to get this change in  now and then iterate, as releasing the gem can take a while. Does that sound ok?

PR #1653, opened 2020-04-03
Update opengraph images

DilwoarH:
https://trello.com/c/dMAEdrCh/145-update-coronavirus-branding-graphics-on-the-social-cards


vanitabarrett:
Do we need to update the image in the /coronavirus page header, or is that a separate card?

DilwoarH:
Yh we do - just waiting for the images.

vanitabarrett:
Sounds like the other images still need sign-off but these are good to go, so let's do those other ones as a separate PR 👍

DilwoarH:
Just added support to not show the coronavirus images for the business page

PR #1656, opened 2020-04-03
Remove related links for now

DilwoarH:


sihugh:
I guess we don't have any tests in for those yet 😬

DilwoarH:
> I guess we don't have any tests in for those yet 😬

We do - I forgot about that

PR #1662, opened 2020-04-06
Add previous video text from content item

DilwoarH:


andysellick:
@DilwoarH the review app page doesn't work - does it require some changes in the yml to be deployed first?

DilwoarH:
> @DilwoarH the review app page doesn't work - does it require some changes in the yml to be deployed first?

Hi Yes - needs new content deployed to live

PR #1664, opened 2020-04-07
Add tracking for accordion open all link

DilwoarH:
We currently track users who click on and interact with accordions but what we don't track is the number of users who interact with the expand and collapse all feature. 

Adding tracking to this will help us determine if users that click this option have their journey altered in some way or don't make it through a large proportion of the page.  

The event structure should follow the accordion structure currently in place. 

Event category: pageElementInteraction
Event action: accordionOpened/accordionClosed
Event label: Expand all/Collapse all

It's done when: 

We can see when users interact with the expand and collapse all portions of the accordion interaction. 

https://trello.com/c/so5IN6ZA/157-add-tracking-to-the-expand-all-option-on-accordions


DilwoarH:
> Looks good. Does it need a test?

You’re right, I’ll get on that

PR #1681, opened 2020-04-18
Changes to how we render live stream

hannako:
- ref: https://github.com/alphagov/collections-publisher/pull/996
- toggle livestream component on presence of new key `live_stream_enabled`
- calculate date instead of fetching from the yaml
[Trello](https://trello.com/c/tsal0J9i/158-create-a-way-for-content-designers-to-switch-the-livestream-on-and-off)


sihugh:
I've just rebased this to pick up changes in https://github.com/alphagov/collections/pull/1683

PR #1682, opened 2020-04-19
Cucumber configuration

hannako:
Adds back in some of the code removed by [6a13dbf52](https://github.com/alphagov/collections/commit/6a13dbf52) as it has been causing load errors eg

```
21:33:44 *** [err :: ip-10-12-4-5.eu-west-1.compute.internal] rake aborted!
21:33:44 *** [err :: ip-10-12-4-5.eu-west-1.compute.internal] LoadError: cannot load such file -- cucumber/rake/task
```


lib/tasks/cucumber.rake
```diff
@@ -1,18 +1,42 @@
...
-    t.fork = true # You may get faster startup if you set this to false
-    t.profile = "default"
-  end
-
-  Cucumber::Rake::Task.new({ wip: "test:prepare" }, "Run features that are being worked on") do |t|
-    t.fork = true # You may get faster startup if you set this to false
-    t.profile = "wip"
-  end
-end
+# IMPORTANT: This file is generated by cucumber-rails - edit at your own peril.	require "cucumber/rake/task"
```
AlanGabbianelli, line 1:
```suggestion
# IMPORTANT: This file is generated by cucumber-rails - edit at your own peril.
```

lib/tasks/cucumber.rake
```diff
@@ -1,18 +1,42 @@
-require "cucumber/rake/task"
+# IMPORTANT: This file is generated by cucumber-rails - edit at your own peril.
+# It is recommended to regenerate this file in the future when you upgrade to a
+# newer version of cucumber-rails. Consider adding your own code to a new file
```
ChrisBAshton, line 3: It looks as though the best approach would be to just generate this file from cucumber-rails (a sort of a return to 'factory settings'), rather than try to edit it, as 6a13dbf52 did that and caused problems.

The original problem 6a13dbf52 was trying to solve was to fix linting errors - but we could add a rule to [.rubocop.yml](https://github.com/alphagov/collections/blob/master/.rubocop.yml) to exclude `lib/tasks/cucumber.rake`, which seems reasonable for a generated file.

AlanGabbianelli: I think that's sensible @ChrisBAshton 👍
@benthorner and @hannako what do you think?

benthorner: Can do. I've also got this, since we're unlikely to ever need to regenerate this file: https://github.com/alphagov/collections/pull/1683.

sihugh: 👋 @ChrisBAshton  I've regenerated the files.  I'll squash if you're happy?

sihugh: Ah, I'll rebase everything

benthorner: Just posting here, since we merged https://github.com/alphagov/collections/pull/1683, which should fix the problem.

benthorner: We could use this one if we want to keep the original file. I've always found it surprising to see such a complicated rake task with all the other ones, but I appreciate that's more personal preference.

sihugh: Yeah, I'll close this, thanks @benthorner

PR #1684, opened 2020-04-20
Implement breadcrumb change

andysellick:
Make the breadcrumb on the [hub page](https://www.gov.uk/coronavirus/business-support) more obvious.

Before:

Screenshot 2020-04-20 at 16 20 17

After:

Screenshot 2020-04-20 at 16 20 22

Trello card: https://trello.com/c/z4DVsyk0/200-make-hub-page-breadcrumbs-more-explicit

### DIFF
--- a/app/assets/stylesheets/views/_covid.scss
+++ b/app/assets/stylesheets/views/_covid.scss
@@ -37,7 +37,7 @@
   @include govuk-media-query($until: tablet) {
     margin-bottom: govuk-spacing(3);
   }
-  
+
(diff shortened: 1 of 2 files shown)


andysellick:
@sihugh I initially put the context in as the link text and it showed that it had those two attributes, so that's what I'm using 🤷 

The full stop at the end isn't a full stop! It's Chrome doing that thing where it breaks up the underline when characters cross it. Yeah, looks weird, doesn't it...

PR #1685, opened 2020-04-21
See all announcements

DilwoarH:
## Context
User research from Italy showed that users struggled to keep track of what they can and can't do as regulations changed over time. Unlike guidance announcements are time-sensitive and foregrounding dates is useful.

## What
Add dates and a [See all announcements](https://www.gov.uk/search/news-and-communications?topical_events%5B%5D=coronavirus-covid-19-uk-government-response) link to the announcements section.

## Why
To help users understand when announcements have been made, and that they are current.

## Done when
Announcement dates and a 'See all announcements' link are live on the homepage.

https://trello.com/c/2dh1maeg/195-update-announcements-section-with-dates-and-see-all-link

### DIFF
--- a/app/assets/stylesheets/views/_covid.scss
+++ b/app/assets/stylesheets/views/_covid.scss
@@ -51,6 +51,18 @@
   }
 }
 
+.covid__page-header-announcement {
+  margin-bottom: govuk-spacing(7);
+
+  @include govuk-media-query($from: desktop) {
+    margin-bottom: govuk-spacing(3);
+  }
+
+  .govuk-body {
+    margin-bottom: govuk-spacing(1);
+  }
+}
+
 .covid__business-list-wrapper {
   margin-bottom: 45px;
 }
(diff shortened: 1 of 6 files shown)


app/presenters/coronavirus_landing_page_presenter.rb
```diff
@@ -1,5 +1,5 @@
 class CoronavirusLandingPagePresenter
-  COMPONENTS = %w(live_stream live_stream_enabled stay_at_home guidance announcements_label announcements nhs_banner sections topic_section country_section notifications find_help).freeze
+  COMPONENTS = %w(live_stream live_stream_enabled stay_at_home guidance announcements_label announcements see_all_announcements_link nhs_banner sections topic_section country_section notifications find_help).freeze
```
andysellick, line 2: This isn't part of your change, but I find the use of the term 'components' here and elsewhere confusing, since they're not components in the sense that we think of components. We should have a conversation about that at some point.

app/assets/stylesheets/views/_covid.scss
```diff
@@ -51,6 +51,18 @@
   }
 }
 
+.covid__page-header-announcement {
+  margin-bottom: govuk-spacing(3);
+
+  @include govuk-media-query($until: desktop) {
+    margin-bottom: govuk-spacing(7);
```
andysellick, line 58: It makes no real difference, but I feel instinctively that we should write our code mobile first, so this would be 

```
.covid__page-header-announcement {
  margin-bottom: govuk-spacing(7);

  @include govuk-media-query($from: desktop) {
    margin-bottom: govuk-spacing(3);
  }
```

app/presenters/coronavirus_landing_page_presenter.rb
DilwoarH: Yh - we probably need to go through all of the hub code and standardise things.

app/assets/stylesheets/views/_covid.scss
DilwoarH: > Seems okay, I'd like to check the review app but it doesn't seem to have deployed yet?

Github was down for a bit yesterday which broke a few things, I'll get it to trigger again.

PR #1699, opened 2020-04-24
Optionally show the ask a question link

sihugh:
We probably want the ability to turn the question link off without needing a deploy.

The question link shows correctly when the content item is configured:

Screenshot 2020-04-24 at 20 25 07

### DIFF
--- a/test/support/coronavirus_landing_page_steps.rb
+++ b/test/support/coronavirus_landing_page_steps.rb
@@ -22,6 +22,10 @@ module CoronavirusLandingPageSteps
     stub_content_store_has_item(CORONAVIRUS_PATH, coronavirus_content_item_with_live_stream_enabled_and_date)
   end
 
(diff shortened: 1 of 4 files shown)


test/support/coronavirus_landing_page_steps.rb
```diff
@@ -65,6 +69,14 @@ def then_i_can_see_the_live_stream_section_with_date_and_time
     assert page.has_text?("5:00pm")
   end
 
+  def then_i_can_see_the_ask_a_question_section
+    assert page.has_link?("Ask a question at the next press conference", href: "https://www.gov.uk")
```
leenagupte, line 73: Should this be "https://www.gov.uk/ask" ?

leenagupte: We've cleared this up on slack. I didn't realise the value is being pulled in from the fixture.

PR #1697, opened 2020-04-24
Landing page accordion updates

danacotoran:
### What
- Update the accordion to add the title: Guidance and support
- Add a line below the accordion with the country specific guidance:

Additional guidance for [Scotland](https://www.gov.uk/guidance/coronavirus-covid-19-information-for-individuals-and-businesses-in-scotland), [Wales](https://gov.wales/coronavirus) and [Northern Ireland](https://www.nidirect.gov.uk/campaigns/coronavirus-covid-19)
- Remove the old 'If you live in Scotland, Wales or Northern Ireland' section

### Why
Titling sections supports scanning in an increasingly long page. We can save a bit of vertical space by condensing this into a line.

Relevant Trello card: https://trello.com/c/WyQxHLMh

PR #1701, opened 2020-04-27
Tweaks post accordion updates

danacotoran:
- Tweak alignment of accordion title to more closely match spec

**Before (note alignment of title to NHS box)**
Screenshot 2020-04-27 at 09 34 27
**After**
Screenshot 2020-04-27 at 09 35 16

- Remove references to `country_section` content from
  `CoronavirusLandingPagePresenter` as we are now pulling country specific
content from `additional_country_guidance` instead (see #1697)

 https://trello.com/c/WyQxHLMh


danacotoran:
> Looks good. Maybe would be cleaner to have the two changes in separate commits, but fine as it is.

@andysellick thank you for approving this, and regarding your comment – that's a good point. 
I broke the changes up into two separate commits

PR #1709, opened 2020-05-01
Render organisation people from links

thomasleese:
This renders the people section of the organisation pages from the links, rather than from the details hash. We've recently started exposing this information in the links so we can rely on the Publishing API to perform dependency resolution for us rather than Whitehall having to manually republish organisation pages whenever a person changes.

Example: https://govuk-collec-organisati-5obtcx.herokuapp.com/government/organisations/cabinet-office

[Trello Card](https://trello.com/c/NRhnejWA/1942-5-send-links-to-people-instead-of-details-hash-in-the-organisations-in-whitehall)

### DIFF
--- a/app/presenters/organisations/people_presenter.rb
+++ b/app/presenters/organisations/people_presenter.rb
@@ -9,11 +9,7 @@ module Organisations
     end
 
     def has_people?
-      all_people.each do |people|
-        return true unless people[:people].empty?
-      end
-
-      false
+      all_people.map { |people| people[:people] }.any?(&:present?)
     end
 
     def all_people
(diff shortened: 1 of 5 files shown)


app/presenters/organisations/people_presenter.rb
```diff
@@ -27,32 +27,6 @@ def all_people
...
-              else
-                all_people_item[:description] = multiple_role_description(all_people_item, person)
-              end
-            end
-          end
-        end
-      end
-
-      all_people
-    end
```
ChrisBAshton, line 58: So glad this block has gone!

thomasleese: Same...!

thomasleese:
> I don't know the implication of moving from s465 images to the default ones

The default ones provided in the links are actually already the s465 ones, so there isn't a change in functionality here, it's just that where the data comes from is different.

PR #1721, opened 2020-05-10
Update alt text to match text in image

quis:


sihugh:
Thanks for noticing @quis, updating that now. 👏

PR #1732, opened 2020-05-14
Risk level indicator

danacotoran:
### What
Build a page element to indicate the national coronavirus risk-level.

Screenshot 2020-05-14 at 14 45 54

### Why
👉 👉 👉 https://trello.com/c/zPDbYoE3

https://github.com/alphagov/govuk-coronavirus-content/pull/234 needs to be reviewed and published first.

### DIFF
--- a/app/views/coronavirus_landing_page/components/landing_page/_risk_level_section.html.erb
+++ b/app/views/coronavirus_landing_page/components/landing_page/_risk_level_section.html.erb
@@ -1,17 +1,17 @@
 <% if data["show_risk_level_section"] %>
   <%= render "govuk_publishing_components/components/govspeak", {
   } do %>
-    <div class="info-notice">
+    <div class="info-notice covid__alert-notice">
       <% if data["heading"].present? %>
         <h3><%= data["heading"] %></h3>
       <% end %>
       <% if data["paragraph"].present? %>
-        <p class="govuk-body">
+        <p class="govuk-body covid__alert-notice-body">
           <%= highlight(data["paragraph"], data["bold_text"], highlighter: '<strong>\1</strong>') %>
         </p>
       <% end %>
       <% if data["link"].present? %>
-        <p class="govuk-body">
+        <p class="govuk-body covid__alert-notice-body">
           <a href="<%= data["link"]["href"] %>"><%= data["link"]["text"] %></a>
         </p>
       <% end %>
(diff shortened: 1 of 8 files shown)


app/helpers/application_helper.rb
```diff
@@ -55,4 +55,12 @@ def t_fallback(key, options = {})
       false
     end
   end
+
+  def make_bold_substring(string, substring)
+    if (string.is_a? String) && (substring.is_a? String) && string.include?(substring)
+      string.gsub(substring, "<strong>#{substring}</strong>").html_safe
+    else
+      string
+    end
```
reggieb, line 64: I'd suggest looking at the[ Rails `highlight` method](https://api.rubyonrails.org/classes/ActionView/Helpers/TextHelper.html#method-i-highlight).

andysellick: Nice 👍

danacotoran:
> Looks fine to me but the `make_bold_substring` method looks a bit verbose, maybe ask @sihugh if there's another way of writing it?

@andysellick thanks for the comment! updated the PR based on @reggieb's suggestion 👍

app/views/coronavirus_landing_page/components/shared/_announcements_section.html.erb
```diff
@@ -24,4 +22,5 @@
       </a>
     </p>
   <% end %>
+  <%= render partial: 'coronavirus_landing_page/components/landing_page/risk_level_section', locals: { data: details.risk_level } if details.risk_level %>
```
hannako, line 25: Once we've added the key to the yaml validation specs on govuk-coronavirus-content we won't need the check here. But for now I guess it's sensible as things are a bit fluid!

app/views/coronavirus_landing_page/components/landing_page/_risk_level_section.html.erb
```diff
@@ -0,0 +1,20 @@
+<% if data["show_risk_level_section"] %>
+  <%= render "govuk_publishing_components/components/govspeak", {
+  } do %>
+    <div class="info-notice">
+      <% if data["heading"].present? %>
+        <h3><%= data["heading"] %></h3>
+      <% end %>
+      <% if data["paragraph"].present? %>
+        <p class="govuk-body">
+          <%= highlight(data["paragraph"], data["bold_text"], highlighter: '<strong>\1</strong>') %>
```
hannako, line 10: nice!

app/views/coronavirus_landing_page/components/shared/_announcements_section.html.erb
danacotoran: **:nod:**

PR #1733, opened 2020-05-14
Featured links automatically showing description

DilwoarH:
## What
Add meta descriptions automatically below hub links in the accordion. (See attached design).

## Why
To give me context on the hub content, and allow for short links.

Screenshot 2020-05-14 at 12 23 33

https://trello.com/c/rBSB0bIQ/281-automatically-add-the-hub-page-metadescription-below-hub-links-in-the-accordion

### DIFF
--- a/app/assets/stylesheets/views/_covid.scss
+++ b/app/assets/stylesheets/views/_covid.scss
@@ -144,6 +144,10 @@
   }
 }
 
+.covid__list-item-description {
+  padding-left: govuk-spacing(6);
+}
+
 .covid__list-item {
   margin-bottom: govuk-spacing(2);
 }
--- a/app/views/coronavirus_landing_page/components/shared/_section.html.erb
+++ b/app/views/coronavirus_landing_page/components/shared/_section.html.erb
@@ -11,6 +11,12 @@
             href: item["url"],
             text: item["label"]
           }%>
+          <% item_description = content_item_description item["url"] %>
+          <% if item_description %>
+            <p class="govuk-body covid__list-item-description">
+              <%= item_description %>
+            </p>
+          <% end %>
         <% else %>
           <%= link_to(
             item["label"],
--- a/app/helpers/content_item_helper.rb
+++ b/app/helpers/content_item_helper.rb
@@ -1,0 +1,6 @@
+module ContentItemHelper
+  def content_item_description(content_item_url)
+    content_item = ContentItem.find!(content_item_url)
+    content_item.description
+  end
+end
--- a/test/unit/content_item_helper_test.rb
+++ b/test/unit/content_item_helper_test.rb
@@ -1,0 +1,16 @@
+require "test_helper"
+
+describe ContentItemHelper do
+  setup do
+    url = content_store_endpoint + "/content/blah/blah"
+    stub_request(:get, url).to_return(status: 200, body: {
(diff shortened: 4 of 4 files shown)


app/views/coronavirus_landing_page/components/shared/_section.html.erb
```diff
@@ -11,6 +11,12 @@
             href: item["url"],
             text: item["label"]
           }%>
+          <p class="govuk-body covid__list-item-description--featured">
+            <%=
+              featured_link_content_item = ContentItem.find!(item["url"])
+              featured_link_content_item.description
```
reggieb, line 17: I'd suggest using a helper method here. I'd add a helper at `app/helpers/content_item_helper.rb`:
```ruby
module ContentItemHelper
  def content_item_description(item)
    content_item = ContentItem.find!(item['url'])
    content_item.description
  end
end
```
Then here I'd just have:
```ruby
<%= content_item_description item %>
```

Oh and I'm assuming `item['description']` doesn't return the description - that is the description isn't already in the `item` hash

app/assets/stylesheets/views/_covid.scss
```diff
@@ -141,6 +141,10 @@
     @include govuk-media-query($from: tablet) {
       margin-bottom: govuk-spacing(3);
     }
+
+    .covid__list-item-description--featured {
```
andysellick, line 145: I think this class should be `covid__list-item-description` as it's an element, not a modifier on an existing element.

andysellick, line 145: Do we need to nest this class like this? We generally avoid nesting with BEM if we can.

app/views/coronavirus_landing_page/components/shared/_section.html.erb
```diff
@@ -11,6 +11,9 @@
             href: item["url"],
             text: item["label"]
           }%>
+          <p class="govuk-body covid__list-item-description--featured">
+            <%= content_item_description item["url"] %>
```
andysellick, line 15: If this is blank, will it render an empty paragraph?

DilwoarH: good spot, was meant to do that but forgot

DilwoarH:
@andysellick updated

app/views/coronavirus_landing_page/components/shared/_section.html.erb
```diff
@@ -11,6 +11,12 @@
             href: item["url"],
             text: item["label"]
           }%>
+          <% item_description = content_item_description item["url"] %>
```
danacotoran, line 14: Should this be moved to the top of the file to make it a bit cleaner?

reggieb: As it's only used once, I think it's nice to have the definition close to where it was used. If it was used multiple times, then I think moving it to the top would be a good idea.

DilwoarH: We can't, unfortunately, move this to the top as it's within a loop and we need to generate this based on the scope

danacotoran: @DilwoarH ah yes – that makes sense

PR #1736, opened 2020-05-14
Add popular questions link

danacotoran:
Add template logic to allow for an optional popular questions link to be added after the other links in the live stream section.

Add (optional) CSS class to the group of links. The CSS class adds additional margin after every 2nd link in the list, for increased legibility. 
The default state will be no extra spacing, unless `spaced_links` is set to `true` in YML

----
Relevant content PR to go with this change 👉  https://github.com/alphagov/govuk-coronavirus-content/pull/236
Trello card 👉 https://trello.com/c/tGEzCbPw

### DIFF
--- a/app/assets/stylesheets/views/_covid.scss
+++ b/app/assets/stylesheets/views/_covid.scss
@@ -125,6 +125,19 @@
     margin-bottom: govuk-spacing(4);
   }
 }
+.covid__spaced-list>li {
+  &:nth-child(2n) {
+    margin-bottom: govuk-spacing(5);
+
+    @include govuk-media-query($from: desktop) {
+      margin-bottom: govuk-spacing(4);
+    }
+  }
+
+  &:last-child {
+    margin-bottom: 0;
+  }
+}
 
 .covid__list--header {
   padding-left: govuk-spacing(4);
--- a/app/views/coronavirus_landing_page/components/landing_page/_live_stream_section.html.erb
+++ b/app/views/coronavirus_landing_page/components/landing_page/_live_stream_section.html.erb
@@ -1,6 +1,10 @@
+<%
+  list_css_classes = %w(govuk-list)
+  list_css_classes << "covid__spaced-list" if live_stream["spaced_links"]
+%>
 <section class="covid__topic-wrapper">
   <%= render "govuk_publishing_components/components/heading", {
-    text: details.live_stream["title"],
+    text: live_stream["title"],
     heading_level: 2,
     margin_bottom: 6,
     padding: true,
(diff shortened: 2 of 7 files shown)


test/support/coronavirus_helper.rb
```diff
@@ -50,6 +50,12 @@ def content_item_with_live_stream_enabled_and_ask_a_question_enabled
   content_item
 end
 
+def content_item_with_live_stream_enabled_and_popular_questions_link_enabled
+  content_item = coronavirus_content_item_with_live_stream_enabled
```
hannako, line 54: You should set `content_item = coronavirus_content_item` because `coronavirus_content_item_with_live_stream_enabled` is actually no longer a thing. Live stream is always enabled, and the key `live_stream_enabled` isn't on the yaml/content item anymore. 

This has made me realise that the tests need tidying up. Not a blocker for this PR, I'll make a ticket!

app/views/coronavirus_landing_page/components/landing_page/_live_stream_section.html.erb
```diff
@@ -28,5 +32,12 @@
         </a>
       </li>
     <% end %>
+    <% if details.live_stream["popular_questions_link_visible"] == true %>
```
hannako, line 35: I think we should be passing in `locals: { live_stream: details.live_stream }` to the live_stream_section partial [from the show page ](https://github.com/alphagov/collections/blob/8f3049d00f2f877fdbe7aaef4173d561efbbf8f6/app/views/coronavirus_landing_page/show.html.erb#L50) so we don't have to keep doing stuff like `details.live_stream["popular_questions_link_visible"]` we could just do `live_stream["popular_questions_link_visible"]`

Not related to your work, just spotted it as something to tidy up if you want to!

test/support/coronavirus_landing_page_steps.rb
```diff
@@ -31,6 +31,10 @@ def given_there_is_a_content_item_with_live_stream_enabled_and_ask_a_question_en
     stub_content_store_has_item(CORONAVIRUS_PATH, content_item_with_live_stream_enabled_and_ask_a_question_enabled)
   end
 
+  def given_there_is_a_content_item_with_live_stream_enabled_and_popular_questions_link_enabled
```
hannako, line 34: related to the comment above, maybe rename this `given_there_is_a_content_item_with_popular_questions_link_enabled`

test/support/coronavirus_helper.rb
danacotoran: > You should set `content_item = coronavirus_content_item` because `coronavirus_content_item_with_live_stream_enabled` is actually no longer a thing.

Done! ✅

app/views/coronavirus_landing_page/components/landing_page/_live_stream_section.html.erb
danacotoran: Done ✅ 
Thanks for the feedback, that was a good suggestion. Made things tidier.

test/support/coronavirus_landing_page_steps.rb
danacotoran: ✅

app/assets/stylesheets/views/_covid.scss
```diff
@@ -125,6 +125,19 @@
     margin-bottom: govuk-spacing(4);
   }
 }
+.covid__spaced-list>li {
```
andysellick, line 128: I'm late to this party but wanted to point out that we always should be applying styles using classes, not element selectors such as `li` here.

PR #1740, opened 2020-05-15
Revert #1733 - Featured links automatically showing description

kevindew:
This reverts https://github.com/alphagov/collections/pull/1733

We need to do a deploy of Collections now (6pm on a Friday) to get a
very small link change. I was worried about deploying this also at this
time as this represented a much bigger change to the Coronavirus page
that I don't think anyone is expecting to be shipped for this weekend.

I was also concerned about the usage of `ContentItem.find!` were this
method to fail in a HTTP request it would take down this page. For
example, if one of the links was to start 404ing this whole page would
404. We'd also get some undesirable outcomes if any time of absolute URL
was used (either internal or external).

Sorry to be a pain here @DilwoarH


DilwoarH:
Hey @kevindew rather than removing the whole thing, would it be better to just comment out the usage on the template?

DilwoarH:
Good point about 404, we should potentially add some error handing for this

DilwoarH:
Sorry on AL, does that make sense? If not we can remove it all @kevindew

kevindew:
> Hey @kevindew rather than removing the whole thing, would it be better to just comment out the usage on the template?

It seemed a bit cleaner to do a full revert then you can just revert this PR to get back to exactly where you were. It's all still here in the git history so should be easy to get back.

kevindew:
Thanks for dropping by despite your leave, you should go back to enjoying it :)

PR #1742, opened 2020-05-18
Show a description on featured links

sihugh:
We're doing this from the content item initially to validate the need.

If we want to do this from content items, we should use links in the /coronavirus content item if we can, rather than requesting lots of extra things from the API. We need to make sure that this page can cope with load.

Relies on alphagov/govuk-coronavirus-content#240 being published before they will show.(Now done ✅)

Resurrected from https://github.com/alphagov/collections/pull/1740

### DIFF
--- a/app/assets/stylesheets/views/_covid.scss
+++ b/app/assets/stylesheets/views/_covid.scss
@@ -157,6 +157,10 @@
   }
 }
 
+.covid__list-item-description {
+  padding-left: govuk-spacing(6);
+}
+
 .covid__list-item {
   margin-bottom: govuk-spacing(2);
 }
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/components/_action-link.scss
```diff
@@ -1,5 +1,6 @@
 .app-c-action-link {
   display: table;
+  margin-bottom: govuk-spacing(2);
```
andysellick, line 3: If you make this change here it affects every instance of this component. So... no. Better to put some equivalent top spacing on the `covid__list-item-description` you're adding below.

sihugh: http://govuk-collec-featured-d-9oli5t.herokuapp.com/coronavirus

PR #1743, opened 2020-05-18
Tweaks to covid featured descriptions

sihugh:
Use the 16px font and halve the margin after feedback from designers.


sihugh:
Thanks @danacotoran 👍

PR #1744, opened 2020-05-18
Fix 'Stay up to date' wrapping text

danacotoran:
On small screens the heading wraps onto a new line underneath the icon, when it should wrap underneath itself.

**Bad:**
Screenshot 2020-05-18 at 16 08 28

**Not bad:**
Screenshot 2020-05-18 at 16 08 13

Use flexbox to ensure the heading and icon align as expected on small mobiles.

https://trello.com/c/0pGcADKT

### DIFF
--- a/app/assets/stylesheets/views/_transition-landing-page.scss
+++ b/app/assets/stylesheets/views/_transition-landing-page.scss
@@ -109,6 +109,7 @@ $red: #E61E32;
 
 .landing-page__email-title {
   margin-bottom: govuk-spacing(2);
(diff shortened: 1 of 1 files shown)


danacotoran:
> This works, but it feels a little over-engineered. Is there a simpler solution? For example, if a margin-left is applied to the `landing-page__email-title` element (in the current version) it seems to solve it.

Yep you're totally right @andysellick. Did not consider that option, for whatever reason.

I've updated the PR ✅

PR #1749, opened 2020-05-20
Change email signup

chao-xian:
Collections is doing too much unnecessary work to enable signups. Email Alert Frontend can handle more of this, and should.

https://trello.com/c/UmDXU47V

[Corresponding PR on email-alert-frontend](https://github.com/alphagov/email-alert-frontend/pull/789)


vanitabarrett:
Can we also delete https://github.com/alphagov/collections/blob/master/app/views/email_signups/new.html.erb ?

app/views/subtopics/_subtopic.html.erb
```diff
@@ -11,8 +11,8 @@
 
     <div class="govuk-grid-column-one-third add-title-margin">
       <%= render "govuk_publishing_components/components/subscription-links", {
-        email_signup_link: email_signup_path(topic_slug: params[:topic_slug], subtopic_slug: params[:subtopic_slug]),
-        email_signup_link_text: "Subscribe to email alerts"
+        email_signup_link: "/email-signup?topic=/topic/#{params[:topic_slug]}/#{params[:subtopic_slug]}",
```
vanitabarrett, line 14: Will we remove/redirect the routes in a separate follow-up PR? https://github.com/alphagov/collections/blob/master/config/routes.rb#L35

benthorner: We could do that in this PR, and add a comment to say they are redundant. [[Example](https://github.com/alphagov/email-alert-frontend/blob/master/config/routes.rb#L40)]

config/routes.rb
```diff
@@ -32,11 +32,7 @@
       to: "latest_changes#index",
       as: :latest_changes
 
-  get "/topic/:topic_slug/:subtopic_slug/email-signup",
-      to: "email_signups#new",
-      as: :email_signup
-  post "/topic/:topic_slug/:subtopic_slug/email-signup",
-       to: "email_signups#create"
+  get "/topic/:topic_slug/:subtopic_slug/email-signup", to: redirect("/email-signup?topic=/topic/%{topic_slug}/%{subtopic_slug}")
```
vanitabarrett, line 35: Can we add a comment here to explain why this redirect exists?

chao-xian: Sorry yep

PR #1754, opened 2020-05-21
Use action link component

andysellick:
**https://github.com/alphagov/static/pull/2192 has now been deployed, so this change is good to go**

This PR removes the action link component from collections and replaces it with the action link component from the gem (itself a modified copy of the component from collections).

This component is also being put into the global bar in static. Because the component will be used in both collections and static (and also will appear on every other page on GOV.UK in the global banner, from static) the styles for the action link component need to be included in static but don't need to be included anywhere else. I don't want to duplicate the component styles on pages served by collections, so I've not included them in collections, instead relying on the styles in static.

This is more efficient but slightly risky (if we stop using the component in static it will appear unstyled in collections, unless we notice) so happy to debate this approach.

Trello card: https://trello.com/c/l9xuuuTB/279-update-emergency-banner-design


vanitabarrett:
@andysellick Is it worth adding a comment in the CSS in Static to warn that removing the action link CSS import will mean needing to add it into Collections if we want the action link to remain on the landing page?

andysellick:
@vanitabarrett good shout, sorry, have corrected this oversight: https://github.com/alphagov/static/pull/2200

PR #1757, opened 2020-05-21
 Email signup for hubs

danacotoran:
### What 
Add an email signup section to the Education and Business hubs similar to that on the https://www.gov.uk/coronavirus page.

### Why
We can now provide more granular sign-up.

--------

### Notes

The email signup box is being introduced on the hub pages, which means that it would have been repeated with the same markup in a few different places.

I created a new email signup component to reduce repetition in the code.

This set of changes has touched:
- Brexit transition page [live](https://www.gov.uk/transition) [preview](https://govuk-collec-email-sign-6qsiut.herokuapp.com/transition)
- COVID landing page [live](https://www.gov.uk/coronavirus) [preview](https://govuk-collec-email-sign-6qsiut.herokuapp.com/coronavirus)
- COVID business support page [live](https://www.gov.uk/coronavirus/business-support) [preview](https://govuk-collec-email-sign-6qsiut.herokuapp.com/coronavirus/business-support)
- COVID education hub page [live](https://www.gov.uk/coronavirus/education-and-childcare) [preview](https://govuk-collec-email-sign-6qsiut.herokuapp.com/coronavirus/education-and-childcare)
- this other education page  and/or every other page using the same template (not sure how many there are) [live](https://www.gov.uk/education) [preview](https://govuk-collec-email-sign-6qsiut.herokuapp.com/education)

There was a very similar component (this one! https://govuk-collections.herokuapp.com/component-guide/email-link ), using much of the same markup, which is currently only being used on one template (this is one of the pages using the template: https://www.gov.uk/education)
I thought it would be a good idea to merge the two together at this stage as they are extremely similar, and there was no reason to keep them separate (not that I could tell)

Very little of the CSS is new – I just put it all in one place and gave it more consistent naming

--------

https://trello.com/c/AFkthgYZ


app/assets/stylesheets/components/_email-signup-section.scss
```diff
@@ -0,0 +1,38 @@
+.app-c-email-signup {
+  @include govuk-responsive-margin(8, "top");
```
andysellick, line 2: We try to aim for a margin-bottom approach to everything, particularly components. Would it be possible to remove this margin and add it to whatever appears above the component? Obviously this'll be more work (if possible), sorry.

app/assets/stylesheets/components/_email-signup-section.scss
```diff
@@ -0,0 +1,38 @@
...
+  padding: govuk-spacing(6);
+  background-color: govuk-colour("light-grey", $legacy: "grey-4");
+  border: 1px solid $govuk-border-colour;
+}
+
+.app-c-email-signup--full-width {
```
andysellick, line 8: I'm not sure this accurately describes the behaviour of this option - the component appears be full width by default. Might be better to call this something else - no sides? No borders? Something like that only better 😁

app/assets/stylesheets/components/_email-signup-section.scss
```diff
@@ -0,0 +1,38 @@
...
+  margin-top: 0;
+  border-left: 0;
+  border-right: 0;
+  padding-left: 0;
+  padding-right: 0;
+}
+
+.app-c-email-signup--transparent_background {
```
andysellick, line 16: Writing this comment mainly to remind myself to check this - when it's transparent, do we still need the left/right padding?

app/views/components/_email-signup-section.html.erb
```diff
@@ -0,0 +1,45 @@
+<%
+  details ||= {}
```
andysellick, line 2: Where is this used? I can't see it.

app/assets/stylesheets/components/_email-signup-section.scss
```diff
@@ -0,0 +1,38 @@
...
+  background: none;
+  border: none;
+}
+
+.app-c-email-signup--no-intro .govuk-link {
+  @include govuk-font(19, $weight: bold);
+  display: block;
```
andysellick, line 23: With this as `block` the focus state is very large, suggest using `inline-block` instead (will need to adjust the margin left as well)

Screenshot 2020-05-26 at 11 39 02

app/assets/stylesheets/components/_email-signup-section.scss
```diff
@@ -0,0 +1,38 @@
...
+  margin-left: govuk-spacing(5);
+  margin-bottom: 0;
+}
+
+.app-c-email-signup__icon {
+  position: relative;
+  float: left;
+  margin-right: .5em;
```
andysellick, line 31: I'm not sure this is needed? Seems to add only a tiny amount of spacing, and the link to the right of it already has left margin, so if we do need that extra bit of space I think it would be better applied in a single place than multiple.

app/views/components/_email-signup-section.html.erb
```diff
@@ -0,0 +1,45 @@
...
+  tracking_data = track_action ? {
+    module: "track-click",
+    "track-category": "emailAlertLinkClicked",
+    "track-action": track_action,
+    "track-label": track_label
+  } : {}
+
+  classes << css_classes if css_classes
+  classes << "app-c-email-signup--full-width" if full_width
+  classes << "app-c-email-signup--no-intro" if !intro
```
andysellick, line 21: This can be also written as `classes << "app-c-email-signup--no-intro" unless intro` which I think is more readable.

app/views/components/_email-signup-section.html.erb
```diff
@@ -0,0 +1,45 @@
...
+  link_text ||= false
+  link_href ||= false
+  intro ||= false
+  full_width ||= false
+  track_action ||= false
+  track_label ||= false
+  transparent_background ||= false
+  css_classes ||= false
+  classes ||= []
+  tracking_data = track_action ? {
```
andysellick, line 12: We generally avoid adding tracking to components by default, I think this is better added per instance.

app/views/components/_email-signup-section.html.erb
```diff
@@ -0,0 +1,45 @@
...
+    "track-label": track_label
+  } : {}
+
+  classes << css_classes if css_classes
+  classes << "app-c-email-signup--full-width" if full_width
+  classes << "app-c-email-signup--no-intro" if !intro
+  classes << "app-c-email-signup--transparent_background" if transparent_background
+
+  #  if full_width is true, wrap content with govuk layout classes for full width styling
+  def full_width_wrapper(is_full_width, &block)
```
andysellick, line 25: What's the requirement for this? Where is it used?

app/views/components/_email-signup-section.html.erb
```diff
@@ -0,0 +1,45 @@
...
+    raw str
+  end
+%>
+<% if link_text && link_href %>
+  <div class="app-c-email-signup dont-print <%= classes.join(' ') %>">
+    <%= full_width_wrapper(full_width) do %>
+      <svg  class="app-c-email-signup__icon" aria-hidden="true" xmlns="http://www.w3.org/2000/svg" height="18" width="18" viewBox="0 0 459.334 459.334">
+        <path fill="black" d="M177.216 404.514c-.001.12-.009.239-.009.359 0 30.078 24.383 54.461 54.461 54.461s54.461-24.383 54.461-54.461c0-.12-.008-.239-.009-.359H175.216zM403.549 336.438l-49.015-72.002v-89.83c0-60.581-43.144-111.079-100.381-122.459V24.485C254.152 10.963 243.19 0 229.667 0s-24.485 10.963-24.485 24.485v27.663c-57.237 11.381-100.381 61.879-100.381 122.459v89.83l-49.015 72.002a24.76 24.76 0 0 0 20.468 38.693H383.08a24.761 24.761 0 0 0 20.469-38.694z"/>
+      </svg>
+      <%= content_tag(:h3, intro, class: "govuk-heading-s app-c-email-signup__title") if intro %>
```
andysellick, line 38: Is H3 always appropriate here, or is it worth adding an option to control the heading level?

app/views/components/docs/email-signup-section.yml
```diff
@@ -0,0 +1,34 @@
+name: Email signup section
+description: Renders a box with a link to sign up for email notifications
+shared_accessibility_criteria:
```
andysellick, line 3: Have you thought about other accessibility criteria? I'm thinking about the contrast of the text against the background(s) and how the icon should appear/not appear to screen readers.

test/components/email_signup_section_test.rb
```diff
@@ -0,0 +1,78 @@
...
+    assert_empty render_component({})
+  end
+
+  test "renders link when link is passed" do
+    render_component(
+      intro: "Stay up to date with GOV.UK",
+      link_href: "/email-signup?topic=/coronavirus-taxon",
+      link_text: "Sign up to get emails when we change any coronavirus information on the GOV.UK website",
+    )
+    assert_select ".app-c-email-signup .govuk-link"
```
andysellick, line 18: We should expand at least some of these tests to be more specific, e.g.

```
assert_select ".app-c-email-signup__title", text: "Stay up to date with GOV.UK"
assert_select ".app-c-email-signup .govuk-link[href='/(etc.)']", text: "Sign up to get emails (etc.)"
```

app/views/components/_email-signup-section.html.erb
```diff
@@ -0,0 +1,45 @@
+<%
+  details ||= {}
+  link_text ||= false
+  link_href ||= false
+  intro ||= false
+  full_width ||= false
+  track_action ||= false
+  track_label ||= false
+  transparent_background ||= false
+  css_classes ||= false
```
andysellick, line 10: We don't allow arbitrary classes to be passed to components because it destroys isolation (an application could style a component, which would make maintaining that component very difficult). Obviously this component is specific to this application, but I think this still applies. What's the use case here?

We do allow classes to be passed that are prefixed with `js-` for JS hooks, see the shared_helper in the components gem for how that's handled.

app/views/coronavirus_landing_page/components/hub/_page_header.html.erb
```diff
@@ -44,4 +44,11 @@
     </div>
   </div>
   <%= render 'coronavirus_landing_page/components/hub/guidance_section', guidance: details.guidance_section, related: details.related_links %>
+  <%= render partial: 'components/email-signup-section', locals: {
```
andysellick, line 47: Can we have a conversation about the design for this page? I see now why the 'full width' option is present, but even if this design goes ahead I think we could achieve less complexity in the component by doing this a different way.

app/views/components/_email-signup-section.html.erb
andysellick: Having looked into this more I can see it's for the desktop version. I've consulted with design and got a new version attached to the card. Basically I think we can strip all of this out and use the transparent background version of the component inside the existing grey box (the box that already wraps the action links on the hub pages).

app/views/components/_email-signup-section.html.erb
andysellick: Incidentally you can use the shared helper from the gem, in the same way that it works for components in the gem...

`shared_helper = GovukPublishingComponents::Presenters::SharedHelper.new(local_assigns)`

You can then use all of the functionality of the helper, including the `classes` functionality for accepting only `js-` prefixed classes.

danacotoran: @andysellick yep makes sense. Thanks for suggesting the shared helper, it's good to know that functionality is available! However I think we do not need it after all – after my most recent updates there is no need to pass through custom classes anymore.

app/views/components/_email-signup-section.html.erb
danacotoran: I think `h3` is appropriate for most instances of this component, however I've added the option to pass through `heading_level` since I think for the hub pages using `h2` would actually be more accurate (let me know what you think though)

app/views/components/_email-signup-section.html.erb
andysellick: Might you also want it for the `margin_bottom` functionality it provides?

app/views/components/docs/email-signup-section.yml
danacotoran: Updated with the criteria you suggested + heading level criteria

app/views/components/_email-signup-section.html.erb
danacotoran: Changed to passing all tracking data from outside the component.

app/views/components/_email-signup-section.html.erb
andysellick: Sounds good. I should have pointed out sooner, the shared helper in the gem handles heading levels as well, might be worth taking a look.

app/views/components/_email-signup-section.html.erb
danacotoran: I've reworked/simplified how the whole thing works. The design tweak makes it much simpler.

danacotoran:
@andysellick thanks a lot for your review.
I've updated the PR – deleted a whole bunch of logic and moved some things around, which has simplified the code a lot.

I think I've managed to address most of your comments. But please feel free to have another look and let me know if there's anything I missed or if you have any additional suggestions. 🙏  Thanks!!

app/assets/stylesheets/components/_email-signup-section.scss
```diff
@@ -0,0 +1,26 @@
+.app-c-email-signup--with-background-and-border {
```
andysellick, line 1: Slight inconsistency here - the component and files are named `Email signup section` but the classes use `email-signup`. Can I suggest `Signup link` instead?

app/assets/stylesheets/components/_email-signup-section.scss
```diff
@@ -0,0 +1,26 @@
...
+  padding: govuk-spacing(6);
+  background-color: govuk-colour("light-grey", $legacy: "grey-4");
+  border: 1px solid $govuk-border-colour;
+}
+
+.app-c-email-signup--no-intro .govuk-link {
```
andysellick, line 7: This class is called `no-intro` but the component guide refers to it as `Link only`. Can we name them something similar for clarity? Would `link-only` work?

app/views/components/_email-signup-section.html.erb
```diff
@@ -0,0 +1,30 @@
...
+%>
+<% if link_text && link_href %>
+  <div class="app-c-email-signup dont-print <%= classes.join(' ') %>">
+    <div class="app-c-email-signup__inner">
+      <svg  class="app-c-email-signup__icon" aria-hidden="true" xmlns="http://www.w3.org/2000/svg" height="18" width="18" viewBox="0 0 459.334 459.334">
+        <path fill="black" d="M177.216 404.514c-.001.12-.009.239-.009.359 0 30.078 24.383 54.461 54.461 54.461s54.461-24.383 54.461-54.461c0-.12-.008-.239-.009-.359H175.216zM403.549 336.438l-49.015-72.002v-89.83c0-60.581-43.144-111.079-100.381-122.459V24.485C254.152 10.963 243.19 0 229.667 0s-24.485 10.963-24.485 24.485v27.663c-57.237 11.381-100.381 61.879-100.381 122.459v89.83l-49.015 72.002a24.76 24.76 0 0 0 20.468 38.693H383.08a24.761 24.761 0 0 0 20.469-38.694z"/>
+      </svg>
+      <%= content_tag(shared_helper.get_heading_level, intro, class: "govuk-heading-s app-c-email-signup__title") if intro %>
+      <%= link_to( link_text, link_href, {
+        class: "govuk-body govuk-link",
```
andysellick, line 25: Does this need the `govuk-body` class on it?

app/assets/stylesheets/components/_email-signup-section.scss
andysellick, line 7: Would it be possible to use a new class here instead of restyling `govuk-link`?

app/views/components/_email-signup-section.html.erb
```diff
@@ -0,0 +1,30 @@
...
+  data ||= false
+
+  shared_helper = GovukPublishingComponents::Presenters::SharedHelper.new(local_assigns)
+
+  classes << shared_helper.get_margin_bottom
+  classes << "app-c-email-signup--no-intro" unless intro
+  classes << "app-c-email-signup--with-background-and-border" if background
+%>
+<% if link_text && link_href %>
+  <div class="app-c-email-signup dont-print <%= classes.join(' ') %>">
```
andysellick, line 18: There are some classes here and some further up. You could group them together by doing something like:

```
classes = %w(app-c-email-signup dont-print)
classes << shared_helper.get_margin_bottom
classes << "app-c-email-signup--no-intro" unless intro
classes << "app-c-email-signup--with-background-and-border" if background

...

<div class="<%= css_classes.join(' ') %>">
```

app/views/components/docs/email-signup-section.yml
```diff
@@ -0,0 +1,39 @@
...
+      link_text: 'Click right here to sign up!!'
+      link_href: '/this-signs-you-up'
+      background: true
+  link_only:
+    description: If no heading text is passed through, a more compact, link only version is rendered.
+    data:
+      link_text: 'Click right here to sign up!!'
+      link_href: '/this-signs-you-up'
+  with_custom_margin_bottom:
+    description: Override default `margin-bottom` of `0` by passing through a unit from the <a href="https://design-system.service.gov.uk/styles/spacing/">GOVUK responsive spacing scale</a>
```
andysellick, line 27: I've been trying to keep descriptions for this consistent (when components use the shared helper). Suggestion:

`The component accepts a number for margin bottom from 0 to 9 (0px to 60px) using the [GOV.UK Frontend spacing scale](https://design-system.service.gov.uk/styles/spacing/#the-responsive-spacing-scale). It defaults to having no margin bottom.`

app/views/coronavirus_landing_page/components/hub/_guidance_section.html.erb
```diff
@@ -8,7 +8,7 @@
           font_size: 19,
           margin_bottom: 5
         } %>
-        <ul class="covid__list">
+        <ul class="covid__list govuk-!-margin-bottom-8">
```
andysellick, line 11: Is it only this `covid__list` that requires this margin bottom or could it applied to the default for this class? We try not to use the spacing override classes unless it's actually needed to override something.

app/views/components/_email-signup-section.html.erb
```diff
@@ -0,0 +1,30 @@
+<%
+  local_assigns[:heading_level] ||= 3
+  local_assigns[:margin_bottom] ||= 0
+  link_text ||= false
+  link_href ||= false
+  intro ||= false
```
andysellick, line 6: Should this variable be called `heading`?

app/views/components/_email-signup-section.html.erb
andysellick, line 18: Apparently you can use `govuk-!-display-none-print` instead of `dont-print` (which might disappear at some point).

app/views/coronavirus_landing_page/components/hub/_guidance_section.html.erb
danacotoran: It is only this `covid__list`, yeah. The default values for `covid__list` margins are different.
An alternative would be to use a modifier class instead of the spacing override class. 
But the idea of adding new CSS to basically do the same  thing as the existing spacing override class doesn't make much sense to me.

app/views/components/_email-signup-section.html.erb
danacotoran: I think I called it `intro` because (for some reason) that's what it's named in the content item. Renamed to `heading` as it makes more sense.

app/views/components/_email-signup-section.html.erb
danacotoran: ✔️

danacotoran:
@andysellick just updated this with my latest changes – do let me know if I missed anything. Cheers 🙌

app/views/coronavirus_landing_page/components/hub/_guidance_section.html.erb
andysellick: Yeah, this is the party line on this but I actually agree. Don't worry.

app/assets/stylesheets/components/_signup-link.scss
```diff
@@ -0,0 +1,34 @@
...
+  padding: govuk-spacing(6);
+  background-color: govuk-colour("light-grey", $legacy: "grey-4");
+  border: 1px solid $govuk-border-colour;
+}
+
+.app-c-signup-link__link {
+  @include govuk-font(19);
+}
+
+.app-c-signup-link--link-only .app-c-signup-link__link {
```
andysellick, line 11: I'm definitely nitpicking now... but could you reorder these classes a little? I'd put all the `__elements` first and the `--modifiers` after that.

app/assets/stylesheets/components/_signup-link.scss
```diff
@@ -0,0 +1,34 @@
...
+.app-c-signup-link__inner {
+  position: relative;
+}
+
+.app-c-signup-link__icon {
+  position: absolute;
+  top: 2px;
+
+  @include govuk-media-query($until: tablet) {
+    top: 0;
```
andysellick, line 27: Could this be rewritten to use `$from:` instead? I think we should be doing this mobile first if possible.

danacotoran:
@andysellick I'm sorry, I went a bit overboard squashing commits. Forgot that makes it harder for you to see what my latest changes are. 
Do give it another look and let me know if you spot anything else? thanks!

PR #1764, opened 2020-05-25
Refactor some coronavirus landing page tests

hannako:
- Remove reference to old `livestream_enabled` key
- Set all optional fields to true in fixture, and turn
off in tests (except risk element tests as that still seems to be somewhat in flight).
[card](https://trello.com/c/02aizrky/287-refactor-and-add-tests-to-collections)

### DIFF
--- a/test/support/coronavirus_landing_page_steps.rb
+++ b/test/support/coronavirus_landing_page_steps.rb
@@ -19,20 +19,16 @@ module CoronavirusLandingPageSteps
     stub_content_store_has_item(CORONAVIRUS_PATH, coronavirus_content_item)
   end
 
-  def given_there_is_a_content_item_with_live_stream_enabled
-    stub_content_store_has_item(CORONAVIRUS_PATH, coronavirus_content_item_with_live_stream_enabled)
+  def given_there_is_a_content_item_with_live_stream_time
+    stub_content_store_has_item(CORONAVIRUS_PATH, coronavirus_content_item_with_live_stream_time)
   end
 
-  def given_there_is_a_content_item_with_live_stream_enabled_and_date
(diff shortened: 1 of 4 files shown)


test/support/coronavirus_landing_page_steps.rb
```diff
@@ -118,22 +114,34 @@ def then_i_can_see_the_live_stream_section_with_streamed_date
   end
 
   def then_i_can_see_the_live_stream_section_with_date_and_time
-    assert page.has_text?("19 April")
-    assert page.has_text?("5:00pm")
+    assert page.has_text?("19 April at 5:00pm")
```
reggieb, line 117: Where does the `5:00pm` come from. I can't see it being set anywhere. Is this an artefact of the test environment? Would it be better to match against a pattern here where what is being set is the content that is set within the test configuration. Would this be better?

```ruby
assert page.has_content(/19 April at \d+:\d+[ap]m/)
```

hannako: Thanks Rob. 

5pm is set in the method `given_there_is_a_content_item_with_live_stream_time` :  
```
def given_there_is_a_content_item_with_live_stream_time
  stub_content_store_has_item(CORONAVIRUS_PATH, coronavirus_content_item_with_live_stream_time)
end

def coronavirus_content_item_with_live_stream_time
  content_item = coronavirus_content_item
  content_item["details"]["live_stream"]["time"] = "5:00pm"
  content_item
end
```

Do you still think it's better to match against a pattern?

reggieb: No - now I can see where it's set I think what you have is fine.

PR #1774, opened 2020-05-28
Initialise Toggle Attribute module after Accordions have been initialised

DilwoarH:
## What
It would appear the tracking on the accordion links broke yesterday (27 May) - with accordionOpen, now reading as accordionClosed.

We should investigate what caused the problem, and either fix it if we can or prod whoever needs to fix it.

https://trello.com/c/qThuLppG/315-investigate-why-the-accordion-track-broke-and-fix-it-if-we-can


DilwoarH:
We will have an incident review to understand why this happened.

PR #1776, opened 2020-05-29
Hubs header spacing adjustment

danacotoran:
## What
We should reduce the blank space underneath the bullets on hub pages to match the space underneath the action link on the landing page.

## Why
Visual consistency and bringing the content further up the page

## Trello 
https://trello.com/c/cBaVFUui

### DIFF
--- a/app/assets/stylesheets/views/_covid.scss
+++ b/app/assets/stylesheets/views/_covid.scss
@@ -48,16 +48,20 @@
   }
 }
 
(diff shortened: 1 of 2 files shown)


danacotoran:
@DilwoarH Yes you are right, I checked with Kate & Mia and we decided to match that space to the padding above the "What you can do now" bit, as it would hopefully look more balanced

PR #1780, opened 2020-06-02
Tweak transition tracking

vanitabarrett:
Trello: https://trello.com/c/9CFX3mvB/306-change-tracking-attributes-on-the-transition-landing-page

## What

- Add missing `data-module='track-click'` attribute which was preventing our tracking event firing on the CTA link
- Change the event label for links in the 'News' and 'What You Can Do Now' sections so they reflect what section they're in.

### DIFF
--- a/app/presenters/transition_landing_page_presenter.rb
+++ b/app/presenters/transition_landing_page_presenter.rb
@@ -66,7 +66,7 @@ private
       data_attributes = {
         track_category: "transition-landing-page",
         track_action: link_item[:link][:path],
-        track_label: link_item[:link][:text],
+        track_label: "News",
       }
 
       link_item[:link][:data_attributes] = data_attributes
--- a/config/locales/cy/transition_landing_page.yml
+++ b/config/locales/cy/transition_landing_page.yml
@@ -10,7 +10,7 @@ cy:
       <p>Bydd y rheolau cyfredol ar fasnach, teithio a busnes ar gyfer y DU a’r UE yn parhau i fod yn berthnasol yn ystod y cyfnod pontio.</p>
       <p>Bydd rheolau newydd yn dod i rym ar 1 Ionawr 2021.</p>
       <p>Dylech baratoi nawr a thanysgrifio i ddiweddariadau ebost am unrhyw drefniadau ychwanegol.</p>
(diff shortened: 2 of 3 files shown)


app/presenters/transition_landing_page_presenter.rb
```diff
@@ -66,7 +66,7 @@ def fetch_comms
       data_attributes = {
         track_category: "transition-landing-page",
         track_action: link_item[:link][:path],
-        track_label: link_item[:link][:text],
+        track_label: "News",
```
benthorner, line 69: This seems like an unusual use for the label; do we know we we're not just changing the category?

config/locales/cy/transition_landing_page.yml
```diff
@@ -49,21 +49,21 @@ cy:
         list_block: |
           Ar ôl 1 Ionawr 2021, bydd angen i chi wneud datganiadau tollau i symud nwyddau i mewn ac allan o’r UE. Dylech chi:
           <ul>
-            <li><a class="govuk-!-font-weight-bold" href="/eori" data-track-category="transition-landing-page" data-track-action="/eori" data-track-label="gael rhif EORI">gael rhif EORI</a> os nad oes gennych un yn barod</li>
-            <li>penderfynu sut yr hoffech wneud datganiadau tollau ac os oes angen i chi gael <a class="govuk-!-font-weight-bold" href="/guidance/appoint-someone-to-deal-with-customs-on-your-behalf" data-track-category="transition-landing-page" data-track-action="/guidance/appoint-someone-to-deal-with-customs-on-your-behalf" data-track-label="rhywun i ddelio â’r tollau ar eich rhan">rhywun i ddelio â’r tollau ar eich rhan</a>.</li>
+            <li><a class="govuk-!-font-weight-bold" href="/eori" data-track-category="transition-landing-page" data-track-action="/eori" data-track-label="What you can do now">gael rhif EORI</a> os nad oes gennych un yn barod</li>
```
benthorner, line 52: Same comment here: we seem to be using the category to group every action on the page together, but that should already be possible using the page path. If we use the category attribute more conventionally, we can keep the existing labels, which provides more data for analytics e.g. if two links have the same path.

vanitabarrett:
@benthorner Spoken to Ash, he's keen to keep the category as it is so it's kept consistent with historical data they've gathered previously.

PR #1792, opened 2020-06-09
Use individual component model

andysellick:
This PR builds on the work in https://github.com/alphagov/collections/pull/1530 following changes to the component gem in https://github.com/alphagov/govuk_publishing_components/pull/1541.

---

Update collections to import the sass and Javascript for only the components it is using, instead of all components, as [documented here](https://github.com/alphagov/govuk_publishing_components/blob/master/docs/install-and-use.md#import-sass-for-individual-components) and [here](https://github.com/alphagov/govuk_publishing_components/blob/master/docs/install-and-use.md#import-javascript-for-individual-components).

## Filesize comparison

Screenshot 2020-06-09 at 09 36 37

CSS before: **739 KB** 

Screenshot 2020-06-09 at 09 28 50

CSS after: **775 KB** 

(this has increased because some components were missing. Note also that we're still not including the action link component CSS, because it's already included in static)

Screenshot 2020-06-09 at 09 27 36

JS before: **1.1 MB** 

Screenshot 2020-06-09 at 09 28 21

JS after: **873 KB** 

Trello card: https://trello.com/c/GWttlBxs/228-upgrade-applications-to-use-individual-component-sass


app/assets/javascripts/application.js
```diff
@@ -1,3 +1,11 @@
+//= require govuk_publishing_components/lib
```
alex-ju, line 1: Can we have these after L9? (`//= require govuk-frontend/govuk/vendor/polyfills/Function/prototype/bind`) to preserve the dependency tree.

andysellick: Updated. Apols, I was following the advice in the component guide `Add this to your application's main js file, before your other local imports.` - does this need clarifying?

alex-ju: Rightly so. I didn't anticipate having applications that would require polyfills from `govuk-frontend` – will update the guidance accordingly.

PR #1808, opened 2020-06-24
Add canonical URL to landing pages

sihugh:
We get this by default on pages that use the machine_readable component, but we've got a custom set up on these pages that doesn't lend itself to that.

We need to add this because these pages are often linked to with campaign or marketing query strings. This causes Google et al to believe that they might behave differently at times. (They don't)

https://trello.com/c/nF93uexi/388-put-canonical-url-on-landing-pages

### DIFF
--- a/app/views/coronavirus_landing_page/components/shared/_meta_tags.html.erb
+++ b/app/views/coronavirus_landing_page/components/shared/_meta_tags.html.erb
@@ -5,9 +5,10 @@
 <% content_for :is_full_width_header, true %>
 <% content_for :title, @content_item["title"] %>
 <% content_for :meta_tags do %>
+  <% page_url = File.join(request.base_url, request.path) %>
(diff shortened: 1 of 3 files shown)


app/views/coronavirus_landing_page/components/shared/_meta_tags.html.erb
```diff
@@ -5,9 +5,10 @@
 <% content_for :is_full_width_header, true %>
 <% content_for :title, @content_item["title"] %>
 <% content_for :meta_tags do %>
+  <% page_url = request.base_url + request.path %>
```
reggieb, line 8: I'd suggest using `File.join` to handle the slash placement.

sihugh: 👍

PR #1809, opened 2020-06-25
Hide live stream section and remove /ask

sihugh:
We may want to turn the livestream section back on again, so we'll use a toggle in the content item.

The /ask page and the FAQ page have already been taken down, and the links disabled, so it's safe to remove these.

https://trello.com/c/r5x97Fni/385-remove-the-press-conferences-and-speeches-section-of-the-landing-page

### DIFF
--- a/app/views/coronavirus_landing_page/components/landing_page/_live_stream_section.html.erb
+++ b/app/views/coronavirus_landing_page/components/landing_page/_live_stream_section.html.erb
@@ -1,3 +1,4 @@
+<% if ActiveModel::Type::Boolean.new.cast(live_stream["show_live_stream"]) %>
 <%
   list_css_classes = %w(govuk-list)
   list_css_classes << "covid__spaced-list" if live_stream["spaced_links"]
@@ -41,3 +42,4 @@
     <% end %>
   </ul>
 </section>
+<% end %>
--- a/app/views/coronavirus_landing_page/components/landing_page/_live_stream_section.html.erb
+++ b/app/views/coronavirus_landing_page/components/landing_page/_live_stream_section.html.erb
@@ -26,20 +26,6 @@
         <%= live_stream["previous_videos"]["previous_videos_text"] %>
       </a>
     </li>
-    <% if live_stream["ask_a_question_visible"] == true %>
-      <li class="covid__topic-list-item">
-        <a href="<%= live_stream["ask_a_question_link"] %>" class="covid__topic-list-link govuk-link">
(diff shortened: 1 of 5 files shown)


app/views/coronavirus_landing_page/components/landing_page/_live_stream_section.html.erb
```diff
@@ -1,3 +1,4 @@
+<% if live_stream["show_live_stream"] %>
```
reggieb, line 1: Would it be worth doing:
```ruby
<% if ActiveModel::Type::Boolean.new.cast(live_stream["show_live_stream"]) %>
```
Otherwise could setting `"show_live_stream": "false"` in the JSON have unexpected consequences?

test/integration/coronavirus_landing_page_test.rb
```diff
@@ -26,26 +26,18 @@ class CoronavirusLandingPageTest < ActionDispatch::IntegrationTest
       given_there_is_a_content_item
       when_i_visit_the_coronavirus_landing_page
       then_i_can_see_the_live_stream_section_with_streamed_date
-      then_i_can_see_the_ask_a_question_section
-      then_i_can_see_the_popular_questions_link
     end
 
-    it "optionally shows the time of a livestream" do
-      given_there_is_a_content_item_with_live_stream_time
```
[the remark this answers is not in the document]
sihugh: they're still in use - unfortunately the diff isn't helpful as it implies that this section has been repurposed, when it's actually shuffled down a few lines :-(

app/views/coronavirus_landing_page/components/landing_page/_live_stream_section.html.erb
sihugh: Nice catch, thanks

sihugh: I went with `live_stream["show_live_stream"] == true` to match the approach used elsewhere

reggieb: I think `"show_live_stream": false` will work as expected as `JSON.parse` will recognise it as the boolean. So the "false" option is a bit of an edge case. I just think it's an easy change to make, and makes the code a little more robust.

sihugh: OK, you've got me :-)

PR #1817, opened 2020-07-08
Amend Transition page metadescription

sihugh:
I'm going to check why this isn't read from the content item. Will make a new story to sort this (and the search index entry) out if they can be.

https://trello.com/c/zYugZiMJ/288-update-meta-description-for-transition-landing-page-for-site-search-purposes


sihugh:
Also see https://github.com/alphagov/search-api/pull/2149

PR #1819, opened 2020-07-13
Import changes from private repo

hannako:
[private repo](https://github.com/alphagov/collections-private) has [only one PR](https://github.com/alphagov/collections-private/pulls?q=is%3Apr+is%3Aclosed) difference to master:

[Add new header, video sections and new campaign styling for the Brexit 2020 campaign](https://github.com/alphagov/collections-private/pull/1/files).

I have followed the steps here: https://docs.publishing.service.gov.uk/manual/make-github-repo-private.html


app/assets/stylesheets/views/_transition-landing-page.scss
```diff
@@ -19,8 +22,123 @@ $red: #E61E32;
...
+.landing-page__header-arrow {
+  width: 90px;
+
+  @include govuk-media-query($from: $header-breakpoint) {
+    width: 165px;
+  }
+}
+
+.landing-page__take-action-button {
+  .govuk-button {
```
andysellick, line 97: I'm late to this, but I notice that this styles a component outside of the gem, which breaks our component isolation rule. Is there a plan to move this style to an option on the component itself?

DilwoarH: I don't think this is a style we want to bring into the component. This was done as an override for this specific button (to match the Brexit buttons).

This is experimental and we are looking at whether we want to go back to the original styling.

andysellick: Thanks. The style should either be removed or moved into the gem - we can't have applications styling components like this, it makes developing components harder and introduces more risk when doing so (for example if the default colour of button text was changed, it could introduce an colour contrast accessibility issue).

PR #1820, opened 2020-07-13
Fix IE issues and accessibility problems

DilwoarH:
## What

Fixes gradient issue on IE

## Before

Screenshot 2020-07-13 at 12 51 07

## After

Screenshot 2020-07-13 at 12 50 31

## IE9

Screenshot 2020-07-13 at 16 00 06

### DIFF
--- a/app/assets/stylesheets/views/_transition-landing-page.scss
+++ b/app/assets/stylesheets/views/_transition-landing-page.scss
@@ -36,6 +36,18 @@ $header-breakpoint: 960px;
     background-position: top;
     min-height: 500px;
   }
+
+  // HIDE BACKGROUND IMAGE ON IE
+
+  // IE 6, 7, 8
+  @media \0screen\,screen\9 {
+    background-image: none;
+  }
+
+  // IE 9
+  @media screen and (min-width:0\0) and (min-resolution: .001dpcm) { 
+    background-image: none;
+  }
 }
 
 .landing-page__header-text-block {
@@ -47,7 +59,6 @@ $header-breakpoint: 960px;
     background: -moz-linear-gradient(90deg, $dark-blue 66.66%, $transparent 66.66%);
     background: -webkit-linear-gradient(90deg, $dark-blue 66.66%, $transparent 66.66%);
     background: linear-gradient(90deg, $dark-blue 66.66%, $transparent 66.66%);
-    filter: progid:DXImageTransform.Microsoft.gradient(startColorstr="#{$dark-blue}", endColorstr="#{$transparent}", GradientType=1);
   }
 }
 
--- a/app/assets/stylesheets/views/_transition-landing-page.scss
+++ b/app/assets/stylesheets/views/_transition-landing-page.scss
@@ -192,6 +192,7 @@ $header-breakpoint: 960px;
   @include govuk-responsive-padding(8, "top");
   @include govuk-responsive-padding(8, "bottom");
   background-color: $dark-blue;
+  margin-top: 0;
 
   h2,
   p {
--- a/app/assets/stylesheets/views/_transition-landing-page.scss
+++ b/app/assets/stylesheets/views/_transition-landing-page.scss
@@ -87,9 +87,11 @@ $header-breakpoint: 960px;
 
 .landing-page__header-arrow {
   width: 90px;
+  height: 81px;
 
   @include govuk-media-query($from: $header-breakpoint) {
     width: 165px;
+    height: 149px;
   }
 }
 
--- a/app/assets/stylesheets/views/_transition-landing-page.scss
+++ b/app/assets/stylesheets/views/_transition-landing-page.scss
@@ -2,7 +2,7 @@
 $red: #FF003B;
 $dark-blue: #0055D2;
 $dark-blue-shadow: #020659;
-$transparent: #ffffff00;
+$transparent: rgba(0,0,0,0);
 $header-breakpoint: 960px;
 
 .large-paragraph {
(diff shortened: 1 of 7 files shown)


app/views/transition_landing_page/_page_header.html.erb
```diff
@@ -40,7 +40,7 @@
       <div class="govuk-grid-row">
         <div class="govuk-grid-column-full">
           <div class="landing-page__header-arrow-container">
-            <%= image_tag 'white-arrow.svg', class: "landing-page__header-arrow" %>
+            <%= image_tag 'white-arrow.svg', class: "landing-page__header-arrow", alt: "" %>
```
huwd, line 43: nice i didn't think ot check those for alt tags before
@injms recently [mentioned an aria tag](https://gds.slack.com/archives/CARGH33JS/p1594629172069500?thread_ts=1594629065.068800&cid=CARGH33JS) that stops these being read out by screen readers, do we need those too?

app/views/transition_landing_page/_take_action.html.erb
```diff
@@ -25,19 +25,19 @@
       <div class="govuk-grid-column-one-third">
         <ul class="govuk-list landing-page__take-action-traffic-lights">
           <li>
-            <%= image_tag 'take-action-red.svg', class: "landing-page__take-action-traffic-lists-icon" %>
+            <%= image_tag 'take-action-red.svg', class: "landing-page__take-action-traffic-lists-icon", alt: "" %>
```
huwd, line 28: Ditto, aria?

app/views/transition_landing_page/_take_action.html.erb
```diff
@@ -25,19 +25,19 @@
...
           <li>
-            <%= image_tag 'take-action-amber.svg', class: "landing-page__take-action-traffic-lists-icon" %>
+            <%= image_tag 'take-action-amber.svg', class: "landing-page__take-action-traffic-lists-icon", alt: ""  %>
             <span class="landing-page__take-action-traffic-lists-text">
               <%= t("transition_landing_page.take_action_list.amber") %>
             </span>
           </li>
           <li>
-            <%= image_tag 'take-action-green.svg', class: "landing-page__take-action-traffic-lists-icon" %>
+            <%= image_tag 'take-action-green.svg', class: "landing-page__take-action-traffic-lists-icon", alt: ""  %>
```
huwd, line 40: Ditto, aria?

app/views/transition_landing_page/_take_action.html.erb
```diff
@@ -25,19 +25,19 @@
...
           <li>
-            <%= image_tag 'take-action-red.svg', class: "landing-page__take-action-traffic-lists-icon" %>
+            <%= image_tag 'take-action-red.svg', class: "landing-page__take-action-traffic-lists-icon", alt: "" %>
             <span class="landing-page__take-action-traffic-lists-text">
               <%= t("transition_landing_page.take_action_list.red") %>
             </span>
           </li>
           <li>
-            <%= image_tag 'take-action-amber.svg', class: "landing-page__take-action-traffic-lists-icon" %>
+            <%= image_tag 'take-action-amber.svg', class: "landing-page__take-action-traffic-lists-icon", alt: ""  %>
```
huwd, line 34: Ditto, aria?

app/views/transition_landing_page/_take_action.html.erb
hannako, line 28: Why `alt: ""` ?

huwd: I saw @DilwoarH [asked about this in frontenders last week](https://gds.slack.com/archives/CARGH33JS/p1594629065068800)

I think @injms said it stops screen readers reading out the source for the image. But as it has no semantic information there's nothing to put in there that would be otherwise useful for a screen reader to read out.

app/views/transition_landing_page/_page_header.html.erb
injms: I don't think an `aria-hidden` attribute is needed - the blank `alt` text should mean that this isn't read aloud by screenreader.

app/views/transition_landing_page/_take_action.html.erb
injms: Without the `alt` attribute some screenreader read out the image `src` instead - a blank `alt` stops this.

app/assets/stylesheets/views/_transition-landing-page.scss
```diff
@@ -42,13 +42,12 @@ $header-breakpoint: 960px;
...
-      $dark-blue 66.66%, 
-      $transparent 66.66%, 
-      $transparent 100%
-    );
+
+    background: $dark-blue;
+    background: -moz-linear-gradient(90deg, $dark-blue 66.66%, $transparent 66.66%);
+    background: -webkit-linear-gradient(90deg, $dark-blue 66.66%, $transparent 66.66%);
+    background: linear-gradient(90deg, $dark-blue 66.66%, $transparent 66.66%);
+    filter: progid:DXImageTransform.Microsoft.gradient(startColorstr="#{$dark-blue}", endColorstr="#{$transparent}", GradientType=1);
```
injms, line 50: I don't think that this filter creates a gradient in the same way that the CSS gradient does - it just gradually goes from blue to transparent. It could be worth removing, since the solid colour background would create a decent enough fallback.

DilwoarH: I might just remove the image for IE9 and below

DilwoarH: Hey - remove this and removed the image completely on IE9 and below

app/assets/stylesheets/views/_transition-landing-page.scss
```diff
@@ -36,19 +36,29 @@ $header-breakpoint: 960px;
     background-position: top;
     min-height: 500px;
   }
+
+  // HIDE BACKGROUND IMAGE ON IE
+
+  // IE 6, 7, 8
+  @media \0screen\,screen\9 {
```
injms, line 43: This could be replaced with the condition comments class if you like:

```suggestion
  .lte-ie8 {
```

DilwoarH: Do we have one for ie9?

DilwoarH:
I'm going to merge as I think further work will be needed to use the lte-ie8 class

app/assets/stylesheets/views/_transition-landing-page.scss
injms: Unfortunately we don't use one for IE9, but it is possible - I think it'd require [a change in static](https://github.com/alphagov/static/blob/master/app/views/layouts/govuk_template.html.erb#L3).

PR #1822, opened 2020-07-13
Add NO_THE exemption for Civil service HR

huwd:
[As requested here](https://gds.slack.com/archives/CADKZN519/p1594650006409800)

Before:
image

After:
image


thomasleese:
I've tested this on integration and it seems to work: https://www-origin.integration.publishing.service.gov.uk/government/organisations/civil-service-hr

@huwd Anything else to be aware of before I merge this in?

PR #1825, opened 2020-07-14
Youtube video update for brexit 2020 campaign

DilwoarH:
- Youtube video recommendations now only showing channel recommendation
- Update to video for Brexit campaign
- Fixes issue with the background showing behind youtube video


DilwoarH:
> As per slack convo, this doesn't seem to work for me when I visit https://www.youtube.com/watch?v=4aO-YTQ94jA&rel=0 directly.

I don't we can do anything about it when you visit directly but we can control what shows when embedded

PR #1827, opened 2020-07-14
Add border to covid 19 find support block

DilwoarH:
# What
We need to look at updating the content, and possibly the design, of the 'Find support if you're struggling' box. The content needs to more accurately reflect the service. 

Update the box with this final design (in cover) and attached 'Final support design'

# Why
Research has shown that some users are either not noticing, or not understanding the content in the 'Find support if you're struggling' box on the landing page.

# Demo
https://govuk-collec-add-border-3uwnln.herokuapp.com/coronavirus

# Before
Screenshot 2020-07-14 at 15 40 00

# After
Screenshot 2020-07-16 at 15 30 11

# IE10 and below
Screenshot 2020-07-16 at 15 29 48

https://trello.com/c/UzmUWkvX/674-update-the-find-support-if-youre-struggling-box

### DIFF
--- a/app/assets/stylesheets/views/_covid.scss
+++ b/app/assets/stylesheets/views/_covid.scss
@@ -123,6 +123,8 @@
   @include govuk-responsive-margin(9, "bottom");
   background-color: $covid-grey;
   padding: govuk-spacing(6) govuk-spacing(5);
+  border-left: 10px solid;
+  border-image: repeating-linear-gradient(25deg, transparent 0, $covid-yellow 1px, $covid-yellow 10px, transparent 11px, transparent 20px) 25;
 }
 
 .covid__list {
--- a/app/assets/stylesheets/views/_covid.scss
+++ b/app/assets/stylesheets/views/_covid.scss
@@ -123,7 +123,7 @@
   @include govuk-responsive-margin(9, "bottom");
   background-color: $covid-grey;
   padding: govuk-spacing(6) govuk-spacing(5);
-  border-left: 10px solid;
+  border-left: 10px solid $covid-yellow;
   border-image: repeating-linear-gradient(25deg, transparent 0, $covid-yellow 1px, $covid-yellow 10px, transparent 11px, transparent 20px) 25;
 }
 


app/assets/stylesheets/views/_covid.scss
```diff
@@ -123,6 +123,9 @@
   @include govuk-responsive-margin(9, "bottom");
   background-color: $covid-grey;
   padding: govuk-spacing(6) govuk-spacing(5);
+  border-left: 10px solid;
+  border-width: 0 0 0 10px;
```
andysellick, line 127: Toggling this line in the browser inspector doesn't seem to change anything - is this for older browsers?

DilwoarH: your right - sorry I forgot to remove, the `border-left` has overridden it

DilwoarH:
> Is it worth putting in some kind of fallback for older browsers? This is IE9, for example...
> 
> Screenshot 2020-07-14 at 16 43 44
> 
> Looks like the border isn't being styled properly but it's still there. I'd suggest providing a colour for the border (which would then be overridden by border-image). Maybe either something bright to make it stand out, or just the colour of the box itself? Might be a question to run by a designer.

Thanks for spotting that - I'll speak with a designer and get back to you

andysellick:
> Thanks for spotting that - I'll speak with a designer and get back to you

Another thought relating to that - if the decision is to not colour the border in this situation, it might look a little out of balance (as it's adding 10px to the left of the box) - might be worth adding an extra 10px of padding to the right to centre the text.

DilwoarH:
> > Thanks for spotting that - I'll speak with a designer and get back to you
> 
> Another thought relating to that - if the decision is to not colour the border in this situation, it might look a little out of balance (as it's adding 10px to the left of the box) - might be worth adding an extra 10px of padding to the right to centre the text.

Hey @andysellick - Had a chat with Kate and she said the yellow solid border is fine on IE10 and below. I have updated the screenshot above for your reference.

PR #1831, opened 2020-07-16
add support for links in NHS box

DilwoarH:
# What
Allows links on NHS box

# Why
We want to start adding links instead of plain text to the list

# Visual diffs
Screenshot 2020-07-16 at 16 47 12

https://trello.com/c/eeSbyTSU/683-rework-nhs-box

### DIFF
--- a/app/views/components/_header-notice.html.erb
+++ b/app/views/components/_header-notice.html.erb
@@ -25,7 +25,13 @@
     <% if list.any? %>
       <ul class="app-c-header-notice__list govuk-list govuk-list--bullet">
         <% list.each do |item| %>
-          <li><%= item %></li>
+          <li>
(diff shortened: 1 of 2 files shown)


app/views/components/_header-notice.html.erb
```diff
@@ -25,7 +25,13 @@
     <% if list.any? %>
       <ul class="app-c-header-notice__list govuk-list govuk-list--bullet">
```
injms, line 26: The gem has a (recently added) component for lists now: https://components.publishing.service.gov.uk/component-guide/list

DilwoarH: Nice - thanks @injms.

When we get a chance, we'll try move our lists over to use the gem

PR #1851, opened 2020-08-04
Hide documents for joining organisations

edwardkerry:
Organisations that are just joining GOV.UK (created as "Coming soon" in [Whitehall](https://github.com/alphagov/whitehall/blob/52aff8f61a29b3999054b5b5c94875a5534eaf9a/app/views/admin/organisations/_form.html.erb#L42) ) currently display all associated documents on their organisation .

This change will prevent these documents from rendering on the organisation page until the organisation is live.

[Trello](https://trello.com/c/7FQYL0w8/2073-investigate-making-content-lists-not-display-on-organisation-pages-with-status-coming-soon)

### DIFF
--- a/app/views/organisations/separate_website.html.erb
+++ b/app/views/organisations/separate_website.html.erb
@@ -19,4 +19,6 @@
   </div>
 </div>
 
-<%= render partial: 'latest_documents_by_supergroup' %>
+<% unless @organisation.is_joining? %>
+  <%= render partial: 'latest_documents_by_supergroup' %>
+<% end %>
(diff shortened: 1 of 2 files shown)


cbaines:
It would be good to say something in the commit message about why this is changing, even if it's a vague "don't confuse users by showing tagging to coming soon organisations".

edwardkerry:
> It would be good to say something in the commit message about why this is changing, even if it's a vague "don't confuse users by showing tagging to coming soon organisations".

I've updated the commit message 👍

PR #1857, opened 2020-08-07
Add lang to organisation contacts section

danacotoran:
## What
Add `lang` attribute to FOI contacts section on organisation pages

## Why
Organisation pages can be translated (often into Welsh), but there are
parts of the page that are never translated and will fall back to the
English version.

The FOI contacts section is one of those areas that will sometimes fall
back to the default English version. This results in a WCAG fail as it
is not clearly indicated that part of the page is in a different
language than the rest of the page, which can be a problem for people
using assistive tech.

Adding lang to the areas that might be falling back to the English
language version will help resolve the problem.

-----

**BEFORE =  bad**: page is in Welsh, FOI contact is in English and not marked with `lang="en"`

Screenshot 2020-08-07 at 10 27 09

**AFTER =  less bad**: page is in Welsh, FOI contact is in English, but marked as `lang="en"`

Screenshot 2020-08-07 at 10 27 50

https://trello.com/c/1sD18QyB

### DIFF
--- a/app/presenters/organisations/contacts_presenter.rb
+++ b/app/presenters/organisations/contacts_presenter.rb
@@ -32,6 +32,7 @@ module Organisations
     def format_contacts(contacts, foi: false)
       contacts.map do |contact|
         {
+          locale: contact["locale"] || "",
           title: foi_title(contact["details"]["title"], foi: foi),
           post_addresses: contact["details"]["post_addresses"]&.map do |post|
             contact_address(post)
--- a/app/presenters/organisations/contacts_presenter.rb
+++ b/app/presenters/organisations/contacts_presenter.rb
@@ -32,7 +32,7 @@ module Organisations
     def format_contacts(contacts, foi: false)
       contacts.map do |contact|
         {
-          locale: contact["locale"] || "",
+          locale: contact["locale"].presence,
           title: foi_title(contact["details"]["title"], foi: foi),
           post_addresses: contact["details"]["post_addresses"]&.map do |post|
             contact_address(post)
--- a/app/views/organisations/_contacts.html.erb
+++ b/app/views/organisations/_contacts.html.erb
@@ -1,6 +1,6 @@
 <% if contacts.present? %>
 <% contacts.each do |contact| %>
-  <div class="organisation__contact-section <%= "organisation__contact-section--border-top" if local_assigns[:border] %>">
+  <div class="organisation__contact-section <%= "organisation__contact-section--border-top" if local_assigns[:border] %>" <%= "lang=#{contact[:locale]}" if contact[:locale] %>>
     <%= render "govuk_publishing_components/components/heading", {
       text: contact[:title],
       padding: true,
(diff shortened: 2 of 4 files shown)


app/presenters/organisations/contacts_presenter.rb
```diff
@@ -32,6 +32,7 @@ def foi_previous_releases_link
     def format_contacts(contacts, foi: false)
       contacts.map do |contact|
         {
+          locale: contact["locale"] || "",
```
injms, line 35: We could use a neat Rails helper here - `presence`. This returns the content of `contact["locale"]` unless it's blank / not there when it returns `nil`.

```suggestion
          locale: contact["locale"].presence,
```

app/views/organisations/_contacts.html.erb
```diff
@@ -1,6 +1,6 @@
 <% if contacts.present? %>
 <% contacts.each do |contact| %>
-  <div class="organisation__contact-section <%= "organisation__contact-section--border-top" if local_assigns[:border] %>">
+  <div class="organisation__contact-section <%= "organisation__contact-section--border-top" if local_assigns[:border] %>" <%= "lang=#{contact[:locale]}" if contact[:locale] %>>
```
injms, line 3: There's a rogue non-English string within this element on line 24. We should either add the lang attribute for the inner element:
```erb
<p lang="<%= :locale %>"><%= t('organisations.contact.email') %></p>
```

or force the `t()` method to retrieve the English version:

```erb
<p><%= t("organisations.contact.email", locale: :en) %></p>
```

(GitHub won't let me comment on line 24, so I'm putting it here.)

danacotoran: Thanks for this comment, totally missed that!
 
The problem with `<p lang="<%= :locale %>"><%= t('organisations.contact.email') %></p>` is that it is inaccurate if `t('organisations.contact.email')` does not find a translation for that locale and falls back to the Eng version.

Similarly, `<p><%= t("organisations.contact.email", locale: :en) %></p>` would be no good when the surrounding block is `cy` or some other language.

So in the end I went with `<p lang="en"><%= t("organisations.contact.email", locale: :en) %></p>`

maxgds: Wouldn't you get the same result and less processing with something like this?
<p lang="en"><%= organisations.contact.email %></p>

danacotoran: @maxgds I think I know what you mean, but trying `<%= organisations.contact.email %>` throws an error. 
I can't seem to find a precedent where we're pulling data straight from a specific locale file as opposed to using that `t()` function

maxgds: Fair enough - ignore me then.

app/views/organisations/_contacts.html.erb
```diff
@@ -21,7 +21,7 @@
         <div class="<%= contact[:post_addresses].any? ? "govuk-grid-column-one-half" : "govuk-grid-column-two-thirds" %>">
           <% if contact[:email_addresses].any? %>
             <div class="organisation__margin-bottom">
-              <p><%= t('organisations.contact.email') %></p>
+              <p lang="en"><%= t('organisations.contact.email', locale: :en) %></p>
```
injms, line 24: Picky and not a blocker - could we change the `lang` attribute to only be displayed if the outer element is not in English?

PR #1867, opened 2020-08-13
Fix translations for documents on organisations pages

owenatgov:
## What
Adds several features to the documents presenter:

- Explicitly passes `locale: "en"` with the document link title to the documents list component
- Passes a `locale` object with the document metadata to the documents list component, including fallback locales for both the document type and document publish time
- Checks for an existing translation of the document type

Related to this PR for govuk_publishing_components <https://github.com/alphagov/govuk_publishing_components/pull/1643>

## Why
This is part of the wider work to fix i18n issues across govuk, which itself is part of the work to make govuk WCAG 2.1 compliant. As the scope of this work is limited and we don't have time to translate everything, passing the locale code as a `lang` html attribute will be enough for these language discrepancies to pass WCAG 2.1. Some more explicit reasoning behind my choices:

- Some research suggests that the govuk search API, where the documents are being pulled from, only return English results currently. I've therefore made the choice to confidently pass an "en" locale code.
- Translations for the vast majority of document types already existed in collections. The code however was being cleaned and passed directly to the frontend instead of via rails' i18n translate method. I've reimplemented this and added both a fallback to pass the locale attribute and if the translation for a particular document type doesn't exist anywhere in collections.

### Before
Screenshot 2020-08-13 at 16 49 54

### After
Screenshot 2020-08-13 at 16 54 52

**Card:** https://trello.com/c/HT0lvX7v/77-fix-part-language-mixed-languages-on-document-lists

### DIFF
--- a/app/presenters/organisations/document_presenter.rb
+++ b/app/presenters/organisations/document_presenter.rb
@@ -4,6 +4,7 @@ module Organisations
   #
   # See https://govuk-publishing-components.herokuapp.com/component-guide/document_list
   class DocumentPresenter
+    include ApplicationHelper
     attr_accessor :include_metadata
 
     def initialize(rummager_result, include_metadata: true)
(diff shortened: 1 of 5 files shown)


injms (approved):

injms:
I'm merging this in, as @owenatgov doesn't have the ~power~ permissions yet.

PR #1871, opened 2020-08-19
Revert "Display localised terms for news document type on org pages"

hannako:
Reverts alphagov/collections#1866

This change is causing a number of errors in sentry at the moment: https://sentry.io/organizations/govuk/issues/1846708045/?project=202213&referrer=slack

Running the master branch locally the same errors can be replicated, and fixed by undoing the change. So reverting for now to give time to find a solution.

PR #1872, opened 2020-08-19
Display localisations for news document type

danacotoran:
Ensure that image card component news stories are displaying the correct translation for document type. E.g displaying 'Stori newyddion' instead of 'News story'.

Don't call the I18n translation function if `document_type` is `null` as this leads to an error.

The original change [here](https://github.com/alphagov/collections/pull/1866) was reverted [here](https://github.com/alphagov/collections/pull/1871) due to causing [errors](https://sentry.io/organizations/govuk/issues/1846708045/events/ef64bf5ed5284790889c1bbf9bdd278b/#exception)  on this page https://www.gov.uk/government/organisations/hm-treasury

A brief investigation into the content item revealed that the **A Plan for Jobs 2020** document  had `"document_type": null` which was leading to an `I18n::InvalidPluralizationData`  error.

------------

## Context from original PR:

On the organisations pages, news stories using the image card component were always displaying the English translation as their document type even though translations existed in the config.

This changes the way document type data is pulled from the locale, to ensure that the localised terms are displayed when an organisation page is viewed in a certain language (e.g displaying 'Stori newyddion' instead of 'News story'

**Before**: the term "Press Release" is in English even though we are viewing the Welsh version of the page

Screenshot 2020-08-13 at 16 50 39

**After**: uses the Welsh version of "Press Release"

Screenshot 2020-08-13 at 16 51 05 

https://trello.com/c/uA7mDUvI

### DIFF
--- a/app/presenters/organisations/documents_presenter.rb
+++ b/app/presenters/organisations/documents_presenter.rb
@@ -93,6 +93,7 @@ module Organisations
 
       featured.each do |news|
         date = Date.parse(news["public_updated_at"]) if news["public_updated_at"]
+        text = I18n.t("organisations.content_item.schema_name.#{news['document_type']&.parameterize(separator: '_')}", count: 1, default: news["document_type"]) if news["document_type"]
 
         news_stories << {
           href: news["href"],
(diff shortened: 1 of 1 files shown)


hannako (approved):
Looks good! 👍

PR #1877, opened 2020-08-21
Make changes to the header notice component

JonathanHallam:
This allows the addition of an extra description to be added underneath the link within the header notice. It also creates the docs for that component.

### DIFF
--- a/app/views/components/docs/header-notice.yml
+++ b/app/views/components/docs/header-notice.yml
@@ -6,8 +6,8 @@ examples:
     data:
       nhs_branding: true
       title_logo: |
-        <svg xmlns="http://www.w3.org/2000/svg" role="img" viewBox="0 0 69.2 28" aria-labelledby="nhs-logo__title">
-          <title id="nhs-logo__title">NHS logo</title>
+        <svg xmlns="http://www.w3.org/2000/svg" role="img" viewBox="0 0 69.2 28" aria-labelledby="nhs-logo__title_0">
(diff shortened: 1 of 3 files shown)


app/views/components/docs/header-notice.yml
```diff
@@ -0,0 +1,76 @@
...
+        </svg>
+      heading: "If you have coronavirus symptoms:"
+      list:
+        - a high temperature
+        - a new, continuous cough
+        - a loss of, or change to, your sense of smell or taste
+      call_to_action:
+        href: "https://www.gov.uk/get-coronavirus-test"
+        title: Get a test now
+        description: do not leave home for at least 10 days after your test
```
injms, line 51:
```suggestion
        description: Do not leave home for at least 10 days after your test
```

app/assets/stylesheets/components/_header-notice.scss
```diff
@@ -51,3 +51,7 @@
 .app-c-header-notice__call-to-action {
   @include govuk-font(16, $weight: bold);
 }
+
+.app-c-header-notice__call-to-action-description {
+  @include govuk-font(16);
```
injms, line 56: If you like, this could be removed completely and in the markup replaced with `govuk-body-s` instead? (https://design-system.service.gov.uk/styles/typography/#body-small)

injms (approved):
One optional suggestion - actually optional - but this looks good to go! 🚀

PR #1882, opened 2020-08-28
Set A to Z element to be an h3

owenatgov:
## What
This PR achieves 2 things: 

1. Reworks the "A to Z" text seen on browse views ([example](https://www.gov.uk/browse/benefits/disability)) so that it now sits within an h3 tag.
2. Removes the `hairspace` helper method and all instances of it.

## Why
It was raised during the accessibility audit of govuk that there are several pages where there are elements that look like headings but aren't, failing WCAG criterion 1.3.1. The reason I've removed `hairspace` is because the hairspace unicode character isn't recognised by screen readers. In this example, the string "A to Z" with the spaces replaced with a hairspace get read out by screen readers as "ah-toz".

### Before
Screenshot 2020-08-28 at 11 11 51

### After
Screenshot 2020-08-28 at 11 14 16

**Card:** https://trello.com/c/g7RVI9Mq/338-browse-pages-text-looks-like-heading-but-isnt-one

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -7,10 +7,6 @@ module ApplicationHelper
     request.path.starts_with?(section.base_path)
   end
 
-  def hairspace(string)
-    string.gsub(/\s/, "\u200A") # \u200A = unicode hairspace
-  end
-
   def current_path_without_query_string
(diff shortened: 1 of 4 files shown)


maxgds (approved):
Looks good.

PR #1883, opened 2020-09-01
Make requested changes to header notice

JonathanHallam:
This makes some changes to how the header notice looks. It slightly increases font size for the description and adds a small amount of padding above the text.

## Before 
Screenshot 2020-09-01 at 14 09 58

## After
Screenshot 2020-09-01 at 14 07 06


injms (approved):
👍

PR #1899, opened 2020-09-10
Make organisation search more accessible

andysellick:

---

## What

Changes the search on https://www.gov.uk/government/organisations to be visually clearer and more accessible. Specifically:

- replaces the existing label that looks like a heading with inline search input with a standard input component
- updates the JS to add a result count and include `aria-live` attribute so screen readers hear the results of their search and know that something is happening
- adds landmarks to the search and results to aid navigation

Before | After
------- | ------
Screenshot 2020-09-10 at 09 37 26 | Screenshot 2020-09-10 at 09 37 06

## Why

The existing search failed accessibility criteria.

Related to: https://github.com/alphagov/whitehall/pull/5788

Trello card: https://trello.com/c/XyOr5lMm/334-in-page-search-unclear-search-functionality-and-text-that-looks-like-headings

### DIFF
--- a/app/assets/javascripts/organisation-list-filter.js
+++ b/app/assets/javascripts/organisation-list-filter.js
@@ -9,6 +9,13 @@
       // Form should only appear if the JS is working
       $('[data-filter="form"]').addClass('filter-organisations-list__form--active');
 
+      this.results = document.createElement('div')
+      this.results.classList.add('filter-organisations-list__results', 'govuk-heading-m', 'js-search-results')
+      this.results.setAttribute('aria-live', 'polite')
+      this.results.innerHTML = window.GOVUK.filter.countInitialDepartments() + ' results found'
+
+      $('#organisations_search_results').prepend(this.results)
+
       // We don't want the form to submit/refresh the page on enter key
       $('[data-filter="form"]').on('submit', function() { return false; })
 
@@ -54,6 +61,10 @@
       }
     },
 
+    countInitialDepartments: function () {
+      return document.querySelectorAll('[data-filter="item"]').length;
+    },
+
     updateDepartmentCount: function(listsToFilter) {
       var totalMatchingOrgs = 0;
 
@@ -74,7 +85,11 @@
         totalMatchingOrgs += matchingOrgCount;
       });
 
-      $('.js-no-filter-matches').toggleClass('js-hidden', totalMatchingOrgs > 0);
+      var text = ' results found'
+      if (totalMatchingOrgs === 1) {
+        text = ' result found'
+      }
+      this.results.innerHTML = totalMatchingOrgs + text
     }
   };
 
--- a/app/assets/stylesheets/views/_organisations.scss
+++ b/app/assets/stylesheets/views/_organisations.scss
@@ -68,39 +68,12 @@
       float: right;
     }
   }
-
-  .organisations__no-filter-match {
-    display: none;
-
-    .js-enabled & {
-      @include govuk-font(19);
-      display: block;
-      text-align: center;
-      margin-bottom: govuk-spacing(9);
-    }
-  }
 }
 
 .filter-organisations-list__form {
   display: none;
-
-  .filter-organisations-list__label {
-    @include govuk-font(36, $weight: bold);
-    margin-right: govuk-spacing(2);
-  }
-
-  .filter-organisations-list__input {
-    @include govuk-font(16);
-    width: 30%;
-    min-width: 150px;
-    vertical-align: text-bottom;
-    padding: 5px;
-    border: 1px solid govuk-colour("mid-grey", $legacy: "grey-2");
-    margin: 0 5px 5px 0;
-  }
 }
 
 .filter-organisations-list__form--active {
   display: block;
-  margin-bottom: 45px;
 }
--- a/app/views/organisations/index.html.erb
+++ b/app/views/organisations/index.html.erb
@@ -9,21 +9,24 @@
     <%= render "govuk_publishing_components/components/title", {
         title: @presented_organisations.title
     } %>
+
+    <form class="filter-organisations-list__form" data-filter="form">
+      <%= render "govuk_publishing_components/components/input", {
+        label: {
+          text: "Search for a department, agency or public body"
+        },
+        name: "search-box",
+        id: "filter-organisations-list",
+        controls: "organisations_search_results"
+      } %>
+    </form>
   </div>
 </div>
 
-<form class="filter-organisations-list__form" data-filter="form">
-  <label class="filter-organisations-list__label" for="filter-organisations-list">What’s the latest<br />from</label>
-  <input class="filter-organisations-list__input govuk-input" type="text" id="filter-organisations-list" placeholder="For example, Home Office">
-  <span class="filter-organisations-list__label" aria-hidden="true">?</span>
-</form>
-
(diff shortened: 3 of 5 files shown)


app/assets/javascripts/organisation-list-filter.js
```diff
@@ -8,6 +8,8 @@
     init: function() {
       // Form should only appear if the JS is working
       $('[data-filter="form"]').addClass('filter-organisations-list__form--active');
+      var $results = $('<div/>').addClass('filter-organisations-list__results govuk-heading-m').attr('id', 'search-results').attr('aria-live', 'polite').html(window.GOVUK.filter.countInitialDepartments() + ' results found');
```
owenatgov, line 11: Couple of things:

- It appears that this code is rendering a div with classes that can contain the results list. Could we render this in the markup itself instead then just grab it via a selector?
- Could we vanillaify the new js? I reckon most of these could stay as single lines, mostly just replacing `$`'s with `querySelector`.

app/assets/javascripts/organisation-list-filter.js
```diff
@@ -75,6 +81,12 @@
       });
 
       $('.js-no-filter-matches').toggleClass('js-hidden', totalMatchingOrgs > 0);
+
+      var text = ' results found'
```
owenatgov, line 85: You could one-line this declaration followed by an if statement. I'm open to if this makes the line more difficult to read. Ideally we'd have ES6 template literals which would make this dreamy to read.

```suggestion
      var text = ' result' + totalMatchingOrgs === 1 ? '' : 's' +  ' found'
```

owenatgov (changes requested):
The search box is pretty cozy to the right of the browser window on smaller screens:
Screenshot 2020-09-10 at 10 27 29

At a glance, there's some styling which changes the box-sizing. Removing this on my browser sorts it out.

Besides that just a few nitpicks. Happy to discuss. Looking good 👍

andysellick:
@owenatgov ah yes, I spotted that layout thing too, forgot to mention this PR: https://github.com/alphagov/govuk_publishing_components/pull/1679

app/assets/javascripts/organisation-list-filter.js
andysellick: This div holds the result count from the user's search and search isn't available when JS is disabled. Since this element is specifically related to search I didn't think it should appear in the markup, as it would appear even when JS is disabled.

I could vanilla this JS, but as the whole of the rest of the file uses jQuery already I think that's outside the scope of this PR.

app/assets/javascripts/organisation-list-filter.js
andysellick: Definitely more efficient, definitely harder to read 😁

app/assets/javascripts/organisation-list-filter.js
owenatgov: I take your point on the search box however whilst it doesn't appear without js it's still stored in the markup, just hidden. Could we hide the results count in the same way? I'm thinking that it'll be easier for a future dev to find and minimise js processing.

Willing to fold on vanilla js.

app/assets/javascripts/organisation-list-filter.js
owenatgov: Point taken. Comment withdrawn.

owenatgov:
@andysellick Ahaaaaaa good old legacy tooling. Will keep an eye on the other PR.

owenatgov:
Ah duh it's been merged. I won't let it stop this one in that case.

app/assets/javascripts/organisation-list-filter.js
andysellick: Oh I see... you're saying that because the search box is in the markup but hidden, the search results should follow the same pattern?

I get that, but the only reason the search box isn't rendered with JS as well is because it's a component. I'd rather try to have as many JS-only elements rendered with JS than make all JS only elements one way or the other.

owenatgov: I don't agree that we should be rendering anything via js if we can avoid it. This ensure we can separate template from functionality, making the code easier to interpret and  minimising the work needed to be done by the js.

I appreciate that this is my opinion though and don't want to perpetuate a stalemate more than I'm already starting to. If there's a case for this, like if it's a common pattern in our apps or there's a performance hit to rendering it in the html etc, then I'll let it go.

andysellick: @injms

app/assets/javascripts/organisation-list-filter.js
```diff
@@ -8,6 +8,8 @@
     init: function() {
       // Form should only appear if the JS is working
       $('[data-filter="form"]').addClass('filter-organisations-list__form--active');
+      var $results = $('<div/>').addClass('filter-organisations-list__results govuk-heading-m').attr('id', 'search-results').attr('aria-live', 'polite').html(window.GOVUK.filter.countInitialDepartments() + ' results found');
+      $('#organisations_search_results').prepend($results);
```
injms, line 12: Suggest using a `js-` class here instead of an id to more clearly show to future people that this is where JavaScript stuff happens in the template.

app/assets/javascripts/organisation-list-filter.js
injms: 🐱 🐦 time.

I think that the element is fine to be added by the JavaScript - but I would suggest changing the selector that it uses to add the new element (`#organisations_search_results`) to a `js-` hook. I've added this as a comment.

In a previous [GOV.UK Frontenders meeting](https://docs.google.com/document/d/1RsB7WQBkt3mH3yOD46VMNw8ZI9VTa04jyS0HXQfeYJI/edit?usp=sharing) there was a decision to not use jQuery when writing JavaScript - and it was fine if that meant mixing JavaScript and jQuery together. It'd be good to see that in this pull request - but not a blocker to it being merged as we're under tight time constraints.

andysellick: Thanks @injms. I've updated to use a `js-` hook as suggested. I've also rewritten the code to use vanilla JS @owenatgov except for one line where I'm using jQuery's `prepend` method - I tried switching to an alternative but it started throwing up weird test failures so I'm quitting while I'm ahead on that, sorry.

I've also changed the code slightly further to remove an element at the bottom of the page that I've only just noticed, that showed with a message when there were no search results, which is redundant now.

andysellick: Will hold off merging for a bit if you both want to take a second look.

app/assets/javascripts/organisation-list-filter.js
```diff
@@ -54,6 +61,10 @@
       }
     },
 
+    countInitialDepartments: function () {
+      return $('[data-filter="item"]').length;
```
owenatgov, line 65: Could un-jquery this line as well please?

```suggestion
      return document.querySelectorAll('[data-filter="item"]').length;
```

owenatgov (commented):
Very nearly good to go from my perspective. Just one stray bit of jQuery 🙏

app/assets/javascripts/organisation-list-filter.js
andysellick: Done, thanks.

owenatgov (approved):

PR #1904, opened 2020-09-11
Update page titles and headings for topic views

owenatgov:

## What
Updates page titles and h1's on subtopic pages and topic pages, to include the suffix "detailed information". This is organised like so:

- Topic page titles: `topic: detailed information - GOV.UK`
- Subtopic page titles: `topic: subtopic - detailed information - GOV.UK`
- Topic headings: `topic: detailed information`
- Subtopic headings: `subtopic: detailed information`

## Why
There are several instances of different pages that feature the same heading and page title in different parts of the site. This PR, using the detailed information title and heading suffix, will provide a clear differentiator between specialist topics and normal browse pages. Examples of specialist topic pages:

- [topic](https://www.gov.uk/topic/business-tax)
- [subtopic](https://www.gov.uk/topic/personal-tax/income-tax)

This was decided by the content support team during an accessibility content blast day.

### Before
Screenshot 2020-09-17 at 17 52 37

### After
Screenshot 2020-09-17 at 17 52 05

**Card:** https://trello.com/c/0ckzTB40/380-update-title-and-heading-pattern-for-specialist-topic-pages

### DIFF
--- a/app/views/subtopics/show.html.erb
+++ b/app/views/subtopics/show.html.erb
@@ -6,7 +6,7 @@
 <% end %>
 <% content_for :page_title do %>
   <%
-    title_params = { title: subtopic.title }
+    title_params = { title: "#{subtopic.title}: detailed information" }
     if subtopic.parent
       title_params[:context] = {
         text: subtopic.parent.title,
--- a/test/integration/topic_browsing_test.rb
+++ b/test/integration/topic_browsing_test.rb
@@ -61,7 +61,7 @@ class TopicBrowsingTest < ActionDispatch::IntegrationTest
     )
 
     visit "/topic/oil-and-gas"
-    assert page.has_title?("Oil and gas - GOV.UK")
+    assert page.has_title?("Oil and gas: detailed information - GOV.UK")
 
     within "header.page-header" do
       assert page.has_content?("Oil and gas")
--- a/app/models/topic.rb
+++ b/app/models/topic.rb
@@ -31,9 +31,9 @@ class Topic
 
   def combined_title
     if parent
-      "#{parent.title}: #{title}"
+      "#{parent.title}: #{title} - detailed information"
     else
-      title
+      "#{title}: detailed information"
     end
   end
 
--- a/app/views/topics/index.html.erb
+++ b/app/views/topics/index.html.erb
@@ -1,4 +1,5 @@
-<% content_for :title, topic.title %>
+<% title_with_suffix = "#{topic.title}: detailed information" %>
+<% content_for :title, title_with_suffix %>
 <%= render 'shared/tag_meta', tag: topic %>
 <% content_for :meta_tags do %>
   <meta name='govuk:navigation-page-type' content='<%= navigation_page_type %>'>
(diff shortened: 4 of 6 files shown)


danacotoran (approved):
Approving in terms of code but I think it would be sensible to also have this checked out by a content designer and/or Tobi to confirm these are the intended changes.

I'm a bit confused with regards to the scope of this ticket, as based on this PR I would think that only pages like [this one](https://govuk-collec-add-detail-jaojwk.herokuapp.com/topic/business-tax) should have the "Detailed information" suffix added to them.
However from reading the ticket, I am not sure if it also includes pages like [this one](https://govuk-collec-add-detail-jaojwk.herokuapp.com/topic/business-tax)... as the ticket mentions that type of topic page as an example. Best to double check I think.

app/views/subtopics/show.html.erb
```diff
@@ -1,20 +1,14 @@
...
-      title_params[:context] = {
-        text: subtopic.parent.title,
-        href: subtopic.parent.base_path
-      }
-    end
-  %>
-  <%= render "govuk_publishing_components/components/title", title_params %>
+  <%= render "govuk_publishing_components/components/title", { 
+    title: subtopic.title,
+    context: "detailed information",
```
injms, line 10: Should `detailed` be capitalised?

```suggestion
    context: "Detailed information",
```

owenatgov: This is something I wasn't 100% on. The card uses lowercase but I'll check with a content person.

test/integration/topic_browsing_test.rb
```diff
@@ -61,7 +61,7 @@ def oil_and_gas_topic_item(params = {})
     )
 
     visit "/topic/oil-and-gas"
-    assert page.has_title?("Oil and gas - GOV.UK")
+    assert page.has_title?("Oil and gas: detailed information - GOV.UK")
```
danacotoran, line 64: Should we add an equivalent test to check subtopic pages title, since that has changed as well?

danacotoran (commented):
A thought I had while reviewing this: are we sure that subtopic names are not repeated between topics?
To explain what I mean: let's say there's **TopicA** which has a subtopic titled **GenericSubtopic**. 
But there's also **TopicB** with a topic titled **GenericSubtopic**, but with different contents which are pertaining to that **TopicB** subject.
Then even though the contents and tagging of the two **GenericSubtopic** pages is different, both of those pages would have the same page title of `GenericSubtopic: detailed information - GOV.UK` when they should really be unique.

This theoretical issue (which may or may not even be possible 😂) should be averted by changing the subtopic page title to something like `TopicA: GenericSubtopic: detailed information - GOV.UK`

owenatgov:
@danacotoran That's a fair point. Let me run it past the content community and see what they say. In prod currently the format seems to be:

- for topic pages: `topicName - GOV.UK`
- for subtopic pages: `parentTopicName: SubtopicName - GOV.UK`

This might be another case of me making an assumption about the work that needs to be done. This card has truly been a whirlwind of my ability to not understand basic instructions 😂

danacotoran (approved):

PR #1909, opened 2020-09-15
Update subtopic semantics to no longer denote sections as navigation

owenatgov:
## What
Removes the `nav` tag surrounding subtopic lists and changes `h1`s within subtopics to `h2`s. Additionally removes some now redundant styling.

## Why
These sections being navigation elements can be confusing to screen readers as they aren't technically navigational items within and what were `h1`s look smaller than `h1`s seen across govuk. This fails WCAG SC 2.4.1 (navigations) and 1.3.1 (headings).

No visual changes.

**Card:** https://trello.com/c/qiAWvH0i/342-unsemantic-topic-pages 

### DIFF
--- a/app/assets/stylesheets/views/_topics.scss
+++ b/app/assets/stylesheets/views/_topics.scss
@@ -23,32 +23,9 @@
     min-height: 500px;
     padding: 0 0 45px 0;
 
-    h2,
-    h3 {
-      @include govuk-font(19, $weight: bold);
-      margin-bottom: 0;
-    }
-
     @include govuk-media-query($from: mobile) {
       padding: 0 0 30px 0;
       min-height: auto;
-    }
-  }
-
-  nav {
-    padding-top: 1em;
-    padding-bottom: 1em;
-    overflow: auto;
-
-    h1 {
-      @include govuk-font(24, $weight: bold);
-
-      margin-bottom: govuk-spacing(2);
-
-      @include govuk-media-query($from: desktop) {
-        float: left;
-        max-width: 12.5em;
-      }
     }
   }
 }
--- a/app/views/subtopics/show.html.erb
+++ b/app/views/subtopics/show.html.erb
@@ -25,13 +25,13 @@
     link_to_latest_feed: true,
   }) do %>
   <% subtopic.lists.each_with_index do |list, list_index| -%>
-    <nav class="govuk-grid-row" aria-labelledby="<%= list.title.parameterize %>">
+    <div class="govuk-grid-row govuk-!-margin-bottom-5">
       <div class="govuk-grid-column-one-third">
-        <h1 id="<%= list.title.parameterize %>"><%= list.title %></h1>
+        <h2 class="govuk-heading-m"><%= list.title %></h2>
       </div>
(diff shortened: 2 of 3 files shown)


app/views/subtopics/_subtopic.html.erb
```diff
@@ -2,7 +2,7 @@
 
 <header class="page-header group">
   <div class="govuk-grid-row">
-    <div class="govuk-grid-column-two-thirds">
+    <div class="govuk-grid-column-two-thirds" id="subtopic-metadata">
```
injms, line 5: Is this `id` just for the test selector?

owenatgov: Yep. I put it on to try and solve my janky test and decided to keep it since before the test was looking for `.gem-c-metadata` which felt kinda fragile to me.

injms: If the test works with the `.gem-c-metadata` selector then I think that selector should be used - otherwise we're adding extra stuff to a page with no benefit the user.

owenatgov: I'd question this. In my opinion this makes the code a little more manageable because it creates a unique id to test against instead of something that's reliant on an external repo, albeit one that the frontend community controls. I personally think that this helps separate the components gem and the specific test being carried out here. I'm also skeptical about how much negative impact this has to the markup as, to my knowledge, it doesn't make a tangible difference to the user.

Is this a point of poor practice?

injms: I agree it has negligible negative impact for the users - but it also doesn't have any positive impact whilst adding bloat (however small that may be) to the page.

The reason I brought this change up is because of the the [Priority of Constituencies](https://www.w3.org/TR/html-design-principles/#priority-of-constituencies):

> In case of conflict, consider users over authors over implementors over specifiers over theoretical purity.

I'd say that this doesn't bring any benefit to the user; and the previous selector worked, so actually brings little benefit to authors / developers.

owenatgov: That's a reasonable point and I've learnt something new with that priority of constituencies. I'm still a little skeptical though that the previous implementation is a good one just because it works. Yes, it works, but in my view that still doesn't make it future proof because it tightly couples itself to the state of the components gem when (again, in my view) the test scope for this codebase should be constrained as much as possible to just this codebase. I'd therefore make the case that it does benefit the author because it prevents false test fails from light changes in the component markup and makes it easier to debug in the event of a breaking change.

Anyway, final begrudging point made. I'll make this change.

injms: I take your point that differentiating tests in `collections` and the components gem does benefit the author. I think it's worth asking around and at the next GOV.UK Frontenders meeting anyone knows a way of doing this that doesn't impact production code.

On thing I should've mentioned earlier - there was an [issue raised on the components gem about adding extra attributes to make testing easier](https://github.com/alphagov/govuk_publishing_components/issues/1285), but it was closed after discussion at the GOV.UK Frontenders meeting.

injms (approved):

PR #1913, opened 2020-09-18
Add show and hide flag for webinar tables

DilwoarH:
# What

DIT have requested to remove reference to webinars from the EU business landing page. The webinars were scheduled to start on Monday but are being rescheduled
It's the section starting with "Additional help and support"

https://www.gov.uk/eubusiness

Reference to the webinar schedule has been removed, something has been put in its place that doesn't confuse users.

Reference https://trello.com/c/T68H7Gc7/464-remove-reference-to-webinars-on-eu-business-page

### DIFF
--- a/app/presenters/table_presenter.rb
+++ b/app/presenters/table_presenter.rb
@@ -3,6 +3,12 @@ class TablePresenter
     @table = table
   end
 
+  def show?
+    return unless valid_input?
+
+    table[:show]
+  end
+
   def rows
     return unless valid_input?
 
@@ -26,7 +32,7 @@ private
   attr_reader :table
 
   def valid_input?
-    return unless %i[headings rows].sort == table.keys.sort
+    return unless %i[show headings rows].sort == table.keys.sort
 
     table[:rows].all? { |e| e.is_a? Array }
   end
--- a/test/integration/dit_landing_page_test.rb
+++ b/test/integration/dit_landing_page_test.rb
@@ -26,5 +26,6 @@ class DitLandingPageTest < ActionDispatch::IntegrationTest
 
   def then_i_can_see_the_training_section
     assert page.has_selector?("h2", text: "Additional help and support")
+    # assert page.has_selector?(".govuk-table__row", text: "Date")
   end
 end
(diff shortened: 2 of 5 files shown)


app/presenters/table_presenter.rb
```diff
@@ -3,6 +3,12 @@ def initialize(table)
     @table = table
   end
 
+  def show
```
leenagupte, line 6: Is this returning a boolean? Should this be called `show?`

app/presenters/table_presenter.rb
```diff
@@ -26,7 +32,7 @@ def headings
   attr_reader :table
 
   def valid_input?
-    return unless %i[headings rows].sort == table.keys.sort
+    return unless %i[show headings rows].sort == table.keys.sort
```
leenagupte, line 35: What is "show" here? is it the new method?

test/integration/dit_landing_page_test.rb
```diff
@@ -26,6 +26,5 @@ def and_i_can_see_the_translation_links
 
   def then_i_can_see_the_training_section
     assert page.has_selector?("h2", text: "Additional help and support")
-    assert page.has_selector?(".govuk-table__row", text: "Date")
```
leenagupte, line 29: Why has this assertion been removed?

leenagupte: Should there be a separate test to check that the table does display when "show=true"?

leenagupte (commented):
I've left some inline comments, mainly about the naming. I think "show" could be a bit confusing as the name clashes with a common controller method.

I haven't seen this working yet, so will try and test it locally.

app/presenters/table_presenter.rb
leenagupte: Is there a test for `valid_input?` that needs to be updated?

DilwoarH: I believe this function checks if the keys in the YAML exist

leenagupte:
Ran it locally, and it looks good: 

Screenshot 2020-09-18 at 12 29 09

alex-ju:
Not that the `eubusiness` was migrated to a more standard page, powered by a publisher app we don't need this functionality anymore.

PR #1931, opened 2020-10-01
Brexit countdown clock

hannako:
## What

We want to run an AB test to see if adding a countdown clock to the transition landing page increases engagement. As a first step we are adding the component, but hiding it behind a flag. The component can be switched on via the AB test as required.

Related PR: https://github.com/alphagov/collections/pull/1932

### show_countdown? = true

Screenshot 2020-10-05 at 10 40 13

### show_countdown? = false

Screenshot 2020-10-05 at 10 40 26

Trello: https://trello.com/c/2xjWj2WN/481-a-b-test-for-urgency-phase-1

### DIFF
--- a/app/assets/stylesheets/views/_transition-countdown.scss
+++ b/app/assets/stylesheets/views/_transition-countdown.scss
@@ -1,0 +1,85 @@
+$dark-blue-2: #003078;
+$desktop-breakpoint-override: 1000px;
+
+.transition-countdown {
+  border: govuk-spacing(1) solid $dark-blue-2;
+  margin: 0;
+
+  .transition-countdown__countdown {
+    padding: govuk-spacing(3);
+    color: $dark-blue-2;
+    text-align: center;
+    width: 100%;
+
+    @include govuk-media-query($from: $desktop-breakpoint-override) {
+      padding: govuk-spacing(4);
+      width: 50%;
+    }
+  }
+
+  .transition-countdown__countdown-days-left {
+    @include govuk-font(80, $weight: bold);
+    display: inline-block;
+    vertical-align: middle;
+
+    @include govuk-media-query($from: tablet) {
+      @include govuk-font(48, $weight: bold);
+    }
+
+    @include govuk-media-query($from: $desktop-breakpoint-override) {
+      @include govuk-font(80, $weight: bold);
+      display: block;
+    }
+  }
+
+  .transition-countdown__countdown-days-text {
+    @include govuk-font($size: 24, $line-height: 1.5);
+    display: inline-block;
+    vertical-align: middle;
+    margin: 0 0 govuk-spacing(1) govuk-spacing(1);
+
+    @include govuk-media-query($from: $desktop-breakpoint-override) {
+      @include govuk-font(19);
+      display: block;
+      margin: 0;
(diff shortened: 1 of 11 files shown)


app/lib/countdown_clock.rb
```diff
@@ -0,0 +1,19 @@
+class CountdownClock
+  END_OF_TRANSITION_PERIOD = Date.new(2021, 1, 1)
+
+  def days_left
+    (END_OF_TRANSITION_PERIOD - today_in_london).to_i
+  end
+
+  def today_in_london
```
sihugh, line 8: could this be a private method?

app/views/components/_countdown.html.erb
```diff
@@ -0,0 +1,3 @@
+<% if clock.show? %>
+  <%= clock.days_left %><%= clock.days_text %> to go
```
sihugh: I'm assuming this is English only for now while we're doing the test.

app/assets/stylesheets/views/_transition-countdown.scss
```diff
@@ -0,0 +1,76 @@
+$dark-blue-2: #003078;
+
+.Transition-countdown {
```
sihugh, line 4: Is this capitalised for a reason? We don't capitalise GOVUK for example?

app/views/transition_landing_page/_countdown.html.erb
```diff
@@ -0,0 +1,9 @@
+<div class="Transition-countdown govuk-grid-row">
+  <div class="govuk-grid-column-one-half Transition-countdown__countdown">
+    <p class="Transition-countdown__countdown-days-left"><%= clock.days_left %></p>
+    <p class="Transition-countdown__countdown-days-text"><%= clock.days_text %> to go</p>
```
sihugh, line 4: Worried that splitting (for example) `92 days to go` up with `p` tags is an accessibility fail, as screen readers will read it out with a big pause in the middle.

sihugh: Should we split using a <br> tag instead?

sihugh (commented):
A few questions, the most important being the possible accessibility issue, though I don't think it's too hard to fix

app/lib/countdown_clock.rb
hannako: thanks yep will do.

app/views/components/_countdown.html.erb
hannako: Yes that's right. We haven't been provided with the `cy` translation of the content, and we're only testing in the `en` locale. But if the AB test shows the clock is a good thing, we would at that point need to have a translation which we might forget. I'll mention to Nick so we don't.

app/assets/stylesheets/views/_transition-countdown.scss
hannako: I don't think so. @DilwoarH is it ok for me to change?

app/views/transition_landing_page/_countdown.html.erb
hannako: roger

sihugh (approved):

PR #1932, opened 2020-10-01
Transition landing page Urgency AB Test 1

hannako:
## Urgency AB Test 1

We would like to test a raft of changes aimed at creating a greater sense of urgency for users visiting the Transition landing page. The hope is that it will result in higher rates of engagement. 

### PAGE HEADER: Default

Screenshot 2020-10-01 at 18 24 10

### PAGE HEADER: Variant B

Screenshot 2020-10-01 at 18 25 24

### TAKE ACTION BOX: Default

Screenshot 2020-10-01 at 18 24 19

### TAKE ACTION BOX: Variant B

Screenshot 2020-10-01 at 18 25 31

### BREXIT COUNTDOWN: Default
Screenshot 2020-10-05 at 12 10 22

### BREXIT COUNTDOWN: Variant B

Screenshot 2020-10-05 at 12 09 48

------

## How to test

To toggle AB tests from the [review app](https://govuk-collec-urgency-ab-ssskzw.herokuapp.com/transition) install [modheader](https://chrome.google.com/webstore/detail/modheader/idgpnmonknjnojddfkpgkljpfnnfcklj?hl=en) or alternative, and then add a new Request Header by clicking the `+` and adding `govuk-abtest-transtitionurgency1: B`

Trello: https://trello.com/c/2xjWj2WN/481-a-b-test-for-urgency-phase-1

### DIFF
--- a/app/assets/stylesheets/views/_transition-countdown.scss
+++ b/app/assets/stylesheets/views/_transition-countdown.scss
@@ -17,7 +17,7 @@ $desktop-breakpoint-override: 1000px;
     }
   }
 
-  .transition-countdown__countdown-days-left {
+  .transition-countdown__days-left {
     @include govuk-font(80, $weight: bold);
     display: inline-block;
     vertical-align: middle;
@@ -32,7 +32,7 @@ $desktop-breakpoint-override: 1000px;
     }
   }
 
-  .transition-countdown__countdown-days-text {
+  .transition-countdown__days-text {
     @include govuk-font($size: 24, $line-height: 1.5);
     display: inline-block;
     vertical-align: middle;
--- a/app/controllers/transition_landing_page_controller.rb
+++ b/app/controllers/transition_landing_page_controller.rb
@@ -7,6 +7,8 @@ class TransitionLandingPageController < ApplicationController
   def show
     setup_content_item_and_navigation_helpers(taxon)
 
+    ab_test_variant.configure_response(response) if page_under_test?
+
     render locals: {
       presented_taxon: presented_taxon,
       presentable_section_items: presentable_section_items,
@@ -45,5 +47,26 @@ private
   def switch_locale(&action)
     locale = params[:locale] || I18n.default_locale
     I18n.with_locale(locale, &action)
+  end
+
+  helper_method :ab_test_variant, :page_under_test?, :show_variant?
+  def ab_test_variant
+    @ab_test_variant ||= begin
+      ab_test = GovukAbTesting::AbTest.new(
+        "TransitionUrgency1",
+        dimension: 44,
+        allowed_variants: %w[A B Z],
+        control_variant: "Z",
+      )
+      ab_test.requested_variant(request.headers)
+    end
+  end
+
+  def page_under_test?
+    request.path == "/transition"
+  end
+
+  def show_variant?
+    page_under_test? && ab_test_variant.variant?("B")
   end
 end
--- a/app/controllers/transition_landing_page_controller.rb
+++ b/app/controllers/transition_landing_page_controller.rb
@@ -41,7 +41,7 @@ private
   end
 
   def show_countdown?
-    show_variant? || false
+    show_variant?
   end
 
   def switch_locale(&action)
--- a/app/controllers/transition_landing_page_controller.rb
+++ b/app/controllers/transition_landing_page_controller.rb
@@ -41,7 +41,7 @@ private
   end
 
   def show_countdown?
-    false
+    show_variant? || false
   end
 
   def switch_locale(&action)
--- a/app/views/transition_landing_page/_countdown.html.erb
+++ b/app/views/transition_landing_page/_countdown.html.erb
@@ -1,8 +1,8 @@
 <div class="transition-countdown govuk-grid-row">
   <div class="govuk-grid-column-one-half transition-countdown__countdown">
     <p>
-      <span class="transition-countdown__countdown-days-left"><%= clock.days_left %></span>
-      <span class="transition-countdown__countdown-days-text"><%= clock.days_text %> to go</span>
+      <span class="transition-countdown__days-left"><%= clock.days_left %></span>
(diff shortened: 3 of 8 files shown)


config/locales/en/transition_landing_page.yml
```diff
@@ -9,7 +9,9 @@ en:
     take_action_title: Take action
     take_action_title_variant_B: Make sure you're ready
     take_action_text: |
-      <p>Answer a few questions to get a personalised list of actions for you, your family, and your business. Then sign up for emails to get updates when things change.</p>
+      <p class="take_action_test_class">Answer a few questions to get a personalised list of actions for you, your family, and your business. Then sign up for emails to get updates when things change.</p>
```
sihugh, line 12: Could we move `<p class="take_action_test_class">` tags into the view? It seems the same for both (though would need to check the Welsh version too).
That would mean we could drop the `html_safe` call and sleep well at night. I think we'd be able to regenerate the brakeman config too, as it wouldn't need an exclusion any more.

app/controllers/transition_landing_page_controller.rb
```diff
@@ -41,7 +41,7 @@ def show_comms?
   end
 
   def show_countdown?
-    false
+    show_variant? || false
```
sihugh, line 44: Do we need the `|| false`? It confused me a little

config/locales/en/transition_landing_page.yml
hannako: Ah yes that would be nicer, will give that a go.

app/controllers/transition_landing_page_controller.rb
hannako: I put it in as I wasn't sure sure where [this error](https://github.com/alphagov/govuk_ab_testing/blob/75d837afa3fc2bde2d06b6a93f5a6db29b42b013/lib/govuk_ab_testing/requested_variant.rb#L47) would bubble up to if the variant in the header isn't A B or Z, in this method:

```
def show_variant?
   page_under_test? && && ab_test_variant.variant?("B")
end
```
But it actually doesn't matter because even if `ab_test_variant.variant?("B")` returns `nil`, `true && nil` is `nil`, so we won't render the component.

So the TLDR is.... we don't need the `|| false` . I shall delete it.

app/assets/stylesheets/views/_transition-countdown.scss
```diff
@@ -0,0 +1,85 @@
...
+    @include govuk-media-query($from: $desktop-breakpoint-override) {
+      @include govuk-font(80, $weight: bold);
+      display: block;
+    }
+  }
+
+  .Transition-countdown__countdown-days-text {
+    @include govuk-font($size: 24, $line-height: 1.5);
+    display: inline-block;
+    vertical-align: middle;
```
andysellick, line 38: I don't see this line making any difference - is this a browser specific thing?

app/assets/stylesheets/views/_transition-countdown.scss
```diff
@@ -0,0 +1,85 @@
...
+      @include govuk-font(19);
+      display: block;
+      margin: 0;
+    }
+  }
+
+  .Transition-countdown__traffic-lights {
+    background-color: $dark-blue-2;
+    padding: govuk-spacing(3) govuk-spacing(2) govuk-spacing(2);
+    width: 100%;
```
andysellick, line 51: Is this necessary? It's a block level element so defaults to this already.

app/views/transition_landing_page/_countdown.html.erb
```diff
@@ -0,0 +1,9 @@
+<div class="Transition-countdown govuk-grid-row">
```
andysellick, line 1: The use of the custom `transition-countdown` classes together with the govuk-frontend grid classes seems to result in a kind of conflict, where the custom classes are having to override aspects of the govuk-frontend classes. Why are the grid classes here at all? It looks like the whole thing could be simpler if they were removed.

andysellick (changes requested):
The frontend seems a little over complicated. I've highlighted a few issues but refrained from going any deeper until they're discussed.

app/assets/stylesheets/views/_transition-countdown.scss
```diff
@@ -0,0 +1,85 @@
+$dark-blue-2: #003078;
+$desktop-breakpoint-override: 1000px;
```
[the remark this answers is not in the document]
DilwoarH: Hey, this was introduced to add in the following style, the current component as it is wasn't working well.

Screenshot 2020-10-02 at 16 17 46

app/assets/stylesheets/views/_transition-countdown.scss
```diff
@@ -0,0 +1,85 @@
...
+
+.Transition-countdown {
+  border: govuk-spacing(1) solid $dark-blue-2;
+  margin: 0;
+
+  .Transition-countdown__countdown {
```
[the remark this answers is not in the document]
DilwoarH: Agreed, we could potentially unnest a lot of these things.

app/assets/stylesheets/views/_transition-countdown.scss
```diff
@@ -0,0 +1,85 @@
+$dark-blue-2: #003078;
+$desktop-breakpoint-override: 1000px;
+
+.Transition-countdown {
```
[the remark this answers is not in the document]
DilwoarH: We've changed it to small caps on a different branch, rebasing will fix this

app/assets/stylesheets/views/_transition-countdown.scss
```diff
@@ -0,0 +1,85 @@
...
+    text-align: center;
+    width: 100%;
+
+    @include govuk-media-query($from: $desktop-breakpoint-override) {
+      padding: govuk-spacing(4);
+      width: 50%;
+    }
+  }
+
+  .Transition-countdown__countdown-days-left {
```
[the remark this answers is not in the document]
hannako: I have made this change.

sihugh (approved):

PR #1936, opened 2020-10-07
[DO NOT MERGE] Lookup partial postcode

chris-gds:
# Context
- As a user I want to lookup local restrictions for Corona Virus to see what rules apply to me, I can visit this page and input my postcode to get relevant information.

Creating a new a postcode lookup partial _similar_ to [find-my-council](https://www.gov.uk/find-local-council) on `finder-frontend`

# Ticket
https://trello.com/c/6mYZd3HQ

# Design
https://www.figma.com/file/GuYNyaQkjKqhR87nqqqRTv/Local-restrictions?node-id=0%3A1

# Code History
Initial PR: https://github.com/alphagov/finder-frontend/pull/2226 

(Draft PR) Feedback integration on-going

Moved from `frontend` to `finder-fronend` to `collections`


app/views/coronavirus_landing_page/components/landing_page/_postcode-lookup.html.erb
```diff
@@ -0,0 +1,42 @@
+<%
+  input_label ||= "Enter your postcode"
+  input_value ||= ""
+  input_error ||= ""
+  hint_text ||= "For example SW1A 2AA"
+  button_text ||= "Find"
+  error_message ||= nil
+  error_description ||= nil
+  autocomplete ||= "postcode"
```
DilwoarH, line 9:
```suggestion
  autocomplete ||= "postal-code"
```

DilwoarH: ref https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete

app/views/coronavirus_landing_page/components/landing_page/postcode-lookup.spec.rb
```diff
@@ -0,0 +1,61 @@
...
+    render_component(hint_text: "For example SW1A 2AA")
+    assert_select ".govuk-hint", text: "For example SW1A 2AA"
+  end
+
+  it "contains lookup button and text" do
+    render_component(button_text: "Find")
+    assert_select ".gem-c-button", text: "Find"
+  end
+
+  it "contains autocomplete" do
```
DilwoarH, line 56: Maybe we should remove this as per Andy's suggestion on the previous PR

app/views/coronavirus_landing_page/components/landing_page/_postcode-lookup.html.erb
DilwoarH: I think we can probably just remove this and just put it in the attribute

app/views/coronavirus_landing_page/components/landing_page/_postcode-lookup.html.erb
```diff
@@ -0,0 +1,42 @@
...
+          label: {
+            text: input_label
+          },
+          heading_size: "s",
+          error_message: input_error,
+          value: input_value,
+          name: "covid__postcode-lookup",
+          hint: hint_text,
+          invalid: error_message ? 'true' : 'false',
+          autocomplete: autocomplete,
```
DilwoarH, line 30:
```suggestion
          autocomplete: "postal-code",
```

app/views/coronavirus_landing_page/components/landing_page/postcode-lookup.spec.rb
```diff
@@ -0,0 +1,61 @@
...
+    render_component(autocomplete: "postcode")
+    assert_select ".gem-c-input.govuk-input"
+    assert_select ".gem-c-input.govuk-input", autocomplete: "postcode"
+  end
```
DilwoarH, line 60:
```suggestion
```

DilwoarH (changes requested):

app/views/coronavirus_landing_page/components/landing_page/_postcode-lookup.html.erb
```diff
@@ -0,0 +1,42 @@
+<%
+  input_label ||= "Enter your postcode"
+  input_value ||= ""
```
DilwoarH, line 3: Should this be nil?

```suggestion
  input_value ||= nil
```

app/views/coronavirus_landing_page/components/landing_page/_postcode-lookup.html.erb
```diff
@@ -0,0 +1,42 @@
...
+  input_error ||= ""
```
DilwoarH, line 4: Same as above

```suggestion
  input_error ||= nil
```

DilwoarH (changes requested):

app/views/coronavirus_landing_page/components/landing_page/_postcode-lookup.html.erb
```diff
@@ -0,0 +1,42 @@
...
+  link ||= { text: "Find a postcode on Royal Mail's postcode finder", href: "http://www.royalmail.com/find-a-postcode"}
+%>
+<% if error_message %>
+  <%= render "govuk_publishing_components/components/error_alert", {
+      message: error_message,
+      description: error_description,
+    } %>
+<% end %>
+<div class="covid__postcode-lookup">
+  <form action="" method="post">
```
DilwoarH, line 19: @leenagupte @Rosa-Fox - do we have a route to post this to yet or are you happy for us to update this once you do?

leenagupte: @DilwoarH Should the form action be here? Shouldn't it be in the view that calls this partial?

leenagupte: Also, I'm not sure about the location of this view. I don't think "landing page" is an appropriate place to put it as the lookup may not end up on the landing page. Could it go in "shared" instead?

DilwoarH: All of the code for the postcode will be within this partial, this will include the form. We've included the components needed for the rest of the pattern within this partial.

DilwoarH: I think shared makes more sense than landing page as this is something that will potentially be used for hub pages too

leenagupte: I'd pictured this being a bit more reusable, so being able to be rendered on any page in collections, that's why I'm surprised that the form has been included as well.

app/views/coronavirus_landing_page/components/landing_page/postcode-lookup.spec.rb
chris-gds: Yeah, I think all tests should be removed this is not needed in this context, the error checking can happen at a page level.

app/views/coronavirus_landing_page/components/landing_page/_postcode-lookup.html.erb
chris-gds: Probably makes more sense to keep as the `autocomplete` as a var with the default being updated, equally happy to remove var - as this is an MVP partial (eg this is unlikely to not be anything but a postcode input)

app/views/coronavirus_landing_page/components/landing_page/_postcode-lookup.html.erb
chris-gds: @leenagupte Reusable in what context? Only within `collections` but not within other frontend repos? Is reusable in scope of for MVP?

leenagupte: Reusable within this app.

leenagupte: It's not a component, but some of the same principles could/should still apply: accessible, testable, isolated..
So, perhaps rather than "resuable" I thought it would be isolated from the coronavirus specific code.

chris-gds: So `collections` app only?  Would that mean it wouldn't be able be reused elsewhere? Reusable in the context of only a _postcode_-look up or all types of possible lookup? 

As this pattern appears on multiple places (eg `find-my-council`) it might make sense to eventually add this as as a component (or pattern) so it can be reused easily.  However this is probably a wider conversation outside the scope of MVP / `collections` only.

leenagupte: Yes, the collections app only, I'm not talking about wider use or considering other postcode lookups that exist elsewhere. But the partial can still follow the principle of being "isolated" within this app.

DilwoarH: > I'd pictured this being a bit more reusable, so being able to be rendered on any page in collections, that's why I'm surprised that the form has been included as well.

I think as it is we can reuse it - I think the only change that will be needed if we do use it in other places is moving it to a more generic place within the app.

We could move the form outside of the partial but I don't think without it, it has any purpose. ie. it will just be an input and button.

Happy to discuss if you are thinking about something I haven't answered.

chris-gds: > I think as it is we can reuse it - I think the only change that will be needed if we do use it in other places is moving it to a more generic place within the app.

> We could move the form outside of the partial but I don't think without it, it has any purpose. ie. it will just be an input and button.

Agreed, if this was used elsewhere it would be moved out of the folder.  Yeah it would be good to discuss, I'd like to understand what possible other use-cases (_aside_ from a postcode lookup) might this be used for within the context of this particular app. (eg last name lookup?) Also how does this link into the scope of MVP delivery.

> I'm not talking about wider use or considering other postcode lookups that exist elsewhere. But the partial can still follow the principle of being "isolated" within this app.

It might be worth considering to be added to the [Component Guide](https://components.publishing.service.gov.uk/component-guide) as it appears this might be have the isolated, reusable properties described rather that investment into just the `collections app` only, especially as other implementations exists on other repos.

app/views/coronavirus_landing_page/components/landing_page/_postcode-lookup.html.erb
chris-gds: Updated to `autocomplete ||= "postal-code"` - `postal-code` fallback but can be passed in a different value if needed

app/views/coronavirus_landing_page/components/landing_page/postcode-lookup.spec.rb
chris-gds: Removed granular test for this context (as per feedback)

app/views/coronavirus_landing_page/components/landing_page/_postcode-lookup.html.erb
chris-gds: Agreed, updated.

app/views/coronavirus_landing_page/components/landing_page/_postcode-lookup.html.erb
chris-gds: Duplicate comment (as per below)
Updated to autocomplete ||= "postal-code" - `postal-code` by default but can be passed in a different value if needed

app/views/coronavirus_landing_page/components/landing_page/_postcode-lookup.html.erb
```diff
@@ -0,0 +1,42 @@
+<%
```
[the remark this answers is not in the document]
DilwoarH: it is fine as it is for now, can you move this under shared please?

app/views/coronavirus_landing_page/show.html.erb
```diff
@@ -41,6 +41,17 @@
     </div>
 
     <div class="govuk-grid-column-two-thirds">
+
+      <%= render partial: 'coronavirus_landing_page/components/landing_page/postcode-lookup', input_error: "This isn't a valid postcode.",
```
DilwoarH, line 45: This should be removed

app/views/coronavirus_landing_page/components/landing_page/_postcode-lookup.html.erb
leenagupte: So when I was taking about reusability, I was thinking about it as a "postcode lookup" but for other things (if need arose). i.e. that submit to a different form, but all still with collections.

app/assets/stylesheets/views/_covid.scss
```diff
@@ -305,3 +305,9 @@ $c19-landing-page-header-background: #0843a1;
 .covid-timeline__item-heading {
   margin-bottom: govuk-spacing(1);
 }
+
+.covid__postcode-lookup {
```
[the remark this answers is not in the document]
chris-gds: separate out due to being move into a potential hub page

chris-gds:
Closed as Draft PR to prompt discussion.

PR #1946, opened 2020-10-09
Changes to the transition page AB test.

sihugh:
We're removing the countdown clock as it's not performing well.

We're keeping the variant text, and making it the default as it _has_ worked well.

We're testing a variant of the bucket section to emphasise business actions. We hope this will mean more people will do both the "things you can do now" and the full checker process.

Further context in each commit.

https://trello.com/c/OCB6pq4D/510-a-b-test-for-urgency-phase-2


hannako (approved):
Lovely!

PR #1947, opened 2020-10-09
Get Lower tier local authority name for postcode

leenagupte:
We need to display the name of the area that a postcode is in if we
can't find any local coronavirus restrictions for the area.
Unfortunately there isn't any consistency in the "type" code for lower
tier areas

The "/find-local-council" service had the same problem, and uses a list
of "known" lower tier codes, that we have borrowed and are reusing.
See: https://github.com/alphagov/frontend/blob/8dbe0f0f6ca0f5a777a0d6ca77b858fe5adc2494/app/controllers/find_local_council_controller.rb#L8-L10

### DIFF
--- a/app/models/mapit_postcode_response.rb
+++ b/app/models/mapit_postcode_response.rb
@@ -17,6 +17,10 @@ class MapitPostcodeResponse
     location["country_name"]
   end
 
+  def area_type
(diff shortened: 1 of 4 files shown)


JonathanHallam (approved):
👍

PR #1950, opened 2020-10-12
Display local restriction results

leenagupte:
Trello: https://trello.com/c/g3eVg2yg

# What?
Display the results of the postcode lookup.
The content will change depending on whether the area:

- Has local restrictions
 - Level: Medium (this will be the lowest level of restrictions and therefore the default)
 - Level: High
 - Level: Very High
- Is in Scotland
- Is in Wales
- Is in Northern Ireland
- there is no information or no result (this shouldn't really happen, but is like the error page) 

# Expected changes

Screenshot 2020-10-12 at 15 42 50
Screenshot 2020-10-12 at 15 44 40
Screenshot 2020-10-12 at 15 44 23
Screenshot 2020-10-12 at 15 43 56
Screenshot 2020-10-12 at 15 43 24
Screenshot 2020-10-12 at 15 43 14
Screenshot 2020-10-12 at 15 42 59

### DIFF
--- a/app/views/coronavirus_local_restrictions/_devolved_nation.html.erb
+++ b/app/views/coronavirus_local_restrictions/_devolved_nation.html.erb
@@ -1,0 +1,19 @@
+<div class="covid-local-restrictions__devolved-nation-guidance">
+  <%= render "govuk_publishing_components/components/title", {
+    title: t("coronavirus_local_restrictions.results.devolved_nations.heading")
+  } %>
+  <% if @location_lookup.data.first.scotland? %>
+    <p class="govuk-body">
+      <%= sanitize(t("coronavirus_local_restrictions.results.devolved_nations.scotland.guidance")) %>
+    </p>
+  <% elsif @location_lookup.data.first.northern_ireland? %>
+    <p class="govuk-body">
+      <%= sanitize(t("coronavirus_local_restrictions.results.devolved_nations.northern_ireland.guidance")) %>
+    </p>
+  <% elsif @location_lookup.data.first.wales? %>
+    <p class="govuk-body">
+      <%= sanitize(t("coronavirus_local_restrictions.results.devolved_nations.wales.guidance")) %>
+    </p>
+  <% end %>
+  <%= render partial: "coronavirus_local_restrictions/travel_guidance", locals: { travel_heading: t("coronavirus_local_restrictions.results.devolved_nations.travel_heading") } %>
+</div>
--- a/app/views/coronavirus_local_restrictions/_tier_two_restrictions.html.erb
+++ b/app/views/coronavirus_local_restrictions/_tier_two_restrictions.html.erb
@@ -1,0 +1,15 @@
+<div class="covid-local-restrictions__local-alert-guidance">
+  <%= render "govuk_publishing_components/components/title", {
+    title: t("coronavirus_local_restrictions.results.level_two.heading")
+  } %>
+  <p class="govuk-body">
+    <%= t("coronavirus_local_restrictions.results.level_two.match") %> <strong><%= @postcode %></strong> to <strong><%= @restriction.area_name %></strong>.
+  </p>
+  <p class="govuk-body">
+    <%= t("coronavirus_local_restrictions.results.level_two.alert_level") %>
+  </p>
+  <p class="govuk-body">
+    <%= sanitize(t("coronavirus_local_restrictions.results.level_two.guidance")) %>
+  </p>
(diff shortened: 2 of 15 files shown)


app/views/coronavirus_local_restrictions/_devolved_nation.html.erb
```diff
@@ -0,0 +1,19 @@
+<div class="covid-local-restrictions__devolved-nation-guidance">
```
andysellick, line 1: Each of these templates has a class at the top that doesn't appear anywhere in this PR - are these classes real? If not I'd suggest removing them.

app/views/coronavirus_local_restrictions/show.html.erb
```diff
@@ -1,22 +1,33 @@
...
 
-      <%= sanitize(t("coronavirus_local_restrictions.lookup.body_content")) %>
+    <%= render "govuk_publishing_components/components/inset_text", text: t('coronavirus_local_restrictions.lookup.inset_text') %>
 
-      <%= render "shared/postcode-lookup" %>
-    </div>
+    <%= sanitize(t("coronavirus_local_restrictions.lookup.body_content")) %>
+    <%= form_for :postcode_lookup, :url => find_coronavirus_local_restrictions_path do |f| %>
+      <%= render "shared/postcode-lookup", {
+        error_message: error_message,
```
andysellick, line 27: Is this level of error message necessary? There's only one field in this form?

Screenshot 2020-10-12 at 17 10 29

leenagupte: Page-level one is supposed to be at the top of the page, not directly above the field. We're aware it's in the wrong place and is on our list of things to fix after go live (this evening! 😱 )

app/views/coronavirus_local_restrictions/_devolved_nation.html.erb
leenagupte: @andysellick Should we keep the `div`?

andysellick: It's probably not doing anything, because by default it has no margin or padding, so probably safe to remove? Try it first, obvs.

leenagupte: We backend devs are a little bit nervous as the pages are working/look as expected at the moment. Will ask @DilwoarH when he gets back.

andysellick: Fair enough. I'm pretty confident it'll be fine - although it's absolutely fine to leave them in for now, could look at it later, in no way a blocking issue.

leenagupte: Thanks! We're at the "if it ain't broke, don't fix it", part of the day!

app/views/coronavirus_local_restrictions/_tier_two_restrictions.html.erb
```diff
@@ -0,0 +1,15 @@
+<div class="covid-local-restrictions__local-alert-guidance">
+  <%= render "govuk_publishing_components/components/title", {
+    title: t("coronavirus_local_restrictions.results.level_two.heading", level: @restriction.alert_level)
+  } %>
+  <p class="govuk-body">
+    <%= t("coronavirus_local_restrictions.results.level_two.match") %> <strong><%= @postcode %></strong> to <strong><%= @restriction.area_name %></strong>.
+  </p>
+  <p class="govuk-body">
+    <%= t("coronavirus_local_restrictions.results.level_two.alert_level", level: @restriction.alert_level) %>
```
Rosa-Fox, line 9: we don't need to pass this variable in any more

app/views/coronavirus_local_restrictions/_tier_two_restrictions.html.erb
```diff
@@ -0,0 +1,15 @@
+<div class="covid-local-restrictions__local-alert-guidance">
+  <%= render "govuk_publishing_components/components/title", {
+    title: t("coronavirus_local_restrictions.results.level_two.heading", level: @restriction.alert_level)
```
Rosa-Fox, line 3: we don't need to pass this variable in any more

Rosa-Fox (approved):

app/views/coronavirus_local_restrictions/_devolved_nation.html.erb
DilwoarH: This can be removed - I had them in there in case we needed some specific styling. We can do this post-release.

PR #1959, opened 2020-10-13
Finish one AB test and start the next

sihugh:
We're finishing TransitionUrgency2 and keeping the changes we made.

In the next test, we're changing the display of the "things you can do now" links to play them down a bit. It's a tricky balance: making sure people know they're there, but also doing the checker as a higher priority.

This replaces the bucket section (which is quite large, bold and contains multiple sections) with a simple list of links.

This test is English only, so the Welsh page should be unaffected.

## A variant (partial screenshot of the section as it's very long)

Screenshot 2020-10-13 at 12 21 43

## B variant (entire section)

Screenshot 2020-10-13 at 12 20 54

https://trello.com/c/3XR9BeaI/511-ab-test-for-urgency-phase-3

### DIFF
--- a/app/controllers/transition_landing_page_controller.rb
+++ b/app/controllers/transition_landing_page_controller.rb
@@ -48,7 +48,7 @@ private
   def ab_test_variant
     @ab_test_variant ||= begin
       ab_test = GovukAbTesting::AbTest.new(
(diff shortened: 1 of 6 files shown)


hannako (approved):
😎

PR #1961, opened 2020-10-13
Tweaks to description on the coronavirus lookup page

sihugh:
## Include standard meta description for SEO

The og:description field is mainly used in social media for sharing links. We should add the description tag too, as this is more commonly used by search engines.

## Update meta description

This will match [the description field in the content item](alphagov/special-route-publisher#123).

The main reason for this change is to include the word "checker" in the text, as we've noted that this is commonly used by people searching for this page. (Suggested by Tara and Hannah)


leenagupte (approved):
Looks good! Thanks Simon! 🌟

PR #1967, opened 2020-10-14
Update LCR link to LCR specific guidance

dracos:
The Liverpool City Region has additional guidance on top of the base very high guidance. The base guidance does not link to the LCR page, but the LCR page does link to the base page, so I think it makes sense to link to the specific page first.

### DIFF
--- a/config/locales/en/lookup.yml
+++ b/config/locales/en/lookup.yml
@@ -52,7 +52,7 @@ en:
       level_three:
         heading: "Local COVID Alert Level: Very High"
         alert_level: "This area is in Local COVID Alert Level: Very High."
-        guidance: <a class="govuk-link" href="/guidance/local-covid-alert-level-very-high">What you can and cannot do.</a>
+        guidance: <a class="govuk-link" href="/guidance/local-covid-alert-level-very-high-liverpool-city-region">What you can and cannot do.</a>
         match: "We've matched the postcode"
       devolved_nations:
         heading: There might be restrictions in this area


dracos:
I'm not sure if these links are actually used; as LCR is the only very high region at present, you could also change the link in https://github.com/alphagov/collections/blob/929ca08d377956ea7d64d997c426c13c34d3a493/config/locales/en/lookup.yml#L43 as an alternative, if that's the link being shown at present.

dracos:
Yes, it looks like you'd need to use `@restriction.guidance.link` in the tier three template (at the least) rather than the lookup.yml string used at present for the change in this PR to do anything. Happy to adjust to either change lookup.yml or change template to use the link from the data, once someone confirms which would be the better approach (or if you've not already done it).

sihugh:
👋  Thanks @dracos, I've flagged this with the team.

dracos:
Have rebased on current master, and adjusted the lookup.yml link instead, as the other links have been removed. This would work until more areas went into VH tier, and so would presumably provide more accurate guidance right now than the current result.
Another solution if this is not wanted would be to have a note on the baseline VH guidance linking to the Liverpool guidance, that would perhaps be easier to implement?

dracos:
Any news? A postcode lookup in Liverpool still does not return any Liverpool specific guidance, it still links to https://www.gov.uk/guidance/local-covid-alert-level-very-high without mentioning https://www.gov.uk/guidance/local-covid-alert-level-very-high-liverpool-city-region even though https://www.gov.uk/guidance/local-covid-alert-levels-what-you-need-to-know#local-covid-alert-level-very-high literally says "You should therefore check whether additional restrictions apply in your area" linking to the postcode lookup.

leenagupte:
Hi @dracos! Thanks for your suggestion.

Currently the guidance for each level (medium, high, very high) links to a single page that then links to the extra restrictions for an area.
As more areas move into very high restrictions, we’re looking into to the best way to show the extra restrictions specific to an area whilst keeping the user journey as simple as possible. At the moment we’re looking at how better to do this in the content for each level, rather than in the postcode checker itself.
As part of this work we've looked at the guidance for the “extra” restrictions, which shows it is not as useful as the very high guidance, so for now we don't think it's the best journey to link to that through the checker. But we will continue to monitor feedback.

dracos:
Thanks for the reply. Think this is out of date now anyway.

PR #1970, opened 2020-10-14
Bug fix: POST route throws error when postcode is nil

leenagupte:
Trello: https://trello.com/c/9DI7SFML

# What?

A 500 error is being thrown if the "results" controller method called is directly without any params. This is because we are not checking if `postcode` is nil before stripping the whitespace.  We should instead be redirecting users back to the show page.

Error in sentry:
https://sentry.io/organizations/govuk/issues/1954271093/?project=202213&query=is%3Aunresolved


Rosa-Fox (approved):
Looks good! I set `params["postcode-lookup"] = ""` at the start of the #results controller action to manually test it.

PR #1973, opened 2020-10-14
Add events to local restrictions

DilwoarH:
## What

We track certain things on pages on GOV.UK but there are certain elements  which won't be tracked unless we set it up.

We use Google Analytics across GOV.UK so we should use the same. 

https://www.gov.uk/find-local-council use events with naming convention so let's copy this.

*Proposal attached but this could change depending on how final design or results page looks*

We should probably capture:

- When the postcode checker is started
- Whether an error message was shown 
- What results are shown to the user: local level alert and local area (?)

Less important but may be something to consider

- What links user clicks on in text 
- Whether user clicks 'check another postcode'

## Why

Without extra tracking we won't know
- How many users saw an error message
- What results are shown to users (Local council and alert level)
- How many users engaged with links on the results page

*Caveat: since December 2019 we do not count everyone that comes to GOV.UK with Google Analytics. We only count users if they have consented to it. That means any figures that we collect is not representative to an unknown degree*

https://trello.com/c/UlXBmBcR/848-add-ga-tracking-to-local-postcode-lookup

## How to test

1. Go to https://www.gov.uk/find-coronavirus-local-restrictions
2. Open console (with OmniBug or something similar to track analytics)
3. Submit the form without any input to test error event
4. Submit the form with a valid postcode to test result event

----

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -18,6 +18,7 @@
 //= require components/accordion
 //= require modules/coronavirus-landing-page
 //= require modules/coronavirus-track-local-restrictions
+//= require modules/coronavirus-track-local-restriction-results
 //= require modules/track-timeline-links
 
 $(document).on('ready', function(){
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -17,6 +17,7 @@
 //= require modules/toggle-attribute
 //= require components/accordion
 //= require modules/coronavirus-landing-page
+//= require modules/coronavirus-track-local-restrictions
 //= require modules/track-timeline-links
 
 $(document).on('ready', function(){
--- a/app/views/coronavirus_local_restrictions/_devolved_nation.html.erb
+++ b/app/views/coronavirus_local_restrictions/_devolved_nation.html.erb
@@ -1,18 +1,23 @@
-<%= render "govuk_publishing_components/components/title", {
-  title: t("coronavirus_local_restrictions.results.devolved_nations.heading")
-} %>
-<% if @location_lookup.data.first.scotland? %>
-  <p class="govuk-body">
-    <%= sanitize(t("coronavirus_local_restrictions.results.devolved_nations.scotland.guidance")) %>
-  </p>
-<% elsif @location_lookup.data.first.northern_ireland? %>
-  <p class="govuk-body">
-    <%= sanitize(t("coronavirus_local_restrictions.results.devolved_nations.northern_ireland.guidance")) %>
-  </p>
-<% elsif @location_lookup.data.first.wales? %>
-  <p class="govuk-body">
-    <%= sanitize(t("coronavirus_local_restrictions.results.devolved_nations.wales.guidance")) %>
-  </p>
-<% end %>
+<%= tag.div :data => {
+  :module => "coronavirus-track-local-restriction-results",
+  "tracking-label" => "Devolved nation: #{@location_lookup.data.first.country}"
+} do %>
+  <%= render "govuk_publishing_components/components/title", {
+    title: t("coronavirus_local_restrictions.results.devolved_nations.heading")
+  } %>
+  <% if @location_lookup.data.first.scotland? %>
+    <p class="govuk-body">
+      <%= sanitize(t("coronavirus_local_restrictions.results.devolved_nations.scotland.guidance")) %>
+    </p>
+  <% elsif @location_lookup.data.first.northern_ireland? %>
+    <p class="govuk-body">
+      <%= sanitize(t("coronavirus_local_restrictions.results.devolved_nations.northern_ireland.guidance")) %>
+    </p>
+  <% elsif @location_lookup.data.first.wales? %>
+    <p class="govuk-body">
+      <%= sanitize(t("coronavirus_local_restrictions.results.devolved_nations.wales.guidance")) %>
+    </p>
+  <% end %>
 
-<%= render partial: "coronavirus_local_restrictions/travel_guidance", locals: { travel_heading: t("coronavirus_local_restrictions.results.devolved_nations.travel_heading") } %>
+  <%= render partial: "coronavirus_local_restrictions/travel_guidance", locals: { travel_heading: t("coronavirus_local_restrictions.results.devolved_nations.travel_heading") } %>
+<% end %>
--- a/spec/javascripts/modules/track-timeline-links_spec.js
+++ b/spec/javascripts/modules/track-timeline-links_spec.js
@@ -23,6 +23,10 @@ describe('Track timeline links', function () {
   var element;
   
   beforeEach(function () {
+    GOVUK.analytics = {
+      trackEvent: function () {
+      }
+    }
     spyOn(GOVUK.analytics, 'trackEvent')
 
     element = document.createElement('div');
--- a/app/assets/javascripts/modules/coronavirus-track-local-restriction-results.js
+++ b/app/assets/javascripts/modules/coronavirus-track-local-restriction-results.js
@@ -1,0 +1,23 @@
+window.GOVUK = window.GOVUK || {}
+window.GOVUK.Modules = window.GOVUK.Modules || {};
+
+(function (Modules) {
+  function CoronavirusTrackLocalRestrictionResults () {}
+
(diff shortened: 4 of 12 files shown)


DilwoarH:
Haven't added tests yet.

Rosa-Fox:
Hi @DilwoarH,

It works!

results-tracking
devolved-nation
postcode-with-errors

="tier-1" src="https://user-images.githubusercontent.com/5963488/96138133-5e7da100-0ef5-11eb-9e51-7c91d79efbfe.png">
tier2 3

I only had one tiny comment about naming of a test. Might be good for @alex-ju or someone with better JS knowledge than me to give the code a look over incase there are ways to improve it... but hopefully the screen shots ^ same them a little bit of time!

Guessing the 'What links user clicks on in text' and 'Whether user clicks 'check another postcode'' will be a future PR?

spec/javascripts/modules/coronavirus-local-restrictions_spec.js
```diff
@@ -0,0 +1,80 @@
...
+      element.querySelector('form').dispatchEvent(new Event('submit'))
+
+      expect(GOVUK.analytics.trackEvent).toHaveBeenCalledWith(
+        'postcodeSearch:local_lockdown', 'postcodeSearchStarted', { transport: 'beacon' }
+      )
+    })
+  })
+
+
+  describe('landing page without errors', function () {
```
Rosa-Fox, line 42: landing page with errors

Rosa-Fox (commented):
Another thing is that this ran really slowly... I think this was more likely due to my local setup/calling mapit, but we should test on integration.

spec/javascripts/modules/coronavirus-local-restrictions_spec.js
DilwoarH: > Another thing is that this ran really slowly... I think this was more likely due to my local setup/calling mapit, but we should test on integration.

That's weird - it should trigger mapit

app/assets/javascripts/application.js
```diff
@@ -17,6 +17,8 @@
 //= require modules/toggle-attribute
 //= require components/accordion
 //= require modules/coronavirus-landing-page
+//= require modules/coronavirus-local-restrictions
+//= require modules/coronavirus-local-restriction-results
```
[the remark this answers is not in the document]
DilwoarH: both get loaded unfortunately

app/views/coronavirus_local_restrictions/_devolved_nation.html.erb
```diff
@@ -1,20 +1,25 @@
...
-  </p>
-<% elsif @location_lookup.data.first.wales? %>
-  <p class="govuk-body">
-    <%= sanitize(t("coronavirus_local_restrictions.results.devolved_nations.wales.guidance")) %>
-  </p>
-<% end %>
+<%= tag.div :data => {
+  :module => "coronavirus-local-restriction-results",
+  "tracking-label" => "Devolved nation: #{@location_lookup.data.first.country}"
+} do %>
```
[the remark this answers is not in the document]
DilwoarH: Do you mean it should be written as HTML?

app/assets/javascripts/modules/coronavirus-local-restrictions.js
```diff
@@ -0,0 +1,37 @@
...
+        var options = {
+          transport: "beacon"
+        }
+
+        GOVUK.analytics.trackEvent("postcodeSearch:local_lockdown", "postcodeSearchStarted", options)
+      })
+  }
+
+  CoronavirusLocalRestrictions.prototype.addErrorMessageEvent = function (element) {
+    var errorMessage = element.querySelector('.govuk-error-summary__body')
```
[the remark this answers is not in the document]
DilwoarH: > Forgive me, looking at this for the first time with very little context - jQuery is loaded by we're using ES6?


We're moving away from jQuery and therefore anything new we write, we try to avoid adding new jQuery in.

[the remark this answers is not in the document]
DilwoarH: > I'm also wondering if data- vs class is a better hook - eg what if the classes change?

Yes you have a point, I will look into whether this is possible with the component as it is

spec/javascripts/modules/track-timeline-links_spec.js
```diff
@@ -23,6 +23,10 @@ describe('Track timeline links', function () {
   var element;
   
   beforeEach(function () {
+    GOVUK.analytics = {
```
leenagupte, line 26: Thank you! ⭐

app/assets/javascripts/modules/coronavirus-local-restrictions.js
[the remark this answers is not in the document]
DilwoarH: Changed.

app/assets/javascripts/modules/coronavirus-local-restriction-results.js
```diff
@@ -0,0 +1,23 @@
+window.GOVUK = window.GOVUK || {}
+window.GOVUK.Modules = window.GOVUK.Modules || {};
+
+(function (Modules) {
+  function CoronavirusLocalRestrictionResults () {}
```
alex-ju, line 5: Since this is only handling event tracking, I would suggest we name it something like `CoronavirusTrackLocalRestrictionResults` for future us to remember that there are no UI enhancement on results and we're only setting up tracking.

app/assets/javascripts/modules/coronavirus-local-restrictions.js
```diff
@@ -0,0 +1,37 @@
+window.GOVUK = window.GOVUK || {}
+window.GOVUK.Modules = window.GOVUK.Modules || {};
+
+(function (Modules) {
+  function CoronavirusLocalRestrictions () {}
```
alex-ju, line 5: Similarly here, I suggest `CoronavirusTrackLocalRestrictions` for the reason mentioned above.

app/assets/javascripts/modules/coronavirus-local-restrictions.js
```diff
@@ -0,0 +1,37 @@
...
+
+  CoronavirusLocalRestrictions.prototype.start = function (element) {
+    this.addPostcodeFormSubmitListener(element[0])
+    this.addErrorMessageEvent(element[0])
+  }
+
+  CoronavirusLocalRestrictions.prototype.addPostcodeFormSubmitListener = function (element) {
+    element.querySelector('.coronavirus-local-restrictions__postcode-form')
```
alex-ju, line 13: Ideally, we would've attached the module script right on the form element without needing to query this element, but I do get that we're going it at a higher level so we can grab the error message too.

alex-ju (approved):
Code looks correct! I added two non-blocking comments around making it more obvious that these two scripts are dealing with tracking.

This whole piece of work reminds me of a wider need for GOV.UK to have a proper analytics framework with consistent ways of tracking across applications. A piece of work that I remember you did investigate in the past in Digital Marketplace at least. It may be worth raising a discussion around this and with some knowledge on the legacy stuff, maybe from Andy, we'd be able to reduce the work required for analytics.

app/assets/javascripts/modules/coronavirus-local-restrictions.js
DilwoarH: I did this initially then realised that had to add in multiple events so changed it to page level

app/assets/javascripts/modules/coronavirus-local-restrictions.js
DilwoarH: Done

app/assets/javascripts/modules/coronavirus-local-restriction-results.js
DilwoarH: Done

DilwoarH:
> Code looks correct! I added two non-blocking comments around making it more obvious that these two scripts are dealing with tracking only.
> 
> This whole piece of work reminds me of a wider need for GOV.UK to have a proper analytics framework with consistent ways of tracking across applications. A piece of work that I remember you did investigate in the past in Digital Marketplace at least. It may be worth raising a discussion around this and with some knowledge on the legacy stuff, maybe from Andy, we'd be able to reduce the work required for analytics.

Definitely agree with the wider stuff - We should investigate

PR #1977, opened 2020-10-15
Remove redundant fields from the yaml file

JonathanHallam:
We added these fields as a preemptive measure. It now looks like we will have to format it differently. This removes all of the things from the yaml file, model and tests that we don't need.

### DIFF
--- a/app/models/local_restriction.rb
+++ b/app/models/local_restriction.rb
@@ -24,20 +24,4 @@ class LocalRestriction
   def alert_level
     restriction["alert_level"]
   end
-
-  def guidance
-    restriction["guidance"]
-  end
-
-  def extra_restrictions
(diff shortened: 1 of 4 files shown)


bevanloon:
Would start and end date not be helpful if we wanted to automate some results based on date? I'm thinking, for example, if something comes into force at midnight that requires a change, having the start and end date would mean we could pre-megre the data and not have someone do a midnight release (with the appropriate logic of course)?

JonathanHallam:
Hi @bevanloon that's what we're doing next! We've had to think a bit about how this file will be structured and it will make the logic a bit easier if this is nested slightly differently. I promise to bring start date back in another form :D

Rosa-Fox (approved):
👍

PR #1982, opened 2020-10-16
Update Essex line

dracos:
The word "Essex" on its own is unclear, so explicitly state it is the county council area, and not the ceremonial county.


JonathanHallam:
Hi @dracos, we're currently looking into how we display this information. Thanks

dracos:
I see this text has been removed.

PR #1983, opened 2020-10-16
Transition landing page header refresh

DilwoarH:
## What

New transition landing page header design to be deploy Friday 16th by 5pm. Both English and Welsh pages are updated. 

[Welsh content](https://docs.google.com/document/d/13PIusFYUfghsGAJLmWS6xfzm0hyMA9JqCihGrhYTkWU/edit#)

Trello: https://trello.com/c/WM8yQVMT/519-updated-landing-page-campaign-refresh-and-phase-4-testing

## Before

Screenshot 2020-10-16 at 16 13 52

## After

Screenshot 2020-10-16 at 16 13 16

Adds in new header styling, countdown flip component and new background image


app/lib/countdown_clock.rb
```diff
@@ -0,0 +1,21 @@
+class CountdownClock
```
DilwoarH, line 1: @hannako - once your stuff is merged - I'll delete this from the branch

app/views/transition_landing_page/_page_header.html.erb
```diff
@@ -1,3 +1,7 @@
+<%
+  clock = CountdownClock.new
```
DilwoarH, line 2: to be removed

DilwoarH:
@hannako - I've removed the clock code - when you merge your stuff - you will need to pass this into the view

danacotoran:
Is there any particular reason why the Welsh version does not have the same red highlight on the header as the English translation? 

**En: has highlight**
Screenshot 2020-10-16 at 16 24 44

**Cy: no highlight**
Screenshot 2020-10-16 at 16 23 53

hannako:
> Is there any particular reason why the Welsh version does not have the same red highlight on the header as the English translation?
> 
> **En: has highlight**
> Screenshot 2020-10-16 at 16 24 44
> 
> **Cy: no highlight**
> Screenshot 2020-10-16 at 16 23 53

Yes - I left it on purpose because the translation file didn't specify which bits to highlight. Wasn't sure if I should risk a google translate! Am just investigating what the opinion is on that...

app/assets/stylesheets/views/_transition-landing-page.scss
```diff
@@ -1,6 +1,7 @@
 
 $red: #FF003B;
 $dark-blue: #0055D2;
+$dark-blue-2: #1E1348;
```
alex-ju, line 4: Can we update the value of `$dark-blue` instead of creating `$dark-blue-2`, it seems like `$dark-blue` is not being used anywhere anymore.

config/locales/en/transition_landing_page.yml
```diff
@@ -2,10 +2,12 @@ en:
   transition_landing_page:
     meta_title: The UK transition
     meta_description: The UK's transition period after Brexit comes to an end this year. Check the new rules from January 2021 and take action now.
-    page_header: The transition period ends in December
+    page_header: "The UK transition:<br/>time is running out"
     page_header_explainer: |
-      The UK has left the EU, and the transition period after Brexit comes to an end this year. Check the new rules from January 2021 and take action now.
-    take_action_title: Make sure you're ready
+      <span class="landing-page__intro-highlight">There are new rules</span> for businesses and citizens from 1 January 2021.
```
injms, line 7: This should be a `strong` tag, since it's being highlighted.
```suggestion
      <strong class="landing-page__intro-highlight">There are new rules</strong> for businesses and citizens from 1 January 2021.
```

danacotoran:
> > Is there any particular reason why the Welsh version does not have the same red highlight on the header as the English translation?
> > **En: has highlight**
> > Screenshot 2020-10-16 at 16 24 44
> > **Cy: no highlight**
> > Screenshot 2020-10-16 at 16 23 53
> 
> Yes - I left it on purpose because the translation file didn't specify which bits to highlight. Wasn't sure if I should risk a google translate! Am just investigating what the opinion is on that...

@hannako ahhh fair enough yeah I totally get that then. Wish we had Welsh speakers around to help us out with stuff like this 😬

I would have thought that **Bydd rheolau newydd** would be highlighted.
Based on the almighty Google: 
"Bydd rheolau newydd" = "There will be new rules"
"Bydd rheolau newydd i fusnesau" = "There will be new rules for businesses"

But I understand not wanting to risk it if we don't know for sure. It's just a detail I noticed.

config/locales/en/transition_landing_page.yml
alex-ju, line 7: Ideally, we would've had this content split into two locale items, so we don't need to hardcode markup in here (especially with styles attached) and we can use a better tag for semantics (such as `<mark>` or `<strong>`)

alex-ju (approved):
I left a few non-blocking comments that can be addressed after release, otherwise it's good to go.

config/locales/en/transition_landing_page.yml
hannako: Thanks

hannako: Thanks, I'll tidy that up in a subsequent PR. 👍

danacotoran:
I have two other Welsh related visual issues. Both are to do with text wrapping in an unsightly way on smaller adaptions.

Screenshot 2020-10-16 at 17 02 17

Screenshot 2020-10-16 at 17 02 00

Not sure these are priority right now considering the very tight timeline we're dealing with here. But they should be recorded and dealt with.

hannako:
@danacotoran Thanks Dana! Will make a card.

bevanloon:
The calculation for `today_in_london` [specifies the time zone as UTC](https://github.com/alphagov/collections/blob/master/app/lib/countdown_clock.rb#L27). Does that mean for the hour between 00:00 and 01:00 each morning while we are still in BST, our calculation will be out by 1? During that hour, UTC will still be the previous day.

The COVID services team bumped into the UTC / BST problem during their recent work and ended up [coercing the time zone to BST](https://github.com/alphagov/collections/blob/b6ed4a9b6e185884295e15f91d594f5b62b051eb/app/models/local_restriction.rb#L36) using an [instance variable](https://github.com/alphagov/collections/blob/b6ed4a9b6e185884295e15f91d594f5b62b051eb/app/models/local_restriction.rb#L6).

I'm not sure whether or not anything else in Collections relies on the time zone being UTC, but might it be a good idea for the teams to have a chat and consider setting the time zone in `config/application` like we do in  [Whitehall](https://github.com/alphagov/whitehall/blob/613f1fc5507e7ee2540329896a8dcaa5c48c8a69/config/application.rb#L28)? That would allow us to rely on `Time.zone.now` to do the right thing.

.cc @Rosa-Fox @leenagupte

leenagupte:
> The calculation for `today_in_london` [specifies the time zone as UTC](https://github.com/alphagov/collections/blob/master/app/lib/countdown_clock.rb#L27). Does that mean for the hour between 00:00 and 01:00 each morning while we are still in BST, our calculation will be out by 1? During that hour, UTC will still be the previous day.
> 
> The COVID services team bumped into the UTC / BST problem during their recent work and ended up [coercing the time zone to BST](https://github.com/alphagov/collections/blob/b6ed4a9b6e185884295e15f91d594f5b62b051eb/app/models/local_restriction.rb#L36) using an [instance variable](https://github.com/alphagov/collections/blob/b6ed4a9b6e185884295e15f91d594f5b62b051eb/app/models/local_restriction.rb#L6).
> 
> I'm not sure whether or not anything else in Collections relies on the time zone being UTC, but might it be a good idea for the teams to have a chat and consider setting the time zone in `config/application` like we do in [Whitehall](https://github.com/alphagov/whitehall/blob/613f1fc5507e7ee2540329896a8dcaa5c48c8a69/config/application.rb#L28)? That would allow us to rely on `Time.zone.now` to do the right thing.
> 
> .cc @Rosa-Fox @leenagupte

I tried setting the time zone in `config/application` on Friday night, but it broken some (unrelated to the postcode checker) atom feed and document presenter tests that I didn't have time to fix, hence the instance variable.

hannako:
Thanks @bevanloon. I'll address this in a fresh PR now. 👍

hannako:
@bevanloon https://github.com/alphagov/collections/pull/1997

PR #1994, opened 2020-10-19
Use My Society's Mapit when running locally

leenagupte:
GOV.UK hosts it's own version of Mapit. Unfortunately, that is only accessibly via the production
and test environments, not locally. Until we can access Mapit locally, add a work around to make it
possibly to manually test changes.

### DIFF
--- a/app/services/location_lookup_service.rb
+++ b/app/services/location_lookup_service.rb
@@ -49,7 +49,7 @@ private
 
   def response
     @response ||= begin
-                    JSON.parse(GdsApi.mapit.location_for_postcode(postcode).to_json)
+                    JSON.parse(mapit.location_for_postcode(postcode).to_json)
                   rescue GdsApi::HTTPNotFound, GdsApi::HTTPClientError => e
                     {
                       error: {
@@ -58,5 +58,11 @@ private
                       },
                     }
                   end
+  end
+
+  def mapit
+    return GdsApi.mapit unless Rails.env.development?
+
+    GdsApi::Mapit.new("https://mapit.mysociety.org/")
   end
 end


app/services/location_lookup_service.rb
```diff
@@ -59,4 +59,10 @@ def response
                     }
                   end
   end
+
+  def mapit
+    return GdsApi.mapit unless Rails.env.development?
+
+    GdsApi::Mapit.new("https://mapit.mysociety.org/")
```
DilwoarH, line 66: Is it possible to do this on a global level? ie. inside of the GDS API adapter?

leenagupte: I kept it like that so it would still throw the same GdsApi errors, i.e. GdsApi::HTTPNotFound and GdsApi::HTTPClientError, otherwise [that bit of the code](https://github.com/alphagov/collections/blob/master/app/services/location_lookup_service.rb#L53) will stop working.

DilwoarH: Correct me if I'm wrong but wouldn't it still keep these errors as we'll be instantiating with the same class?

Anyways, I think, for now, we can localise the change until we have something better.

DilwoarH (approved):
Not a blocker but  should we make this an environment variable?

app/services/location_lookup_service.rb
leenagupte: Oh! sorry! I mis-read! You meant, put this inside of the gem! 🤦 Yes, it should be possible to do that, but I only envisioned this as a stop-gap. There's a bigger problem around how do we access Mapit data locally.

leenagupte:
> Not a blocker but should we make this an environment variable?

Could do, but I'm not convinced it's worth the effort at the moment.

PR #1995, opened 2020-10-19
Bug Fix: Remove special characters from postcode before calling Mapit

leenagupte:
Trello: https://trello.com/c/Sw0IjxOP

# What?

Sanitise the postcode entered by the user to remove all special characters.

# Why?
If you add a "." to the beginning of the postcode, e.g. ".CH89RS" the lookup throws a "500" error. This is not being caught be the postcode validation regex,  although "." by itself is being caught correctly.

https://sentry.io/organizations/govuk/issues/1954296277/?project=202213&referrer=slack

### DIFF
--- a/app/controllers/coronavirus_local_restrictions_controller.rb
+++ b/app/controllers/coronavirus_local_restrictions_controller.rb
@@ -20,7 +20,7 @@ class CoronavirusLocalRestrictionsController < ApplicationController
 
     if @postcode.blank?
       return render_no_postcode_error
-    elsif !postcode_validation
+    elsif !PostcodeService.new(@postcode).valid?
       return render_invalid_postcode_error
     end
 
@@ -86,10 +86,6 @@ private
         url: "/coronavirus",
       },
     ]
-  end
-
-  def postcode_validation
-    @postcode =~ /^([Gg][Ii][Rr] 0[Aa]{2})|((([A-Za-z][0-9]{1,2})|(([A-Za-z][A-Ha-hJ-Yj-y][0-9]{1,2})|(([A-Za-z][0-9][A-Za-z])|([A-Za-z][A-Ha-hJ-Yj-y][0-9][A-Za-z]?))))\s?[0-9][A-Za-z]{2})$/
   end
 
   def content_item
--- a/app/controllers/coronavirus_local_restrictions_controller.rb
+++ b/app/controllers/coronavirus_local_restrictions_controller.rb
@@ -16,7 +16,7 @@ class CoronavirusLocalRestrictionsController < ApplicationController
       return render_no_postcode_error
     end
 
-    @postcode = params["postcode-lookup"].gsub(/\s+/, "").upcase
+    @postcode = PostcodeService.new(params["postcode-lookup"]).sanitize
 
     if @postcode.blank?
       return render_no_postcode_error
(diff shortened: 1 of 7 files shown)


test/services/postcode_service_test.rb
```diff
@@ -0,0 +1,37 @@
...
+    it "transposes O/0 and I/1 if necessary" do
+      assert_equal "W1A 0AA", described_class.new("WIA OAA").sanitize
+    end
+
+    it "returns nil if the postcode parameter is nil or an empty string" do
+      assert_nil described_class.new("").sanitize
+    end
+
+    it "removes special characters from the start of the postcode" do
+      assert_equal "WC2B 6NH", described_class.new(".WC2B6NH").sanitize
```
sihugh, line 30: I wonder if it's worth testing specifically for the non-breaking space that causes https://sentry.io/organizations/govuk/issues/1954296277/?project=202213&query=is%3Aunresolved? 

(Calls `<MAPIT>/postcode/%C2%A0WC1A1AAjson`)

leenagupte: Interestingly, "%C2%A0WC1A1AA" is sanitized as "C2A0WC1A1AA" 😅 , though I don't think that's quite the same as the value the user is typing in.

sihugh: Sorry 😬

sihugh: Looks great otherwise :-D

leenagupte: That's ok, I found the right character! `\u00A0` I'll add a test for it.

leenagupte:
@sihugh Thanks for your review ❤️. I've added a test for non-breaking spaces and fixed up.

app/controllers/coronavirus_local_restrictions_controller.rb
```diff
@@ -16,11 +16,11 @@ def results
       return render_no_postcode_error
     end
 
-    @postcode = params["postcode-lookup"].gsub(/\s+/, "").upcase
+    @postcode = PostcodeService.new(params["postcode-lookup"]).sanitize
 
     if @postcode.blank?
       return render_no_postcode_error
-    elsif !postcode_validation
+    elsif !PostcodeService.new(@postcode).valid?
```
DilwoarH, line 23: Hey according to the library (https://github.com/threedaymonk/uk_postcode) partial postcodes are valid, is this fine?

DilwoarH: ignore - I can see you're using the same regex.

Should we be testing for partial postcodes?

sihugh: It might be a nice route to handle, but OK to defer for now? On local transactions I think we tracked the different sorts of errors, like valid postcode but no location associated with it, partial postcode, not a postcode, those sorts of things.

sihugh (approved):
Works on my machine 🚀

app/controllers/coronavirus_local_restrictions_controller.rb
leenagupte: I can add a test to check partial postcodes fail validation.

leenagupte:
@sihugh @DilwoarH Thanks for the reviews! I've added a test for a partial postcode and rebased against master.

sihugh (approved):

PR #1996, opened 2020-10-19
New video link and updated transcripts

sihugh:
There is a new transition promo video, and we have updated the transcripts
accordingly.

Affects:
- www.gov.uk/transition
- www.gov.uk/transition.cy


hannako (approved):
Looks good to me!

PR #1997, opened 2020-10-19
Fix time zone to London

hannako:
## What

- Set config.time_zone for collections to London
- Update some tests 
- Refactor the Transition Countdown Clock to use local time, rather than UTC.

## Context

- If no time_zone is set, Rails uses UTC by default. This proved problematic in a recent change to the postcode checker, which had changes that were to be made [live at a specific time](https://github.com/alphagov/collections/blob/master/lib/local_restrictions/local-restrictions.yaml), and because in the UK as we are still in BST, they would have been an hour late. 

## Consequences

- Collections compiles [two atom feeds](https://github.com/alphagov/collections/blob/master/app/controllers/feeds_controller.rb). It creates an ID for each entry in the feed by parsing the entry's public_timestamp. `public_timestamp` == `public_updated_at` [which is set by the by the content store](https://github.com/alphagov/search-api/blob/58986cdaa8c8cc7c3e36198a5b225065b9192d83/lib/govuk_index/presenters/common_fields_presenter.rb#L23) which runs in UTC. To ensure we maintain backwards compatibility, the Feed Entry Presenter needs to sets time zone to UTC before creating the ID. 

- Collections also builds the Organisation pages. The Organisation Document Presenter uses `Date.parse` to present the `public_updated_at` time stamp, but the test was using `Time.zone.parse`. This was passing when the default time zone was UTC, but now Time.zone is set, the two methods return a different string.

---

### DIFF
--- a/app/presenters/feed_entry_presenter.rb
+++ b/app/presenters/feed_entry_presenter.rb
@@ -18,7 +18,11 @@ class FeedEntryPresenter
   end
 
   def updated
-    Time.zone.parse(result["public_timestamp"])
+    # Collections now uses London time. Convert it back to it's former zone
+    # of UTC in the domain of atom feeds to ensure backwards compatibility.
+
+    @utc_time = Time.find_zone("UTC")
+    @utc_time.parse(result["public_timestamp"])
   end
 
   def title
--- a/app/presenters/feed_entry_presenter.rb
+++ b/app/presenters/feed_entry_presenter.rb
@@ -21,8 +21,7 @@ class FeedEntryPresenter
     # Collections now uses London time. Convert it back to it's former zone
     # of UTC in the domain of atom feeds to ensure backwards compatibility.
 
-    @utc_time = Time.find_zone("UTC")
-    @utc_time.parse(result["public_timestamp"])
+    Time.find_zone("UTC").parse(result["public_timestamp"])
   end
 
   def title
--- a/app/lib/countdown_clock.rb
+++ b/app/lib/countdown_clock.rb
@@ -24,6 +24,6 @@ private
   end
 
   def today_in_london
-    Time.zone.now.utc.to_date
+    Time.zone.today
   end
 end
--- a/config/application.rb
+++ b/config/application.rb
@@ -19,6 +19,7 @@ module Collections
   class Application < Rails::Application
     # Initialize configuration defaults for originally generated Rails version.
     config.load_defaults 6.0
+    config.time_zone = "London"
 
     # Application configuration can go into files in config/initializers
     # -- all .rb files in that directory are automatically loaded after loading
--- a/test/integration/organisation_test.rb
+++ b/test/integration/organisation_test.rb
@@ -677,9 +677,9 @@ class OrganisationTest < ActionDispatch::IntegrationTest
     visit "/government/organisations/attorney-generals-office"
     assert page.has_css?("section#latest-documents")
     assert page.has_css?(".gem-c-heading", text: "Latest from the Attorney General's Office")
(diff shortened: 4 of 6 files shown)


pixeltrix (approved):

bevanloon:
@hannako 👍 thanks for raising this.

Sorry to be a pain, but I think we should leave the atom feed in UTC. It's already established in the feed and changing it does break backward compatibility. Also we do have a few other atom feeds in UTC around GOV.UK, so if we aim for consistency that would be good.

hannako:
> @hannako 👍 thanks for raising this.
> 
> Sorry to be a pain, but I think we should leave the atom feed in UTC. It's already established in the feed and changing it does break backward compatibility. Also we do have a few other atom feeds in UTC around GOV.UK, so if we aim for consistency that would be good.

@sihugh just said the same thing in a DM - great minds!! Just amending now @bevanloon  👍

app/presenters/feed_entry_presenter.rb
```diff
@@ -18,7 +18,11 @@ def atom_url
   end
 
   def updated
-    Time.zone.parse(result["public_timestamp"])
+    # Collections now uses London time. Convert it back to it's former zone
+    # of UTC in the domain of atom feeds to ensure backwards compatibility.
+
+    @utc_time = Time.find_zone("UTC")
+    @utc_time.parse(result["public_timestamp"])
```
bevanloon, line 25: do we need instance variables here? could we get away with

```
Time.find_zone("UTC").parse(result["public_timestamp"])
```

hannako: yep! Will change

app/presenters/feed_entry_presenter.rb
```diff
@@ -18,7 +18,10 @@ def atom_url
   end
 
   def updated
-    Time.zone.parse(result["public_timestamp"])
+    # Collections now uses London time. Convert it back to it's former zone
+    # of UTC in the domain of atom feeds to ensure backwards compatibility.
+
+    Time.find_zone("UTC").parse(result["public_timestamp"])
```
bevanloon, line 24: sorry @hannako - I think we need `.iso8601` appended to this to bring it in line with the [iso standards](https://tools.ietf.org/html/rfc3339) for [date constructs](https://tools.ietf.org/html/rfc4287#section-3.3) in atom feeds.

hannako: Thanks for Bevan.  

So I was following [this](https://validator.w3.org/feed/docs/atom.html#aboutThisDocument) which says all atom dates need to be in RFC3339.

We create the ID like so:

```
def id
   "#{url}##{updated.rfc3339}"
end
```

And [on the review app](https://govuk-collec-fix-time-z-aucqvo.herokuapp.com/government/organisations/hm-revenue-customs.atom) the updated field within the ID's, and the updated field itself are identical. 

Also, adding `.rfc3339` to `Time.find_zone("UTC").parse(result["public_timestamp"])` locally raises an error when the feed.atom.builder (tries to present it)[https://github.com/alphagov/collections/blob/9f339b6a8b78f43b5720aebd0e62c7af1e7bb7ea/app/views/feeds/feed.atom.builder#L8]. 

These two things lead me to suspect that Rails's Atom feed helpers do some magic to format the date as `.rfc3339`. Will see if I can find the source code to confirm.....

(Ah here it is)[https://github.com/rails/rails/blob/0a608bd987fde4f9bc5bcf4bcebdb181d199cf4f/actionview/lib/action_view/helpers/atom_feed_helper.rb#L170]

bevanloon: 👍 magical - thanks! Learned something for the day 😄

bevanloon (approved):

PR #2023, opened 2020-10-23
Add related content to the 'Find local restrictions' page

DilwoarH:
## What

Add related content to the first page of the postcode lookup - we don't need to show it on the results page as it distracts from the main call to action

Related content should be:
gov.uk/coronavirus
gov.uk/find-coronavirus-support
https://www.gov.uk/guidance/making-a-support-bubble-with-another-household

## Why

To help users who can't find what they need on the page

Screenshot 2020-10-28 at 09 44 42

[Trello](https://trello.com/c/CXVGa7zY/861-add-related-content-to-the-find-local-restrictions-page)

### DIFF
--- a/app/views/coronavirus_local_restrictions/show.html.erb
+++ b/app/views/coronavirus_local_restrictions/show.html.erb
@@ -50,4 +50,21 @@
       } %>
     <% end %>
   </div>
+  <% if t("coronavirus_local_restrictions.related_navigation").is_a?(Array) %>
+    <div class="govuk-grid-column-one-third">
+      <%= render 'govuk_publishing_components/components/related_navigation', {
+        "content_item": {
+          "links" => {
+            "ordered_related_items" => t("coronavirus_local_restrictions.related_navigation").map do | item |
+              {
+                "title" => item[:title],
+                "base_path" => item[:base_path]
+              }
+            end
+          }
+        },
+        context: :sidebar
+      } %>
+    </div>
+  <% end %>
 </div>
--- a/app/views/coronavirus_local_restrictions/show.html.erb
+++ b/app/views/coronavirus_local_restrictions/show.html.erb
@@ -50,21 +50,14 @@
       } %>
     <% end %>
   </div>
-  <% if t("coronavirus_local_restrictions.related_navigation").is_a?(Array) %>
-    <div class="govuk-grid-column-one-third">
-      <%= render 'govuk_publishing_components/components/related_navigation', {
-        "content_item": {
-          "links" => {
-            "ordered_related_items" => t("coronavirus_local_restrictions.related_navigation").map do | item |
-              {
-                "title" => item[:title],
-                "base_path" => item[:base_path]
-              }
-            end
-          }
-        },
-        context: :sidebar
-      } %>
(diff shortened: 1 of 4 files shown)


app/views/coronavirus_local_restrictions/show.html.erb
```diff
@@ -50,4 +50,21 @@
       } %>
     <% end %>
   </div>
+  <% if t("coronavirus_local_restrictions.related_navigation").is_a?(Array) %>
+    <div class="govuk-grid-column-one-third">
+      <%= render 'govuk_publishing_components/components/related_navigation', {
+        "content_item": {
+          "links" => {
+            "ordered_related_items" => t("coronavirus_local_restrictions.related_navigation").map do | item |
```
JonathanHallam, line 58: Are there any legs in moving this logic into the controller? Just thinking it could clean up the view a bit

test/integration/coronavirus_local_restrictions_test.rb
```diff
@@ -327,4 +334,12 @@ def level_two_area
       "country_name" => "England",
     }
   end
+
+  def then_i_see_the_related_links_for_coronavirus
+    within ".gem-c-related-navigation" do
+      assert page.has_link?("Coronavirus (COVID-19): guidance and support", href: "/coronavirus")
```
JonathanHallam, line 340: Can we use a reference to the I18n instead of hardcoding this?

JonathanHallam (dismissed):
I'm happy that this is displaying properly. Tests look good but I'd like them using the I18n so that we can change the copy without changing the tests. Ignore my comments on logic in views if you want, just a suggestion :) Looks good

app/views/coronavirus_local_restrictions/show.html.erb
DilwoarH: We should ideally move this into a presenter if possible

JonathanHallam: good shout. That'd be great if we can :)

test/integration/coronavirus_local_restrictions_test.rb
GDSNewt: I'm trying to figure out how we could do this but it's tricky given how we iterate through related navigation in the yaml. Open to suggestions 🤔

JonathanHallam: you could replace the assertions with a loop;
```
      I18n.t("coronavirus_local_restrictions.related_navigation").each do |link|
        assert page.has_link?(link[:title], href: link[:base_path])
      end
```

GDSNewt: Brilliant Jon 😄 I've never seen that before!

GDSNewt:
> I'm happy that this is displaying properly. Tests look good but I'd like them using the I18n so that we can change the copy without changing the tests. Ignore my comments on logic in views if you want, just a suggestion :) Looks good

I think in the interest of rolling this out I'll hold off on the presenter logic stuff and add that to the backlog

Rosa-Fox (approved):
👍

PR #2026, opened 2020-10-23
Improve enter postcode box

GDSNewt:
This content change should make it clearer to users what is required of them, for the postcode input, by making the heading and hints less ambiguous.

It also adds a h2 tag to the input label to improve the accessibility for users of screen readers.

**BEFORE**

Screenshot 2020-10-23 at 12 47 19

**AFTER**

Screenshot 2020-11-12 at 10 42 48

[Trello card](https://trello.com/c/2MANqkBI/880-improve-enter-postcode-box)

### DIFF
--- a/app/assets/stylesheets/shared/_postcode-lookup.scss
+++ b/app/assets/stylesheets/shared/_postcode-lookup.scss
@@ -2,4 +2,8 @@
   padding: govuk-spacing(3);
   margin: govuk-spacing(6) 0;
   background: govuk-colour("light-grey");
+
+  .govuk-hint .gem-c-govspeak {
+    color: $govuk-secondary-text-colour;
+  }
 }
--- a/app/views/shared/_postcode-lookup.html.erb
+++ b/app/views/shared/_postcode-lookup.html.erb
@@ -2,7 +2,7 @@
   input_label ||= t("coronavirus_local_restrictions.lookup.input_label")
   input_value ||= nil
   input_error ||= nil
-  hint_text ||= t("coronavirus_local_restrictions.lookup.hint_text")
+  hint_text ||= render_govspeak(t("coronavirus_local_restrictions.lookup.hint_text"))
   button_text ||= t("coronavirus_local_restrictions.lookup.button_text")
   link ||= { text: t("coronavirus_local_restrictions.royal_mail_lookup.link_text"), href: t("coronavirus_local_restrictions.royal_mail_lookup.href")}
 %>
@@ -12,6 +12,7 @@
           label: {
             text: input_label
           },
+          heading_level: 2,
           heading_size: "s",
           error_message: input_error,
           value: input_value,
--- a/config/locales/en/lookup.yml
+++ b/config/locales/en/lookup.yml
@@ -10,8 +10,11 @@ en:
         input_error: Enter a valid postcode
         error_description: Enter a valid postcode
     lookup:
-      input_label: Enter your postcode
-      hint_text: For example LS1 1UR
+      input_label: Enter a full postcode 
+      hint_text: |
+        For example, LS1 1UR
+
+        We use this to give you more accurate information about the area. 
       button_text: Find
       title: Find out the coronavirus restrictions in a local area
       inset_text: |
--- a/test/integration/coronavirus_local_restrictions_test.rb
+++ b/test/integration/coronavirus_local_restrictions_test.rb
@@ -153,7 +153,7 @@ class CoronavirusLocalRestrictionsTest < ActionDispatch::IntegrationTest
     ]
     stub_mapit_has_a_postcode_and_areas(postcode, [], areas)
 
-    fill_in "Enter your postcode", with: postcode
+    fill_in "Enter a full postcode", with: postcode
   end
 
   def then_i_enter_a_valid_english_postcode_with_a_future_level_two_restriction
@@ -168,7 +168,7 @@ class CoronavirusLocalRestrictionsTest < ActionDispatch::IntegrationTest
     ]
     stub_mapit_has_a_postcode_and_areas(postcode, [], areas)
 
-    fill_in "Enter your postcode", with: postcode
+    fill_in "Enter a full postcode", with: postcode
   end
 
   def then_i_enter_a_valid_english_postcode_with_a_future_level_three_restriction
@@ -183,20 +183,20 @@ class CoronavirusLocalRestrictionsTest < ActionDispatch::IntegrationTest
     ]
     stub_mapit_has_a_postcode_and_areas(postcode, [], areas)
 
-    fill_in "Enter your postcode", with: postcode
+    fill_in "Enter a full postcode", with: postcode
   end
 
   def then_i_enter_a_valid_english_postcode_with_an_extra_special_character
     stub_mapit_has_a_postcode_and_areas("E1 8QS", [], [level_two_area])
 
-    fill_in "Enter your postcode", with: ".e18qs"
+    fill_in "Enter a full postcode", with: ".e18qs"
   end
 
   def then_i_enter_a_valid_english_postcode_in_tier_two
     postcode = "E1 8QS"
     stub_mapit_has_a_postcode_and_areas(postcode, [], [level_two_area])
 
-    fill_in "Enter your postcode", with: postcode
+    fill_in "Enter a full postcode", with: postcode
   end
 
   def then_i_enter_a_valid_english_postcode_in_tier_three
@@ -211,7 +211,7 @@ class CoronavirusLocalRestrictionsTest < ActionDispatch::IntegrationTest
     ]
     stub_mapit_has_a_postcode_and_areas(postcode, [], areas)
 
-    fill_in "Enter your postcode", with: postcode
+    fill_in "Enter a full postcode", with: postcode
   end
 
   def then_i_enter_a_valid_welsh_postcode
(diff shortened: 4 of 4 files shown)


leenagupte (commented):
It all works, which is great!

The only thing I'm concerned about is the extra css styling that's being added to the locale file entries as I feel that this makes it harder to update the content. Is there anyway to add this to the view instead?

config/locales/en/lookup.yml
```diff
@@ -10,8 +10,10 @@ en:
         input_error: Enter a valid postcode
         error_description: Enter a valid postcode
     lookup:
-      input_label: Enter your postcode
-      hint_text: For example LS1 1UR
+      input_label: <h2 class="govuk-heading-s"> Enter a full postcode </h2>
```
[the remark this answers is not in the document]
andysellick: Will this put a `h2` tag inside a `label`? Only I'm pretty sure that's invalid HTML.

What I think you need is this: https://github.com/alphagov/govuk_publishing_components/pull/1747 which will go out in the next version of the components gem early next week, might be worth hanging on and waiting for that.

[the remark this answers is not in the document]
GDSNewt: > Has this css class been added because we want to make the label an "H2" but not actually change the visible size of it?

Hi Leena, yes it was. The default heading level for a h2 tag in the design system has the css "govuk-heading-l" whereas the card specified a small size, not ideal I agree.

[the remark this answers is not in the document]
GDSNewt: > Will this put a `h2` tag inside a `label`? Only I'm pretty sure that's invalid HTML.
> 
> What I think you need is this: [alphagov/govuk_publishing_components#1747](https://github.com/alphagov/govuk_publishing_components/pull/1747) which will go out in the next version of the components gem early next week, might be worth hanging on and waiting for that.

Hi Andy, that is great timing haha this is exactly what I needed! I will wait for the update 😄

config/locales/en/lookup.yml
```diff
@@ -10,8 +10,10 @@ en:
...
+      hint_text: |
+        <p class="govuk-hint"> We use this to give you more accurate information about the area.</p>
```
DilwoarH, line 15: we should just have the text here and let the view handle the classes etc

config/locales/en/lookup.yml
```diff
@@ -10,7 +10,7 @@ en:
         input_error: Enter a valid postcode
         error_description: Enter a valid postcode
     lookup:
-      input_label: <h2 class="govuk-heading-s"> Enter a full postcode </h2>
+      input_label: Enter a full postcode
```
leenagupte, line 13: This is much better! Can you fix up this commit to the first one?

GDSNewt: Thanks Leena, i've fixed up the commits now 🙂

config/locales/en/lookup.yml
leenagupte: Was this resolved via a conversation? Why is the `govuk-hint` class needed here?

leenagupte: Ok, I played around with this locally, and figured out why the `govuk-hint` class is needed. Can you add a note to the commit message about this @GDSNewt?

GDSNewt: > Was this resolved via a conversation? Why is the `govuk-hint` class needed here?

Ahhh I prematurely resolved the conversation when I removed the html from the input_label yaml. As for the gov-hint class in the hint text, yeah it is needed to have the correct spacing between the two lines of the hint. We adopt a similar approach for the body text and inset text on this page so I thought this was ok to do so in this instance. I'm not really sure how you would do it via the view, but happy to look into it and change it for this, the body text and the inset text if that is the approach we are taking.

leenagupte: I wonder if someone like @DilwoarH or @andysellick knows of another way to add multiline hint text?

DilwoarH: I'll look into it

app/views/shared/_postcode-lookup.html.erb
```diff
@@ -12,11 +12,12 @@
           label: {
             text: input_label
           },
+          heading_level: 2,
           heading_size: "s",
           error_message: input_error,
           value: input_value,
           name: "postcode-lookup",
-          hint: hint_text,
+          hint: sanitize(t("coronavirus_local_restrictions.lookup.hint_text")),
```
DilwoarH, line 20: Is there a reason why we don't use the variable above https://github.com/alphagov/collections/pull/2026/files#diff-cd642c40804c0a24c6e913fda95a4ba7cd90b1918f2585c36749622ec36f59b2R5

config/locales/en/lookup.yml
```diff
@@ -10,8 +10,10 @@ en:
         input_error: Enter a valid postcode
         error_description: Enter a valid postcode
     lookup:
-      input_label: Enter your postcode
-      hint_text: For example LS1 1UR
+      input_label: Enter a full postcode 
+      hint_text: |
+        <p class="govuk-hint"> We use this to give you more accurate information about the area.</p>
+        <p class="govuk-hint"> For example, LS1 1UR</p>
```
DilwoarH, line 16:
```suggestion
      hint_text: |
        We use this to give you more accurate information about the area.<br/>
        For example, LS1 1UR
```

I think this is probably best way of going about it

app/views/shared/_postcode-lookup.html.erb
DilwoarH:
```suggestion
          hint: hint_text,
```

DilwoarH: Change line 5 to: ```  hint_text ||= sanitize(t("coronavirus_local_restrictions.lookup.hint_text"))```

config/locales/en/lookup.yml
```diff
@@ -10,8 +10,10 @@ en:
         input_error: Enter a valid postcode
         error_description: Enter a valid postcode
     lookup:
-      input_label: Enter your postcode
-      hint_text: For example LS1 1UR
+      input_label: Enter a full postcode 
+      hint_text: |
+        We use this to give you more accurate information about the area.<br/>
+        For example, LS1 1UR
```
leenagupte, line 16: This doesn't work, I'm afraid:

Screenshot 2020-08-27 at 09 30 20

I tried it with a `<p>` instead of the `<br/>` and that didn't work either.

cc @DilwoarH

DilwoarH, line 16:
```suggestion
        <div class="govuk-!-margin-bottom-2">We use this to give you more accurate information about the area.</div>
        <div>For example, LS1 1UR</div>
```

leenagupte: To be fair, that's harder than using the hint class.

app/assets/stylesheets/shared/_postcode-lookup.scss
```diff
@@ -2,4 +2,8 @@
   padding: govuk-spacing(3);
   margin: govuk-spacing(6) 0;
   background: govuk-colour("light-grey");
+
+  .govuk-hint .gem-c-govspeak {
```
injms, line 6: It could be worth a comment to explain why this override is in place - I'm guessing it's because you render something in Govspeak which makes the text black, and you want to revert it to make the hint colour grey?

app/assets/stylesheets/shared/_postcode-lookup.scss
```diff
@@ -2,4 +2,8 @@
...
+    color: govuk-colour("dark-grey");
```
injms, line 7: I'd suggest using `$govuk-secondary-text-colour` - it'll have the same output, but means that if the secondary text colour changes in a future release of the Design System then this will keep up with that change as well.

```suggestion
    color: $govuk-secondary-text-colour;
```

leenagupte:
@GDSNewt Could you update the screenshots in the PR so we can see if these changes render correctly?

app/views/shared/_postcode-lookup.html.erb
```diff
@@ -2,7 +2,7 @@
   input_label ||= t("coronavirus_local_restrictions.lookup.input_label")
   input_value ||= nil
   input_error ||= nil
-  hint_text ||= sanitize(t("coronavirus_local_restrictions.lookup.hint_text"))
+  hint_text ||= render_govspeak(t("coronavirus_local_restrictions.lookup.hint_text"))
```
leenagupte, line 5: Could you either squash the two commits together, or re-arrange them so that this change, and the change to locale file are in the right place? This commit should only have the test changes in it.

config/locales/en/lookup.yml
```diff
@@ -10,8 +10,11 @@ en:
         input_error: Enter a valid postcode
         error_description: Enter a valid postcode
     lookup:
-      input_label: Enter your postcode
-      hint_text: For example LS1 1UR
+      input_label: Enter a full postcode 
+      hint_text: |
```
maxgds, line 14: Has a content designer approved this? It looks confusing to me, it could be read as LS1 1UR being an example of accurate information about the area. If the sentences were the other way around it would work better.

GDSNewt: I see what you mean, it was content approved but I can go back to them with your point and see what they say.

leenagupte: Hi @maxgds! It was a content designer who added the ticket with the mocked up example. The content here matches what they asked for.

GDSNewt: They have agreed to your suggestion, I will make the change and ask for re-review 🙂

benthorner:
Removed CD warning.

maxgds (approved):

PR #2027, opened 2020-10-23
Use Local Authority name from Mapit

leenagupte:
When the restrictions data in "lib/local_restrictions/local-restrictions.yaml" is updated,
a developer is calling Mapit manually to get the name of the local authority to populate
the "name" field.

This seems pointless as we are already get the local authority name
when we do the lookup:

https://github.com/alphagov/collections/blob/master/app/services/location_lookup_service.rb#L37-L40

Asking people to call mapit like this, makes the restrictions data harder to update. Instead we should use
the "name" field in the YAML file as something human-readable for the person updating to look as gss codes are
hard to read, but not something that is used in the code.

### DIFF
--- a/app/views/coronavirus_local_restrictions/_tier_three_restrictions.html.erb
+++ b/app/views/coronavirus_local_restrictions/_tier_three_restrictions.html.erb
@@ -17,7 +17,7 @@
   } %>
 
   <p class="govuk-body">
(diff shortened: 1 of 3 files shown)


JonathanHallam (approved):
Looks good to me, ran locally with the names field removed and it displayed properly 👍

benthorner:
Removed CD warning.

leenagupte:
Closing this as we have no guarantee that the restrictions will always be determined by local authority boundaries. If we want to use the name from Mapit, we need to get the one associated with the appropriate GSS code which requires a bit more thought.

PR #2028, opened 2020-10-23
Add docs for updating COVID Local Restriction Data

Rosa-Fox:


docs/update-covid-local-restriction-data.md
```diff
@@ -0,0 +1,67 @@
...
+
+Otherwise we just display the current restrictions. This is determined by the `start_date` and `start_time`.
+
+![Image of results page without changing alert level](docs/alert-level-no-upcoming-changes.png)
+
+Example of somewhere that just has current restrictions:
+
+```
+E08000012:
+  name: Liverpool City Council
```
leenagupte, line 30: One thing we should mention is that we use Mapit to look up the "name" we enter here.

leenagupte (commented):
One comment. Otherwise looks good!

docs/update-covid-local-restriction-data.md
```diff
@@ -0,0 +1,67 @@
...
+    - alert_level: 3
+      start_date: 2020-10-23
+      start_time: "00:01"
+```
+
+To update the restriction data, you will need to [add a list to the YAML](https://github.com/alphagov/collections/blob/master/lib/local_restrictions/local-restrictions.yaml) for the given GSS codes that contain:
+
+- the new `alert_level`.
+- the `start_date` for when the restrictions will be introduced.
+- the `start_time` for when the restrictions will be introduced.
```
leenagupte, line 55: We should also mention, that a "new" entry should be added for the upcoming changes. Almost overwrote some existing ones not 5 minutes ago! 😱

docs/update-covid-local-restriction-data.md
Rosa-Fox: thanks, have added :)

docs/update-covid-local-restriction-data.md
Rosa-Fox: Thanks @leenagupte! have added something about adding rather than overwriting!

leenagupte (approved):
Looks good! 🎉 ⭐

PR #1987, opened 2020-10-16
Adjusting fallback copy

chris-gds:
## What?
Adjusting copy so it reads within one sentence

## Why?
Fail of WCAG SC 1.3.1
_Screen reader and user style users will not see the text as one sentence. Screen reader users might get confused by the incomplete heading and not understand it out of context._

[Ticket](https://trello.com/c/xZaPl0v7)

## Visuals
No visual change
Screenshot 2020-10-16 at 14 26 08

-----
:warning: This application is Continuously Deployed: :warning:

- Merged changes are automatically deployed to staging and production.

- Make sure you follow [the guidance for deployments](https://docs.publishing.service.gov.uk/manual/development-pipeline.html#merge-your-own-pull-request) **before** you merge.

- Check your branch is being deployed in the [Release app](https://release.publishing.service.gov.uk/applications/collections), after merging.

PR #2036, opened 2020-10-27
Adjusting fallback copy

chris-gds:
## What?

When cookies are not accepted, adjusting Content to ensure that screen-readers do not read the _YouTube_ fallback Content in a disjointed manner. Partially, maintaining existing Design.  

`govspeak` replaces code when a video is detected inside a `p` tag to render a video. There is also an update to improve the use of bold copy to be semantic

Change impacting the [coronavirus](https://www.gov.uk/coronavirus) and [transition](https://www.gov.uk/transition) page only.

## Why? 

([Ticket](https://trello.com/c/xZaPl0v7)) - A user has to accept cookies to view a _YouTube_ video, if they don't consent the video cannot be displayed as _YouTube_ will track the user, therefore breaching GPDR.  The fallback Content currently doesn't read in a coherent manner for screen readers.

_Screen reader and user style users will not see the text as one sentence. Screen reader users might get confused by the incomplete heading and not understand it out of context._

`[P]Change your cookie settings[/P] [P]to watch the video[/P] [P]or[/P] Watch on YouTube`

 **Fail of WCAG SC 1.3.1**
present on [coronavirus](https://www.gov.uk/coronavirus) and [transition](https://www.gov.uk/transition) page.

## Anything else?

 - The [original PR](https://github.com/alphagov/collections/pull/1987) was [reverted](https://github.com/alphagov/collections/pull/2017/files) after a visual issue went live, this led to some deeper investigation:

- Testing reveals that `br` [tags are problematic for VoiceOver](https://axesslab.com/text-splitting/) they cause disjointed text splitting when being read out.  

- [A similar PR using `br` as solution was previously created](https://github.com/alphagov/collections/pull/1849)

 - There is a [similar issue](https://github.com/alphagov/govuk_publishing_components/issues/1256) that references this fallback within the publishing components.

- There is no dedicated video component, `govspeak` has a feature to include _YouTube_ video which is then enhanced, to increase its level of accessibility.

- the Design System have [explored various considerations to add video ](https://github.com/alphagov/govuk-design-system-backlog/issues/107)

- [Coronavirus](https://www.gov.uk/coronavirus) and [transition](https://www.gov.uk/transition) pages are seemingly the only pages that build upon the `govspeak` component to provide this fallback Design / Content

- Pragmatic fix to resolve issue however, aware this does introduce tech debt as additional development diverges from _component gem_ based architecture. 

A visual representation of how the text now sounds can be seen below:

```
Link, Change your cookie settings to watch the video

or 

Link, Watch on YouTube
```

## Visuals

Fallback **before**
Screenshot 2020-10-26 at 12 59 12

Fallback **after**
Screenshot 2020-10-29 at 11 44 57

 
Cookies Accepted

Screenshot 2020-10-26 at 10 52 13

---

Fallback 

Screenshot 2020-10-29 at 11 45 19

Cookies Accepted:

Screenshot 2020-10-26 at 12 57 06

----

### DIFF
--- a/app/assets/stylesheets/components/_youtube-player.scss
+++ b/app/assets/stylesheets/components/_youtube-player.scss
@@ -5,15 +5,6 @@
   border: solid 2px govuk-colour("black");
   text-align: center;
   margin-bottom: govuk-spacing(2);
-
-  .govuk-body {
-    margin: 0;
-    color: govuk-colour("black");
-  }
-
-  .youtube__bold {
-    font-weight: bold;
-  }
 }
 
 .youtube__video-logo {
@@ -42,6 +33,10 @@
 }
 
 .youtube__video-wrapper--inverse {
+  .youtube__video-placeholder {
+    color: govuk-colour("black");
+  }
+
   .govuk-details__summary {
     color: govuk-colour('white');
 
--- a/app/views/shared/_youtube_player.html.erb
+++ b/app/views/shared/_youtube_player.html.erb
@@ -12,27 +12,28 @@
 <%= tag.div class: css_classes do %>
   <%= render "govuk_publishing_components/components/govspeak", {
   } do %>
-    <div class="youtube__video-placeholder">
-      <div class="youtube__video-logo"><%= no_cookies_icon_text %></div>
-      <p class="govuk-body">
-        <a href="<%= no_cookies_change_settings_link %>" class="govuk-link youtube__page-header-link youtube__bold">
+    <p class="youtube__video-placeholder">
+      <span class="youtube__video-logo">
+        <%= no_cookies_icon_text %>
+      </span>
+      <strong>
+        <a href="<%= no_cookies_change_settings_link %>" class="govuk-link">
           <%= no_cookies_change_settings_text %>
-        </a>
-        <br>
-        <span class="youtube__bold">
+          <br aria-hidden="true">
           <%= no_cookies_to_watch_text %>
-        </span>
-        <br>
-        <%= no_cookies_or_text %>
-      </p>
+        </a>
+      </strong>
+      <br aria-hidden="true">
+      <%= no_cookies_or_text %>
+      <br aria-hidden="true">
       <a
-        href="<%= youtube_video_url %>"
-        class="govuk-body govuk-link"
-        data-youtube-player-analytics="true"
-        data-youtube-player-analytics-category="EmbeddedYoutube">
+          href="<%= youtube_video_url %>"
+          class="govuk-body govuk-link"
+          data-youtube-player-analytics="true"
+          data-youtube-player-analytics-category="EmbeddedYoutube">
         <%= no_cookies_watch_link_text %>
       </a>
-    </div>
+    </p>
   <% end %>
   <%= render "govuk_publishing_components/components/details", { title: transcription_summary} do %>
     <%= transcription_text.html_safe %>


danacotoran:
Sorry have not gone into the code yet, just popping this comment in to say I've tested in browser and it still reads disjointed to me using VoiceOver. It currently reads as:


```
Link, Change your cookie settings

To watch the video or (you are currently on a text element)

Link, Watch on YouTube
```

Would it be clearer for screen reader users if it read like this instead?

```
Link, Change your cookie settings to watch the video

or (you are currently on a text element)

Link, Watch on YouTube
```

No matter what we do, I think this code will still read somewhat disjointed due to links being interspersed with text. It's a tricky one.

Also I had a quick glance and I think your checks might be failing due to your use of `const` – they should pass if you use  `var` instead.

chris-gds:
> Sorry have not gone into the code yet, just popping this comment in to say I've tested in browser and it still reads disjointed to me using VoiceOver. It currently reads as:
> 
> ```
> Link, Change your cookie settings
> 
> To watch the video or (you are currently on a text element)
> 
> Link, Watch on YouTube
> ```
> 
> Would it be clearer for screen reader users if it read like this instead?
> 
> ```
> Link, Change your cookie settings to watch the video
> 
> or (you are currently on a text element)
> 
> Link, Watch on YouTube
> ```
> 
> No matter what we do, I think this code will still read somewhat disjointed due to links being interspersed with text. It's a tricky one.

I see your thinking - this changes the visual as now a link is "Change your cookie settings to watch the video" however I do agree, this is more desirable from many angles. 

> 
> Also I had a quick glance and I think your checks might be failing due to your use of `const` – they should pass if you use `var` instead.

Interesting I had not considered this, i'll double check this.  This might be a webbook + jenkins issue.

danacotoran:
> I see your thinking - this changes the visual and the Context as now a link is "Change your cookie settings to watch the video" however I do agree, this is more desirable from many angles.

It doesn't have to change the visual necessarily, we could technically achieve that with **aria-label**s, so that the information would read differently only for screenreader users.
I'm not sure whether that would be just as problematic in terms of WCAG as the current state of things though. Knowing WCAG, the answers is probably yes 😅

chris-gds:
> > I see your thinking - this changes the visual and the Context as now a link is "Change your cookie settings to watch the video" however I do agree, this is more desirable from many angles.
> 
> It doesn't have to change the visual necessarily, we could technically achieve that with **aria-label**s, so that the information would read differently only for screenreader users.

> I'm not sure whether that would be just as problematic in terms of WCAG as the current state of things though. Knowing WCAG, the answers is probably yes 😅

There a few general issues here:
1. When a user lands on this aspect of Content there is implied knowledge of what GDPR laws are and why this Content is being displayed.
2. Aside from hearing: “VIDEO” there is no real contextual understanding of what this is before the user is presented with choices.
3. In this visual presentation of bold, there is a nudge for the user to accept cookies on govuk _rather_ than a neutral option being presented to also watch on YouTube - unclear is this intentional or desirable.

The reasoning to the user hearing (below) is partially based on details from 1.3.1

_information and relationships that are implied by visual or auditory formatting are preserved when the presentation format changes._

_When such relationships are perceivable to one set of users, those relationships can be made to be perceivable to all_

```
VIDEO, image.

Link, Change your cookie settings

To watch the video or (you are currently on a text element)

Link, Watch on YouTube
```

A lot of the context is this Design relationship, I think to factor in both - changing the link and visuals to 

`Link, Change your cookie settings (visual break only) to watch the video` is desirable even if this adds a visual change -  will explore.

chris-gds:
Content review also provided similar feedback and questioning around this fallback content in general.  

- How many users see this?
- The meaning of visual Design of "VIDEO and the red circle
- The use of video in general
- All of the accessibility concerns around video + Whitehall.

Many elements out of scope of the ticket.  _Suggested_ recommendations:

_You cannot watch this video as you've rejected cookie settings. Change your cookie settings to watch this video, or watch it on Youtube._

danacotoran:
I think this solution, while it does work, is a little bit over-engineered.
I think we can work around the limitations of this govspeak youtube link enhancement functionality without adding extra javascript. I'm really of the mindset that we should avoid Javascript unless absolutely necessary/beneficial. 🤷🏻‍♀️

I'd suggest that for instance, we could shift the markup around a little bit to have something like this:

```
  <%= render "govuk_publishing_components/components/govspeak", {
  } do %>
    <p class="youtube__video-placeholder">
      <span class="youtube__video-logo"><%= no_cookies_icon_text %></span>
      <strong>
        <a href="/help/cookies" class="govuk-link">
          Change your cookie settings  <br aria-hidden="true"> to watch the video
        </a>
      </strong>
      <br aria-hidden="true">
      or
      <br aria-hidden="true">
      <a href="https://www.youtube.com/watch?v=1CUrxdTd1bc&amp;rel=0" class="govuk-body govuk-link" data-youtube-player-analytics="true" data-youtube-player-analytics-category="EmbeddedYoutube">
        Watch on YouTube
      </a>
    </p>
  <% end %>
```
....or similar.

Although arguably having that "VIDEO" inside the paragraph is not great, restructuring things like this should remove the need for new JS and I think the additional CSS can go too?

chris-gds:
> I think this solution, while it does work, is a little bit over-engineered.
 > I'm really of the mindset that we should avoid Javascript unless absolutely necessary/beneficial. 

I 100% agree with all of this, changing the Design changed the possible approach.  This has been removed.

> Although arguably having that "VIDEO" inside the paragraph is not great, restructuring things like this should remove the need for new JS and I think the additional CSS can go too?

Agreed, it's not great, this is the Content the user needs almost needs to hear:  _"this is a YouTube video, to watch this on govuk please accept cookies to allow marketing from Google or you can watch externally on YouTube"_  however this isn't really reflected but out of scope so is this "VIDEO" issue.

danacotoran (approved):
Thanks for getting it done 👏 💯

PR #2041, opened 2020-10-28
Use A/B test machinery to show signed in / out header

barrucadu:
See also https://github.com/alphagov/slimmer/pull/255 & https://github.com/alphagov/govuk-cdn-config/pull/300

---

[Trello card](https://trello.com/c/vdZrgBRy/363-make-changes-to-transition-for-trial-launch)


brucebolt (approved):
Based on my limited understanding of our A/B test framework, this looks ok to me

PR #2047, opened 2020-10-29
Fix promoted news - tweaks

sihugh:
Note - this is to merge into the `fix-promoted-news` branch to fix some tests

The problem this fixes is to configure a test so that search returns only 1 result for the news and communications section.

I've refactored a little, in reasonably small steps to hopefully make it followable. There's further tidy up work that could be done, but it would affect other tests in the app and I wanted to keep the changes small.


injms (approved):
Thank you @sihugh 🎉

PR #1952, opened 2020-10-12
Remove constraint around local restrictions

leenagupte:
Trello: https://trello.com/c/UDdYKNd9

We added a constraint around local restrictions so that we could deploy the bulk
of the code, but prevent people from looking up the guidance prematurely.

The lookup is almost ready to be turned on, so we can remove these constraints.

Removes changes added here: https://github.com/alphagov/collections/commit/621314db74bde1d15e37beff5774dd544702e912

:warning: This application is Continuously Deployed: :warning:

- Merged changes are automatically deployed to staging and production.

- Make sure you follow [the guidance for deployments](https://docs.publishing.service.gov.uk/manual/development-pipeline.html#merge-your-own-pull-request) **before** you merge.

- Check your branch is being deployed in the [Release app](https://release.publishing.service.gov.uk/applications/collections), after merging.

PR #2057, opened 2020-11-02
Add route constraint to redirect postcode checker to guidance page in production

Rosa-Fox:
**DO NOT MERGE/DEPLOY until 00:00 05/11/2020**

- From 05/11/2020 - 02/12/2020 England will go into a national lockdown.
- As a result we will redirect from the postcode checker to the guidance.
- We have added the constraint depending on ENV so that we can bring the postcode checker back towards the end of the national lockdown.
- We still want to be able to continue development work, without users being to access the postcode checker.

[Original constraint adding PR!](https://github.com/alphagov/collections/pull/1952/files)

**DO NOT MERGE/DEPLOY until 00:00 05/11/2020**

**Will need to be merged after [this PR](https://github.com/alphagov/smokey/pull/736) to remove the smoke test has been deployed!**

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -28,8 +28,14 @@ Rails.application.routes.draw do
   end
 
   # Routes for local restrictions postcode lookup
-  get "/find-coronavirus-local-restrictions" => "coronavirus_local_restrictions#show"
-  post "/find-coronavirus-local-restrictions" => "coronavirus_local_restrictions#results"
+  constraints CoronavirusLocalRestrictionsConstraint.new do
+    get "/find-coronavirus-local-restrictions" => "coronavirus_local_restrictions#show"
+    post "/find-coronavirus-local-restrictions" => "coronavirus_local_restrictions#results"
+  end
+
+  # Will redirect for the production environment.
+  # This is because the postcode checker will not be in use during November national lockdown!
+  get "/find-coronavirus-local-restrictions", to: redirect("https://www.gov.uk/guidance/new-national-restrictions-from-5-november")
 
   # TODO: this redirect causes the request to be routed to Whitehall where
   # the country A-Z currently lives. This needs to be removed when the world index
--- a/app/lib/coronavirus_local_restrictions_constraint.rb
+++ b/app/lib/coronavirus_local_restrictions_constraint.rb
@@ -1,0 +1,7 @@
+class CoronavirusLocalRestrictionsConstraint
+  def matches?(*)
+    return false if Rails.env.production?
+
+    true
+  end
+end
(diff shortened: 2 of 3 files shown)


leenagupte (approved):
LGTM! 🎉 

I did wonder if we should use short-url-manager for the redirect, but then reconsidered as it's possible that this redirect will only be needed for a few weeks. Also, using short-url manager would require us to unpublish and re-publish the content item at midnight on two separate occasions, which feel like a lot of hassle.

Rosa-Fox:
> LGTM! 🎉
> 
> I did wonder if we should use short-url-manager for the redirect, but then reconsidered as it's possible that this redirect will only be needed for a few weeks. Also, using short-url manager would require us to unpublish and re-publish the content item at midnight on two separate occasions, which feel like a lot of hassle.
> 
> The smokey test will also need to be updated: https://github.com/alphagov/smokey/blob/7bd44a59e90aaaee8055b9b81120b8b533cd91a5/features/step_definitions/coronavirus_local_restrictions_steps.rb#L2

Thanks 

> LGTM! 🎉
> 
> I did wonder if we should use short-url-manager for the redirect, but then reconsidered as it's possible that this redirect will only be needed for a few weeks. Also, using short-url manager would require us to unpublish and re-publish the content item at midnight on two separate occasions, which feel like a lot of hassle.
> 
> The smokey test will also need to be updated: https://github.com/alphagov/smokey/blob/7bd44a59e90aaaee8055b9b81120b8b533cd91a5/features/step_definitions/coronavirus_local_restrictions_steps.rb#L2

thanks @leenagupte!

I made a PR for removing the smoke tests, if you could review that would be fab: https://github.com/alphagov/smokey/pull/736.

I also updated the link to the finalised guidance link so should be ready to go for midnight Thurs.

config/routes.rb
```diff
@@ -28,8 +28,14 @@
...
-  get "/find-coronavirus-local-restrictions" => "coronavirus_local_restrictions#show"
-  post "/find-coronavirus-local-restrictions" => "coronavirus_local_restrictions#results"
+  constraints CoronavirusLocalRestrictionsConstraint.new do
+    get "/find-coronavirus-local-restrictions" => "coronavirus_local_restrictions#show"
+    post "/find-coronavirus-local-restrictions" => "coronavirus_local_restrictions#results"
+  end
+
+  # Will redirect for the production environment.
+  # This is because the postcode checker will not be in use during November national lockdown!
+  get "/find-coronavirus-local-restrictions", to: redirect("https://www.gov.uk/guidance/new-national-restrictions-from-5-november")
```
leenagupte, line 38: I wonder if we can use the relative path here?
e.g. 

```
get "/find-coronavirus-local-restrictions", to: redirect("/guidance/new-national-restrictions-from-5-november")
```

leenagupte: Apparently not. It tries to render the taxon page.

PR #2059, opened 2020-11-05
Remove PR template for CD

benthorner:
https://trello.com/c/epNWudCR/176-write-an-rfc-for-continuous-deployment-and-seek-review

We are disabling CD for this app as it does not meet the new safety
criteria for automatic deployments.

PR #2074, opened 2020-11-11
Send errors to sentry to see why are users shown an error message

leenagupte:
Trello: https://trello.com/c/dZz2sedX

# What?

Send postcode errors the user is seeing to Sentry we can see if there are any common problems.

# Why?

We know that between 12 - 15% of users will see an error message when they enter their postcode. (Tracking began on 20 October)
We also receive a fair amount of feedback about this. This seems high.
However, we don't know why users are seeing these errors. 

# Results from testing in integration

## Submitting a blank postcode
Screenshot 2020-11-11 at 12 21 28
Screenshot 2020-11-11 at 12 22 15

## Submitting a postcode with a special character in it
Screenshot 2020-11-11 at 12 24 03
Screenshot 2020-11-11 at 12 30 28

### DIFF
--- a/app/controllers/coronavirus_local_restrictions_controller.rb
+++ b/app/controllers/coronavirus_local_restrictions_controller.rb
@@ -15,14 +15,29 @@ class CoronavirusLocalRestrictionsController < ApplicationController
 
   def results
     if params["postcode-lookup"].blank?
+      GovukError.notify(
+        "Postcode left blank",
+        extra: { error_message: "Postcode field left blank" },
+      )
+
       return render_no_postcode_error
     end
 
     @postcode = PostcodeService.new(params["postcode-lookup"]).sanitize
 
     if @postcode.blank?
+      GovukError.notify(
+        "Sanitized postcode is blank",
+        extra: { error_message: "Postcode is blank after being sanitized" },
+      )
+
       return render_no_postcode_error
     elsif !PostcodeService.new(@postcode).valid?
+      GovukError.notify(
+        "Postcode failed validation",
+        extra: { error_message: "Postcode #{@postcode} failed validation after being sanitized" },
+      )
+
       return render_invalid_postcode_error
     end
 
@@ -36,8 +51,18 @@ class CoronavirusLocalRestrictionsController < ApplicationController
                       breadcrumbs: breadcrumbs,
                     }
     elsif @location_lookup.invalid_postcode?
+      GovukError.notify(
+        "Postcode failed Mapit validation",
+        extra: { error_message: "Postcode #{@postcode} failed validation with Mapit" },
+      )
+
       return render_invalid_postcode_error
     elsif @location_lookup.postcode_not_found?
+      GovukError.notify(
+        "Mapit couldn't find postcode",
+        extra: { error_message: "Postcode #{@postcode} couldn't be found in Mapit" },
+      )
+
       return render_no_postcode_error
     end
 


leenagupte:
@thomasleese Do you think this could cause a problem with Sentry's rate limiting? Is there somewhere else I should be sending the errors?

leenagupte:
@DilwoarH Is there anywhere I can add more specific error messages (naturally I wouldn't include the postcode) to the GA error tracking? I had a look at the [error summary component](https://components.publishing.service.gov.uk/component-guide/error_summary) but nothing jumped out at me. I was hoping to add more information [here](https://github.com/alphagov/collections/blob/master/app/views/coronavirus_local_restrictions/show.html.erb#L34).

app/controllers/coronavirus_local_restrictions_controller.rb
```diff
@@ -36,8 +51,18 @@ def results
                       breadcrumbs: breadcrumbs,
                     }
     elsif @location_lookup.invalid_postcode?
+      GovukError.notify(
+        "Postcode failed Mapit validation",
+        extra: { error_message: "Postcode #{@postcode} failed validation with Mapit" },
```
DilwoarH, line 56: Should we include the postcode to sentry? what if it includes PII?

leenagupte: That's a tough one. We have the postcode in the Kibana logs when the checker is running, because we're logging every POST request. I suppose if we don't log it, how will we know what the actual problem is? Do we need to know?

DilwoarH: correct me if my understanding is wrong but should we only log to sentry when there is an actual error - invalid postcodes are handled appropriately. I think sentry should only get errors when there is an error with the API or another unknown error

leenagupte: One of the questions in the trello card is:

>What we don't know
  We don't know what users are putting in the postcode look up box to trigger an error message
  We don't know whether this is a tech or content problem. Is there a problem with mapit? Is it out of date?

I was hoping that if we log a more detailed error, we could answer some of those questions.

thomasleese:
> @thomasleese Do you think this could cause a problem with Sentry's rate limiting? Is there somewhere else I should be sending the errors?

There's certainly a risk with it hitting the rate limits. We already pass the limits most hours, so it's likely you'll get most of these through to Sentry but perhaps not all of them.

I would say that reporting to Sentry on user error does seem a little unorthodox where we could be logging these to Logit instead. However, if it'll make it easier to identify the problem, fix the issue, and therefore remove the Sentry reporting then it's probably fine as a temporary measure.

leenagupte:
> > @thomasleese Do you think this could cause a problem with Sentry's rate limiting? Is there somewhere else I should be sending the errors?
> 
> There's certainly a risk with it hitting the rate limits. We already pass the limits most hours, so it's likely you'll get most of these through to Sentry but perhaps not all of them.
> 
> I would say that reporting to Sentry on user error does seem a little unorthodox where we could be logging these to Logit instead. However, if it'll make it easier to identify the problem, fix the issue, and therefore remove the Sentry reporting then it's probably fine as a temporary measure.

Thanks @thomasleese. Kibana/Logit and I are not friends. I'm not sure how to 2) get the errors into Kibana and b) query Kibana to get them out again.

benthorner:
Removed CD warning.

leenagupte:
Thanks for all of the comments and suggestions. 

I've talked to the performance analyst on the team, and we're going to start by sending more descriptive errors to GA than the one generic one that we send at the moment. 

If we see that more users are failing in for one reason (e.g. validation) that the others, we will add in Sentry logging at the point of failure for a few days so we can properly investigate.

PR #2076, opened 2020-11-11
Render breadcrumbs dynamically

JonathanHallam:
This removes the hardcoded breadcrumbs now that the `find-coronavirus-local-restrictions` page has a content item. There are some other code changes that are in place either to allow the tests to pass or to ensure that the landing page can be accessed without the hardcoded breadcrumbs.

The only visual difference is the addition of the footer. The breadcrumbs and the sidebar will remain the same
#### Computer
Screenshot 2020-11-12 at 12 01 39

#### Phone
Screenshot 2020-11-12 at 12 01 46

#### Computer
Screenshot 2020-11-12 at 12 00 43

#### Phone
Screenshot 2020-11-12 at 12 01 02

### DIFF
--- a/app/views/coronavirus_local_restrictions/show.html.erb
+++ b/app/views/coronavirus_local_restrictions/show.html.erb
@@ -4,10 +4,7 @@
 %>
 
 <% content_for :breadcrumbs do %>
-  <%= render 'govuk_publishing_components/components/breadcrumbs', {
-    breadcrumbs: breadcrumbs,
-    collapse_on_mobile: true
-  } %>
+  <%= render 'govuk_publishing_components/components/contextual_breadcrumbs', content_item: @content_item %>
 <% end %>
 
 <% content_for :title, title %>
--- a/app/views/coronavirus_local_restrictions/show.html.erb
+++ b/app/views/coronavirus_local_restrictions/show.html.erb
@@ -49,14 +49,7 @@
   </div>
   <div class="govuk-grid-column-one-third">
     <div class="coronavirus-local-restriction__related-navigation">
-      <%= render 'govuk_publishing_components/components/related_navigation', {
-        "content_item": {
-          "links" => {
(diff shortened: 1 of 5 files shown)


leenagupte (commented):
Looks good to me! 🎉 

I haven't approved this yet, because there's still the lingering question of whether we need to  add the contextual sidebar as well.

benthorner:
Removed CD warning.

hannahjump (commented):
Footer looks good from my perspective

app/views/coronavirus_local_restrictions/show.html.erb
```diff
@@ -49,7 +49,7 @@
   </div>
   <div class="govuk-grid-column-one-third">
     <div class="coronavirus-local-restriction__related-navigation">
-      <%= render 'govuk_publishing_components/components/related_navigation', {
+      <%= render 'govuk_publishing_components/components/contextual_sidebar', {
```
leenagupte, line 52: Could we tag the related content in content-tagger and pass in the whole content instead, like we do for the breadcrumb?

For example: https://github.com/alphagov/government-frontend/blob/960e79d99c05f48cb110d007b11ad31e7d802970/app/views/shared/_sidebar_navigation.html.erb#L6

leenagupte (commented):
Looks really good! 🎉 
I just have lingering question about the sidebar.

app/views/coronavirus_local_restrictions/show.html.erb
JonathanHallam: I did try this and it wouldn't render anything. I wonder if the `parsed_content` thing in your example is doing something special. I'll have a look :)

JonathanHallam: Sorted now that it's been properly tagged in content tagger

leenagupte (approved):
Looks good! 👍 💯 
Yay for deleting snowflake config! 🥇

PR #2083, opened 2020-11-18
Adjust social links hidden text

danacotoran:
## What
Amend the hidden text for organisation social media links. So that AT users can have a better experience.

## Why
There are instances where it does not make sense to have "Follow on" hidden text before the social media link text. Prime example here: https://www.gov.uk/government/organisations/land-registry

Screenshot 2020-11-18 at 11 46 03

Screen reader software would read these social links as 
- Follow on Sign up for email alerts
- Follow on Read our blog posts
- Follow on Follow us on Twitter @HMLandRegistry

...etc etc

It makes sense to add "Follow on" for more context when the link text is just one word (for instance "Follow on Facebook" would make more sense to a screen reader user than just "Facebook"). 
It does not make sense to add hidden text when the social link text already contains a verb like "Watch our videos on YouTube".

https://trello.com/c/shAQ6AQu

### DIFF
--- a/app/presenters/organisations/what_we_do_presenter.rb
+++ b/app/presenters/organisations/what_we_do_presenter.rb
@@ -17,7 +17,7 @@ module Organisations
         links << {
           href: link["href"],
           text: link["title"],
-          hidden_text: "Follow on",
+          hidden_text: link["title"].include?('Sign up') ? "" : "Follow on",
           icon: link["service_type"],
         }
       end
--- a/app/presenters/organisations/what_we_do_presenter.rb
+++ b/app/presenters/organisations/what_we_do_presenter.rb
@@ -14,10 +14,11 @@ module Organisations
       links = []
 
       org.social_media_links.each do |link|
+        link_has_cta = ["Sign up", "Follow", "Watch", "Read"].any? { |cta| link["title"].include?(cta) }
         links << {
           href: link["href"],
           text: link["title"],
-          hidden_text: link["title"].include?('Sign up') ? "" : "Follow on",
+          hidden_text: link_has_cta ? "" : "Follow on",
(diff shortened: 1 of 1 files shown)


app/presenters/organisations/what_we_do_presenter.rb
```diff
@@ -14,10 +14,11 @@ def share_links
       links = []
 
       org.social_media_links.each do |link|
+        link_has_cta = ["Sign up", "Follow", "Watch", "Read"].any? { |cta| link["title"].include?(cta) }
```
chris-gds, line 17: @danacotoran are these Icon Content's likely to change? Probably not, but I'm wondering if they update and this gets skipped.  Eg the gem has ["Get email alerts"](https://components.publishing.service.gov.uk/component-guide/subscription-links) rather than "Sign up"

danacotoran: Yep as you've pointed out, it's possible the links might be updated to use different wording.  But I've checked a fairly large number of organisation pages with social media links and from what I could see if any "call to action" verbs are used for their social media links, it's usually the ones I listed.

As you can tell, this is somewhat of a hack/workaround to make sure that the issue is fixed in most cases.
Since we're dealing with language, which is so fluid, I think it would be tricky to implement a catch-all solution, so my objective with this was more in the vein of "fix most instances where the issue would occur" than "make it perfect".

chris-gds: >  I think it would be tricky to implement a catch-all solution, so my objective with this was more in the vein of "fix most instances where the issue would occur" than "make it perfect".

Fair.

chris-gds (approved):

PR #2090, opened 2020-11-25
Update content on postcode checker results pages before it goes live again

leenagupte:
Trello: https://trello.com/c/xMzgVArN

# What's changed and why?

We need to iterate the content in the postcode checker before it goes live again, including:
- 'tiers' wording
- information about searching for where you live 
- information about following national restrictions until 2nd dec

# Expected changes
### Before
Screenshot 2020-11-25 at 12 45 30
### After
Screenshot 2020-11-25 at 12 28 49

### Before
Screenshot 2020-11-25 at 12 45 03
### After
Screenshot 2020-11-25 at 12 27 59

### Before
Screenshot 2020-11-25 at 12 44 36
### After
Screenshot 2020-11-25 at 12 29 56

# Note to reviewer
I haven't removed all of the labels (for example for future restrictions) from lookup.yml or the logic in the views as I'm conscious of the fact that a lot of this will have to be put back after 2nd December and I'm trying to make that as easy as possible.

### DIFF
--- a/app/views/coronavirus_local_restrictions/_national_restrictions.html.erb
+++ b/app/views/coronavirus_local_restrictions/_national_restrictions.html.erb
@@ -1,12 +1,12 @@
-<%= render "govuk_publishing_components/components/heading", {
-    text: t("coronavirus_local_restrictions.results.national_restrictions.heading"),
-    font_size: "m",
-    margin_bottom: 4
-} %>
-
 <%= render "govuk_publishing_components/components/action_link", {
     text: t("coronavirus_local_restrictions.results.national_restrictions.action_label"),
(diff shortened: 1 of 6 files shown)


hannahjump:
Looks good!

Rosa-Fox (commented):
[This line](https://github.com/alphagov/collections/blob/477f31d1de0480f8f886bd8e1566b11e65568bb9/app/views/coronavirus_local_restrictions/_tier_three_restrictions.html.erb#L36) needs to be

`text: t("coronavirus_local_restrictions.results.guidance_label", level: 3)`

Otherwise it displays `Find out what the tier %{level} rules are` on the view but doesn't substitute the variable.

Rosa-Fox (approved):

PR #2092, opened 2020-11-25
Adding full tier name

hannahjump:
The first time a tier is mentioned it should be the full name

Tier 1: medium alert
Tier 2: high alert
Tier 3: very high alert


leenagupte (approved):
💯

PR #2095, opened 2020-11-26
Added link to tiers overview guidance

hannahjump:
The other links are old, specific guidance for each tier is not yet published and might not be today

### DIFF
--- a/config/locales/en/lookup.yml
+++ b/config/locales/en/lookup.yml
@@ -65,7 +65,7 @@ en:
         wales:
           guidance:
             label: Find out what you can or cannot do (Welsh Government)
-            link: "https://gov.wales/local-lockdown"
+            link: "https://gov.wales/coronavirus-regulations-guidance"
         northern_ireland:
           guidance:
             label: Find out what you can or cannot do (Northern Ireland Government)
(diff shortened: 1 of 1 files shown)


config/locales/en/lookup.yml
```diff
@@ -65,16 +65,16 @@ en:
...
           guidance:
             label: Find out what you can or cannot do (Northern Ireland Government)
             link: "https://www.nidirect.gov.uk/articles/coronavirus-covid-19-regulations-and-localised-restrictions"
       no_information:
         heading: There is no information about the restrictions in this area
         guidance:
-          label: Full list of Local COVID Alert Levels by area
-          link: "/guidance/full-list-of-local-covid-alert-levels-by-area"
+          label: Full list of tiers by area
+          link: "https://www.gov.uk/guidance/full-list-of-local-restriction-tiers-by-area"
```
GDSNewt, line 77: This link "https://www.gov.uk/guidance/full-list-of-local-restriction-tiers-by-area" currently gives a 404, I assume this is because we don't have the full list ready yet?

hannahjump: ah yes, it's not yet published... hmmmm let me change that

GDSNewt:
The labels are not quite accurate for the links, but given the rapidly changing nature of this happy to approve for now 👍

GDSNewt (approved):

PR #2096, opened 2020-11-26
Update restrictions yaml with new data

JonathanHallam:


Rosa-Fox:
Cherry picked onto https://github.com/alphagov/collections/pull/2093.

PR #2097, opened 2020-11-26
Image optimisation for the DIT training page

Nooshu:
Images used on this page were a little big in terms of filesize so have optimised them.

| Filename        | Before (KB) | After (KB) | Diff | Converted to jpeg |
|-----------------|-------------|------------|------|-------------------|
| DIT-banner@2x   | 1000        | 176        | -82% | Yes               |
| DIT-small-image | 73.7        | 21.8       | -70% | Yes               |

Images converted to jpeg, but very hard to spot the difference. And given they are background images 100% clarity isn't a requirement for the images for users anyway.


alex-ju (approved):

andysellick (approved):

hannako (approved):
When I added the original images I think I put them through https://imageoptim.com/. What do you use to make them so tiny? Or is it that they are now jpegs that makes the difference?

Nooshu:
Imageoptim is an excellent tool. But if it encounters a PNG I believe it will still compress it in a lossless format. So the file will be smaller, but still huge. As you say, It's the converting to lossy jpeg that has reduced the size here.

My go to tool for this is https://squoosh.app/, an in-browser app where you can easily adjust settings and compare the original vs the optimised version.

PR #2100, opened 2020-11-26
Cache content store requests for the checker

sihugh:
This doesn't change that often. 1 minute should relieve some load.

This app is already set up to use memcached as its cache.


brucebolt (approved):

bilbof (approved):
LGTM

thomasleese (approved):
Makes sense to me

PR #2105, opened 2020-11-27
Devolved nation content update and add buttons

leenagupte:
Trello: https://trello.com/c/MJzQaElt

# Expected changes

| Before             |  After|
:-------------------------:|:-------------------------:
Screenshot 2020-11-27 at 16 49 57  |  Screenshot 2020-11-27 at 17 02 22
Screenshot 2020-11-27 at 16 49 01  |  Screenshot 2020-11-27 at 16 48 49
Screenshot 2020-11-27 at 16 00 38 |  Screenshot 2020-11-27 at 16 32 25

### DIFF
--- a/test/integration/coronavirus_local_restrictions_test.rb
+++ b/test/integration/coronavirus_local_restrictions_test.rb
@@ -289,15 +289,15 @@ class CoronavirusLocalRestrictionsTest < ActionDispatch::IntegrationTest
   end
 
   def then_i_see_the_results_for_wales
-    assert page.has_text?("Welsh Government")
+    assert page.has_text?(I18n.t("coronavirus_local_restrictions.results.devolved_nations.wales.guidance.label"))
   end
 
   def then_i_see_the_results_for_scotland
-    assert page.has_text?("Scottish Government")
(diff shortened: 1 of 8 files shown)


test/integration/coronavirus_local_restrictions_test.rb
```diff
@@ -289,15 +289,15 @@ def then_i_see_details_of_national_restrictions
   end
 
   def then_i_see_the_results_for_wales
-    assert page.has_text?("Welsh Government")
```
JonathanHallam, line 292: Is there a reason we shouldn't use the I18n here? Figure it would avoid having to update this in the future

JonathanHallam:
Nice one, happy with this other than small question re. I18n

test/integration/coronavirus_local_restrictions_test.rb
leenagupte: @JonathanHallam That's a good point. I've updated the test and fixed up.

JonathanHallam: nice one, thanks :)

leenagupte:
@JonathanHallam Thanks for your review. I've updated the tests to use the I18n translation and fixed up.

JonathanHallam (approved):
Looks good, thank you :)

PR #2111, opened 2020-11-30
End of national restrictions changes

leenagupte:
Trello: https://trello.com/c/SfF6xuKh

# Expected changes
## Tier 1
current_tier_1

## Tier 2
current_tier_2

## Tier 3
current_tier_3

## Tier 1 to 2
future_tier_1_to_2

## Tier 1 to 3
future_tier_1_to_3

## Tier 2 to 1
future_tier_2_to_1

## Tier 2 to 3
future_tier_2_to_3

## Tier 3 to 1
future_tier_3_to_1

## Tier 3 to 2
future_tier_3_to_2

# How to test locally
Pick an area in the [restrictions file](https://github.com/alphagov/collections/blob/master/lib/local_restrictions/local-restrictions.yaml) and change the entry in the YAML to match each of the scenarios above.


config/locales/en/lookup.yml
```diff
@@ -71,8 +71,8 @@ en:
         heading: From %{date}
         guidance_label: What you can and cannot do from %{date}
         level_one:
-          alert_level: "From %{date} this area will be in Tier 1: Medium alert."
+          alert_level: "%{area} will be in Tier 1: Medium alert."
         level_two:
-          alert_level: "From %{date} this area will be in Tier 2: High alert."
+          alert_level: "%{area} will be in Tier 2: High alert."
```
Rosa-Fox, line 76: This has a full stop but `alert_level` doesn't for the other tiers

leenagupte: I'm not sure what this means, however it looks like this content is being duplicated on line 43, so I'll DRY-up it.

app/views/coronavirus_local_restrictions/_future_restrictions.html.erb
```diff
@@ -2,43 +2,29 @@
...
-
-    <%= render "govuk_publishing_components/components/action_link", {
-      text: t("coronavirus_local_restrictions.results.future.guidance_label", date: date),
-      href: t("coronavirus_local_restrictions.results.level_three.guidance_link"),
-      simple: true,
-      margin_bottom: 9
-    } %>
   <% end %>
+
+  <%= render "govuk_publishing_components/components/button", {
```
Rosa-Fox, line 24: The button needs to be within the conditional as well, because the `href` changes depending on the tier. At the moment the future restrictions button always links to tier 3 guidance. I guess we could put rendering the button component into a partial and then pass in the relevant href and render the partial in each of the conditions?

leenagupte: Oooh good spot! Will fix!

leenagupte:
@Rosa-Fox Thank you for being so thorough in your reviews. I've fixed the mistake and pushed up.

Rosa-Fox (approved):

PR #2120, opened 2020-12-04
Update wales link

hannahjump:
Wales link from start page is old and 404ing. Changing to same link as in checker.


JonathanHallam (approved):
looks good, thanks

PR #2122, opened 2020-12-04
Implement cross-domain tracking on accounts redirects

barrucadu:
There are some links to www.gov.uk which may redirect to the accounts
domain.  We want cross-domain tracking to work in the case where there
is a redirect, which means we need to pass along the cross-domain _ga
parameter.

Since these are links to www.gov.uk, they don't get decorated
automatically.  So instead, decorate links with a specific data
attribute.

See also github.com/alphagov/static/pull/2346

---

[Trello card](https://trello.com/c/ISYymNsg/426-fix-cross-domain-tracking-on-redirects)

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -12,6 +12,7 @@
 //= require support
 //= require browse-columns
 //= require organisation-list-filter
+//= require ga_cross_domain_accounts_tracking
 //= require modules/current-location
 //= require modules/feeds.js
 //= require modules/toggle-attribute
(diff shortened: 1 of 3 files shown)


andysellick:
Since this is going to be more widespread, would it be better to put this script into static (or even the components gem, and call it in static) instead?

barrucadu:
@andysellick Sounds good to me, I'll move this (and the form logic from the account manager & finder frontend) into the components gem

barrucadu:
This is going to be done through govuk_publishing_components so the only change needed in this app is upgrading the gem after the new version comes out.

PR #2088, opened 2020-11-24
Update DIT landing page

alex-ju:
## What

Update the `/eubusiness` layout and prepare it for a multi-lingual setup.

## Why

To unclutter the page and streamline the update process.

## Ready when

- [x] design was reviewed
- [x] links were added
- [x] content was reviewed

## Visual changes

We currently have some issues with aligning the content and the image to the classic 2/3 – 1/3 containers at any breaking points. The approach we took for the `/transition` layout feels a bit hacky, as we're serving [an image with 2/3 white on the left-hand side](https://www.gov.uk/assets/collections/transition-2020-header-background-8fd574542d5960200a694fbbbff2a4f0fac92b6f809db66b1bbc5dcf38573e1c.jpg) so I come up with an alternative approach in this PR that we can re-use in the `/transition` page.

👉 [Preview the page](https://govuk-collec-update-dit-gthtvd.herokuapp.com/eubusiness)

### Before

Screenshot 2020-11-24 at 14 41 45

### After
Screenshot 2020-11-24 at 14 59 08

Note: The capture above shows how the page would look like with translation navigation, however, we might need to roll this out in English only first, then update with languages as the content becomes available.

### Preview
localhost_3070_eubusiness

[Trello card](https://trello.com/c/u7gFwzY3)

PR #2123, opened 2020-12-04
Revert #2088

alex-ju:
A revert of #2088 was requested 🙃

As the commits in #2088 ended up quite scattered in the git history, GitHub was unable to do a UI-based revert, so I did it using `git revert`.

To double-check the revert I compared [the version we had before #2088](https://web.archive.org/web/20201201234049/https://www.gov.uk/eubusiness) with [the result of this PR](https://govuk-collec-revert-pr--wsutvu.herokuapp.com/eubusiness).

Before merging this we'll need to re-publish the old translation pages and remove the redirects.

[Trello card](https://trello.com/c/u7gFwzY3)

### DIFF
--- a/app/assets/stylesheets/views/_dit-landing-page.scss
+++ b/app/assets/stylesheets/views/_dit-landing-page.scss
@@ -119,12 +119,6 @@ $dark-grey: #282828;
   max-width: 100%;
 }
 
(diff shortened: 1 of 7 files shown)


sihugh (approved):
👍  🔙

PR #2124, opened 2020-12-04
Audit headings throughout collections

owenatgov:
## What
This PR:

- Replaces instances (where possible) of raw html h tags with the [govuk publishing components heading component](https://components.publishing.service.gov.uk/component-guide/heading)
- Replaces instances of visually hidden headings using the `visuallyhidden` or `visually-hidden` classes with the most up to date `govuk-visually-hidden` class
- Removes `h3` tags from the list items in the latest changes view
- Fixes a stray broken `p` tag on the covid landing page

## Why
As part of the work by the govuk accessibility team, we are ensuring that where possible, the govuk publishing components are being used. This ensures that any bug fixes or changes to our markup can be rolled out quickly and with greater confidence that a given problem has been fixed everywhere within our apps.

The latter 3 changes are just some little fixes whilst I'm in the code. The change to latest changes in particular is because the `h3` is overly verbose, especially because the view is styling the heading to not look like a heading. Screen reader users can navigate this page via tabbing or links and this change makes no difference to sighted users.

I've omitted a couple of instances of raw headings (such as visually hidden headings) that have some styling specific to their view, where headings aren't being used the way that the heading component should be used, making said component a poor use case. Probably these pages need a design review but that's not within the scope of this work.

## Pages tested against

- https://www.gov.uk/eubusiness
- https://www.gov.uk/transition
- https://www.gov.uk/browse/benefits/universal-credit
- https://www.gov.uk/coronavirus
- https://www.gov.uk/government/organisations
- https://www.gov.uk/topic/business-tax/alcohol-duties/latest

No visual changes.

[Card](https://trello.com/c/dxbqoY7q/545-use-heading-component-in-collections-app)

### DIFF
--- a/app/views/dit_landing_page/_guidance.html.erb
+++ b/app/views/dit_landing_page/_guidance.html.erb
@@ -2,7 +2,12 @@
   <div class="govuk-width-container">
     <div class="govuk-grid-row">
       <div class="govuk-grid-column-two-thirds">
-        <%= tag.h2 sanitize(t("dit_landing_page.ribbon_title")), class: "govuk-heading-l" %>
+        <%= render "govuk_publishing_components/components/heading", {
+          text: sanitize(t("dit_landing_page.ribbon_title")),
+          font_size: "l",
+          heading_level: 2,
+          margin_bottom: 6,
+        } %>
       </div>
     </div>
     <div class="govuk-grid-row">
--- a/app/views/transition_landing_page/_brexit_finders.html.erb
+++ b/app/views/transition_landing_page/_brexit_finders.html.erb
@@ -1,8 +1,11 @@
 <div class="govuk-grid-row govuk-!-margin-bottom-8">
   <div class="govuk-grid-column-one-third">
-    <h2 class="govuk-heading-m">
-      <%= t("transition_landing_page.topic_section_header") %>
-    </h2>
+    <%= render "govuk_publishing_components/components/heading", {
+      text: t("transition_landing_page.topic_section_header"),
+      heading_level: 2,
+      font_size: "m",
+      margin_bottom: 3,
+    } %>
   </div>
   <div class="govuk-body govuk-grid-column-two-thirds landing-page__finders">
     <p class="govuk-body"><%= t("transition_landing_page.topic_section_subheading") %></p>
--- a/test/components/header_notice_test.rb
+++ b/test/components/header_notice_test.rb
@@ -78,12 +78,12 @@ class HeaderNoticeTest < ComponentTestCase
 
   test "does not renders a heading if no heading is passed in" do
     render_component(branded_notice.without(:heading))
-    assert_select ".app-c-header-notice__heading", count: 0
+    assert_select ".gem-c-heading", count: 0
   end
 
   test "renders a heading if heading is passed in" do
     render_component(branded_notice)
-    assert_select ".app-c-header-notice__heading", text: branded_notice[:heading]
+    assert_select ".gem-c-heading", text: branded_notice[:heading]
   end
 
   test "does not renders a list if no list is passed in" do
--- a/test/support/coronavirus_landing_page_steps.rb
+++ b/test/support/coronavirus_landing_page_steps.rb
@@ -166,16 +166,16 @@ module CoronavirusLandingPageSteps
   def then_i_can_see_the_timeline
     assert page.has_selector?("h2", text: "Recent and upcoming changes")
 
-    assert page.has_selector?(".covid-timeline__item-heading", text: "18 September")
+    assert page.has_selector?(".covid-timeline .gem-c-heading", text: "18 September")
     assert page.has_selector?(".covid-timeline__item .gem-c-govspeak", text: "If you live, work or travel in the North East, you need to follow different covid rules")
 
-    assert page.has_selector?(".covid-timeline__item-heading", text: "15 September")
+    assert page.has_selector?(".covid-timeline .gem-c-heading", text: "15 September")
     assert page.has_selector?(".covid-timeline__item .gem-c-govspeak", text: "If you live, work or visit Bolton, you need to follow different covid rules")
 
-    assert page.has_selector?(".covid-timeline__item-heading", text: "14 September")
+    assert page.has_selector?(".covid-timeline .gem-c-heading", text: "14 September")
     assert page.has_selector?(".covid-timeline__item .gem-c-govspeak", text: "People must not meet in groups larger than 6 in England. There are exceptions to this ‘rule of 6’")
 
-    assert page.has_selector?(".covid-timeline__item-heading", text: "24 July")
+    assert page.has_selector?(".covid-timeline .gem-c-heading", text: "24 July")
     assert page.has_selector?(".covid-timeline__item .gem-c-govspeak", text: "Face coverings are mandatory in shops")
   end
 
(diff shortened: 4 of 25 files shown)


app/assets/stylesheets/views/_transition-landing-page.scss
```diff
@@ -168,11 +168,8 @@ $green: #00703c;
   padding: 0 3px;
 }
 
-.landing-page__page_title {
-  @include govuk-font($size: 48, $weight: bold);
-  color: $dark-blue;
+.landing-page__title-container {
```
chris-gds, line 171: Potentially this could be removed and added directly on the markup using the Design system:
 `govuk-!-margin-top-7`

app/views/transition_landing_page/_take_action.html.erb
```diff
@@ -2,9 +2,12 @@
...
       <div class="govuk-grid-column-full">
-        <h2 class="govuk-heading-l">
-          <%= t("transition_landing_page.take_action_title") %>
-        </h2>
+        <%= render "govuk_publishing_components/components/heading", {
+          text: t("transition_landing_page.take_action_title"),
+          heading_level: 2,
+          font_size: "l",
+          margin_bottom: 6,
+        } %>
```
chris-gds, line 10: 👍🏼

app/views/dit_landing_page/_guidance.html.erb
```diff
@@ -2,9 +2,12 @@
...
       <div class="govuk-grid-column-two-thirds">
-        <h2 class="govuk-heading-l">
-          <%= t("dit_landing_page.guidance_title") %>
-        </h2>
+        <%= render "govuk_publishing_components/components/heading", {
+          text: t("dit_landing_page.guidance_title"),
+          font_size: "l",
+          heading_level: 2,
+          margin_bottom: 6,
+        } %>
```
chris-gds, line 10: 👍🏼

test/components/header_notice_test.rb
```diff
@@ -78,12 +78,12 @@ def list_with_links
 
   test "does not renders a heading if no heading is passed in" do
     render_component(branded_notice.without(:heading))
-    assert_select ".app-c-header-notice__heading", count: 0
+    assert_select "#header-notice-heading", count: 0
```
chris-gds, line 81: I prefer this patten of testing using ID. Maybe even with `test-`
Part of me thinks this is worth updating like this in more places / adding notes on the gem.

chris-gds (changes requested):

test/components/header_notice_test.rb
maxgds: I disagree here. We generally follow BEM guidelines for our CSS and they advocate avoiding IDs - when used as selectors they cause problems with specificity and make override attempts difficult. It's not an issue in the way it's being used right here, but it opens the door to problems further down the line. I also feel it's also rarely a good idea to pollute user facing code with hooks meant solely for testing. We have a convention of the outer wrapper of our components being .app-c-thing or .gem-c-thing and every other class within cascades from that so we should be able to rely on these when it comes to the testing side.

maxgds: Hmm, having looked at where the ID is introduced I withdraw this objection! However I'm not convinced we're doing the right thing by introducing IDs like that, going to take it out for discussion elsewhere.

chris-gds: > I disagree here. We generally follow BEM guidelines for our CSS and they advocate avoiding IDs - when used as selectors they cause problems with specificity and make override attempts difficult. It's not an issue in the way it's being used right here, but it opens the door to problems further down the line. 

Ah yes 100%, but _not_ for styling _only_ for testing purposes. This way the tests are independent of any gem class updates, or tag changes - they are hooked onto the element itself and are clearly prefixed.  If the `gem-` class changed to something else many tests would fail and need to be manually updated (this is happening here) Also, it's clearer to Devs at a glance updating files and tests.

Typically I see something like `[data-test]` ([rather than just ID](https://kentcdodds.com/blog/making-your-ui-tests-resilient-to-change)) being used within this context so that testing selectors and styling are separated.  If something like this was done, the Dev experience becomes simplified a testing selector is clear and the pattern is consisent.

> I also feel it's also rarely a good idea to pollute user facing code with hooks meant solely for testing. 

Hmm I'm not sure I share this concern, the extra code might be a little undesirable but potentially this could even stripped out on production at build if it's all a consistent. The benefits would be a clear consistent pattern around testing hooks that ideally work independent of gem updates.

chris-gds: @maxgds just seen the update, 100% this is worthy of a FE discussion f'sure.

owenatgov: Reading this thread, it reminded me of a time when I had it out a bit with Ian on an identical issue on a different PR. I think Max's point makes sense, making user-facing markup less verbose and ensuring that everything we send the user has an intent for that user is a good thing. However I do feel like this is a tricky situation because I share Chris' concern that hooking tests up to `gem-` classnames is a risk. In theory this is just a case of updating tests before you merge a dependabot PR but this seems to be a problem in govuk and a much wider discussion. I'm happy with either using the gem classnames or creating a quick `app-c` classname instead of an id.

I do think this is something worth discussing in a govuk frontenders catchup as both implementations feel imperfect. I wonder if we should be interrogating our tests and how they interact with the gem components.

maxgds: Yeah, you add 1 ID and it's a few bytes of code, but when that code gets downloaded several million times it has impact.. If we're changing the gem classes that's pretty fundamental then I'd expect tests to break and need attention - that's a feature not a bug, we ensure we haven't changed the expectations of the app when we look at what it thought it would receive. And we shouldn't be able to merge code that's breaking our tests. I don't think I want tests that don't break when the thing that's being tested fundamentally does. I could maybe be convinced with examples. I say go either way right now (although I prefer avoiding the ID) and drop this conversation onto an agenda somewhere, whether frontenders general catchup or the specific gov.uk frontenders one.

chris-gds: Yeah let's pop it up for discussion, at the moment I _tend_ to see chained classes  + tags being used within testing files, so I would stick to this convention for this context of getting this PR merged.  

Maybe raise a a RFC / general topic within one of the frontend meetings??

app/views/transition_landing_page/_brexit_finders.html.erb
```diff
@@ -1,8 +1,10 @@
 <div class="govuk-grid-row govuk-!-margin-bottom-8">
   <div class="govuk-grid-column-one-third">
-    <h2 class="govuk-heading-m">
-      <%= t("transition_landing_page.topic_section_header") %>
-    </h2>
+    <%= render "govuk_publishing_components/components/heading", {
+      text: t("transition_landing_page.topic_section_header"),
+      heading_level: 2,
+      font_size: "m",
+    } %>
```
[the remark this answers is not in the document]
owenatgov: Good spot!

test/support/coronavirus_landing_page_steps.rb
```diff
@@ -166,16 +166,16 @@ def then_i_can_see_the_nhs_banner
   def then_i_can_see_the_timeline
     assert page.has_selector?("h2", text: "Recent and upcoming changes")
 
-    assert page.has_selector?(".covid-timeline__item-heading", text: "18 September")
+    assert page.has_selector?("#covid-timeline h3", text: "18 September")
```
chris-gds, line 169: Thinking about the other conversation about testing and IDs, should this be swapped to class in this instance?

injms (approved):

PR #2127, opened 2020-12-07
Make the postcode checker heading line break in the right place

DilwoarH:
## What
The results of the postcode checker have the heading 'This area is in [full tier name]'.

We want to make a line break so the full tier name is always on one line so people can easily understand it. So the options will display as:

This area is in
Tier 1: Medium alert

This area is in
Tier 2: High alert

This area is in
Tier 3: Very High alert

## Why
So people can easily identify their tier

## Before

Screenshot 2020-12-08 at 10 36 14

## After

Screenshot 2020-12-08 at 10 36 27

https://trello.com/c/LCDIMFk5/967-make-the-postcode-checker-heading-line-break-in-the-right-place

### DIFF
--- a/app/views/coronavirus_local_restrictions/_tier_one_restrictions.html.erb
+++ b/app/views/coronavirus_local_restrictions/_tier_one_restrictions.html.erb
@@ -1,19 +1,26 @@
+<% tier_heading = capture do %>
+  <%= t("coronavirus_local_restrictions.results.level_one.heading_pretext") %>
+  <span class="coronavirus-local-restriction__heading-tier-label">
+    <%= t("coronavirus_local_restrictions.results.level_one.heading_tier_label") %>
+  </span>
+<% end %>
+
 <%
   title = if @restriction.present? && @restriction.future.present?
     t("coronavirus_local_restrictions.results.changing_levels_title")
   else
-    t("coronavirus_local_restrictions.results.level_one.heading")
+    tier_heading
   end
 %>
 
-<% content_for :title, title %>
+<% content_for :title, strip_tags(title) %>
 
 <%= tag.div :data => {
   :module => "coronavirus-track-local-restriction-results",
   "tracking-label" => "Tier one: #{@location_lookup.lower_tier_area_name}"
 } do %>
   <%= render "govuk_publishing_components/components/title", {
-    title: title
+    title: sanitize(title)
   } %>
 
   <p class="govuk-body">
@@ -22,7 +29,7 @@
 
   <% if @restriction.present? && @restriction.future.present? %>
     <%= render "govuk_publishing_components/components/heading", {
-      text: t("coronavirus_local_restrictions.results.level_one.heading"),
+      text: sanitize(tier_heading),
       font_size: "m",
       margin_bottom: 4
     } %>
--- a/config/locales/en/lookup.yml
+++ b/config/locales/en/lookup.yml
@@ -36,15 +36,18 @@ en:
       guidance_label: Find out what the Tier %{level} rules are
       match: "We've matched the postcode"
       level_one:
-        heading: "This area is in Tier 1: Medium alert"
+        heading_pretext: This area is in
+        heading_tier_label: "Tier 1: Medium alert"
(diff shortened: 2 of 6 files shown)


config/locales/en/lookup.yml
```diff
@@ -36,15 +36,15 @@ en:
...
         changing_alert_level: "At the moment %{area} is in Tier 1."
         guidance_link: "/guidance/tier-1-medium-alert"
       level_two:
-        heading: "This area is in Tier 2: High alert"
+        heading: "This area is in <div>Tier 2: High alert</div>"
         changing_alert_level: "At the moment %{area} is in Tier 2."
         guidance_link: "/guidance/tier-2-high-alert"
       level_three:
-        heading: "This area is in Tier 3: Very High alert"
+        heading: "This area is in <div>Tier 3: Very High alert</div>"
```
andysellick, line 47: Not sure that having a `div` inside a heading is valid HTML. Could this be a span with a class to prevent wrapping on the whitespace instead? (might be something in the DS that does this already).

andysellick: `white-space: nowrap`

DilwoarH: @andysellick - updated

andysellick: Great.

app/views/coronavirus_local_restrictions/_tier_one_restrictions.html.erb
```diff
@@ -1,19 +1,23 @@
 <%
+  tier_heading_pretext = t("coronavirus_local_restrictions.results.level_one.heading_pretext")
+  tier_heading_label = t("coronavirus_local_restrictions.results.level_one.heading_tier_label")
+  tier_heading = "#{tier_heading_pretext} <span class=\"coronavirus-local-restriction__heading-tier-label\">#{tier_heading_label}</span>"
+
```
leenagupte, line 5: This feels like a lot of logic. As `tier_heading_pretext` and `tier_heading_label` are not used anywhere else in the view would it be easier to just have one assignment? 

Also, rather interpolating CSS in a string, which feels weird, could the `tier_heading` be assigned in a `capture` block like:

```
<% tier_heading = capture do %>
  <%= t("coronavirus_local_restrictions.results.level_one.heading_pretext") %>
  <span class="coronavirus-local-restriction__heading-tier-label">
    <%= t("coronavirus_local_restrictions.results.level_one.heading_tier_label") %>
  </span>
<% end %>
```

leenagupte:
@DilwoarH Can you add some screenshots? For example, what does the page look like if there are future restrictions?

app/views/coronavirus_local_restrictions/_tier_one_restrictions.html.erb
```diff
@@ -1,19 +1,23 @@
...
   title = if @restriction.present? && @restriction.future.present?
     t("coronavirus_local_restrictions.results.changing_levels_title")
   else
-    t("coronavirus_local_restrictions.results.level_one.heading")
+    tier_heading
   end
 %>
 
-<% content_for :title, title %>
+<% content_for :title, strip_tags(title) %>
```
leenagupte, line 13: I'm a bit confused about why `strip_tags` is needed here and not on line 29. Surely the natural wrapping needs to be allowed in both places? Is line 29 a fluke because the text for `tier_heading` happens to fit on one line?

leenagupte (commented):
I've run it locally and all looks good. I've left some inline comments questions and suggestions.

app/views/coronavirus_local_restrictions/_tier_one_restrictions.html.erb
DilwoarH: Didn't know this - thanks Leena!

app/views/coronavirus_local_restrictions/_tier_one_restrictions.html.erb
DilwoarH: this is for the browser title so any HTML will display as text

leenagupte (approved):
👍

PR #2132, opened 2020-12-08
Fix leaky locale assignment

kevindew:
A number of controllers in this app were using a before_action to set
the value of `I18n.locale` based on the request locale. This changes the
locale for the entire application, including subsequent requests. [1]

We learnt of this becoming a problem when we heard reports that users of
"/find-coronavirus-local-restrictions" were occasionally seeing a
back button in Welsh (Yn ôl). Investigating further we learnt that this
problem could be replicated by visiting a page in Welsh in this app (such as
[2]) and then visiting the postcode checker. The reason it was only
the back button that exhibited this problem is because it's the only
content with an available Welsh translation on this page.

To resolve this we need to change the approach used to change locale to
be encapsulated to only the particular action that is running. The Rails
documentation explains how this is done [1].

I've taken this as an opportunity to DRY up the locale setting so we
have one action on the superclass that other controllers can use.

[1]: https://guides.rubyonrails.org/i18n.html#managing-the-locale-across-requests
[2]: https://www.gov.uk/government/organisations/office-of-the-secretary-of-state-for-wales.cy

----

Before:
welsh-leak

After:

welsh-fixed-leak

### DIFF
--- a/test/controllers/application_controller_test.rb
+++ b/test/controllers/application_controller_test.rb
@@ -3,9 +3,14 @@ require "test_helper"
 class ConcreteTestController < ApplicationController
   enable_request_formats json: :json, js_or_atom: %i[js atom]
   protect_from_forgery except: :js_or_atom
+  around_action :switch_locale
 
   def test
     render html: "ok"
+  end
+
+  def locale_check
+    render html: I18n.locale
   end
 
   def json
@@ -29,6 +34,7 @@ describe ConcreteTestController do
     with_routing do |map|
       map.draw do
         get "/test", to: "concrete_test#test"
+        get "/locale", to: "concrete_test#locale_check"
         get "/test", to: "concrete_test#json"
         get "/test", to: "concrete_test#js_or_atom"
       end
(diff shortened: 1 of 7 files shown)


test/controllers/application_controller_test.rb
```diff
@@ -105,4 +111,14 @@ def with_test_routing
...
+
+  it "changes locale only for the duration of the request" do
+    with_test_routing do
+      get :locale_check, params: { locale: "cy" }
+      assert_response :success
+      assert_equal "cy", response.body
+
+      assert_equal :en, I18n.locale
+    end
+  end
```
leenagupte, line 123: This is nice! 💯  Who knew we had Application controller tests!

leenagupte (approved):
LGTM 👍 

Thanks for fixing this, and for the comprehensive commit message. It really made the problems easy to understand. 💯

DilwoarH:
https://trello.com/c/Hnr1asLX/492-investigate-caching-issue-on-back-link

PR #2133, opened 2020-12-08
Add Christmas rules to postcode checker

DilwoarH:
## What

Add Christmas rules to postcode checker results.

## Why

So the checker shows the most up to date information on what rules to follow

## NOTE

Please note that the button component could not be used at this time as there is no support for the style of secondary button that we need. 

## Screenshot

Screenshot 2020-12-08 at 22 57 32

### With future restrictions

Screenshot 2020-12-09 at 20 26 19

https://trello.com/c/zjaqrsBT/971-add-christmas-rules-to-postcode-checker

### DIFF
--- a/app/views/coronavirus_local_restrictions/_christmas_rules.html.erb
+++ b/app/views/coronavirus_local_restrictions/_christmas_rules.html.erb
@@ -1,0 +1,17 @@
+<%= render "govuk_publishing_components/components/heading", {
+    text: t("coronavirus_local_restrictions.results.christmas_rules.heading"),
+    font_size: "m",
+    margin_bottom: 4,
+    mobile_top_margin: true,
+  } %>
+
+<p class="govuk-body">
+  <%= t("coronavirus_local_restrictions.results.christmas_rules.body") %>
+</p>
+
+<%= render "govuk_publishing_components/components/button", {
+    text: t("coronavirus_local_restrictions.results.christmas_rules.guidance.label"),
+    href: t("coronavirus_local_restrictions.results.christmas_rules.guidance.link"),
+    margin_bottom: true,
+    secondary_solid: true,
+  } %>
--- a/test/integration/coronavirus_local_restrictions_test.rb
+++ b/test/integration/coronavirus_local_restrictions_test.rb
@@ -17,6 +17,7 @@ class CoronavirusLocalRestrictionsTest < ActionDispatch::IntegrationTest
       then_i_enter_a_valid_english_postcode
       then_i_click_on_find
       then_i_see_the_results_page_for_level_one
+      then_i_see_details_of_christmas_rules
     end
 
     it "displays the tier two restrictions" do
@@ -25,6 +26,7 @@ class CoronavirusLocalRestrictionsTest < ActionDispatch::IntegrationTest
       then_i_enter_a_valid_english_postcode_in_tier_two
       then_i_click_on_find
       then_i_see_the_results_page_for_level_two
+      then_i_see_details_of_christmas_rules
     end
 
     it "displays the tier three restrictions" do
@@ -33,6 +35,7 @@ class CoronavirusLocalRestrictionsTest < ActionDispatch::IntegrationTest
       then_i_enter_a_valid_english_postcode_in_tier_three
       then_i_click_on_find
       then_i_see_the_results_page_for_level_three
+      then_i_see_details_of_christmas_rules
     end
   end
 
@@ -319,6 +322,10 @@ class CoronavirusLocalRestrictionsTest < ActionDispatch::IntegrationTest
 
     assert page.has_text?(area)
     assert page.has_text?(I18n.t("coronavirus_local_restrictions.results.level_two.changing_alert_level", area: area))
+  end
+
+  def then_i_see_details_of_christmas_rules
+    assert page.has_text?(I18n.t("coronavirus_local_restrictions.results.christmas_rules.heading"))
   end
 
   def level_two_area
--- a/app/views/coronavirus_local_restrictions/_future_restrictions.html.erb
+++ b/app/views/coronavirus_local_restrictions/_future_restrictions.html.erb
@@ -16,7 +16,7 @@
     <%= render "govuk_publishing_components/components/button", {
       text: t("coronavirus_local_restrictions.results.guidance_label", level: @restriction.future_alert_level),
       href: t("coronavirus_local_restrictions.results.level_one.guidance_link"),
-      secondary: true,
+      secondary_solid: true,
       margin_bottom: 9
     } %>
   <% elsif @restriction.future_alert_level == 2 %>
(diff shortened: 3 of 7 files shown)


app/views/coronavirus_local_restrictions/_christmas_rules.html.erb
```diff
@@ -0,0 +1,15 @@
...
+  } %>
+
+<p class="govuk-body">
+  <%= t("coronavirus_local_restrictions.results.christmas_rules.body") %>
+</p>
+
+<a href="<%= t("coronavirus_local_restrictions.results.christmas_rules.guidance.link") %>"
+  class="govuk-button govuk-button--secondary govuk-!-margin-bottom-8" role="button" data-module="govuk-button">
+  <%= t("coronavirus_local_restrictions.results.christmas_rules.guidance.label") %>
+</a>
```
kevindew, line 16: Can we use the button component for this?

app/views/coronavirus_local_restrictions/_christmas_rules.html.erb
```diff
@@ -0,0 +1,15 @@
...
+    margin_bottom: 4,
+    mobile_top_margin: true,
+  } %>
+
+<p class="govuk-body">
+  <%= t("coronavirus_local_restrictions.results.christmas_rules.body") %>
+</p>
+
+<a href="<%= t("coronavirus_local_restrictions.results.christmas_rules.guidance.link") %>"
+  class="govuk-button govuk-button--secondary govuk-!-margin-bottom-8" role="button" data-module="govuk-button">
```
kevindew, line 14: What's the purpose of the data-module ?

kevindew (commented):
Code looks good - it looks like you have some errors in the commit message to resolve. Also while you're editing the commit message, you might want to think more about explaining the why of some of the decisions you've made in this change - there seems to be some supplementary information about dates in there whereas this code doesn't seem to do anything with dates?

Just a minor thing but a screenshot would have been helpful in the description, as otherwise I had to run the code manually to see what it looked like (as the heroku version doesn't work for local restrictions).

app/views/coronavirus_local_restrictions/_christmas_rules.html.erb
DilwoarH: That is the plan but at the moment the button component doesn't support this style of a secondary button.

Once the support has been added in, this will be replaced by the component

app/views/coronavirus_local_restrictions/_christmas_rules.html.erb
DilwoarH: Good spot - should have removed - copy and paste error from https://design-system.service.gov.uk/components/button/#secondary-buttons

app/views/coronavirus_local_restrictions/_christmas_rules.html.erb
DilwoarH: > Code looks good - it looks like you have some errors in the commit message to resolve. Also while you're editing the commit message, you might want to think more about explaining the why of some of the decisions you've made in this change - there seems to be some supplementary information about dates in there whereas this code doesn't seem to do anything with dates?
> 
> Just a minor thing but a screenshot would have been helpful in the description, as otherwise I had to run the code manually to see what it looked like (as the heroku version doesn't work for local restrictions).

Hey @kevindew - I've updated the things that you mentioned, happy to chat if you need to talk further.

leenagupte (commented):
@DilwoarH What does this look like if there are future restrictions? At the moment the future restrictions uses the `secondary` button from the component guide, i.e. green text with a green border.

I'm a bit concerned that the two different styles of secondary button will look odd together. Perhaps the old style could be used here rather than the work around, and a proper updated applied to the publishing components gem?

DilwoarH:
> @DilwoarH What does this look like if there are future restrictions? At the moment the future restrictions uses the `secondary` button from the component guide, i.e. green text with a green border.
> 
> I'm a bit concerned that the two different styles of secondary button will look odd together. Perhaps the old style could be used here rather than the work around, and a proper updated applied to the publishing components gem?

I'll take a screenshot 

PR created for button https://github.com/alphagov/govuk_publishing_components/pull/1814

DilwoarH:
@leenagupte 

### With future restrictions

Screenshot 2020-12-09 at 12 20 15

DilwoarH:
@leenagupte - I have added in the component now

app/views/coronavirus_local_restrictions/_christmas_rules.html.erb
DilwoarH: The HTML has been removed as support for this styling has now been added to the component

test/integration/coronavirus_local_restrictions_test.rb
```diff
@@ -321,6 +324,13 @@ def then_i_see_the_results_page_for_level_two_with_changing_restriction_levels
     assert page.has_text?(I18n.t("coronavirus_local_restrictions.results.level_two.changing_alert_level", area: area))
   end
 
+  def then_i_see_details_of_christmas_rules
+    assert page.has_text?(I18n.t("coronavirus_local_restrictions.results.christmas_rules.heading"))
+    assert page.has_text?(I18n.t("coronavirus_local_restrictions.results.christmas_rules.body"))
+    assert page.has_text?(I18n.t("coronavirus_local_restrictions.results.christmas_rules.guidance.label"))
+    assert page.has_css?("a[href='#{I18n.t('coronavirus_local_restrictions.results.christmas_rules.guidance.link')}']")
```
kevindew, line 331: It seems at odds with the rest of this test file to assert for every single translation phrase. Looking at [no information](https://github.com/alphagov/collections/blob/4a6e24b8ec9c5606f0d09d82dae52b3443beb20a/test/integration/coronavirus_local_restrictions_test.rb#L310) a precedent would be testing just the header.

Gemfile.lock
```diff
@@ -159,7 +159,7 @@ GEM
       statsd-ruby (~> 1.4.0)
       unicorn (>= 5.4, < 5.8)
     govuk_document_types (0.9.2)
-    govuk_publishing_components (23.7.7)
+    govuk_publishing_components (23.9.0)
```
kevindew, line 162: It's worth rebasing this PR as this looks already commited to the repo.

I was going to suggest that the commit made a reference to this dependency change, but since it's already in the project it seems simpler to just rebase it away.

kevindew (commented):
Looks good, just a couple of minor comments to address.

One thing I wondered is why do we prefer a secondary solid button to a secondary one is there any rationale?

test/integration/coronavirus_local_restrictions_test.rb
DilwoarH: Make sense - I'll get these additional checks removed

Gemfile.lock
DilwoarH: Done

test/integration/coronavirus_local_restrictions_test.rb
DilwoarH: Done

kevindew (approved):

PR #2134, opened 2020-12-08
Update DIT landing page (take 2)

alex-ju:
This PR consists of commits from #2088 (which were temporarily rolled back over the weekend in #2123).

The changes are made to reflect the ~~[updated design](https://trello.com/c/ZfMI7LNc/676-eu-business-page-new-design-additional-translation-pages#action-5fcf983241288f6bb1d978a4)~~ [updated design](https://trello.com/c/ZfMI7LNc/676-eu-business-page-new-design-additional-translation-pages#action-5fc9226375e51932f7bf04dd) – meant to unclutter the page and streamline the update process by moving sections to dedicated pages maintained by departments.

👉 [Preview the `/eubusiness` page](https://govuk-collec-update-dit-vgqkx8.herokuapp.com/eubusiness)

<table>
<tr><th>Before</th><th>In beween</th><th>After</th></tr>
<tr><td valign="top">

www gov uk_eubusiness

</td><td valign="top">

localhost_3070_eubusiness

</td><td valign="top">

localhost_3070_eubusiness

</td></tr>
</table>

[Trello card](https://trello.com/c/ZfMI7LNc)

### DIFF
--- a/app/assets/stylesheets/views/_dit-landing-page.scss
+++ b/app/assets/stylesheets/views/_dit-landing-page.scss
@@ -2,11 +2,13 @@ $red: #ff003b;
 $dark-grey: #282828;
 
 .dit-landing-page__kbm-logo {
-  max-height: 70px;
+  height: 50px;
+  max-height: 50px;
   margin-top: govuk-spacing(3);
 
   @include govuk-media-query($from: desktop) {
-    max-height: 100px;
+    max-height: 70px;
+    height: 70px;
     margin-top: 0;
   }
 }
--- a/app/assets/stylesheets/views/_dit-landing-page.scss
+++ b/app/assets/stylesheets/views/_dit-landing-page.scss
@@ -15,19 +15,13 @@ $dark-grey: #282828;
 }
 
 .dit-landing-page__header--with-bg-image {
-  padding-bottom: govuk-spacing(4);
-  padding-top: govuk-spacing(3);
   background-color: $dark-red;
   border-top: govuk-spacing(2) solid $red;
-  border-bottom: govuk-spacing(2) solid $red;
 
   @include govuk-media-query($from: $header-breakpoint) {
-    padding-bottom: govuk-spacing(4);
-    padding-top: govuk-spacing(8);
     background-image: image-url('DIT-banner@2x.jpg');
     background-size: cover;
-    background-position: top;
-    min-height: 500px;
+    background-position: right;
   }
 
   // HIDE BACKGROUND IMAGE ON IE
@@ -43,13 +37,26 @@ $dark-grey: #282828;
   }
 }
 
+.dit-landing-page_header-text-container {
+  background-color: $dark-red;
+  box-shadow: 15px 0 0 0 $dark-red;
+  padding-bottom: govuk-spacing(4);
+
+  @include govuk-media-query($from: tablet) {
+    padding-bottom: govuk-spacing(9);
+  }
+}
+
+.dit-landing-page_header-text-container--no-translation-nav {
+  padding-top: govuk-spacing(9);
+  padding-bottom: govuk-spacing(9);
+}
+
 .dit-landing-page__header-text-block {
   @include govuk-media-query($from: $header-breakpoint) {
-    padding-bottom: 80px;
-    padding-top: govuk-spacing(5);
-    background: $red;
-    background: -webkit-linear-gradient(90deg, $dark-red 66.66%, $transparent 66.66%);
-    background: linear-gradient(90deg, $dark-red 66.66%, $transparent 66.66%);
+    background: $dark-red;
+    background: -webkit-linear-gradient(90deg, $dark-red 50%, $transparent 50%);
+    background: linear-gradient(90deg, $dark-red 50%, $transparent 50%);
   }
 }
 
--- a/app/assets/stylesheets/views/_dit-landing-page.scss
+++ b/app/assets/stylesheets/views/_dit-landing-page.scss
@@ -39,7 +39,7 @@ $dark-grey: #282828;
 
 .dit-landing-page_header-text-container {
   background-color: $dark-red;
-  box-shadow: 15px 0 0 0 $dark-red;
+  box-shadow: 15px 0 0 0 $dark-red, 4px 0 0 0 $dark-red, 3px 0 0 0 $dark-red, 1px 0 0 0 $dark-red;
   padding-bottom: govuk-spacing(4);
 
   @include govuk-media-query($from: tablet) {
--- a/app/assets/stylesheets/views/_dit-landing-page.scss
+++ b/app/assets/stylesheets/views/_dit-landing-page.scss
@@ -3,7 +3,13 @@ $red: #ff003b;
 $dark-grey: #282828;
 
 .dit-landing-page__kbm-logo {
-  max-height: 100px;
+  max-height: 70px;
+  margin-top: govuk-spacing(3);
+
+  @include govuk-media-query($from: desktop) {
+    max-height: 100px;
+    margin-top: 0;
+  }
 }
 
 .dit-landing-page__header {
--- a/app/assets/stylesheets/views/_dit-landing-page.scss
+++ b/app/assets/stylesheets/views/_dit-landing-page.scss
@@ -1,4 +1,3 @@
-$dark-red: #c21d26;
 $red: #ff003b;
 $dark-grey: #282828;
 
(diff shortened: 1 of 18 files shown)


app/views/dit_landing_page/_page_header.html.erb
```diff
@@ -15,19 +15,26 @@
 
 <header class="dit-landing-page__header--with-bg-image">
   <div class="dit-landing-page__header-text-block">
-    <div class="govuk-width-container landing-page_header-text-container">
+    <div class="govuk-width-container">
       <div class="govuk-grid-row">
-        <div class="govuk-grid-column-two-thirds">
+        <div class="govuk-grid-column-two-thirds dit-landing-page_header-text-container <%= 'dit-landing-page_header-text-container--no-translation-nav' unless defined?(presented_taxon) %>">
```
benthorner, line 20: Could you help me understand the styling variation for no translation nav, and why it depends on a presented taxon being defined? The two sounds quite unrelated to me - maybe you could add something in the commit message? (Perhaps we could check `presented_taxon.translation_links` instead.)

app/views/dit_landing_page/_page_header.html.erb
```diff
@@ -15,19 +15,26 @@
...
+          <% if defined?(presented_taxon) %>
+            <%= render "govuk_publishing_components/components/translation-nav", {
+              translations: presented_taxon.translation_links,
+              inverse: true,
+            } %>
+          <% end %>
+
           <h1 class="dit-landing-page__page_title">
-            <%= t("dit_landing_page.page_header") %>
+            <%= sanitize(t("dit_landing_page.page_header")) %>
```
benthorner, line 29: Minor: how come we need to sanitize now? Ah, maybe it's because of the `&nbsp`?

app/assets/stylesheets/views/_dit-landing-page.scss
```diff
@@ -43,13 +37,26 @@ $dark-grey: #282828;
   }
 }
 
+.dit-landing-page_header-text-container {
+  background-color: $dark-red;
+  box-shadow: 15px 0 0 0 $dark-red;
```
benthorner: Where does the shadow appear?

config/locales/en/dit_landing_page.yml
```diff
@@ -13,17 +13,17 @@ en:
       <p>If you run an EU-based business, you need to check the new rules and prepare for the changes so that you can continue trading with the UK from January 2021.</p>
     guidance_links:
       - label: Importing from the UK
-        url:
+        url: "/guidance/eu-business-importing-from-the-uk"
```
benthorner, line 15: Minor: this could be merged with the previous commit.

app/views/dit_landing_page/_training.html.erb
```diff
@@ -5,7 +5,7 @@
         <h2 class="govuk-heading-l">
           <%= t("dit_landing_page.training_section_title") %>
         </h2>
-        <p class="govuk-body"><%= t("dit_landing_page.training_section_description") %></p>
+        <p class="govuk-body"><%= sanitize(t("dit_landing_page.training_section_description", url: t("dit_landing_page.training_section_url"))) %></p>
```
benthorner, line 8: Could we embed the URL in the description instead? It's a bit surprising to see us passing it back through in code.

benthorner (commented):
Looks good to me - definitely a much cleaner design, with less hard-coded content for us to update 🎉. I've left a few minor comments, the only more major one is the conditional around translations - would be good to understand how that's working before this goes in.

(Very minor: I noticed the "after" screenshot still has the dark grey background 😄.)

app/views/dit_landing_page/_page_header.html.erb
alex-ju: Yes, we use an HTML entity in the page_header so I'm using sanitize to escape it otherwise it will get rendered as text.

config/locales/en/dit_landing_page.yml
```diff
@@ -3,191 +3,36 @@ en:
...
-      [Check the new rules for unregistered design protection in the UK and EU from 1 January 2021.](/guidance/changes-to-unregistered-designs-after-the-transition-period)
-
-    training_section_title: Additional help and support
+      If you run an EU-based business, you need to check the new rules and prepare for the changes so that you can continue trading with the UK from January 2021.
+    guidance_links:
+      - label: Importing from the UK
+        url: "/guidance/eu-business-importing-from-the-uk"
+        text: Find out what you need to do to import from the UK, including customs procedures and importing goods such as food, drink, animal, and plant products.
+      - label: Exporting to the UK
+        url: "/guidance/eu-business-exporting-to-the-uk"
```
hannako, line 18: Each of the `guidance_links` is for a page that has a translation nav itself. Not a blocker for now, but it would be nice if when a user on https://www.gov.uk/eubusiness.de clicks on this link, they are taken to https://www.gov.uk//guidance/eu-business-exporting-to-the-uk.de instead of the english language version. I haven't tried it locally but I reckon we could do that quite easily by appending the locale to the urls.

app/views/dit_landing_page/_page_header.html.erb
alex-ju: To prevent the page layout from moving around, when we don't have the translation navigation we add extra padding to the header container. Indeed `presented_taxon.translation_links` might be a better check. I'm rather sceptical to introducing the translation navigation with this work, as it doesn't seem related. So I'll amend the commit to remove it from this PR. The modifier class `dit-landing-page_header-text-container--no-translation-nav` would still be valuable until we get the navigation in.

app/assets/stylesheets/views/_dit-landing-page.scss
alex-ju: We're using the `box-shadow` as a technique to render a border outside the element boundaries and we do this to cover the standard margin we have after the `govuk-grid-column-two-thirds` container.

Screenshot 2020-12-08 at 17 07 21

hannako (approved):
Removing the translation nav code commit is a good plan. 

Other than that looks great to me. ✂️ ✂️ ✂️

app/assets/stylesheets/views/_dit-landing-page.scss
benthorner: Yeah, I noticed it "extends the redness" a bit to the right. Would we expect to see some "white" if we removed it? It's a minor thing - I'm just struggling to understand why we need it, given that it also looks OK to me without.

Thanks for the diagram ⭐.

benthorner (approved):
Thanks for addressing all the comments @alex-ju 💯. I've still got one about that `box-shadow`, but definitely not a blocker for this going in.

app/assets/stylesheets/views/_dit-landing-page.scss
alex-ju: If we remove it we'll see more of the background image (the one that appears on the right-hand side). We need it to align the text container with the search box at the top (and with any content sitting in the one-third column; for example the 'keep business moving' logo or the image down below the page).

benthorner: Ahhh, so _that's_ what it's for. Thanks for explaining. Could we add that to the commit? I think it would be helpful for someone in the future, in case it's not obvious to someone else too.

app/controllers/dit_landing_page_controller.rb
```diff
@@ -1,3 +1,31 @@
 class DitLandingPageController < ApplicationController
+  around_action :switch_locale
+
   def show; end
+
+  helper_method :translation_links
```
alex-ju, line 6: Functions in this controller certainly need tests, but I wanted to check the approach before committing to it.

hannako: Once we merge, and run [this](https://github.com/alphagov/collections-publisher/pull/1196) I believe the content item will contain the available translations in the [available translations array](https://www.gov.uk/api/content/eubusiness) so we won't need to hardcode them in the controller. ie we can fetch them from the content item and pass them to the translation nav component. But in terms of roll out, my feeling is that hardcoding is fine for now. We can open a refactor PR after this one is merged and live.  If you added a controller test eg

```
it "renders the page for the es locale" do
  get :show, params: { locale: "es" }
  assert_response :success
  assert_select "h1", "Comercio con el Reino Unido a partir del 1 de enero de 2021 como empresa con sede en la UE"
end
```

That would cover us for now, rather than extensively testing the `translation links` method which will be removed.

config/locales/de/dit_landing_page.yml
```diff
@@ -15,27 +15,27 @@ de:
       - title: Einfuhren aus dem Vereinigten Königreich
         text: Informieren Sie sich, was Sie als Unternehmen mit Sitz in der EU im Zusammenhang mit Einfuhren aus dem Vereinigten Königreich beachten müssen, u.a. hinsichtlich Zollverfahren und Import von Waren wie Lebensmitteln, Getränken, Tier- und Pflanzenprodukten.
         link: Mehr zu Einfuhren aus dem Vereinigten Königreich
-        url: "/guidance/eu-business-importing-from-the-uk"
+        url: "/guidance/eu-business-importing-from-the-uk.de"
       - title: Ausfuhren in das Vereinigte Königreich
```
hannako, line 19: 🎉 🥇

PR #2140, opened 2020-12-10
Rename coronavirus local restrictions locale file

kevindew:
I'm assuming the name of lookup for this file was a mistake, the established
convention for this repo is to name these files after the controller
they reference.


DilwoarH:
Do we need to update any code?

leenagupte (approved):

leenagupte:
> Do we need to update any code?

No, the keys in the file start "en.coronavirus_local_restrictions" etc, so the file has just been renamed to match.

PR #2142, opened 2020-12-11
Replace postcode partial and show invalid entry

kevindew:
Trello: https://trello.com/c/qMKzsOVO/975-change-form-submission-from-post-to-get

This is a draft pull request as it is built upon https://github.com/alphagov/collections/pull/2136. Once that is merged I'll change this PR's base branch and mark it as ready for review.

This removes the `shared/postcode-lookup` partial and moves it into the `coronavirus_local_restrictions/show` view reflecting that this shared partial is not shared with any other views and has lots of coupling to the existing `coronavirus_local_restrictions` views. 

Previously we didn't show postcodes back to users if they input them incorrectly, this changes that to show their original input back.

Before:

Screenshot 2020-12-11 at 11 12 58

After: 

Screenshot 2020-12-11 at 11 15 28


DilwoarH:
I remember the reasoning behind this was so that we can put the postcode look up in other places. This hasn't happened so happy to approve once rebased.

leenagupte (commented):
I think the user-input needs to be stripped from the GA tracking in the lookup page as well now.

Event info on lookup page (when there's an error)
lookup_page_ga

Event info on results page (where there isn't an error)
results_page_ga

My concern is people accidentally putting in their whole address, or enough of it to make them identifiable.

kevindew:
> I think the user-input needs to be stripped from the GA tracking in the lookup page as well now.
> 
> Event info on lookup page (when there's an error)
> lookup_page_ga
> 
> Event info on results page (where there isn't an error)
> results_page_ga
> 
> My concern is people accidentally putting in their whole address, or enough of it to make them identifiable. I assumed this would be covered by the `meta_tags` partial 🤔

It's stripped on all the pages however the pre-existing approach to do this matches a postcode by a regex: https://github.com/alphagov/static/blob/ea01e2fab465e4d98c1fe97c35b1201f6fcd6420/app/assets/javascripts/analytics/pii.js#L6 so your example of ut8 doesn't match the regex.

It'd be nicer if we had means to tell GA to ignore a particular parameter (as was actually discussed the other week) https://github.com/alphagov/govuk_publishing_components/pull/1807#issuecomment-737874416 but we're unlikely to be able to make that change in a hurry (Smart Answers will have the same issue).

leenagupte:
> > I think the user-input needs to be stripped from the GA tracking in the lookup page as well now.
> > Event info on lookup page (when there's an error)
> > lookup_page_ga
> > Event info on results page (where there isn't an error)
> > results_page_ga
> > My concern is people accidentally putting in their whole address, or enough of it to make them identifiable. I assumed this would be covered by the `meta_tags` partial 🤔
> 
> It's stripped on all the pages however the pre-existing approach to do this matches a postcode by a regex: https://github.com/alphagov/static/blob/ea01e2fab465e4d98c1fe97c35b1201f6fcd6420/app/assets/javascripts/analytics/pii.js#L6 so your example of ut8 doesn't match the regex.
> 
> It'd be nicer if we had means to tell GA to ignore a particular parameter (as was actually discussed the other week) [alphagov/govuk_publishing_components#1807 (comment)](https://github.com/alphagov/govuk_publishing_components/pull/1807#issuecomment-737874416) but we're unlikely to be able to make that change in a hurry (Smart Answers will have the same issue).

Ah, ok, that makes sense. One for the backlog I guess.

leenagupte (approved):

PR #2143, opened 2020-12-11
Update email link text

1pretz1:
We're removing the 'alert' terminology from our email links. This text
has been content reviewed.

Example pages:

* https://www.gov.uk/government/organisations/attorney-generals-office
* https://www.gov.uk/world/usa

Trello:
https://trello.com/c/ssSnzp2L/640-remove-alert-from-all-the-sign-up-links

### DIFF
--- a/app/views/components/docs/accordion.yml
+++ b/app/views/components/docs/accordion.yml
@@ -64,7 +64,7 @@ examples:
           panel: '<div class="govuk-govspeak">
             <div class="subscriptions">
               <a href="/email-signup/?topic=/education/alternative-provision-censuses" class="email-alerts">
-                Get email alerts for this topic
+                Get emails for this topic
                 <span class="visuallyhidden">Alternative provision censuses</span>
               </a>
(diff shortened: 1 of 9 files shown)


kevindew (approved):
Code looks good to me 👍 

The only thing I'm a little concerned about is that there may people quite opinionated about the transition landing page, so may need a check from that team. Having looked at https://www.gov.uk/transition I can't actually see where the copy is used though.

benthorner:
> I can't actually see where the copy is used though.

It looks like it was [never actually used](https://github.com/alphagov/collections/commit/8ce94224d5b2bcd466f4cd2e5a013eb50daf1d33#diff-6b1fe0e2753009ad118fdd4a51ed6b5cfd6f67541a0c91af6162749d392579f4)...so we can just delete it?!

1pretz1:
> It looks like it was never actually used...so we can just delete it?!

Oh interesting - I've removed the partial @benthorner.

PR #2145, opened 2020-12-11
Return cache headers for Coronavirus local restrictions

kevindew:
Trello: https://trello.com/c/qMKzsOVO/975-change-form-submission-from-post-to-get

This is a draft pull request as it is built upon #2136. Once that is merged I'll change this PR's base branch and mark it as ready for review.

This returns a default cache header of 30 minutes public for all
responses except when there is an upcoming future restriction.

By default Rails will return a cache header of max-age=0, public=false.
Somewhere in the GOV.UK stack this seems to get changes to a public 30
minute cache implicitly. This feels a bit dubious as we don't know what
part of the stack is controlling the cache so it seems better to set
this explicitly.

When a future restriction is coming we don't want users to be served a
page telling them the wrong information. When the time until a new
restriction is less than 30 minutes we don't want to tell users they can
cache the page for 30 mins, instead we return the time until the
restriction takes effect. So for a future restriction starting in 5
minutes, we tell users they can cache the page for 5 minutes.

### DIFF
--- a/app/controllers/coronavirus_local_restrictions_controller.rb
+++ b/app/controllers/coronavirus_local_restrictions_controller.rb
@@ -1,7 +1,9 @@
 class CoronavirusLocalRestrictionsController < ApplicationController
+  MAX_CACHE_TIME = 30.minutes
   skip_before_action :verify_authenticity_token, only: [:legacy]
 
   def show
+    expires_in(MAX_CACHE_TIME, public: true)
     @content_item = content_item.to_hash
 
     if params[:postcode].nil?
@@ -16,6 +18,8 @@ class CoronavirusLocalRestrictionsController < ApplicationController
     elsif @search.no_information?
       render :no_information
     else
+      expires_in(restriction_expiry(@search), public: true)
+
       render :results
     end
   end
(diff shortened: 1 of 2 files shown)


app/controllers/coronavirus_local_restrictions_controller.rb
```diff
@@ -36,4 +40,14 @@ def content_item
       end
     end
   end
+
+  def restiction_expiry(search)
```
leenagupte, line 44: The Cache time stuff looks good. There's a teeny tiny typo

kevindew: Oops - thanks for catching.

leenagupte (approved):
I'm not sure how to test the change in cache times locally, how the page still works, and the test looks good.

kevindew:
> I'm not sure how to test the change in cache times locally, how the page still works, and the test looks good.

Thanks Leena, I tested it out by adding a future restriction and checking the response headers. For example: `Cache-Control: max-age=564, public` compared to the usual `Cache-Control: max-age=1800, public`

PR #2148, opened 2020-12-14
Update to component

chris-gds:
## What?

Updating various UI elements within collections to use [component gem](https://components.publishing.service.gov.uk/component-guide) architecture
 
## Why?

Updating to gem aligns with Frontend architecture, improves the consistency of a11y and Design, also allows updates cascaded across all apps.

[Trello ticket](https://trello.com/c/C8BumP2B)

## Visuals

Before (Left) / After (right) 

[#1 Live link](https://www.gov.uk/coronavirus)

No perceivable difference 

1-logo-change

[#2 Live link](https://www.gov.uk/government/organisations)

2-search

[#3 Live link](https://www.gov.uk/transition)

3-lead-para

[#4 Live link](https://www.gov.uk/coronavirus/business-support?priority-taxon=774cee22-d896-44c1-a611-e3109cce8eae)

5-list-mob

5-list-desktop

[#5 Live link](https://www.gov.uk/world/afghanistan)

6-accordian-desktop

6-accordian-mob

(Overlay diff of before and after - changes in spacing.)
6-accoridan-diff

## Anything else?

The NHS logo was explored to add to the gem for further use in this [draft PR](https://github.com/alphagov/govuk_publishing_components/pull/1818), however, as this asset is localised in it's use, decided to keep within the app until a time where there are multiple uses. 

[IE8](
https://www.gov.uk/service-manual/technology/designing-for-different-browsers-and-devices) has rendering issues on this logo as it does not have a fallback PNG.  
(Other icons do not render within IE8)

Screenshot 2020-12-11 at 17 12 01


app/views/components/_accordion.html.erb
```diff
@@ -1,28 +1,30 @@
...
   sections ||= false
+  accordion_contents = []
 %>
 <% if sections %>
-  <div data-module="collections-accordion" class="app-c-accordion js-hidden">
-    <% sections.each_with_index do |section, index| %>
-      <%
-        id = section[:id] || section[:title].downcase.tr(" ","-")
-      %>
-      <div class="app-c-accordion__section js-section" id="<%= id %>" data-track-count="accordionSection">
```
chris-gds, line 10: Querying how to reimplement this within in gem version.

chris-gds:
@maxgds flagged build issues, moved back to draft.

chris-gds:
Note: update to [this element](https://github.com/alphagov/collections/blob/b1c55a17e53fa199f7cfcfeb7f94be1f0c02944f/app/views/step_nav/show.html.erb#L18) to replace with [lead paragraph](https://components.publishing.service.gov.uk/component-guide/lead_paragraph)

Resulted in a failed test (only on integration, tests pass locally) as the actual Content entered contains invalid HTML on different environments. Removed as this needs to be change at a publisher level from `govspeak` to text only + the actual Content of every instance.

[#4 Live link](https://www.gov.uk/report-treasure-step-by-step)

4-lead-para

app/views/organisations/index.html.erb
```diff
@@ -11,13 +11,11 @@
     } %>
 
     <form class="filter-organisations-list__form" data-filter="form" role="search" aria-label="Departments, agencies and public bodies">
-      <%= render "govuk_publishing_components/components/input", {
-        label: {
-          text: "Search for a department, agency or public body"
-        },
-        name: "search-box",
+      <%= render "govuk_publishing_components/components/search", {
```
maxgds, line 14: What was the justification for this change? Is it to do a better job of presenting it to assistive tech as a search field? I'm not sure it's a great change as it introduces a search button that wasn't there before, but the searching is interactive/live and the button serves no purpose. It doesn't seem ideal to have an effectively dead button on the page. @injms thoughts?

chris-gds: This is _great_ point, it felt like this worked but of course this button is redundant, I know it's not user-facing but the [search in the gem](https://components.publishing.service.gov.uk/component-guide) works a similar way acting a _dynamic-filter-search_ rather than a _click-search_ 

I feel like the Design is correct in the sense the search _icon_ should be there but it should of change to clear icon upon keydown (of course this doesn't exist).

🤔 

Currently exploring other uses of search across govuk / Design system (this is probably worth reverting)

chris-gds: Reverted (will rebase before merge to keep clarity for review)
Will raise an enhancement suggestion on the gem and ref on the [Design System backlog](https://github.com/alphagov/govuk-design-system-backlog/issues/132)

chris-gds: Similar use on blur within search:

ezgif-7-4d6d10c0e4d6

maxgds (approved):
Approving this as it seems sound BUT please add the newline that's needed. 

I think reverting the searchbox change was the right thing here. 

Possibly we need a separate ticket to deal with the lead paragaph on step-by-steps so we can get the content into shape if it's invalid - although I think lead paragraphs don't tend to have links (would need to check that theory) and I think step by step intros almost always do. If I'm right maybe we don't want to be switching it out anyway.

PR #2149, opened 2020-12-14
Redirect to postcode checker to full guidance page

leenagupte:
Trello: https://trello.com/c/cr8l3ddo

# What's changed and why?

We're expecting potential spike in traffic. This a standby change that's
only to be used if we need to take the postcode checker offline quickly,
to take pressure of the system whilst we increase capacity to respond to
the spike.

### DIFF
--- a/app/lib/coronavirus_local_restrictions_constraint.rb
+++ b/app/lib/coronavirus_local_restrictions_constraint.rb
@@ -1,0 +1,5 @@
+class CoronavirusLocalRestrictionsConstraint
+  def matches?(*)
+    !Rails.env.production?
+  end
+end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -28,7 +28,12 @@ Rails.application.routes.draw do
   end
 
   # Route for local restrictions postcode lookup
-  get "/find-coronavirus-local-restrictions" => "coronavirus_local_restrictions#show"
+  constraints CoronavirusLocalRestrictionsConstraint.new do
+    get "/find-coronavirus-local-restrictions" => "coronavirus_local_restrictions#show"
+  end
+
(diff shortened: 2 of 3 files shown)


app/lib/coronavirus_local_restrictions_constraint.rb
```diff
@@ -0,0 +1,7 @@
+class CoronavirusLocalRestrictionsConstraint
+  def matches?(*)
+    return false if Rails.env.production?
```
kevindew, line 3: Nit, could be `!Rails.env.production?` but don't worry about changing given how short lived this might be.

kevindew (approved):
Looks good, lets hope we don't need it 😄

app/lib/coronavirus_local_restrictions_constraint.rb
```diff
@@ -0,0 +1,9 @@
+class CoronavirusLocalRestrictionsConstraint
+  def matches?(*)
+    !Rails.env.production? || redirect_from_date&.future?
```
kevindew, line 3: Minor - doesn't seem necessary to have the safe navigation operator since this will only return a time value.

kevindew (commented):
Looks good - guess we'll keep this on ice until we know of a redirect destination. It doesn't look like this need to work with: https://github.com/alphagov/collections/pull/2205 anymore so I'll close that.

app/lib/coronavirus_local_restrictions_constraint.rb
leenagupte: I added that in preparation in switching out the hard-coded time for config. However we won't be needing this anymore, so I'll remove the commit.

kevindew (approved):

PR #2151, opened 2020-12-14
Add a 3 hour Rails cache to mapit postcode lookup

huwd:
This line returns the area a postcode is in. Area boundaries are
infrequently updated, (rarely more than twice a year).

Repeated queries for the same postcode will return the same area, making
this a good candidate for caching to boost performance.

We use a three hour expiry because long enough that it should help cover
spikes in traffic, but not require cache busting when there is a rare
area boundary update.

### DIFF
--- a/app/services/location_lookup_service.rb
+++ b/app/services/location_lookup_service.rb
@@ -49,7 +49,9 @@ private
 
   def response
     @response ||= begin
-                    JSON.parse(mapit.location_for_postcode(postcode).to_json)
+                    Rails.cache.fetch("mapit-location-#{postcode}", expires_in: 3.hours) do
+                      JSON.parse(mapit.location_for_postcode(postcode).to_json)
+                    end
                   rescue GdsApi::HTTPNotFound, GdsApi::HTTPClientError => e
                     {
                       error: {


app/services/location_lookup_service.rb
```diff
@@ -49,7 +49,7 @@ def areas
 
   def response
     @response ||= begin
-                    JSON.parse(mapit.location_for_postcode(postcode).to_json)
+                    Rails.cache.fetch(JSON.parse(mapit.location_for_postcode(postcode).to_json), expires_in: 3.hours)
```
kevindew, line 52: This doesn't look like correct usage of the API. As I understand the usage is along the lines of:

`Rails.cache.fetch(key_name, expires_in: 3.hours) { code_to_run_to_populate_cache }`

This should provide an example: https://guides.rubyonrails.org/caching_with_rails.html#low-level-caching

huwd: ah ace yeah, sorry we're just figuring out the API - preaturely tagged you in this!

huwd: sorry swaped that over now, we're using the postcode as the key, that seemed like a good idea?

app/services/location_lookup_service.rb
```diff
@@ -49,7 +49,9 @@ def areas
 
   def response
     @response ||= begin
-                    JSON.parse(mapit.location_for_postcode(postcode).to_json)
+                    Rails.cache.fetch(postcode, expires_in: 3.hours) do
```
kevindew, line 52: I'd give this a slightly more distinct key, perhaps "mapit-location-#{postcode}".

kevindew (commented):
This looks good and the commit message is nice and explanatory - just a small suggestion.

app/services/location_lookup_service.rb
huwd: makes sense, I that just to prevent potential clashes, or is there somewhere that we regularly read out of the cache and so would want to know what this "postcode" thing actually is?

kevindew: Yeah just to prevent any potential clashes. It doesn't seem like the only thing we could cache with a postcode.

kevindew (approved):

PR #2153, opened 2020-12-14
Flip the partials for christmas rules and future restrictions

JonathanHallam:
All restrictions are meant to be in date order. The new restrictions that come into being before the christmas rule changes are in the wrong place. This swaps the christmas rules and future restrictions so it displays properly. This will have to be reverted before any new restrictions that come in after christmas are made live.

### Desktop
Screenshot 2020-12-14 at 16 06 08
### Mobile
Screenshot 2020-12-14 at 16 07 10

### Desktop
Screenshot 2020-12-14 at 16 06 36
### Mobile
Screenshot 2020-12-14 at 16 06 56

### DIFF
--- a/app/views/coronavirus_local_restrictions/_tier_one_restrictions.html.erb
+++ b/app/views/coronavirus_local_restrictions/_tier_one_restrictions.html.erb
@@ -44,7 +44,7 @@
     margin_bottom: 9
   } %>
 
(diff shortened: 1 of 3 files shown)


kevindew (approved):
Nice one

AlanGabbianelli (approved):

hannahjump:
Fab!

PR #2154, opened 2020-12-15
Remove POST route in postcode checker

leenagupte:
# What's changed and why?

In https://github.com/alphagov/collections/pull/2136 the postcode
checker was changed to use GET requests rather than POST requests
to find and display the results.

To make sure that people who were half way through the journey weren't
affected by this change, we changed the POST route to redirect to the
GET. But this was only meant to be temporary. It should now be safe to
remove the POST route completely.


kevindew (approved):
Nice one

PR #2156, opened 2020-12-15
Normalise postcodes with JavaScript

kevindew:

---

TL;DR - if you go to `/find-coronavirus-local-restrictions` and enter a postcode such as 'e18qs' this will be automatically converted to 'E1 8QS' at the point of the form submission. 

---

This adds JavaScript which intercepts the submission of the Coronavirus
local restriction postcode form and converts an input that looks like a
postcode into a formatted approach (uppercase with a space between
inward and onward codes).

This has been done to increase the chance that the same postcodes
entered produce cachable HTTP responses. It is intended as a less aggressive
replacement to [our CDN code](https://github.com/alphagov/govuk-cdn-config/pull/312/files#diff-94c0786370f6c095cf75ced830bd83a978ee352c1a5724c223f812f256699350R254-R265) which will produce a nicer user
experience.

To avoid conflicting with the postcode validation in the app I've taken
a reasonably lax approach to checking for a postcode. This just checks
and normalises something that looks like a common UK postcode. It leaves
erroneous input intact which should be beneficial in helping users
correct their errors.


DilwoarH (approved):

config/initializers/assets.rb
```diff
@@ -9,4 +9,4 @@
 # Precompile additional assets.
 # application.js, application.css, and all non-JS/CSS in the app/assets
 # folder are already added.
-Rails.application.config.assets.precompile += %w[print.css]
+Rails.application.config.assets.precompile += %w[print.css test-dependencies.js]
```
DilwoarH, line 12: will this add it to the compiled js in the frontend?

kevindew: Nope, it just means that one of the files built will be a `/assets/collections/test-dependencies-{hash}.js` however that asset would only be included if the HTML references it - which only happens in the test environment.

spec/javascripts/modules/coronavirus-local-restrictions-postcode-form_spec.js
```diff
@@ -0,0 +1,50 @@
...
+    var form = container.querySelector('form')
+    form.addEventListener('submit', function (e) {
+      e.preventDefault()
+    })
+
+    input = form.querySelector('input')
+    button = form.querySelector('button')
+
+    var postcodeForm = new GOVUK.Modules.CoronavirusLocalRestrictionsPostcodeForm()
+    postcodeForm.start($(form))
```
DilwoarH, line 24:
```suggestion
    postcodeForm.start(form)
```

We stopped using jQuery in our JS now - does this still work as expected

DilwoarH (changes requested):
Need to remove some one the jQuery stuff

spec/javascripts/modules/coronavirus-local-restrictions-postcode-form_spec.js
kevindew: It would work, but it doesn't feel very realistic to fake it as a different type of object when we can use the actual type.

DilwoarH:
Do you think we should just add it this whole thing to this listener @kevindew https://github.com/alphagov/collections/blob/master/app/assets/javascripts/modules/coronavirus-track-local-restrictions.js

spec/javascripts/modules/coronavirus-local-restrictions-postcode-form_spec.js
DilwoarH: We've started doing it in a few places now - https://github.com/alphagov/collections/blob/master/spec/javascripts/modules/coronavirus-track-local-restrictions_spec.js#L25

I'm happy either way but seeing as you went through the trouble of adding vanilla JS everywhere - it would be good to just not use it completedly.

kevindew: Sure, I feel like when we have a jQuery object in use it makes sense to use it, otherwise your tests are less realistic while not actually bringing us closer to removing jQuery - given that we'll need to edit the interface of the `GOVUK.Modules.start` pattern to remove jQuery.

kevindew:
> Do you think we should just add it this whole thing to this listener @kevindew https://github.com/alphagov/collections/blob/master/app/assets/javascripts/modules/coronavirus-track-local-restrictions.js

Sure, that was a good idea. I've extended this to include that now.

app/views/coronavirus_local_restrictions/show.html.erb
```diff
@@ -40,7 +36,14 @@
...
     <%= sanitize(t("coronavirus_local_restrictions.lookup.body_content")) %>
 
-    <%= form_tag find_coronavirus_local_restrictions_path, method: :get, class: "coronavirus-local-restrictions__postcode-form" do |f| %>
+    <%= form_tag find_coronavirus_local_restrictions_path,
+      method: :get,
+      class: "coronavirus-local-restrictions__postcode-form",
+      data: {
+        module: "coronavirus-local-restrictions-postcode-form",
+        "error-code": @search&.error_code,
+        "error-message": error_key ? t("coronavirus_local_restrictions.errors.#{error_key}.error_description") : nil,
```
DilwoarH, line 45: Should we put this logic on the top so we don't have to repeat?

DilwoarH (approved):
Non blocking comment - LGTM 👍

app/views/coronavirus_local_restrictions/show.html.erb
kevindew: Thanks, thats a good suggestion. On reflection I think that it is ok to remain where it is as we don't repeat the same logic and it'd probably be harder code to read if we had an error_message variable created in a completely different context.

PR #2163, opened 2020-12-16
Add tier 1 areas to local restrictions data

Rosa-Fox:
[Trello](https://trello.com/c/K69U9Y1V/989-add-tier-1-places-to-the-yaml-file)

- Add tier 1 restriction data to local_restrictions.yml.
- Display 'No tier information' result page to user if a postcode is valid but the restriction is not defined within the YML data.

## Tier 1 result

tier-1-results

## Tier 1 with future restrictions (please note... this is just test data)

tier-1-with-future-restrictions

## No tier information

no-tier-information-results

### DIFF
--- a/app/controllers/coronavirus_local_restrictions_controller.rb
+++ b/app/controllers/coronavirus_local_restrictions_controller.rb
@@ -14,7 +14,7 @@ class CoronavirusLocalRestrictionsController < ApplicationController
 
     if @search.blank_postcode? || @search.invalid_postcode?
       render
-    elsif @search.no_information?
+    elsif @search.no_restriction? || @search.no_information?
       render :no_information
     else
       expires_in(restriction_expiry(@search), public: true)
--- a/app/models/postcode_local_restriction_search.rb
+++ b/app/models/postcode_local_restriction_search.rb
@@ -42,6 +42,10 @@ class PostcodeLocalRestrictionSearch
     %w[invalidPostcodeFormat fullPostcodeNoMapitValidation].include?(error_code)
   end
 
+  def no_restriction?
+    location_lookup.data.first&.england? && local_restriction.nil?
+  end
+
   def location_lookup
     @location_lookup ||= LocationLookupService.new(sanitised_postcode)
   end
--- a/app/views/coronavirus_local_restrictions/_tier_one_restrictions.html.erb
+++ b/app/views/coronavirus_local_restrictions/_tier_one_restrictions.html.erb
@@ -17,14 +17,14 @@
 
 <%= tag.div :data => {
   :module => "coronavirus-track-local-restriction-results",
-  "tracking-label" => "Tier one: #{location_lookup.lower_tier_area_name}"
+  "tracking-label" => "Tier one: #{restriction.area_name}"
 } do %>
   <%= render "govuk_publishing_components/components/title", {
     title: sanitize(title)
   } %>
 
   <p class="govuk-body">
-    <%= t("coronavirus_local_restrictions.results.match") %> <strong><%= postcode %></strong> to <strong><%= location_lookup.lower_tier_area_name %></strong>.
+    <%= t("coronavirus_local_restrictions.results.match") %> <strong><%= postcode %></strong> to <strong><%= restriction.area_name %></strong>.
   </p>
 
   <% if restriction.present? && restriction.future.present? %>
--- a/app/views/coronavirus_local_restrictions/results.html.erb
+++ b/app/views/coronavirus_local_restrictions/results.html.erb
@@ -17,14 +17,14 @@
     %>
     <% if !@search.location_lookup.data.first.england? %>
       <%= render partial: "coronavirus_local_restrictions/devolved_nation", locals: locals %>
+    <% elsif @search.local_restriction&.tier_one? %>
+      <%= render partial: "coronavirus_local_restrictions/tier_one_restrictions", locals: locals %>
     <% elsif @search.local_restriction&.tier_two? %>
       <%= render partial: "coronavirus_local_restrictions/tier_two_restrictions", locals: locals %>
     <% elsif @search.local_restriction&.tier_three? %>
       <%= render partial: "coronavirus_local_restrictions/tier_three_restrictions", locals: locals %>
     <% elsif @search.local_restriction&.tier_four? %>
       <%= render partial: "coronavirus_local_restrictions/tier_four_restrictions", locals: locals %>
-    <% else %>
-      <%= render partial: "coronavirus_local_restrictions/tier_one_restrictions", locals: locals %>
     <% end %>
   </div>
 </div>
--- a/config/local_restrictions.yml
+++ b/config/local_restrictions.yml
@@ -2196,3 +2196,21 @@ E07000202:
     - alert_level: 2
       start_date: 2020-12-02
       start_time: "00:01"
+E06000052:
+  name: Cornwall Council
+  restrictions:
+    - alert_level: 1
+      start_date: 2020-12-02
+      start_time: "00:01"
+E06000046:
+  name: Isle of Wight Council
+  restrictions:
+    - alert_level: 1
+      start_date: 2020-12-02
+      start_time: "00:01"
+E06000053:
+  name: Council of the Isles of Scilly
+  restrictions:
+    - alert_level: 1
+      start_date: 2020-12-02
+      start_time: "00:01"
--- a/test/integration/coronavirus_local_restrictions_test.rb
+++ b/test/integration/coronavirus_local_restrictions_test.rb
@@ -10,7 +10,7 @@ class CoronavirusLocalRestrictionsTest < ActionDispatch::IntegrationTest
       given_i_am_on_the_local_restrictions_page
       then_i_can_see_the_postcode_lookup_form
       and_there_is_metadata
-      then_i_enter_a_valid_english_postcode
+      then_i_enter_a_valid_english_postcode_in_tier_one
       then_i_click_on_find
       then_i_see_the_results_page_for_level_one
       then_i_see_details_of_christmas_rules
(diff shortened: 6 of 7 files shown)


app/views/coronavirus_local_restrictions/results.html.erb
```diff
@@ -17,12 +17,14 @@
...
       <%= render partial: "coronavirus_local_restrictions/devolved_nation", locals: locals %>
+    <% elsif @search.local_restriction&.tier_one? %>
+      <%= render partial: "coronavirus_local_restrictions/tier_one_restrictions", locals: locals %>
     <% elsif @search.local_restriction&.tier_two? %>
       <%= render partial: "coronavirus_local_restrictions/tier_two_restrictions", locals: locals %>
     <% elsif @search.local_restriction&.tier_three? %>
       <%= render partial: "coronavirus_local_restrictions/tier_three_restrictions", locals: locals %>
     <% else %>
-      <%= render partial: "coronavirus_local_restrictions/tier_one_restrictions", locals: locals %>
+      <%= render partial: "coronavirus_local_restrictions/no_tier_information" %>
```
leenagupte, line 27: There's already a template for no tier information. Can that be used instead? Perhaps some of the common elements could be moved into a partial?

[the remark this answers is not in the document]
Rosa-Fox: @kevindew see commit message!

config/locales/en/coronavirus_local_restrictions.yml
```diff
@@ -69,6 +69,11 @@ en:
         guidance:
           label: Full list of areas and tiers
           link: "/guidance/full-list-of-local-restriction-tiers-by-area"
+      no_tier_information:
+        heading: There is no information about the restrictions in this area
+        guidance:
+          label: Full list of areas and tiers
+          link: "/guidance/full-list-of-local-restriction-tiers-by-area"
```
leenagupte, line 76: This identical to the content for `no_information`. Could that be used instead.

app/views/coronavirus_local_restrictions/_tier_one_restrictions.html.erb
```diff
@@ -17,14 +17,14 @@
 
 <%= tag.div :data => {
   :module => "coronavirus-track-local-restriction-results",
-  "tracking-label" => "Tier one: #{location_lookup.lower_tier_area_name}"
+  "tracking-label" => "Tier one: #{restriction.area_name}"
```
leenagupte, line 20: Is the `lower_tier_area_name` method in the LookupService still needed? Should we get rid of it?

leenagupte (commented):
I haven't had a chance to test this locally yet, but before I do I wonder if it's possible to reuse the existing "no_information" template and content?

app/views/coronavirus_local_restrictions/results.html.erb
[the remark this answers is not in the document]
kevindew: Sure, I've given it a read. It does seem an unnecessary overhead to have two different fallback views - particularly if we 

Can the `search.no_information?` consider the scenario? It's a bit awkward that we even have devolved_nation mixed in as a result hmm.

app/views/coronavirus_local_restrictions/_tier_one_restrictions.html.erb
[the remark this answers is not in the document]
Rosa-Fox: thanks! yeah that can go, i'll change it now.

app/services/location_lookup_service.rb
```diff
@@ -34,11 +34,6 @@ def error
     response[:error]
   end
 
-  def lower_tier_area_name
```
leenagupte, line 37: I'm so sorry to do this to you, but it looks like we're going to need this method in this ticket for devolved nations: https://trello.com/c/Tz6It6TX/991-improve-content-for-devolved-nations, so can we not delete this method after all? So sorry!

app/views/coronavirus_local_restrictions/results.html.erb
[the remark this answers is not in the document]
Rosa-Fox: I made a separate method `no_restriction_information_england_postcode?` and then used that in the controller to render the view, rather than rendering it as a partial from the results view.

Rosa-Fox: I made a separate method no_restriction_information_england_postcode? and then used that in the controller to render the view, rather than rendering it as a partial from the results view.

app/services/location_lookup_service.rb
Rosa-Fox: Have removed the commit that deleted it :)

Rosa-Fox:
@leenagupte @kevindew ready for re-review!

app/views/coronavirus_local_restrictions/results.html.erb
```diff
@@ -17,12 +17,18 @@
     %>
     <% if !@search.location_lookup.data.first.england? %>
       <%= render partial: "coronavirus_local_restrictions/devolved_nation", locals: locals %>
-    <% elsif @search.local_restriction&.tier_two? %>
+    <% end %>
+
+    <% if @search.local_restriction&.tier_one? %>
```
leenagupte, line 22: I think these conditions should be kept as if/elsif so we reduce the need of always having to evaluate all of them.

app/controllers/coronavirus_local_restrictions_controller.rb
```diff
@@ -14,7 +14,7 @@ def show
 
     if @search.blank_postcode? || @search.invalid_postcode?
       render
-    elsif @search.no_information?
+    elsif @search.no_restriction_information_england_postcode? || @search.no_information?
```
kevindew, line 17: Nice one solving the problem here 🎉 

I feel like there's an element of a fight here to fit this in which makes it a little tricky to understand. I don't think that is too big a deal as I've been working on a wider refactor that I think makes it easier:

```
    if @search.blank_postcode? || @search.invalid_postcode?
      render
    elsif @search.devolved_nation_result
      @result = @search.devolved_nation_result

      render :devolved_nation_result
    elsif @search.england_result
      expires_in(restriction_expiry(@search.england_result), public: true)
      @result = @search.england_result

      render :england_result
    else
      render :no_information
    end
```

config/local_restrictions.yml
```diff
@@ -2033,3 +2033,21 @@ E07000202:
...
+  restrictions:
+    - alert_level: 1
+      start_date: 2020-12-02
+      start_time: "00:01"
+E06000053:
+  name: Council of the Isles of Scilly
+  restrictions:
+    - alert_level: 1
+      start_date: 2020-12-02
+      start_time: "00:01"
```
kevindew, line 2053: As a future step should we consider having LocalRestriction raise an exception if there is an area here which doesn't have a restriction? It feels like there's places where the current_restiction method could return nil based on the data in this YAML file but is actually something that would break the application logic.

app/models/postcode_local_restriction_search.rb
```diff
@@ -42,6 +42,10 @@ def invalid_postcode?
     %w[invalidPostcodeFormat fullPostcodeNoMapitValidation].include?(error_code)
   end
 
+  def no_restriction_information_england_postcode?
```
leenagupte, line 45: I wonder if this could be called `restriction?`.  Then when we call it, it would read as `@search.restriction?` or `!@search.restriction?`

I don't think we need to specify that it's only for England in the name as the whole checker is only for England.

test/integration/coronavirus_local_restrictions_test.rb
```diff
@@ -33,6 +33,14 @@ class CoronavirusLocalRestrictionsTest < ActionDispatch::IntegrationTest
       then_i_see_the_results_page_for_level_three
       then_i_see_details_of_christmas_rules
     end
+
+    it "displays no tier information" do
```
leenagupte, line 37: Nice!

test/models/postcode_local_restriction_search_test.rb
```diff
@@ -84,6 +84,21 @@
     end
   end
 
+  describe "#no_restriction_information_england_postcode?" do
```
leenagupte, line 87: Could we add a test for the country that isn't England?

leenagupte (commented):
I think it's great that we're no longer assuming people are in a low tier just because we don't have any information about their postcode. 💯 

I think the approach in general is good, I just have a few small requests/suggestions in the inline comments.

app/controllers/coronavirus_local_restrictions_controller.rb
Rosa-Fox: nice!

config/local_restrictions.yml
Rosa-Fox: I will add a card to the backlog!

app/views/coronavirus_local_restrictions/results.html.erb
Rosa-Fox: ah yeah! I have changed back.

test/models/postcode_local_restriction_search_test.rb
Rosa-Fox: have added!

app/models/postcode_local_restriction_search.rb
Rosa-Fox: @leenagupte I changed this to `no_restriction?` as I struggled with `!@search.restriction?` due to the devolved nation needing to be true for `location_lookup.data.first&.england?`... also naming sort of works due to being similar to `no_information?`. I think Kevin will somewhat overwrite this anyway with his refactor.

test/integration/coronavirus_local_restrictions_test.rb
```diff
@@ -181,6 +189,7 @@ def then_i_enter_a_valid_english_postcode_with_a_future_level_two_restriction
     postcode = "E1 8QS"
     stub_local_restriction(postcode: postcode,
                            name: @area,
+                           current_alert_level: 1,
```
leenagupte, line 192: Ah! This is interesting! Yes, we're assuming that every local authority has an entry in the yaml now, so they should all have a current restriction.

leenagupte (approved):
LGTM 👍

PR #2168, opened 2020-12-17
Improve content for devolved nations

leenagupte:
Trello: https://trello.com/c/Tz6It6TX

# What's changed and why?
Entering a postcode for one of the devolved nations will take you to the generic page for that nation. This means that users in nations other than England may have to enter their postcode twice and visit two different websites to get the information they need.

# Expected changes

- Display council name for postcode in devolved nation
- Add extra line of content to display the country
- Link directly to results page of Scottish postcode checker on results page for Scotland

| Before             |  After|
:-------------------------:|:-------------------------:
Screenshot 2020-12-17 at 14 20 28|Screenshot 2020-12-17 at 14 07 23
Screenshot 2020-12-17 at 14 21 48|Screenshot 2020-12-17 at 13 53 25
Screenshot 2020-12-17 at 14 23 08|Screenshot 2020-12-17 at 13 52 23

### DIFF
--- a/app/views/coronavirus_local_restrictions/_devolved_nation.html.erb
+++ b/app/views/coronavirus_local_restrictions/_devolved_nation.html.erb
@@ -22,7 +22,7 @@
   <% if location.scotland? %>
     <%= render "govuk_publishing_components/components/button", {
       text: t("coronavirus_local_restrictions.results.devolved_nations.scotland.guidance.label"),
-      href: t("coronavirus_local_restrictions.results.devolved_nations.scotland.guidance.link"),
+      href: t("coronavirus_local_restrictions.results.devolved_nations.scotland.guidance.link", postcode: postcode.gsub(/\s+/, '')),
       margin_bottom: 9
     } %>
   <% elsif location.northern_ireland? %>
--- a/app/views/coronavirus_local_restrictions/_devolved_nation.html.erb
+++ b/app/views/coronavirus_local_restrictions/_devolved_nation.html.erb
@@ -15,6 +15,10 @@
   <p class="govuk-body">
     <%= t("coronavirus_local_restrictions.results.match") %> <strong><%= postcode %></strong> to <strong><%= location_lookup.lower_tier_area_name %></strong>.
   </p>
+  <p class="govuk-body">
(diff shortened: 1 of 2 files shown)


GDSNewt (approved):
Looks great! 

I checked the NI and Wales journeys, which are fine,  and confirmed Scotland retains the postcode as seen below :shipit: 

**GOV.UK POSTCODE CHECKER**
Screenshot 2020-12-17 at 14 37 18

**SCOT GOV**
Screenshot 2020-12-17 at 14 37 12

hannahjump:
Looks great!

app/views/coronavirus_local_restrictions/_devolved_nation.html.erb
```diff
@@ -13,12 +13,16 @@
     title: title
   } %>
   <p class="govuk-body">
-    <%= t("coronavirus_local_restrictions.results.match") %> <strong><%= postcode %></strong> to <strong><%= location.country %></strong>.
+    <%= t("coronavirus_local_restrictions.results.match") %> <strong><%= postcode %></strong> to <strong><%= location_lookup.lower_tier_area_name %></strong>.
```
kevindew, line 16: More a query than a code concern, Is it possible that this area name doesn't return anything? I know we look up using: 

https://github.com/alphagov/collections/blob/15a5df7af75faa265d62063220a7685915970497/app/services/location_lookup_service.rb#L4

and I wondered if Mapit pretty much guarantees a result to one of those or if we have cases where it may be nil and we should fallback to the country name?

leenagupte: I "borrowed" this list from [find-local-council](https://github.com/alphagov/frontend/blob/master/app/controllers/find_local_council_controller.rb#L8-L10). I've tested various supermarkets in the devolved nations and it looks to be ok. From what I can tell, if the postcode matches in Mapit, it will have one of these area codes.

kevindew: 👍  Cool cool - looks like it'll only be an empty string rather than a crash if that was to occur

PR #2172, opened 2020-12-19
Add Tier 4

JonathanHallam:
This adds tier 4 into the postcode checker. As of yet this is just the logic to enable it. Full list of areas to follow. 

## Before
Screenshot 2020-12-19 at 18 17 01

## After (AREA USED AS AN EXAMPLE)
Screenshot 2020-12-19 at 18 20 31

### DIFF
--- a/app/views/coronavirus_local_restrictions/_tier_four_restrictions.html.erb
+++ b/app/views/coronavirus_local_restrictions/_tier_four_restrictions.html.erb
@@ -1,0 +1,49 @@
+<% tier_heading = capture do %>
+  <%= t("coronavirus_local_restrictions.results.level_four.heading_pretext") %>
+  <span class="coronavirus-local-restriction__heading-tier-label">
+    <%= t("coronavirus_local_restrictions.results.level_four.heading_tier_label") %>
+  </span>
+<% end %>
+
+<%
+  title = if restriction.present? && restriction.future.present?
+    t("coronavirus_local_restrictions.results.changing_levels_title")
+  else
+    tier_heading
+  end
+%>
+
+<% content_for :title, strip_tags(title) %>
+
+<%= tag.div :data => {
+  :module => "coronavirus-track-local-restriction-results",
+  "tracking-label" => "Tier four: #{restriction.area_name}"
+} do %>
+  <%= render "govuk_publishing_components/components/title", {
+    title: sanitize(title)
+  } %>
+
+  <p class="govuk-body">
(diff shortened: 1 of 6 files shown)


app/views/coronavirus_local_restrictions/_tier_four_restrictions.html.erb
```diff
@@ -0,0 +1,50 @@
...
+  <% end %>
+
+  <%= render "govuk_publishing_components/components/button", {
+    text: t("coronavirus_local_restrictions.results.guidance_label", level: 4),
+    href: t("coronavirus_local_restrictions.results.level_four.guidance_link"),
+    margin_bottom: 9
+  } %>
+
+  <%= render partial: "coronavirus_local_restrictions/future_restrictions", locals: { restriction: restriction } %>
+  <%= render partial: "coronavirus_local_restrictions/christmas_rules" %>
```
leenagupte, line 48: I think Christmas rules don't apply to tier 4.

richardTowers (approved):
This looks good to me (code review, plus a bit of manual testing tweaking the restrictions myself). Other than @leenagupte's comment about the christmas rules, which I think you might be looking at separately?

app/views/coronavirus_local_restrictions/_tier_four_restrictions.html.erb
JonathanHallam: So this doesn't matter any more. We'll have some work to do to introduce some logic for the other tiers once that goes live again.

test/integration/coronavirus_local_restrictions_test.rb
```diff
@@ -198,6 +224,14 @@ def then_i_enter_a_valid_english_postcode_in_tier_three
     fill_in "Enter a full postcode", with: postcode
   end
 
+  def then_i_enter_a_valid_english_postcode_in_tier_four
+    @area = "Anoat"
```
leenagupte, line 228: Nice!

test/integration/coronavirus_local_restrictions_test.rb
```diff
@@ -175,6 +190,17 @@ def then_i_enter_a_valid_english_postcode_with_a_future_level_three_restriction
     fill_in "Enter a full postcode", with: postcode
   end
 
+  def then_i_enter_a_valid_english_postcode_with_a_future_level_four_restriction
+    @area = "Kashyyyk"
```
leenagupte, line 194: ❤️

leenagupte (approved):
Looks great! Tested locally and it works like a charm. Guidance links work too! Awesome work! 🥇

PR #2175, opened 2020-12-19
Add back content for Christmas rules

leenagupte:
# What's changed and why?

The Christmas rules were removed from the postcode checker whilst the guidance is being updated as we didn't want to be unintentionally giving people the wrong information.

The Christmas rules only apply to tiers 1 - 3, they do not apply to tier 4, nor do they apply to the places about to go into tier 4.

# Expected changes
Screenshot 2020-12-19 at 18 52 23

Screenshot 2020-12-19 at 18 53 17

# No change
Screenshot 2020-12-19 at 18 54 29

### DIFF
--- a/app/views/coronavirus_local_restrictions/_tier_one_restrictions.html.erb
+++ b/app/views/coronavirus_local_restrictions/_tier_one_restrictions.html.erb
@@ -45,5 +45,10 @@
   } %>
 
   <%= render partial: "coronavirus_local_restrictions/future_restrictions", locals: { restriction: restriction } %>
+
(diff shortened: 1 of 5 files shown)


richardTowers (approved):
Code review wise, this looks good to me - exactly how I would have done it, only slightly better 😅

I've tested it locally, and it does what it says on the tin. Works correctly for test postcodes which are moving to tier 4, and for postcodes that aren't.

Happy for this to be merged once the guidance is updated.

JonathanHallam (approved):
Looks good to me!

PR #2176, opened 2020-12-19
Add tier 4 area data

JonathanHallam:
Update to the local restriction data with tier 4 areas.

### DIFF
--- a/config/local_restrictions.yml
+++ b/config/local_restrictions.yml
@@ -179,38 +179,38 @@ E06000030:
 E06000031:
   name: Peterborough City Council
   restrictions:
-    - alert_level: 2
-      start_date: 2020-12-02
-      start_time: "00:01"
     - alert_level: 3
       start_date: 2020-12-19
+      start_time: "00:01"
+    - alert_level: 4
+      start_date: 2020-12-20
       start_time: "00:01"
 E06000032:
   name: Luton Borough Council
   restrictions:
-    - alert_level: 2
-      start_date: 2020-12-02
-      start_time: "00:01"
     - alert_level: 3
       start_date: 2020-12-19
+      start_time: "00:01"
+    - alert_level: 4
+      start_date: 2020-12-20
       start_time: "00:01"
 E06000033:
   name: Southend-on-Sea Borough Council
   restrictions:
-    - alert_level: 2
-      start_date: 2020-12-02
-      start_time: "00:01"
     - alert_level: 3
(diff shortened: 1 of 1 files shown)


config/local_restrictions.yml
```diff
@@ -1557,165 +1563,201 @@ E07000094:
...
+      start_time: "00:01"
 E07000114:
   name: Thanet District Council
   restrictions:
     - alert_level: 3
       start_date: 2020-12-02
       start_time: "00:01"
+      - alert_level: 4
+        start_date: 2020-12-21
+        start_time: "00:01"
```
richardTowers, line 1742: This is indented one level too far.

```suggestion
    - alert_level: 4
      start_date: 2020-12-21
      start_time: "00:01"
```

config/local_restrictions.yml
```diff
@@ -179,105 +179,111 @@ E06000030:
...
   name: Peterborough City Council
   restrictions:
-    - alert_level: 2
-      start_date: 2020-12-02
-      start_time: "00:01"
     - alert_level: 3
       start_date: 2020-12-19
       start_time: "00:01"
+    - alert_level: 4
+      start_date: 2020-12-21
```
leenagupte, line 186: I think this needs to be `2020-12-20`

config/local_restrictions.yml
JonathanHallam: good spot, thanks

config/local_restrictions.yml
JonathanHallam: 🤦 I copied it from an earlier notes file after my editor crashed. Replaced them all now

richardTowers:
Some `jq` sneaky tricks that I'm using to review this:

```
ruby -rjson -ryaml -e 'puts YAML.load_file("config/local_restrictions.yml").to_json' |
  jq -r '.[] | select(.restrictions[].alert_level == 4) | .name'
```

Returns a list of 98 LTLAs which are moving into tier 4. I'm just matching them up with the spreadsheet.

richardTowers (approved):
I've checked that:

* The LTLAs moving to tier 4 are the same as in the spreadsheet that was signed off
* Every tier 4 alert level has the same configuration
* The tiers being removed are all in the past

richardTowers (approved):
Also tested manually for a few postcodes.

leenagupte (approved):
🥇

PR #2178, opened 2020-12-21
Fix bug for tier 1 postcodes

JonathanHallam:
Over the weekend we introduced some code to conditionally display christmas rules. Whilst this worked for the overwhelming majority of places, those areas in tier 1 don't technically have a restriction according to the code. This means that for those users things won't have displayed. This adds an extra check for the presence of a restriction before checking the future level of it which fixes the issue.


kevindew (approved):
Nice fix. It’d be good to have some tests updated too - as seemingly code paths aren’t executed in them the same way as live.

JonathanHallam:
@kevindew agreed! I looked into the tests and figured it should have been picked up, think you must be right on the execution. I'll take a closer look once this is in and people can see things again.

PR #2179, opened 2020-12-21
Refactor the postcode search modelling

kevindew:

Oh my.. this has taken a surprisingly long time. I wrote the code for this last Friday and never seemed to find a moment to get it finished until the subsequent Friday - and by then the world of postcode checking had changed somewhat.

This was initially started following feedback on  https://github.com/alphagov/collections/pull/2136 which led to some pondering about how things should be modelled. The aim of this work was to see if the concepts involved in modelling the postcode search could be more specific and modular, with the aim of different contexts (such as search or results) having more specific objects to work with in the views.

The Tier 1 change in https://github.com/alphagov/collections/pull/2163 is quite significant, so I cherry picked the Tier 1 changes into this so they could be considered.

So anyway what does this do?

- Renames  LocalRestriction to CoronavirusRestrictionArea and refactors it to use a nested class for a Restriction
  - **Why:** this class didn't represent a restriction it represented a geographic region, I found it was confusing talking about an object as a restriction when it contained a collection of restrictions. Moving to a Restriction object resolves the need to understand the hash and [parse time](https://github.com/alphagov/collections/blob/f57e49c188b997026331dc6ced375857f235f34b/app/controllers/coronavirus_local_restrictions_controller.rb#L41) in disparate places.
  - Note: this does rename the YAML file again (I can keep the old name if we're worried that is disruptive, though I still think of it as part of a private API)
- Rename MapitPostcodeResponse to MapitLocation and add a static method to load data
  - **Why:** MapitLocation seemed to better capture this concept as a model. The code to load Mapit data is relatively simple so this code felt a bit neater ported from LocationLookupService to here.
- Rename PostcodeLocalRestrictionSearch to CoronavirusRestrictionSearch
  - **Why:** postcode seemed an odd choice of naming as a user isn't searching for a Postcode local restriction they're searching for Coronavirus restrictions, postcode is the input.
- Create CoronavirusRestrictionSearch::DevolvedNationResult and CoronavirusRestrictionSearch::EnglandResult classes
  - **Why:** the former PostcodeLocalRestrictionSearch class felt like it had multiple personalities modelling both a search and a result. Creating result classes allows it to model a search and for it to produce result objects. England and devolved nation results seem to have very little in common so they were a simple split without duplication of logic.
- Create a devolved_nation_result and england_result view
  - **Why:** This simplifies dealing with results by having devolved nations handled entirely separate
- Remove LocationLookupService
  - **Why:** the responsibilities of this class have moved to MapitLocation and DevolvedNationResult classes
- Raise an error when a CoronavirusRestrictionArea doesn't have any current restrictions
  - **Why:** since we've moved to modelling tier 1 restrictions it doesn't make any sense for an area to not have a restriction, and we don't code the UI for this situation. Raising an error allows us to be confident to not use any guard clauses when working with a current_restriction.
  
I'm sorry this is rather long. Reviewing it is best to do commit by commit, they're hopefully pretty straight forward and there's plenty more context in the commit messages.

There were a few additional things I wondered about too (which would be separate PRs)

- The YAML seems a bit unnecessarily verbose with start_date and start_time as separate fields (which convert to Ruby a bit strangely with start_date being a Date object and start_time being a String). I didn't know if we had an advantage in setting it up this way (perhaps it's related to how we receive the data. If not I'd propose we change to it being either `start_time: 2020-1-1 00:01` or `start: 2020-1-1 00:01`. What do people think? I guess this might be disruptive :(
- We seem to have a mix of locale keys where some reference level_* and others reference tier_*. Would it be good to make these all be tier_*? Would that have risks of ruining any content designers working patterns?
- Should we consider a no mapit match an invalid postcode (for example: [SE1 9XX](https://www.gov.uk/find-coronavirus-local-restrictions?postcode=SE1+9XX). It seems a bit confusing that we tell a user "Enter a postcode" when they have done so?


app/models/coronavirus_restriction_search.rb
```diff
@@ -38,19 +35,54 @@ def invalid_postcode?
     %w[invalidPostcodeFormat fullPostcodeNoMapitValidation].include?(error_code)
   end
 
-  def no_restriction?
-    location_lookup.data.first&.england? && local_restriction.nil?
+  def devolved_nation?
+    ["Northern Ireland", "Scotland", "Wales"].include?(locations&.first&.country)
```
leenagupte, line 39: We have helper methods for the country in MapitLocation, e.g. `northern_ireland?` etc. Can they be used instead. e.g

```
location = locations&.first
location.wales? || location.scotland? || location.northern_ireland?
```

Or 

```
!locations&.first&.england?
```

app/models/coronavirus_restriction_search.rb
```diff
@@ -38,19 +35,54 @@ def invalid_postcode?
...
+  end
+
+  def england_result
+    return if error_code || locations.empty? || devolved_nation?
+
+    @england_result ||= begin
+      restriction_areas = locations.filter_map do |area|
+        CoronavirusRestrictionArea.find(area.gss)
```
leenagupte, line 47: I'd forgotten what `CoronavirusRestrictionArea` is and had to go and look it up again. 
The confusion with the class would be reduced if the list was called `local_restrictions` rather than `restriction_areas` to remind us of what we we're getting back.

app/models/coronavirus_restriction_search/devolved_nation_result.rb
```diff
@@ -0,0 +1,36 @@
...
+    country == "Northern Ireland"
+  end
+
+  def scotland?
+    country == "Scotland"
+  end
+
+  def wales?
+    country == "Wales"
+  end
```
leenagupte, line 31: Why have these been repeated here, when they are already in the MapitLocation response?

app/models/coronavirus_restriction_search/devolved_nation_result.rb
```diff
@@ -0,0 +1,36 @@
+class CoronavirusRestrictionSearch::DevolvedNationResult
+  AREA_TYPES_FOR_NAME = %w[COI LBO LGD MTD UTA DIS].freeze
```
leenagupte, line 2: `AREA_TYPES_FOR_NAME ` is too generic  as these are specifically lower tier area names. So can the old name, `LOWER_TIER_AREA_CODES` be reinstated as I think that better describes what they are.

leenagupte (commented):
Hi! I got to commit "Refactor CoronavirusRestrictionSearch" and had to stop. There are a lot of changes happening that are very large, and it would be really helpful if this PR could be split into smaller changes to reduce the overhead of the review.

Some things I've noticed are that a lot of things are being renamed potentially unnecessarily, for example the the data for the restrictions, which is the most edited file in the checker.

Also, I stopped at "Refactor CoronavirusRestrictionSearch" because I think the way the Results is being modelled needs to be re-thought. 
I think separating out England and what English results have access to is too rigid. 
As we saw recently, being able to add more tiers etc, or change the data that can be displayed on different types of results pages at short notice needs to be as easy as possible. 
Being able to find country information one place helps that. Ditto being able to get information like the lower tier area name.

I think it's easier to follow if the controller checks for `result.england?` etc and the views picks out the information it needs from a full Results object rather than being limited to just its area name (I'm thinking about the lower tier area name here) as it's entirely possible we could need it. 
For example, up until recently, the "lower tier area name" area name was only being used on the results page for Tier 1. If this refactor was done a few weeks ago then it's more than probable that the "lower tier area name" would have been restricted to the EnglandResult object and not the DevolvedNationResult, and would have made the change to display the council name osn the devolved nations page much harder to implement.

It's great that you are trying to improve the code base, but any changes need to keep flexibility at the heart of them. Changes are expected quickly, so the easier it is to get at the data needed the better.

kevindew:
Thanks @leenagupte - hmm this seems a lot more divisive than I anticipated, sorry to rock the boat. I've done as you suggested and opened up a separate smaller PR for just the results part: https://github.com/alphagov/collections/pull/2199

PR #2181, opened 2020-12-21
Fix, test and simplify the postcode pattern

kevindew:
It turned out that this pattern wasn't matching the infamously weird GIR
0AA postcode because of the unescaped white space. Whilst fixing this I
thought it would be good if this had some tests, and then I saw the
opportunity to make it case insensitive to cut out a lot of the
matching.

### DIFF
--- a/app/models/postcode_local_restriction_search.rb
+++ b/app/models/postcode_local_restriction_search.rb
@@ -1,16 +1,12 @@
 class PostcodeLocalRestrictionSearch
   UK_POSTCODE_PATTERN = %r{
     \A
-    ([Gg][Ii][Rr] 0[Aa]{2}) # Unusual postcodes
-    |
-    (
-      # Outward code, for example SW1A
-      (([A-Za-z][0-9]{1,2})|(([A-Za-z][A-Ha-hJ-Yj-y][0-9]{1,2})|(([A-Za-z][0-9][A-Za-z])|([A-Za-z][A-Ha-hJ-Yj-y][0-9][A-Za-z]?))))
-      \s?
-      [0-9][A-Za-z]{2} # Inward code, for example 2AA
-    )
+    # Outward code, for example SW1A
+    (([A-Z][0-9]{1,2})|(([A-Z][A-HJ-Y][0-9]{1,2})|(([A-Z][0-9][A-Z])|([A-Z][A-HJ-Y][0-9][A-Z]?))))
+    \s?
+    [0-9][A-Z]{2} # Inward code, for example 2AA
     \Z
-  }x.freeze
+  }xi.freeze
 
   attr_reader :postcode
   delegate :no_information?, to: :location_lookup
--- a/test/models/postcode_local_restriction_search_test.rb
+++ b/test/models/postcode_local_restriction_search_test.rb
@@ -13,6 +13,26 @@ describe PostcodeLocalRestrictionSearch do
     }])
   end
 
+  describe "UK_POSTCODE_PATTERN" do
+    let(:pattern) { described_class::UK_POSTCODE_PATTERN }
+
+    it "matches valid postcodes" do
+      assert_match pattern, "sw1a 2aa"
+      assert_match pattern, "LS11UR"
+      assert_match pattern, "BT15 3JX"
+    end
+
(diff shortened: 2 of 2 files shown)


leenagupte:
@kevindew Is "GIR0AA" a valid postcode? Google Maps and Mapit don't have results for it:

https://mapit.mysociety.org/postcode/GIR0AA

app/models/postcode_local_restriction_search.rb
```diff
@@ -1,16 +1,16 @@
...
       # Outward code, for example SW1A
-      (([A-Za-z][0-9]{1,2})|(([A-Za-z][A-Ha-hJ-Yj-y][0-9]{1,2})|(([A-Za-z][0-9][A-Za-z])|([A-Za-z][A-Ha-hJ-Yj-y][0-9][A-Za-z]?))))
+      (([A-Z][0-9]{1,2})|(([A-Z][A-HJ-Y][0-9]{1,2})|(([A-Z][0-9][A-Z])|([A-Z][A-HJ-Y][0-9][A-Z]?))))
       \s?
-      [0-9][A-Za-z]{2} # Inward code, for example 2AA
+      [0-9][A-Z]{2} # Inward code, for example 2AA
     )
     \Z
-  }x.freeze
+  }xi.freeze
```
leenagupte, line 13: Is it the `i` that makes the regex case-insensitive?

app/models/postcode_local_restriction_search.rb
```diff
@@ -1,16 +1,16 @@
 class PostcodeLocalRestrictionSearch
   UK_POSTCODE_PATTERN = %r{
     \A
-    ([Gg][Ii][Rr] 0[Aa]{2}) # Unusual postcodes
+    (GIR\s?0AA) # Unusual postcodes
```
leenagupte, line 4: Oh! So this isn't actually a postcode, it's for Girobank! https://www.streetcheck.co.uk/postcode/gir0aa
Could that be added to the commit message/comment?

kevindew:
> @kevindew Is "GIR0AA" a valid postcode? Google Maps and Mapit don't have results for it:
> 
> https://mapit.mysociety.org/postcode/GIR0AA

It is valid if we consider "GIR 0AA" valid - the space is optional, which is what the regexp we've got included.

It's a rather infamously quirkly postcode which I don't really understand the current status of. Some more information is here: https://www.ianvisits.co.uk/blog/2020/09/02/londons-non-geographic-postcodes/. It sounds like it probably still exists but has no use for this checker since it doesn't represent a geographical region.

We could remove the first part of the regexp that checks for it? Would just mean the regex is less a UK postcode regex and more one tailored to our needs.

app/models/postcode_local_restriction_search.rb
leenagupte: As it's not a geographical area, do we need to handle it? Mapit returns an error for this postcode regardless, so we're not really saving the user anything.

leenagupte (commented):
> It's a rather infamously quirkly postcode which I don't really understand the current status of. Some more information is here: https://www.ianvisits.co.uk/blog/2020/09/02/londons-non-geographic-postcodes/. It sounds like it probably still exists but has no use for this checker since it doesn't represent a geographical region.
> 
> We could remove the first part of the regexp that checks for it? Would just mean the regex is less a UK postcode regex and more one tailored to our needs.

We're not handling the other non-geographic postcodes in the article, like E50, so I think it's fine to exclude this one. As mentioned in the inline comments, the user is still going to see an error if they search for "GIR0AA", so I think it's fine to exclude it.

The rest looks good though. 💯

kevindew:
Thanks @leenagupte that sounds reasonable. I'll update this later today or tomorrow.

app/models/postcode_local_restriction_search.rb
kevindew: Yup, I've updated the commit message to explain that.

kevindew:
Thanks for the review @leenagupte, this is now updated.

leenagupte (approved):
Looks good! 💯 
Thanks for updating the commit message. And also the regex is much easier to follow now that it's case-insensitive. Thanks! ❤️

PR #2182, opened 2020-12-22
Remove Brexit button AB test

sihugh:
The B variant wins, so we change the default button text to "Brexit checker: start now"

https://trello.com/c/YVC3PxSt/730-tear-down-ab-test-and-set-start-button-to-brexit-checker-start-now


hannako (approved):
🎉

PR #2183, opened 2020-12-22
Stop showing Christmas rules after 25th December

leenagupte:
Trello: https://trello.com/c/cZ7cBdfi

# What's changed and why?

At the moment the Christmas rules only apply to 25 December, so we
should stop showing the rules after that date. However the 25th is a
Friday, and the next working day is the 29th December. That means we
would be showing out of date Christmas rules on the checker for 4 days.

This change uses the date to hide the Christmas rules after 25 December.

The built-in `after?` method is used because that returns true for the
26th December onwards, whereas the `before?` method only returns true
for everything on 24 December and before, not 25 December.

The integration tests that look for Christmas content have also been
consolidated to make them easier to find and remove later on.

# Expected changes

## On and before Christmas Day 🎄 
Screenshot 2020-12-22 at 15 39 07

## After Christmas
Screenshot 2020-12-22 at 15 40 02

### DIFF
--- a/app/controllers/coronavirus_local_restrictions_controller.rb
+++ b/app/controllers/coronavirus_local_restrictions_controller.rb
@@ -43,4 +43,12 @@ private
 
     [time_until_restriction, MAX_CACHE_TIME].min
   end
+
+  def after_christmas?
+    christmas = Time.zone.parse("2020-12-25")
+
+    Time.zone.today.after?(christmas)
+  end
+
+  helper_method :after_christmas?
 end
--- a/app/views/coronavirus_local_restrictions/_tier_one_restrictions.html.erb
+++ b/app/views/coronavirus_local_restrictions/_tier_one_restrictions.html.erb
@@ -46,7 +46,7 @@
 
   <%= render partial: "coronavirus_local_restrictions/future_restrictions", locals: { restriction: restriction } %>
 
-  <% unless restriction.present? && restriction.future_alert_level == 4 %>
+  <% unless restriction.future_alert_level == 4 || after_christmas? %>
     <%= render partial: "coronavirus_local_restrictions/christmas_rules" %>
   <% end %>
 
--- a/test/integration/coronavirus_local_restrictions_test.rb
+++ b/test/integration/coronavirus_local_restrictions_test.rb
@@ -4,6 +4,7 @@ require "gds_api/test_helpers/mapit"
 class CoronavirusLocalRestrictionsTest < ActionDispatch::IntegrationTest
   include GdsApi::TestHelpers::Mapit
   include GdsApi::TestHelpers::ContentStore
+  include ActiveSupport::Testing::TimeHelpers
 
   describe "current restrictions" do
     it "displays the tier one restrictions" do
@@ -13,7 +14,6 @@ class CoronavirusLocalRestrictionsTest < ActionDispatch::IntegrationTest
       then_i_enter_a_valid_english_postcode_in_tier_one
       then_i_click_on_find
       then_i_see_the_results_page_for_level_one
-      then_i_see_details_of_christmas_rules
     end
 
     it "displays the tier two restrictions" do
@@ -22,7 +22,6 @@ class CoronavirusLocalRestrictionsTest < ActionDispatch::IntegrationTest
       then_i_enter_a_valid_english_postcode_in_tier_two
       then_i_click_on_find
       then_i_see_the_results_page_for_level_two
-      then_i_see_details_of_christmas_rules
     end
 
     it "displays the tier three restrictions" do
@@ -31,7 +30,6 @@ class CoronavirusLocalRestrictionsTest < ActionDispatch::IntegrationTest
       then_i_enter_a_valid_english_postcode_in_tier_three
       then_i_click_on_find
       then_i_see_the_results_page_for_level_three
-      then_i_see_details_of_christmas_rules
     end
 
     it "displays the tier four restrictions" do
@@ -40,7 +38,6 @@ class CoronavirusLocalRestrictionsTest < ActionDispatch::IntegrationTest
       then_i_enter_a_valid_english_postcode_in_tier_four
       then_i_click_on_find
       then_i_see_the_results_page_for_level_four
-      then_i_do_not_see_details_of_christmas_rules
     end
 
     it "displays no tier information" do
@@ -132,7 +129,6 @@ class CoronavirusLocalRestrictionsTest < ActionDispatch::IntegrationTest
       then_i_enter_a_valid_english_postcode_with_a_future_level_two_restriction
       then_i_click_on_find
       then_i_see_the_results_page_for_level_one_with_changing_restriction_levels
-      then_i_see_details_of_christmas_rules
     end
 
     it "displays restrictions changing from level two to level three" do
@@ -140,7 +136,6 @@ class CoronavirusLocalRestrictionsTest < ActionDispatch::IntegrationTest
       then_i_enter_a_valid_english_postcode_with_a_future_level_three_restriction
       then_i_click_on_find
       then_i_see_the_results_page_for_level_two_with_changing_restriction_levels
-      then_i_see_details_of_christmas_rules
     end
 
     it "displays restrictions changing from level three to level four" do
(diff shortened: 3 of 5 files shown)


app/controllers/coronavirus_local_restrictions_controller.rb
```diff
@@ -43,4 +43,12 @@ def restriction_expiry(search)
 
     [time_until_restriction, MAX_CACHE_TIME].min
   end
+
+  def before_christmas?
+    christmas = Time.zone.parse("2020-12-25")
+
+    !Time.zone.today.after?(christmas)
```
kevindew, line 50: This might be simpler to read as `Time.zone.now < Time.zone.parse("2020-12-25")`. I find myself a bit thrown off reading this with `.after` in the condition.

test/integration/coronavirus_local_restrictions_test.rb
```diff
@@ -132,22 +129,130 @@ class CoronavirusLocalRestrictionsTest < ActionDispatch::IntegrationTest
...
       given_i_am_on_the_local_restrictions_page
       then_i_enter_a_valid_english_postcode_with_a_future_level_four_restriction
       then_i_click_on_find
       then_i_see_the_results_page_for_level_three_with_changing_restriction_levels
+    end
+  end
+
+  describe "before christmas" do
+    before do
+      travel_to Time.zone.parse("2020-12-25")
```
kevindew, line 151: I'd have thought this is exactly Christmas day rather than being before Christmas? as the description suggests?

app/views/coronavirus_local_restrictions/_tier_one_restrictions.html.erb
```diff
@@ -46,7 +46,7 @@
 
   <%= render partial: "coronavirus_local_restrictions/future_restrictions", locals: { restriction: restriction } %>
 
-  <% unless restriction.present? && restriction.future_alert_level == 4 %>
+  <% if restriction.present? && restriction.future_alert_level != 4 && before_christmas? %>
```
kevindew, line 49: I found myself a bit confused by the inverting of this conditional wondering about the `restriction.present?` side of things. Is that needed anymore since the Tier 1 change? I'd guess we expect every request to this view to have a present restriction.

kevindew (commented):
The conditions look good here - it'll be great to have no-one manually editing this on Christmas day. I'm a little bit confused by the "before Christmas" though because of it actually testing the date.

I do wonder if we've got a testing smell going on with this utility. The LOC for this change is less than 10 lines but that transpires to be over 100 lines of integration tests. Perhaps we're doing too much at integration level. I'm not too sure, at this point, what to suggest though - has this been a concern raised before?

test/integration/coronavirus_local_restrictions_test.rb
leenagupte: It is. I suppose I was being lazy and not calling it "on or before christmas". I'll update the test name though.

app/views/coronavirus_local_restrictions/_tier_one_restrictions.html.erb
leenagupte: You're right, the `restriction.present?` shouldn't be needed anymore, and the view ill fail on `area_name` if the restriction isn't present. I'll take it out.

app/controllers/coronavirus_local_restrictions_controller.rb
leenagupte: I'm never sure if `<` will include Christmas day or not. I'll have a go and see what happens in tests.

leenagupte:
> The conditions look good here - it'll be great to have no-one manually editing this on Christmas day. I'm a little bit confused by the "before Christmas" though because of it actually testing the date.
> 
> I do wonder if we've got a testing smell going on with this utility. The LOC for this change is less than 10 lines but that transpires to be over 100 lines of integration tests. Perhaps we're doing too much at integration level. I'm not too sure, at this point, what to suggest though - has this been a concern raised before?

I guess I was being overly thorough with the integration tests. We don't have any view tests set up though, and I couldn't think of another way to check that the Christmas content wasn't being displayed on the page.

These tests are only temporary as I expect (hope) someone will go and remove everything Christmas related on December 29th.

Rosa-Fox (commented):
Works well :) Wonder if `def before_christmas?` should be within `local_restriction.rb` and then the partials could use `<% if restriction.present? && restriction.future_alert_level != 4 && restriction.before_christmas? %>`? As I know you love a unit test... is it just because it won't be around for long?

app/controllers/coronavirus_local_restrictions_controller.rb
DilwoarH: <= ?

kevindew: Sure, it won't include Christmas. When you do Time.zone.parse without specifying a time it'll default to midnight. So the moment it hits midnight it'll return false.

```
irb(main):005:0> Time.zone.parse('2020-12-24 23:59:59.999') < Time.zone.parse('2020-12-25')
=> true
irb(main):006:0> Time.zone.parse('2020-12-25 00:00') < Time.zone.parse('2020-12-25')
=> false
```

A thing to consider is that for cached responses they'll get this Christmas info for the first 30 mins - but that doesn't seem too big a deal.

test/integration/coronavirus_local_restrictions_test.rb
kevindew: Ah sure, sorry I thought it was more a logical flaw than a moment of laziness.

app/controllers/coronavirus_local_restrictions_controller.rb
leenagupte: So, I'm going to change the logic back to use an `unless` and change the method name to `after_christmas?` which I think reads better. As I'm going that, I think it's only to continue using `Time.zone.today.after?(christmas)`

kevindew:
> > The conditions look good here - it'll be great to have no-one manually editing this on Christmas day. I'm a little bit confused by the "before Christmas" though because of it actually testing the date.
> > I do wonder if we've got a testing smell going on with this utility. The LOC for this change is less than 10 lines but that transpires to be over 100 lines of integration tests. Perhaps we're doing too much at integration level. I'm not too sure, at this point, what to suggest though - has this been a concern raised before?
> 
> I guess I was being overly thorough with the integration tests. We don't have any view tests set up though, and I couldn't think of another way to check that the Christmas content wasn't being displayed on the page.
> 
> These tests are only temporary as I expect (hope) someone will go and remove everything Christmas related on December 29th.

Ok cool - shall we put a card in for that removal? yeah it feels like view tests are probably the way to go to cover this. The integration test file is already quite a beast to navigate and it seems like there is more to come, there seems something wrong if that's trying to exhaust all scenarios.

app/controllers/coronavirus_local_restrictions_controller.rb
kevindew: Sure thing. Would `after_christmas?` mean >= 00:00 on Boxing day?

leenagupte:
> Works well :) Wonder if `def before_christmas?` should be within `local_restriction.rb` and then the partials could use `<% if restriction.present? && restriction.future_alert_level != 4 && restriction.before_christmas? %>`? As I know you love a unit test... is it just because it won't be around for long?

Exactly that, because it won't be around very long. My first attempt had a new helper class set up, which I thought was overkill for something so temporary.

app/controllers/coronavirus_local_restrictions_controller.rb
leenagupte: Yes. Exactly that.

leenagupte: Yes!

```
irb(main):001:0> christmas = Time.zone.parse("2020-12-25")
=> Fri, 25 Dec 2020 00:00:00 GMT +00:00

irb(main):003:0> boxing_day = Time.zone.local(2020, 12, 26, 0, 0, 0)
=> Sat, 26 Dec 2020 00:00:00 GMT +00:00
irb(main):004:0> boxing_day.after?(christmas)
=> true

```

leenagupte:
> > > The conditions look good here - it'll be great to have no-one manually editing this on Christmas day. I'm a little bit confused by the "before Christmas" though because of it actually testing the date.
> > > I do wonder if we've got a testing smell going on with this utility. The LOC for this change is less than 10 lines but that transpires to be over 100 lines of integration tests. Perhaps we're doing too much at integration level. I'm not too sure, at this point, what to suggest though - has this been a concern raised before?
> > 
> > 
> > I guess I was being overly thorough with the integration tests. We don't have any view tests set up though, and I couldn't think of another way to check that the Christmas content wasn't being displayed on the page.
> > These tests are only temporary as I expect (hope) someone will go and remove everything Christmas related on December 29th.
> 
> Ok cool - shall we put a card in for that removal? yeah it feels like view tests are probably the way to go to cover this. The integration test file is already quite a beast to navigate and it seems like there is more to come, there seems something wrong if that's trying to exhaust all scenarios.

I think there's already a card for this, but I'll update. And yes, view tests would be great. I'll add another card to the backlog to convert some of the integration tests to view tests.

app/controllers/coronavirus_local_restrictions_controller.rb
kevindew: Sure gotcha - I found myself a bit confused by the Time.zone.today - my brain struggles to realise that you're actually comparing a date object with a time object.

kevindew (approved):

PR #2185, opened 2020-12-22
Add partial that displays "an out of date warning"

leenagupte:
Trello: https://trello.com/c/NYHZel2U

# What's changed and why?
Sometimes restrictions are announced before we have had a chance to
update the postcode checker. It can be hours before the checker is
updated with the latest restriction data.

Users coming to the checker will expect to see up to date information.
We need a way to let them know that their restriction may be changing
and that they should come back later.

A flag on the LocalRestriction model is being used to set whether or not
the data is up to date. This flag is stopgap measure so that the warning
can be displayed in time for the latest set of updates. The upside to
this approach is that it is fast to implement. The downside is that the
application needs to be deployed every time the flag changes.

The long term plan is to use a rake task to set a value in Redis that
determines when the warning should be shown in a similar manner to the
emergency banner.

# Expected changes
# Restrictions are out of date
Screenshot 2020-12-22 at 17 36 29
Screenshot 2020-12-22 at 17 36 44

# Restrictions are up to date
Screenshot 2020-12-22 at 17 37 21
Screenshot 2020-12-22 at 17 37 38

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -40,6 +40,7 @@ $covid-yellow: #fff500;
 @import 'govuk_publishing_components/components/translation-nav';
 @import "govuk_publishing_components/components/error-summary";
 @import "govuk_publishing_components/components/back-link";
+@import "govuk_publishing_components/components/warning-text";
 
 @import "helpers/layouts";
 @import "helpers/margins";
--- a/app/controllers/coronavirus_local_restrictions_controller.rb
+++ b/app/controllers/coronavirus_local_restrictions_controller.rb
@@ -51,4 +51,10 @@ private
   end
 
   helper_method :after_christmas?
+
+  def out_of_date?
+    false
+  end
+
+  helper_method :out_of_date?
 end
--- a/app/views/coronavirus_local_restrictions/_out_of_date_warning.html.erb
+++ b/app/views/coronavirus_local_restrictions/_out_of_date_warning.html.erb
@@ -1,0 +1,5 @@
+<% if out_of_date? %>
+  <%= render "govuk_publishing_components/components/warning_text", {
+    text: t("coronavirus_local_restrictions.out_of_date_warning")
+  } %>
+<% end %>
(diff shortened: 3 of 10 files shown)


app/models/local_restriction.rb
```diff
@@ -1,4 +1,6 @@
 class LocalRestriction
+  OUT_OF_DATE = false
```
kevindew, line 2: This doesn't really feel the domain of the restriction, given it's not the restriction that is out of date it's the whole system - for something quick it'd probably make most sense as a helper in the controller.

kevindew (commented):
Nice one, this looks pragmatic. It seems a tricky problem to solve in the longer term but this looks like it'd be a great UX improvement in the short term.

app/models/local_restriction.rb
leenagupte: I um'ed and ah'ed about where to put this. A helper in the controller was one of the options I tried, but wasn't sure if that was appropriate. I picked LocalRestriction because it is the restriction data that is out of date, but I'm happy to move it.

leenagupte: I'm not sure how to stub the helper method in the integration tests. That is another bonus of putting the flag in the model!

kevindew:
I think we should also consider dropping the cache time to be a max of 3 mins (or something similar) when this is in effect, that should save people seeing the warning for ages on the /find-coronavirus-local-restriction page when most of the results actually come back up to date.

I also had a thought that it might be good to put the data about whether restrictions are in and out of date as data in the content item - since that is already a place where external information is fed in. That would save having to add Redis as a dependency to this app and mean no extra lookups.

kevindew (approved):
Thanks 👍

leenagupte:
> I think we should also consider dropping the cache time to be a max of 3 mins (or something similar) when this is in effect, that should save people seeing the warning for ages on the /find-coronavirus-local-restriction page when most of the results actually come back up to date.
> 
> I also had a thought that it might be good to put the data about whether restrictions are in and out of date as data in the content item - since that is already a place where external information is fed in. That would save having to add Redis as a dependency to this app and mean no extra lookups.

@kevindew Thanks for your review. I've been thinking about the cache-times. As it can take hours for the data to be updated, I do wonder if 3 minutes is long enough. As for using Redis as the longer term plan, it was only an idea as something we could turn on without having to deploy the app, but if adding data to the content item is simpler, I'm happy to go with that.

PR #2186, opened 2020-12-23
Make Christmas partial changes

JonathanHallam:
Now that we can be confident that there are no changes before to tiers before Christmas we need to make some changes to how things are displayed. This flips the partials for Christmas rules and future restrictions to ensure that things are displayed in chronological order. It also changes the conditional showing of the Christmas rules because there will be no new areas in tier 4 before Christmas. There was a test associated with this that has also been changed to reflect the fact that Christmas rules should always display unless an area is currently in Tier 4 or it's after Christmas.

## Before (POSTCODE USED AS AN EXAMPLE, NOT REAL DATA)
Screenshot 2020-12-23 at 08 51 54

## After (POSTCODE USED AS AN EXAMPLE, NOT REAL DATA)
Screenshot 2020-12-23 at 08 53 17

### DIFF
--- a/test/integration/coronavirus_local_restrictions_test.rb
+++ b/test/integration/coronavirus_local_restrictions_test.rb
@@ -197,11 +197,11 @@ class CoronavirusLocalRestrictionsTest < ActionDispatch::IntegrationTest
       then_i_see_details_of_christmas_rules
     end
 
-    it "does not display christmas rules for future tier four restrictions" do
+    it "displays christmas rules for future tier four restrictions" do
       given_i_am_on_the_local_restrictions_page
       then_i_enter_a_valid_english_postcode_with_a_future_level_four_restriction
       then_i_click_on_find
-      then_i_do_not_see_details_of_christmas_rules
+      then_i_see_details_of_christmas_rules
     end
   end
 
--- a/app/views/coronavirus_local_restrictions/_tier_one_restrictions.html.erb
+++ b/app/views/coronavirus_local_restrictions/_tier_one_restrictions.html.erb
@@ -44,11 +44,10 @@
     margin_bottom: 9
   } %>
 
-  <%= render partial: "coronavirus_local_restrictions/future_restrictions", locals: { restriction: restriction } %>
-
-  <% unless restriction.future_alert_level == 4 || after_christmas? %>
+  <% unless after_christmas? %>
     <%= render partial: "coronavirus_local_restrictions/christmas_rules" %>
   <% end %>
 
+  <%= render partial: "coronavirus_local_restrictions/future_restrictions", locals: { restriction: restriction } %>
   <%= render partial: "coronavirus_local_restrictions/travel_guidance" %>
 <% end %>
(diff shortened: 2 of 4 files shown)


test/integration/coronavirus_local_restrictions_test.rb
```diff
@@ -148,7 +148,7 @@ class CoronavirusLocalRestrictionsTest < ActionDispatch::IntegrationTest
       then_i_enter_a_valid_english_postcode_with_a_future_level_four_restriction
       then_i_click_on_find
       then_i_see_the_results_page_for_level_three_with_changing_restriction_levels
-      then_i_do_not_see_details_of_christmas_rules
+      then_i_see_details_of_christmas_rules
```
DilwoarH, line 151: Are we testing to see if you're in tier 4 that you don't see Christmas rules?

DilwoarH (approved):
Looks good to me

test/integration/coronavirus_local_restrictions_test.rb
kevindew, line 151: This confused me a bit as I wondered why it had changed. Having a little think I wonder if it's best to remove the Christmas assertions from these restriction changing test since the restriction changing has no impact on that and it's already tested in the various ["displays the tier x restrictions"](https://github.com/alphagov/collections/blob/540fe2c2f0411ab41dccb487bec3fc6a13dd55fd/test/integration/coronavirus_local_restrictions_test.rb#L19) tests.

kevindew (commented):
Looks good, it's nice to be simpler. You might need to co-ordinate this with https://github.com/alphagov/collections/pull/2183 as there'll need to be some rebasing.

test/integration/coronavirus_local_restrictions_test.rb
JonathanHallam: yes! https://github.com/alphagov/collections/blob/master/test/integration/coronavirus_local_restrictions_test.rb#L43

I removed it here because all we care about now that there are no tier 4 things coming in before christmas

JonathanHallam: yeah, excellent shot. Have taken them out

test/integration/coronavirus_local_restrictions_test.rb
```diff
@@ -197,11 +197,11 @@ class CoronavirusLocalRestrictionsTest < ActionDispatch::IntegrationTest
       then_i_see_details_of_christmas_rules
     end
 
-    it "does not display christmas rules for future tier four restrictions" do
+    it "displays christmas rules for future tier four restrictions" do
```
leenagupte, line 200: I don't understand. I thought Christmas rules never apply to a place going into tier 4?

JonathanHallam: We should now because we know that no new tier 4 restrictions are coming into effect before Christmas. In its current state it won't show the Christmas rules to someone currently in Tier 3 going into Tier 4 after Christmas.

leenagupte: Ok. That makes sense.

JonathanHallam: nice one, thanks :)

leenagupte (approved):
LGTM 👍

PR #2187, opened 2020-12-23
Adjust the cache time while the service is being updated

kevindew:
Trello: https://trello.com/c/NYHZel2U/1017-callout-box-for-when-service-is-being-updated-quick-version

When we're updating the service we shouldn't be returning a 30 minute
cache time otherwise users will have a rather odd staggered experience
where the /find-coronavirus-local-restrictions start page shows a
warning long after most of the results are showing updated data.

Setting this to 5 minutes should offer a reasonable balance of not
having our servers hit too frequently during the change, with only a
modest delay to wait for the change to take effect.

### DIFF
--- a/app/controllers/coronavirus_local_restrictions_controller.rb
+++ b/app/controllers/coronavirus_local_restrictions_controller.rb
@@ -1,8 +1,9 @@
 class CoronavirusLocalRestrictionsController < ApplicationController
+  OUT_OF_DATE_CACHE_TIME = 5.minutes
   MAX_CACHE_TIME = 30.minutes
 
   def show
-    expires_in(MAX_CACHE_TIME, public: true)
+    expires_in(cache_time, public: true)
     @content_item = content_item.to_hash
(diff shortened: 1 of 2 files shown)


leenagupte (approved):
This looks good. 👍  
The only concern I have is the cache time. Is 5 minutes enough? 
Usually our traffic spikes in the hours between the new tiers being announced and the checker being updated.

kevindew:
> This looks good. 👍
> The only concern I have is the cache time. Is 5 minutes enough?
> Usually our traffic spikes in the hours between the new tiers being announced and the checker being updated.

I get your concern. We have to balance it based on the cost of telling people out of date information. I think ideally it'd be shorter than 5 minutes - maybe even 1 minute, to be sufficiently informative. Seems a balance of risk of confusion for users vs risk of higher load.

This should be offset by using a shared cache for postcodes so hopefully we can avoid mapit having many hits.

PR #2193, opened 2020-12-24
Update /transition content as a result of a deal with the EU

Rosa-Fox:
## Before

landing-page-before

landing-page-announcements-before

## After

landing-page

landing-page-announcements

### DIFF
--- a/config/locales/cy/transition_landing_page.yml
+++ b/config/locales/cy/transition_landing_page.yml
@@ -7,9 +7,9 @@ cy:
     page_header_explainer: i fusnesau a dinasyddion o 1 Ionawr 2021.
     countdown_days_to_go: diwrnod i fynd
     countdown_day_to_go: diwrnod i fynd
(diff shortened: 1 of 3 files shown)


config/locales/cy/transition_landing_page.yml
```diff
@@ -57,6 +57,13 @@ cy:
     comms_header: "Cyhoeddiadau"
     comms:
       links:
+      - link:
+          text: "The UK and the EU announce an outcome to negotiations"
+          path: "/url-for-announcement"
```
richardTowers, line 62: TODO

richardTowers (changes requested):
Looks good to me, other than `/url-for-announcement` which needs updating.

"Request changes" just so we don't merge it by mistake!

bilbof (approved):

richardTowers (approved):

PR #2199, opened 2020-12-31
Separate postcode results into England and devolved nation results

kevindew:

This is a more modest spin off of https://github.com/alphagov/collections/pull/2179 which focuses just on separating the treating of England and devolved nation results with appropriate modelling of those particular results so there are less concerns to understand in the views. There are further changes in https://github.com/alphagov/collections/pull/2179 that make the modelling simpler and less abstract but I'll see how this PR goes before considering opening those. If this proves divisive I'll just close this up.

A brief summary of the changes here are:

So anyway what does this do?

- Create PostcodeLocalRestrictionSearch::DevolvedNationResult and PostcodeLocalRestrictionRestrictionSearch::EnglandResult classes
  - **Why:** the PostcodeLocalRestrictionSearch class felt like it had multiple personalities modelling both a search and a result. Creating result classes allows it to model a search and for it to produce result objects, which allow views to be only concerned with one model. England and devolved nation results seem to have very little in common so they were a simple split without duplication of logic.
- Create a devolved_nation_result and england_result view
  - **Why:** This simplifies dealing with results by having devolved nations handled entirely separate

There's more info in the commits


app/views/coronavirus_local_restrictions/devolved_nation_result.html.erb
```diff
@@ -0,0 +1,50 @@
...
+    } do %>
+      <%= render "govuk_publishing_components/components/title", {
+        title: title
+      } %>
+      <p class="govuk-body">
+        <%= t("coronavirus_local_restrictions.results.match") %>
+        <strong><%= @result.postcode %></strong>
+        to
+        <strong><%= @result.area_name %></strong>.
+      </p>
```
leenagupte, line 25: This content is missing from the page: 

`<%= t("coronavirus_local_restrictions.results.devolved_nations.country_rules", country: location.country) %>`

leenagupte (commented):
Thanks for splitting this out into a smaller PR. It made it much easier to review.

I haven't had a chance to test this locally yet, but more I do, I wanted mention a concern I have. I'm still not convinced about how the results have been split. 
My main worry is that all of the data from Mapit is not available to both the England and the Devolved Nations results, and I worry that this will make it harder to update the checker (usually at the last minute) in the future.
I think the only difference between the EnglandResult and the DevolvedNationResult should be that the EnglandResult also has Local Restrictions data for the postcode.

I've added some other smaller comments inline, but that is the main concern. As ever the postcode checker needs to be a flexible and as easy to update as possible. For example, can I easily pluck more data out of Mapit? Is it easy to display that data anywhere I want?

kevindew:
> Thanks for splitting this out into a smaller PR. It made it much easier to review.

No problem

> I haven't had a chance to test this locally yet, but before I do, I wanted mention a concern I have. I'm still not convinced about how the results have been split.
> My main worry is that all of the data from Mapit is not available to both the England and the Devolved Nations results, and I worry that this will make it harder to update the checker (usually at the last minute) in the future.
> I think the only difference between the EnglandResult and the DevolvedNationResult should be that the EnglandResult also has Local Restrictions data for the postcode.

Sure, I don't really see it being valuable to have the data there available when nothing is using it. It's just extra code to test that nothing actually executes. It seems fairly trivial to include mapit data in an England result model should it be needed by injecting the location_lookup. 

I understand your concern about flexibility but paradoxically I think in ways this approach makes it easier and safer to adapt as you have just one thing that is used in the view, the result, and that just needs to interface to what is needed. Whereas the current approach we have various concepts in the view that you have to consider. This seemed to catch us out before when we had to consider whether any restriction existed or not and broke all the tier 1 results.

> I've added some other smaller comments inline, but that is the main concern. As ever the postcode checker needs to be a flexible and as easy to update as possible. For example, can I easily pluck more data out of Mapit? Is it easy to display that data anywhere I want?

It should be easy and there should probably be a consistent interface to do so. I think that is an area where there is room for improvement as it took me a long time to get my head around what we're doing there and still find it rather nuanced (why only care about first item in location_lookup.data sometimes, I don't know what a lower tier area is) - I think there's possibly a missing model there that already makes the mapit stuff rather confused.

leenagupte:
@kevindew I think we've reached a bit of an impasse. Whilst I definitely agree that the routing logic should be moved out of the view and into the controller, and that the `@search` object should be split up to return a separate `@result` I don't agree with how the `result` is being constructed, or in this case split up. 
I feel that for the postcode checker it is important to make it possible to just pluck out that data you need as and when you need it, without the extra overhead of having to modify the results classes as well. Honestly I'd be happy with one type of results object and letting the controller figure out the routing by using `@result.devolved_nation?` etc.

> Whereas the current approach we have various concepts in the view that you have to consider
These routing concerns should definitely be moved from the view into the controller. If we were routing results for England in the controller, and checking for a restriction there, I think we would have avoided the tier 1 problems.

> I don't know what a lower tier area is
Lower tier is the local authority. I think I called it "lower tier" because Mapit also returns an "upper tier" which confusingly is also the local authority. The vernacular in the team at the time was also around LTLAs - lower tier local authorities.

kevindew:
Since this still seems divisive and given the postcode checker is currently offline I've closed this in favour of a Trello card to look into improving the modelling: https://trello.com/c/Ra3whanR/502-refactoring-coronavirus-postcode-checker

PR #2202, opened 2021-01-04
Strip all postcode parameters from being sent to Analytics

kevindew:
Trello: https://trello.com/c/LiAOrW5F/1000-remove-or-redact-values-at-the-end-of-postcode-look-up-string-that-dont-match-postcode-format

This depends on https://github.com/alphagov/static/pull/2367 and can't be merged until that is released.

This changes our approach to stripping postcodes out of strings that are
sent to Google Analytics. Rather than matching a postcode pattern we
instead match any instances of a query string that has a postcode=
parameter and strip the value of that parameter.

This has been done as we've had occurrences of users entering PII that
is not a postcode and, as it doesn't match our postcode pattern, it is not
stripped. This instead strips any user input to the postcode field.

## Difference:

Given a URL of `https://www.integration.publishing.service.gov.uk/find-coronavirus-local-restrictions?postcode=52+made+up+address+London+E1+8QS`

Before this change this would be submitted to Analytics as: `https://www.integration.publishing.service.gov.uk/find-coronavirus-local-restrictions?postcode=52+made+up+address+London+[postcode]`
After this change this would be submitted to Analytics as: `https://www.integration.publishing.service.gov.uk/find-coronavirus-local-restrictions?postcode=[postcode]`

### DIFF
--- a/app/views/coronavirus_local_restrictions/_meta_tags.html.erb
+++ b/app/views/coronavirus_local_restrictions/_meta_tags.html.erb
@@ -6,6 +6,6 @@
   <meta property="og:title" content="<%= t("coronavirus_local_restrictions.lookup.meta_title") %>">
   <meta property="og:description" content="<%= t("coronavirus_local_restrictions.lookup.meta_description") %>">
   <meta name="description" content="<%= t("coronavirus_local_restrictions.lookup.meta_description") %>">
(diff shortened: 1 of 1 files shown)


JonathanHallam (approved):
Given the static PR has tests and has been through the review process and the variable names match up I'm happy. Thanks for this :)

PR #2205, opened 2021-01-04
Cache time variable

kevindew:


kevindew:
Doesn't look like this will be needed anymore.

leenagupte:
> Doesn't look like this will be needed anymore.

I'm removing it. It took ages to figure out the tests too!

leenagupte:
> Doesn't look like this will be needed anymore.

😞

PR #2206, opened 2021-01-04
Prepare C19 page for an update

alex-ju:
This makes the frontend code for the `/coronavirus` lading page a bit more resilient to content change – that comes from https://github.com/alphagov/govuk-coronavirus-content/blob/master/content/coronavirus_landing_page.yml – as follows:

 - Hide the bespoke instruction list when no content is presented (we expect [a few content lines](https://github.com/alphagov/govuk-coronavirus-content/blob/master/content/coronavirus_landing_page.yml#L7-L13) to be removed and we're updating the frontend code to cope with this)
 - Allow a custom description under the level 2 heading (we expect some description in form of paragraphs or list items to be added)
 - Increase size for level 2 heading (as requested by design, this is the only change that will be visible before any updates will be made via the [govuk-coronavirus-content](https://github.com/alphagov/govuk-coronavirus-content/blob/master/content/coronavirus_landing_page.yml#L7-L13)

<table>
<tr><th>Before</th><th>After</th></tr>
<tr><td valign="top">

localhost_3070_coronavirus (1)

</td><td valign="top">

localhost_3070_coronavirus

</td></tr>
</table>

### DIFF
--- a/app/assets/stylesheets/views/_covid.scss
+++ b/app/assets/stylesheets/views/_covid.scss
@@ -1,4 +1,4 @@
-$c19-landing-page-header-background: #0843a1;
+$c19-landing-page-header-background: #272828;
 
 .covid__page {
   .govuk-grid-column-one-third,
@@ -22,7 +22,7 @@ $c19-landing-page-header-background: #0843a1;
 .covid__page-header {
   margin-bottom: govuk-spacing(8);
   border-top: solid govuk-spacing(2) $covid-yellow;
-  background-color: $c19-landing-page-header-background;
+  background-color: govuk-colour("black");
   color: govuk-colour("white");
 
   .covid__page-header-link {
@@ -132,8 +132,6 @@ $c19-landing-page-header-background: #0843a1;
 }
 
 .covid__page-header-title-wrapper {
-  border-bottom: 1px solid $covid-yellow;
-
   @include govuk-media-query($from: tablet) {
     position: relative;
   }
--- a/app/assets/stylesheets/views/_covid.scss
+++ b/app/assets/stylesheets/views/_covid.scss
@@ -66,6 +66,10 @@ $c19-landing-page-header-background: #0843a1;
   @include govuk-media-query($from: tablet) {
     padding: 45px 0 40px;
   }
+
+  .gem-c-govspeak {
+    color: govuk-colour("white");
+  }
 }
 
 .covid__instruction-list {
--- a/app/views/coronavirus_landing_page/components/landing_page/_page_header.html.erb
+++ b/app/views/coronavirus_landing_page/components/landing_page/_page_header.html.erb
@@ -23,41 +23,13 @@
       <div class="govuk-grid-row">
         <div class="govuk-grid-column-two-thirds">
           <%= render "govuk_publishing_components/components/heading", {
-            text: sanitize(header["title"]),
+            text: t("coronavirus_landing_page.header_title"),
             font_size: 24,
             margin_bottom: 6,
             inverse: true
           } %>
 
-          <% if header["description"] %>
-            <%= render "govuk_publishing_components/components/govspeak", {
-            } do %>
-              <%= sanitize(header["description"]) %>
-            <% end %>
-          <% end %>
-
-          <% if header["list"] %>
-            <ul class="covid__instruction-list">
-              <% icons = [
-                "icon-hand-wash",
-                "icon-face-cover",
-                "icon-make-space"
-              ] %>
-              <% header["list"].each_with_index do | item, index | %>
-                <li class="covid__instruction-list-item">
-                  <% if item.is_a?(Hash) %>
-                    <%= render "coronavirus_landing_page/components/landing_page/#{icons[index]}.svg" if icons[index] %>
-                    <span class="covid__instruction-text">
-                      <strong class="covid__instruction"><%= item["instruction"].capitalize %></strong>
-                      <%= item["detail"] %>
-                    </span>
-                  <% else %>
-                    <%= item %>
-                  <% end %>
-                </li>
-              <% end %>
-            </ul>
-          <% end %>
+          <%= render_govspeak(t("coronavirus_landing_page.header_intro")) %>
         </div>
       </div>
 
@@ -95,8 +67,8 @@
             <%= render 'govuk_publishing_components/components/action_link', {
               light_text: true,
               transparent_icon: true,
-              href: header["link"]["href"],
-              text: header["link"]["link_text"],
+              href: t("coronavirus_landing_page.action_link_url"),
+              text: t("coronavirus_landing_page.action_link_text"),
               subtext: header["link"]["subtext"],
               mobile_subtext: true,
               nowrap_text: header["link"]["link_nowrap_text"],
@@ -104,7 +76,7 @@
                 module: "track-click",
                 track_category: "pageElementInteraction",
                 track_action: "Announcements",
-                track_label: header["link"]["href"]
+                track_label: t("coronavirus_landing_page.action_link_url")
               }
             } %>
           <% end %>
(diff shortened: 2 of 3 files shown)


app/views/coronavirus_landing_page/components/landing_page/_page_header.html.erb
```diff
@@ -24,30 +24,40 @@
...
-                  </span>
-                <% else %>
-                  <%= item %>
-                <% end %>
-              </li>
+
+          <% if header["description"] %>
+            <%= render "govuk_publishing_components/components/govspeak", {
+            } do %>
+              <%= sanitize(header["description"]) %>
```
kevindew, line 35: Do you mean to use sanitize here? Is this used somewhere else with HTML?

app/assets/stylesheets/views/_covid.scss
```diff
@@ -66,6 +66,10 @@ $c19-landing-page-header-background: #0843a1;
   @include govuk-media-query($from: tablet) {
     padding: 45px 0 40px;
   }
+
+  .gem-c-govspeak {
+    color: govuk-colour("white");
+  }
```
kevindew, line 72: Eep - is there a container div we could apply this to? it seems a bit of a no-no to override components directly.

app/views/coronavirus_landing_page/components/landing_page/_page_header.html.erb
```diff
@@ -24,30 +24,40 @@
         <div class="govuk-grid-column-two-thirds">
           <%= render "govuk_publishing_components/components/heading", {
             text: sanitize(header["title"]),
-            font_size: 19,
+            font_size: 24,
```
kevindew, line 27: Should this be changing in both if scenarios?

kevindew (commented):
Looks good - mostly concerned the santize isn't what you're aiming for

app/assets/stylesheets/views/_covid.scss
alex-ju: We seem to be doing it [rather](https://github.com/alphagov/collections/pull/2206/files/2323664398cdaf7715b61ae7e29d25d28a136fc4#diff-b599aa1768841fa36cef988b2e58f4bcc5fb93fe4681f31f5f43d8bd81510aa2R47-L49) [often](https://github.com/alphagov/collections/pull/2206/files/2323664398cdaf7715b61ae7e29d25d28a136fc4#diff-b599aa1768841fa36cef988b2e58f4bcc5fb93fe4681f31f5f43d8bd81510aa2L209-L211) – although I've heard voices saying we don't.

app/views/coronavirus_landing_page/components/landing_page/_page_header.html.erb
alex-ju: I think sanitizing everything that comes from a yaml is good practice – trust but verify.

app/views/coronavirus_landing_page/components/landing_page/_page_header.html.erb
alex-ju: Design is happy for this to go in as is, affecting the current design too, but equally happy to render them differently based on the existence of `header["list"]` or `header["description"]`

app/assets/stylesheets/views/_covid.scss
alex-ju: Adding extra containers to apply the colour on won't work as `gem-c-govspeak` defines its own colour and will override it.

app/views/coronavirus_landing_page/components/landing_page/_page_header.html.erb
kevindew: Hmm I think this seems a bit fishy. The string will already have html entities stripped because it's not marked as html_safe.

However when we're using the govspeak component we actually expect the content to be markdown that is converted to HTML whereas this was just be a string of content so there'd be no govspeak elements to style?

app/views/coronavirus_landing_page/components/landing_page/_page_header.html.erb
kevindew: Sure, so long as it's intentional.

app/assets/stylesheets/views/_covid.scss
kevindew: Ah bummer - hmm well I guess there isn't time for a better approach now. I wonder what a better approach would be though.

app/views/coronavirus_landing_page/components/landing_page/_page_header.html.erb
alex-ju: As far as I know, the govspeak component is yielding a block, it doesn't know or care what that is and we can write anything in that yaml file in `coronavirus_landing_page` – hence I felt that sanitizing that might not be a bad idea. Otherwise, someone can add a script tag successfully and we'll enjoy the show.

app/assets/stylesheets/views/_covid.scss
alex-ju: I raised the overriding topic in the #govukfrontenders a few times now – people are not interested – but happy to raise it again in the hope to reach a consensus. You probably know my stance by now: modifier classes should be allowed in components 🙂

app/views/coronavirus_landing_page/components/landing_page/_page_header.html.erb
kevindew: I pinged you on Slack as I'm not sure we're on the same wavelength here but given Slack's problems today I'll try explain in a comment too.

So first concern is the info going into this isn't Govspeak HTML.

For example if `header["description"] = "This is some text to show"`

Then the output will be:

```
<div class="gem-c-govspeak govuk-govspeak " data-module="govspeak">This is some text to show</div>
```

Note that is not any paragraph elements like we'd expect.

---

Less of a concern is the sanitize front. If `header["description"] = "This is some malicious html <script>alert</script>"`

Then the output without calling the sanitize function would be:

```
<div class="gem-c-govspeak govuk-govspeak " data-module="govspeak">
    
              This is some malicious html &lt;script&gt;alert&lt;/script&gt;

</div>
```

because we've never run html_safe on `header["description"]` it'll have all HTML escaped by default.

alex-ju: Not sure where at which point the HTML is escaped from the yml to the output, but if that's the case, happy to drop the sanitize.

alex-ju: Just tested now if I'm not using `sanitize` then the following

```
<% header["description"] = "<p>Some content</p>" %> 
<%= render "govuk_publishing_components/components/govspeak", { } do %>
  <%= sanitize(header["description"]) %>
<% end %>
```

 will result in `<p>Some content</p>` instead of `Some content`

alex-ju: Note my comment in the PR description that by design, we are required to have a few <p> tags in the yaml

alex-ju:
Update to use the app-defined content instead of https://github.com/alphagov/govuk-coronavirus-content/blob/master/content/coronavirus_landing_page.yml to ease the release.

app/views/coronavirus_landing_page/components/landing_page/_page_header.html.erb
alex-ju: Clarified this offline and changed the approach to use locally defined content and `render_govspeak`.

app/views/coronavirus_landing_page/components/landing_page/_page_header.html.erb
```diff
@@ -23,31 +23,16 @@
...
-                    <%= item["detail"] %>
-                  </span>
-                <% else %>
-                  <%= item %>
-                <% end %>
-              </li>
-            <% end %>
-          </ul>
+
+          <%= render "govuk_publishing_components/components/govspeak", {
```
kevindew, line 32: We shouldn't need this as `render_govspeak` will spit out the component as per: https://github.com/alphagov/collections/blob/aa0ea47e99e3da23d631ec1badfab182a6038748/app/helpers/application_helper.rb#L56

alex-ju: Indeed

kevindew (approved):
There's a thing to change but happy for you to merge this in since it doesn't seem to be a noticeable change.

PR #2210, opened 2021-01-04
Use content from `govuk-coronavirus-content`

alex-ju:
To be released after https://github.com/alphagov/govuk-coronavirus-content/pull/539


kevindew (approved):
Nice one

PR #2215, opened 2021-01-05
Replace JPG stay-at-home-social-share with a PNG

kevindew:
Trello: https://trello.com/c/IR6EVlaj/1031-replace-sharing-image-when-coronavirus-is-shared

This is produced from a vector graphic so is sharper and larger but
is a smaller file size.


leenagupte (approved):
👍

DilwoarH (approved):

PR #2226, opened 2021-01-08
Configure /find-coronavirus-local-restrictions redirect in controller

kevindew:
Trello: https://trello.com/c/HzeMkOac/1038-return-a-temporary-cachable-redirect-for-the-postcode-checker-while-its-unavailable

We noticed that this redirect isn't cached and a fresh origin request is
made for every request. In order to add caching and to make this be
testable I moved this concern to the controller. I used the same cache
method as is used in the controller so that this can consider whether
out of date is set or not for a shorter cache time.

As we expect to change this redirect back to the destination page I've
marked this redirect with a temporary redirect (307) status code.

This is ported with the same behaviour as the previous constraint of
basing this condition on Rails.env.production?

---

Before:

```
➜  curl -sI https://www.integration.publishing.service.gov.uk/find-coronavirus-local-restrictions | grep "HTTP\|cache"
HTTP/2 301
cache-control: no-cache
x-served-by: cache-lhr7365-LHR
x-cache: MISS, MISS
x-cache-hits: 0
```

After:

```
➜   curl -sI https://www.integration.publishing.service.gov.uk/find-coronavirus-local-restrictions | grep "HTTP\|cache"
HTTP/2 307
cache-control: max-age=1800, public
x-served-by: cache-lon4269-LON
x-cache: HIT, HIT
x-cache-hits: 1
```


leenagupte (approved):
👍 LGTM!

Thanks for leaving up the branch in integration to make testing easier.

This is a much nicer solution than adding a constraint to the routes file.

PR #1928, opened 2020-09-29
Add event tracking to timeline

DilwoarH:
## What

Add Google Analytics event tracking to the links within timeline section of the landing page.

This should use the following:
Event Category: pageElementInteraction
Event Action: Timeline
Event Label: [Page URL of destination page]

## Why

So we can track engagement with the timeline element of the landing page.

Trello: https://trello.com/c/cwkmPkIN/822-add-event-tracking-to-timeline

:warning: This application is Continuously Deployed: :warning:

- Merged changes are automatically deployed to staging and production.

- Make sure you follow [the guidance for deployments](https://docs.publishing.service.gov.uk/manual/development-pipeline.html#merge-your-own-pull-request) **before** you merge.

- Check your branch is being deployed in the [Release app](https://release.publishing.service.gov.uk/applications/collections), after merging.

PR #2227, opened 2021-01-08
Add event tracking to covid

DilwoarH:
## What

Add event tracking clicks so we know which content is clicked and where on the page. 

We need to add tracking to clicks on

- Accordion content
- Statistics and all guidance sections
- The call out boxes to the testing service and find coronavirus support

More detail here 
https://docs.google.com/document/d/1YvBU48g-tsY_HqaBE4z6BIP9c8W3W4MW5i0dhjAyZy8/edit?usp=sharing

## Why

We can sort of do it using page pathing information but it can get a little bit complicated when the link exists in more than one location on the page

https://trello.com/c/qUfhSJU9/1003-add-event-tracking-to-landing-page-content-clicks

### DIFF
--- a/app/assets/javascripts/modules/track-links.js
+++ b/app/assets/javascripts/modules/track-links.js
@@ -4,9 +4,13 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
 (function (Modules) {
   'use strict'
 
-  Modules.TrackTimelineLinks = function () {
+  Modules.TrackLinks = function () {
     this.start = function (element) {
+      var category = element[0].getAttribute('data-track-category')
+      var action = element[0].getAttribute('data-track-action')
       var links = element[0].getElementsByTagName('a')
+
+      if (!category) return
 
       for (var i = 0; i < links.length; i++) {
         var link = links[i]
@@ -16,7 +20,9 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
             label: e.target.getAttribute('href')
           }
 
-          GOVUK.analytics.trackEvent('pageElementInteraction', 'Timeline', options)
+          if (!action) action = e.target.innerText
+
+          GOVUK.analytics.trackEvent(category, action, options)
         })
       }
     }
--- a/app/views/coronavirus_landing_page/components/shared/_section.html.erb
+++ b/app/views/coronavirus_landing_page/components/shared/_section.html.erb
@@ -15,7 +15,12 @@
             simple: true,
             href: item["url"],
             text: item["label"],
-            margin_bottom: 1
+            margin_bottom: 1,
+            data: {
+              track_category: "pageElementInteraction",
+              track_action: "AccordionFeatureClick",
+              track_label: item["url"],
+            }
           }%>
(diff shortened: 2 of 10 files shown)


app/assets/javascripts/modules/track-links.js
```diff
@@ -4,10 +4,14 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
 (function (Modules) {
   'use strict'
 
-  Modules.TrackTimelineLinks = function () {
+  Modules.TrackLinks = function () {
```
alex-ju, line 7: I'm not sure what the `TrackLinks` does that [`TrackClick`](https://github.com/alphagov/static/blob/master/app/assets/javascripts/modules/track-click.js) can not cover. Would you be able to help me understand the need for this script or what can't be achieved by the `track-click` script? I'm thinking we might have been able to avoid [introducing](https://github.com/alphagov/collections/pull/1928) it in the first place.

alex-ju (commented):
Great to see tracking done using the existing scripts and data attributes only – I even wonder if we actually need the bespoke `track-timeline-links.js` (or `track-links.js` after the refactor)

app/views/coronavirus_landing_page/components/shared/_section.html.erb
```diff
@@ -27,7 +27,12 @@
           <%= link_to(
             item["label"],
             item["url"],
-            class: "govuk-link covid__accordion-link"
+            class: "govuk-link covid__accordion-link",
+            data: {
+              track_category: "pageElementInteraction",
+              track_action: "AccordionClick",
+              track_label: item["url"],
+            }
```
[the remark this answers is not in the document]
DilwoarH: I'll update the commit but the module was included in the parent - therefore wasn't needed here.

app/assets/javascripts/modules/track-links.js
DilwoarH: The reason why I can't use track click is that the behaviour of track-click requires you to add the category and action attributes to the link itself. Because we use markdown, we are unable to add attributes to links and this stops us from tracking it.

TrackLinks allows us to apply a module on the wrapper element and track all links within this scope (not just ones with the attributes). I'm going to look into refactoring track-click to support these scenarios but as it is now, it won't work.

alex-ju: Makes sense. Thanks for explaining this, Dilwoar! Let's get this in then, and we can consider extending TrackClick to cover this, indeed.

alex-ju (approved):

app/assets/javascripts/modules/track-links.js
andysellick: Just came across this while looking into a problem with the tracking on these links. Since this PR was raised the `gem-track-click` script has been extended to include the functionality required - you can specify the attributes on a parent element and the tracking label will be set to the URL of the link: https://github.com/alphagov/govuk_publishing_components/blob/master/docs/analytics/track-click.md

I'll look into whether `TrackLinks` is still needed as part of fixing the tracking issue.

PR #2162, opened 2020-12-16
Remove redundant file

chris-gds:
## What?

As part of [this PR](https://github.com/alphagov/collections/pull/2148/) to change list item noticed file was unused. 

## Why?

Most likely no longer needed.

## Anything else?

[Original PR](https://github.com/alphagov/collections/pull/1652) adding file:

PR #2231, opened 2021-01-13
Update list components

chris-gds:
## What?

Update to use component gem.

## Why?

As part of on-going work auditing apps to use component gem, this improves consistency and a11y / [Ticket](https://trello.com/c/nAMZtgQ7)

## Visuals 
Before (Left) After (Right)

[Live link](https://www.gov.uk/topic/business-tax/alcohol-duties/latest)
This element has been updated to use a [document list](https://components.publishing.service.gov.uk/component-guide/document_list) this Design has changed.

Screenshot 2021-01-12 at 20 23 56

Screenshot 2021-01-12 at 20 23 04

---

Updated to [list component](https://components.publishing.service.gov.uk/component-guide/list).
[Live link](https://www.gov.uk/transition)

Screenshot 2021-01-13 at 17 59 31

Screenshot 2021-01-13 at 17 58 14

---

## Anything else?

[`_related_links`](https://github.com/alphagov/collections/blob/b2f746715a600d5ee38d1e91f0b974921b8f4b7a/app/views/coronavirus_landing_page/components/shared/_related_links.html.erb) was due to be updated however this file _looks_ as though it's redundant, this was removed as part of [this PR](https://github.com/alphagov/collections/pull/2162) - additional references to this file have also been removed.

The list as part of `_buckets.html.erb` has GA tracking included, the list component within the gem does not have functionality in-built to allow tracking on this nature so an additional wrapper tag had been added (less than ideal).  Will flag to consider updating the gem to allow data attributes to be included.

Link list was updated on `_section.html` / covid landing page however has been reverted due to tracking being added at a later date.  This particular implementation of tracking is not support on this component . 

[Enhancement raised on gem to add tracking.](https://github.com/alphagov/govuk_publishing_components/issues/1887)


app/views/latest_changes/index.html.erb
```diff
@@ -25,25 +27,24 @@
...
+      link: {
+      text: "#{document.title}",
+      path: "#{document.base_path}",
+      },
+    metadata: {
+      public_updated_at: document.public_updated_at,
+      },
+      subtext: "#{document.change_note}"
+      }
+    %>
```
injms, line 43: Could this use `map` to avoid setting a variable and then pushing into it?
```suggestion
  <% document_list_contents = subtopic.changed_documents.map do |document|
    {
      link: {
        text: "#{document.title}",
        path: "#{document.base_path}",
      },
      metadata: {
        public_updated_at: document.public_updated_at,
      },
      subtext: "#{document.change_note}"
    }
  %>
```

app/views/transition_landing_page/_buckets.html.erb
```diff
@@ -30,12 +31,15 @@
     <div class="govuk-grid-column-two-thirds">
       <p class='govuk-body'><%= campaign_links[:lead_in] %></p>
 
-      <%= tag.ul class: %w[govuk-list govuk-list--bullet govuk-list--spaced], data: { module: "track-click" } do %>
+      <%= tag.div data: { module: "track-click" } do %>
         <% links.each do |link| %>
```
injms, line 35: Could `map` be used here?

app/views/latest_changes/index.html.erb
chris-gds: Good shout!

app/views/transition_landing_page/_buckets.html.erb
```diff
@@ -15,7 +14,6 @@
       } %>
     </div>
   </div>
-
   <% campaign_links = I18n.t("transition_landing_page.campaign_links") %>
   <%
     links = campaign_links[:links].map do |link|
```
chris-gds, line 19: Ah @injms this is as the link items are being built higher up using `map` so that it has the relevant elements.  Potentially this could be merged 🤔

owenatgov (approved):
Looking good. One little opinion but not a blocker for getting this in.

app/views/transition_landing_page/_buckets.html.erb
chris-gds: Makes sense, done 👍🏼

PR #2244, opened 2021-01-19
Replace govuk prefix meta tags with the govuk meta tag component

owenatgov:
## What
Replace instances of meta tags prefixed with `govuk` in their name eg: `govuk:section` with the [govuk meta tags component](https://components.publishing.service.gov.uk/component-guide/meta_tags).

## Why
This is part of ongoing work to ensure our frontend apps are using components from our components gem as often as possible. Whilst this change makes the code more verbose, it allows us to easily extend our custom metadata ecosystem which is currently used for things like analytics and the govuk browser extension. This change hasn't included meta tags without the `govuk` prefix like opengraph tags because we currently don't cover these in the component.

No visual changes

[Card](https://trello.com/c/AomyVLLP/551-investigate-and-update-collections-to-use-meta-tag-component)

### DIFF
--- a/app/views/browse/index.html.erb
+++ b/app/views/browse/index.html.erb
@@ -1,7 +1,11 @@
 <% content_for :title, "All categories" %>
 <% content_for :page_class, "browse" %>
 <% content_for :meta_tags do %>
-  <meta name='govuk:navigation-page-type' content='Browse Index'>
+  <%= render "govuk_publishing_components/components/meta_tags", {
+    content_item: {
+      navigation_page_type: "Browse Index",
+    }
+  } %>
 <% end %>
 
 <h1 class="govuk-visually-hidden"><%= t("shared.browse.title") %></h1>
--- a/app/views/transition_landing_page/_page_header.html.erb
+++ b/app/views/transition_landing_page/_page_header.html.erb
@@ -11,7 +11,13 @@
   <meta name="twitter:card" content="summary_large_image">
   <meta name="twitter:image" content="<%= image_url("transition-period.png") %>">
   <meta name="description" content="<%= t("transition_landing_page.meta_description") %>">
-  <meta name="govuk:navigation-page-type" content="Taxon Page" />
+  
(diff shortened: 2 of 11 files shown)


app/views/browse/index.html.erb
```diff
@@ -1,7 +1,11 @@
 <% content_for :title, "All categories" %>
 <% content_for :page_class, "browse" %>
 <% content_for :meta_tags do %>
-  <meta name='govuk:navigation-page-type' content='Browse Index'>
+  <%= render "govuk_publishing_components/components/meta_tags", {
+    content_item: {
+      navigation_page_type: "Browse Index",
+    }
+  } %>
```
chris-gds, line 8: Nice, liking this meta gem works like this now.

app/views/transition_landing_page/_page_header.html.erb
```diff
@@ -11,7 +11,13 @@
   <meta name="twitter:card" content="summary_large_image">
   <meta name="twitter:image" content="<%= image_url("transition-period.png") %>">
   <meta name="description" content="<%= t("transition_landing_page.meta_description") %>">
```
chris-gds, line 13: Should we convert these tags to the gem too?

chris-gds (approved):

app/views/transition_landing_page/_page_header.html.erb
owenatgov: I asked about this in govuk frontenders as it would be nice and tidy. Here's aquote from Alex J on the subject:

>my understanding is that the reason for having a simple tag (meta that is) as a component is to ensure consistency and have control over the name attribute which is being used as a contract between the values (provided by the backend) and the script that picks these values up. based on this rationale I wouldn’t find much value in extending/rescoping the component to cover all meta tags (opengraph and such) – unless we have a plan to standardise the open graph tags across gov.uk, in which case we could consider it

With this in mind I think for now it makes sense to me to keep the meta tags component as purely a delivery means for `govuk` prefixed tags instead of trying to make it all encompassing.

PR #2245, opened 2021-01-19
Start converting Minitest specs to RSpec 

hannako:
## What

Start converting Minitest specs to RSpec. See individual commit messages for further details.

## Why

- Collections currently uses Minitest. 
- If we [implement pact testing](https://github.com/alphagov/collections/pull/2230) with gds-api adaptors, we add rspec as a dependency.
- It is a bit hacky and complicated having two test suites running in one repository. Although not unique. See [here](https://github.com/alphagov/smart-answers/pull/5000) and briefly [here](https://github.com/alphagov/frontend/pull/2486)

See issue #2259 for further context and to do list. 

---
Trello https://trello.com/c/DOUWmQkC/1326-convert-collections-component-and-unit-tests-to-rspec


spec/views/components/accordion.html.erb_spec.rb
```diff
@@ -0,0 +1,71 @@
+require "rails_helper"
+include ComponentTestHelper
```
benthorner, line 2: This and the method both need to be inside the `describe` block, so they're isolated to this file. Otherwise, the component tests all fail when run together, as the methods override each other.

spec/views/components/accordion.html.erb_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+    descriptions_accordion[0][:description] = "1st section description"
+
+    render_component(sections: descriptions_accordion)
+
+    assert_select ".govuk-accordion__section-header:nth-child(1) .govuk-accordion__section-summary", text: "1st section description"
+    assert_select ".govuk-accordion__section-header:nth-child(2) .govuk-accordion__section-summary", false
+  end
+
+  xit "renders an accordion with custom section ids" do
+    # FIXME: On the original minitest specs this test is passing but is a false positive.
```
benthorner, line 59: Good spot. This seems to be because the original `assert_select` doesn't support an `id` param, so we were never checking for the ID. The [component](https://github.com/alphagov/govuk_publishing_components/blob/master/app/views/govuk_publishing_components/components/_accordion.html.erb) doesn't seem to support this, so arguably we could just get rid of this test.

spec/views/components/header_notice.html.erb_spec.rb
```diff
@@ -0,0 +1,145 @@
...
+  end
+
+  it "renders a title and logo if title and logo are both passed in" do
+    notice = branded_notice
+    notice[:title] = "this is a title"
+
+    render_component(notice)
+    expect(rendered).to have_selector(".app-c-header-notice__title", text: notice[:title])
+    expect(rendered).to have_selector(".app-c-header-notice__title-logo")
+    expect(rendered).to match Regexp.new(notice[:title_logo])
```
benthorner, line 81: Similarly here.

spec/helpers/organisation_helper_spec.rb
```diff
@@ -1,15 +1,15 @@
-require "test_helper"
+require "rails_helper"
 
 describe OrganisationHelper do
   include OrganisationHelpers
```
benthorner, line 4: I had to manually copy both of these files from `test/support`. That seems reasonable, since we'll need to migrate all the helpers over anyway.

benthorner (commented):
Phew, these are also quite a chore to review 😩. Great work getting this started, as at first it seems like quite a daunting task 🥇 ⭐. I had some issues running it myself, which I've explained in two comments.

The other thing we should do here is modify the Rakefile so that RSpec runs as part of the `default` task. The actual tests look good 👍. If we want to pursue merging this in, then I think we should:

- Post about it in #govuk-developers, so people can raise any concerns.
- Open an issue to explain the overall change and the work left to do.

spec/rails_helper.rb
```diff
@@ -0,0 +1,55 @@
+# This file is copied to spec/ when you run 'rails generate rspec:install'
+require "spec_helper"
```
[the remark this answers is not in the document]
hannako: There is some variety amongst our repos, eg I followed [publishing-api](https://github.com/alphagov/publishing-api/blob/b401c159d760e6a28ff68de0ada5901ed6018216/spec/rails_helper.rb#L11). But I am happy to combine - less searching for needles in hack-stacks!

spec/helpers/email_helper_spec.rb
```diff
@@ -0,0 +1,42 @@
+require "rails_helper"
+
+describe EmailHelper do
+  class DummyClass
```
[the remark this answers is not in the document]
hannako: Yes we can, thanks 👍

spec/views/components/accordion.html.erb_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+    expect(rendered).to have_selector(section2 + " .govuk-accordion__section-content", text: "2nd panel contents")
+  end
+
+  it "renders accordion with description in a header" do
+    descriptions_accordion = simple_accordion
+    descriptions_accordion[0][:description] = "1st section description"
+
+    render_component(sections: descriptions_accordion)
+
+    assert_select ".govuk-accordion__section-header:nth-child(1) .govuk-accordion__section-summary", text: "1st section description"
```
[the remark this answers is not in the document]
hannako: whoops good spot 😅

test/components/header_notice_test.rb
```diff
@@ -58,6 +58,8 @@ def list_with_links
 
   test "renders a logo if title logo is passed in" do
     render_component(branded_notice)
+    ## this is a false positive
```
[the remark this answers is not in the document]
hannako: Yes, I just left it in to help the review process - so you could match up the comments to see the bogus assertion.

spec/views/components/header_notice.html.erb_spec.rb
```diff
@@ -0,0 +1,145 @@
...
+
+  it "renders a branded notice if branding is passed in" do
+    render_component(branded_notice)
+    expect(rendered).to have_selector(".app-c-header-notice.app-c-header-notice__branding--nhs")
+  end
+
+  it "renders a logo if title logo is passed in" do
+    render_component(branded_notice)
+    expect(rendered).to have_selector(".app-c-header-notice__title-logo", count: 1)
+    expect(rendered).to match Regexp.new(branded_notice[:title_logo])
```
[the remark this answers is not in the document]
hannako: thanks 👍

spec/helpers/organisation_helper_spec.rb
hannako: Ah yep sorry. The OrganisationHelpers module is used in specs that aren't yet converted, ie integration specs. Rather than migrate them over, I'll add `Dir[Rails.root.join("test/support/**/*.rb")].sort.each { |f| require f }` in addition to Dir[Rails.root.join("spec/support/**/*.rb")].sort.each { |f| require f }` for the time being, so we don't end up with two copies of the same file that might then diverge.

benthorner: Sounds good - I couldn't decide whether it would be better to copy or reference. Time will tell 🤞.

Gemfile
```diff
@@ -22,9 +22,11 @@ group :test do
   gem "govuk_schemas"
   gem "minitest-spec-rails"
   gem "mocha"
+  gem "parser", require: false
   gem "rails-controller-testing"
   gem "simplecov"
   gem "simplecov-rcov"
+  gem "unparser", require: false
```
benthorner, line 29: We only actually need this one, as it depends on `parser`. These also seem to have leaked into this commit from the one where we add the converter. If they stay here, could we explain why they're being added in the commit message? This would help someone easily identify them for removal in future.

benthorner (commented):
Since I won't be on GOV.UK much longer, I thought it'd be helpful to do another round of review: only a few more minor comments, based on some of the changes made since the first review.

spec/views/components/accordion.html.erb_spec.rb
hannako: I checked in with @alex-ju 
- The custom ID is not being passed through, so we can remove the test.
- It looks like we can replace this custom component with the regular govuk-publishing-component gem instead. He will pick that work up. Once the partial is removed, we will remove this test.

lib/parsers/convert_to_rspec.rb
```diff
@@ -0,0 +1,229 @@
...
+  end
+
+  def go!
+    Dir["#{minitest_folder}/*_test.rb"].each do |test|
+      puts test
+      body = File.read(test)
+      test.gsub!(minitest_folder.to_s, rspec_folder.to_s)
+      new_path = test.gsub!(/_test\.rb$/, "_spec.rb")
+      # custom_fixes_before!(body, new_path)
+      rails_helper!(body, new_path)
```
leenagupte, line 22: As `rails_helper` as been merged with `spec_helper`, I don't think this is needed.

lib/tasks/converter.rake
```diff
@@ -0,0 +1,13 @@
+require_relative "../parsers/convert_to_rspec.rb"
+
+# example usage: bundle exec rake convert:to_rspec\["test/components","spec/views/components"\]
+namespace :convert do
+  desc "task to do some basic conversion of minitest syntax to rspec syntax"
+  task :to_rspec, %i[minitest_folder rspec_folder] => [:environment] do |_, args|
+    message = "minitest_folder and rspec_folder args are required"
```
leenagupte, line 7: I think the example "usage" comment would be really useful here as it's unclear whether or not the whole path to the test file is required.

leenagupte (commented):
This is looking good! I've added a couple of inline comments, but nothing major.

hannako:
Thanks for the comments @leenagupte

alex-ju (approved):
Impressed by the MiniTest to RSpec conversion approach – hopefully we won't be needing it for long.

sihugh (approved):

PR #2262, opened 2021-02-03
Error messages for postcode checker

leenagupte:
Trello: https://trello.com/c/vFwLB8Dz

# What's changed and why?

We have stats on the types of postcode errors that users are putting in to the checker.  We can make specific error messages and summaries for these errors, so users know how to rectify their mistakes. 

Creates 2 new error messages and summaries to replace the ones that already exist. 

## 1. Postcode not found 
Error message and summary: 
The postcode could not be found, enter a full postcode

Show for types of error: 
 Postcode couldn’t be found in Mapit

## 2.  Invalid format and other errors
Error message and summary: 
Enter a full postcode, like LS1 1UR

Show for types of error:
- Postcode invalid format
- Postcode failed validation in Mapit
- Postcode left blank
- Other errors

# Expected changes
## 1. Postcode not found 
screenshot-collections dev gov uk-2021 02 03-12_54_24

## 2.  Invalid format and other errors
screenshot-collections dev gov uk-2021 02 03-12_51_35

### DIFF
--- a/config/locales/en/coronavirus_local_restrictions.yml
+++ b/config/locales/en/coronavirus_local_restrictions.yml
@@ -1,7 +1,7 @@
 en:
   coronavirus_local_restrictions:
     errors:
-      no_postcode:
+      postcode_not_found:
         error_message: There is a problem
         input_error: The postcode could not be found, enter a full postcode
         error_description: The postcode could not be found, enter a full postcode
--- a/config/locales/en/coronavirus_local_restrictions.yml
+++ b/config/locales/en/coronavirus_local_restrictions.yml
@@ -3,12 +3,12 @@ en:
     errors:
       no_postcode:
         error_message: There is a problem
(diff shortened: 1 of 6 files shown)


config/locales/en/coronavirus_local_restrictions.yml
```diff
@@ -3,12 +3,12 @@ en:
     errors:
       no_postcode:
```
kevindew, line 4: I feel like this could benefit from being renamed to `postcode_not_found` to match the condition. The current name makes me think it refers to a blank postcode, which is not what this condition means anymore.

kevindew (commented):
This looks good - and nicer for the user - I've just got a single suggestion.

leenagupte:
@kevindew Thanks for your review. I've added a new commit with your recommended change.

kevindew (approved):
Nice one 👍

PR #2264, opened 2021-02-04
Replace references to Rummager with SearchAPI

hannako:
## What

Update collections to refer to SearchApi by its true name, rather than by its legacy name.

## Why

It is time!

### DIFF
--- a/app/lib/services.rb
+++ b/app/lib/services.rb
@@ -17,12 +17,12 @@ module Services
   def self.cached_search(params, metric_key: "search.request_time")
     Rails.cache.fetch(params, expires_in: 5.minutes) do
       GovukStatsd.time(metric_key) do
-        rummager.search(params).to_h
+        search_api.search(params).to_h
       end
     end
   end
 
-  def self.rummager
-    @rummager ||= GdsApi::Search.new(Plek.new.find("search"))
+  def self.search_api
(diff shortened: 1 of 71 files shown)


app/lib/services.rb
```diff
@@ -17,12 +17,12 @@ def self.content_store
...
       GovukStatsd.time(metric_key) do
-        rummager.search(params).to_h
+        search_api.search(params).to_h
       end
     end
   end
 
-  def self.rummager
-    @rummager ||= GdsApi::Search.new(Plek.new.find("search"))
+  def self.search_api
```
leenagupte, line 25: I don't think this method is needed. I think you can call `GdsApi.search` instead of `Services.search_api`: https://github.com/alphagov/gds-api-adapters/blob/master/lib/gds_api.rb#L159

sihugh (approved):
🇷🇺 🧙

leenagupte (approved):
This is amazing! Thanks for doing all this 🏅 ⭐ 

I left one inline comment, but it's not a blocker.

app/lib/services.rb
hannako: Ah that's cool thank you. I will make that change in a separate PR I think as there are quite a lot of tests that will need new stubs if I make the change here.

PR #2267, opened 2021-02-08
View tests

Rosa-Fox:
[Trello](https://trello.com/c/8Sa7OXXA/1070-add-view-tests-for-the-postcode-checker)

View tests allow us to test what would be rendered depending on different data, without having to run through all of the integration specs, which can be slow and difficult to maintain.

This PR replaces some of the integration tests with view tests.

[This PR is a result of a spike](https://github.com/alphagov/collections/compare/view-tests-local-restrictions-kevin) by myself and @kevindew.


test/views/coronavirus_local_restrictions_test.rb
```diff
@@ -0,0 +1,90 @@
+require "test_helper"
+require "gds_api/test_helpers/mapit"
+
+class CoronavirusLocalRestrictionTest < ActionView::TestCase
```
kevindew, line 4: It'll be good to think about how the name (and filename) of a view test maps into the wider test structure.

With [RSpec's approach to view specs](https://relishapp.com/rspec/rspec-rails/v/2-3/docs/view-specs/view-spec) the tests would be in a file of `spec/views/coronavirus_local_restrictions/results.html.erb_spec.rb` so there is an individual test per view.

I think here we should decide if an individual view has it's own test or whether each view inside a directory has a describe block (as per controllers)

e.g.

```
# tests/views/coronavirus_local_restrictions/results_html_test.rb
class CoronavirusLocalRestrictions::ResultsHtmlTest < ActionView::TestCase
  test "renders postcode match tier 4" 
end
```

or 

```
# tests/views/coronavirus_local_restrictions_test.rb
class CoronavirusLocalRestrictionsTest < ActionView::TestCase
  describe "results.html.erb" do
    test "renders postcode match tier 4" 
  end
end
```

I think the first approach of test per view is probably better as it is less likely to have a namespace collision, can scale easier and produces more short tests compared to one big test. It'd also make it easier to port to RSpec.

test/views/coronavirus_local_restrictions_test.rb
```diff
@@ -0,0 +1,90 @@
...
+                             "start_date" => 1.week.from_now.to_date,
+                             "start_time" => "10:00" }
+                         end
+
+    local_restriction = LocalRestriction.new(gss, {
+      "name" => name,
+      "restrictions" => [current_restriction, future_restriction].compact,
+    })
+    LocalRestriction.stubs(:find).with(gss).returns(local_restriction)
+  end
```
kevindew, line 142: Since this is basically the same as the helper in https://github.com/alphagov/collections/blob/66a5aab214bf800194b5ab6ef2d011dbaaf807b3/test/integration/coronavirus_local_restrictions_test.rb#L442-L475 I'd wonder if this would be better as a shared helper.

test/views/coronavirus_local_restrictions_test.rb
```diff
@@ -0,0 +1,90 @@
...
+
+class CoronavirusLocalRestrictionTest < ActionView::TestCase
+  include GdsApi::TestHelpers::Mapit
+
+  test "renders postcode match tier 4" do
+    render_tier_results(4)
+
+    assert_includes rendered, I18n.t("coronavirus_local_restrictions.results.level_four.heading_pretext")
+    assert_includes rendered, I18n.t("coronavirus_local_restrictions.results.level_four.heading_tier_label")
+    assert_includes rendered, area
```
kevindew: This area variable seems to appear by surprise. Perhaps it could be specified in an argument of the render method. That could be something like `render_view(tier: 4, area: "Tattooine")` and then you'd not need to find the area method.

test/views/coronavirus_local_restrictions_test.rb
```diff
@@ -0,0 +1,90 @@
...
+  end
+
+  def area
+    "Tattooine"
+  end
+
+  def postcode
+    "E1 8QS"
+  end
```
kevindew, line 88: This method doesn't seem to be used outside of `render_tier_results` - I wonder if it can either be just hardcoded in there until something else is needed or be set as an optional argument to the method, for example `render_tier_result(postcode: "E1 8QS")`

test/integration/coronavirus_local_restrictions_test.rb
```diff
@@ -15,30 +15,6 @@ class CoronavirusLocalRestrictionsTest < ActionDispatch::IntegrationTest
...
-    end
-
-    it "displays the tier four restrictions" do
-      given_i_am_on_the_local_restrictions_page
-      then_i_can_see_the_postcode_lookup_form
-      then_i_enter_a_valid_english_postcode_in_tier_four
-      then_i_click_on_find
-      then_i_see_the_results_page_for_level_four
-    end
-
```
kevindew, line 41: Yay 🎉

test/views/coronavirus_local_restrictions_test.rb
```diff
@@ -50,6 +50,35 @@ class CoronavirusLocalRestrictionTest < ActionView::TestCase
...
+
+  test "renders guidance for Northern Ireland" do
+    postcode = "BT48 7PX"
+    country_name = "Northern Ireland"
+
+    render_devolved_nation_guidance(postcode, country_name)
+
+    assert_includes rendered, I18n.t("coronavirus_local_restrictions.results.devolved_nations.northern_ireland.guidance.label")
+  end
```
kevindew, line 80: Minor - this has slipped out of your describe.

test/views/coronavirus_local_restrictions_test.rb
```diff
@@ -0,0 +1,90 @@
+require "test_helper"
+require "gds_api/test_helpers/mapit"
+
+class CoronavirusLocalRestrictionTest < ActionView::TestCase
+  include GdsApi::TestHelpers::Mapit
+
+  test "renders postcode match tier 4" do
```
kevindew: This might need a little more information. It seems a little unclear to me quite what this is testing. is it `test "rendering tier 4 guidance for a postcode in tier 4"` ?

kevindew (commented):
Great work on this Rosa it looks like it's all heading in a good direction 👍

Rosa-Fox:
thanks @kevindew. I'll get rid of the `area` and `postcode` methods, make a shared helper and split the tests in separate files tomorrow morning.

test/views/coronavirus_local_restrictions_test.rb
Rosa-Fox: Thanks! I have updated.

test/views/coronavirus_local_restrictions_test.rb
Rosa-Fox: Thanks! Have updated.

test/views/coronavirus_local_restrictions_test.rb
Rosa-Fox: Thanks... I refactored!

test/views/coronavirus_local_restrictions_test.rb
Rosa-Fox: Thanks, I refactored this.

test/views/coronavirus_local_restrictions_test.rb
Rosa-Fox: thanks, I updated

test/views/coronavirus_local_restrictions/results_html_test.rb
```diff
@@ -0,0 +1,169 @@
...
+
+    test "rendering tier 1 results for a postcode in tier 1" do
+      render_tier_results(current_alert_level: 1)
+
+      assert_includes rendered, I18n.t("coronavirus_local_restrictions.results.level_one.heading_pretext")
+      assert_includes rendered, I18n.t("coronavirus_local_restrictions.results.level_one.heading_tier_label")
+      assert_includes rendered, area
+    end
+
+    test "rendering postcode match if the postcode is correct except for a special character" do
```
kevindew, line 43: I don't think we need this test as there isn't any logic in the view that we're actually exercising with this - the behaviour is already [tested in the model](https://github.com/alphagov/collections/blob/3733d391d81ab1ce23561dfa2bdeb9f024db64d0/test/models/postcode_local_restriction_search_test.rb#L42-L45)

test/views/coronavirus_local_restrictions/results_html_test.rb
```diff
@@ -0,0 +1,169 @@
...
+  include CoronavirusLocalRestrictionsHelpers
+  helper Rails.application.helpers
+
+  describe "current restrictions" do
+    test "rendering tier 4 results for a postcode in tier 4" do
+      render_tier_results(current_alert_level: 4)
+
+      assert_includes rendered, I18n.t("coronavirus_local_restrictions.results.level_four.heading_pretext")
+      assert_includes rendered, I18n.t("coronavirus_local_restrictions.results.level_four.heading_tier_label")
+      assert_includes rendered, area
```
kevindew, line 16: I don't think we need to assert on area here. I find it a bit of a surprising guest in the code as the variable isn't mentioned until this assertion.

test/views/coronavirus_local_restrictions/results_html_test.rb
```diff
@@ -0,0 +1,169 @@
...
+
+      assert_includes rendered, I18n.t("coronavirus_local_restrictions.results.level_two.heading_pretext")
+      assert_includes rendered, I18n.t("coronavirus_local_restrictions.results.level_two.heading_tier_label")
+      assert_includes rendered, area
+    end
+  end
+
+  describe "devolved nations" do
+    test "rendering results for a Welsh postcode" do
+      postcode = "LL11 0BY"
```
kevindew, line 63: does this postcode impact the test at all? I'm wondering if it's superfluous.

test/views/coronavirus_local_restrictions/results_html_test.rb
```diff
@@ -0,0 +1,169 @@
...
+    end
+
+    test "rendering an out of date warning on the tier four page" do
+      render_tier_results_out_of_date_warning(current_alert_level: 4)
+
+      assert_includes rendered, I18n.t("coronavirus_local_restrictions.out_of_date_warning")
+    end
+  end
+
+  def render_tier_results(current_alert_level: nil, future_alert_level: nil)
```
kevindew, line 136: As there is a lot of similarity between these methods I wonder if we could have one method that covers them all?

For example:

```
def render_results_view(local_restriction: {}, out_of_date: false)
  local_restriction = { postcode: "E1 8QS" }.merge(local_restriction)
  @search = PostcodeLocalRestrictionSearch.new(local_restriction[:postcode])
  stub_local_restriction(**local_restriction)
  view.stubs(:out_of_date?).returns(out_of_date)

  render template: "coronavirus_local_restrictions/results"
end
```

And then you'd have to adapt the calls to pass a local_restriction argument with a hash:

```
render_results_view(local_restriction: { current_alert_level: 4 })
render_results_view(local_restriction: { country_name: "Northern Ireland" })
```

test/views/coronavirus_local_restrictions/show_html_test.rb
```diff
@@ -0,0 +1,42 @@
...
+  include GdsApi::TestHelpers::Mapit
+  include GdsApi::TestHelpers::ContentItemHelpers
+  include CoronavirusLocalRestrictionsHelpers
+  helper Rails.application.helpers
+
+  describe "errors" do
+    test "rendering error when invalid postcode is entered" do
+      postcode = "hello"
+
+      stub_local_restriction(postcode: postcode)
```
kevindew, line 16: I don't think this line is necessary.

test/views/coronavirus_local_restrictions/show_html_test.rb
```diff
@@ -0,0 +1,42 @@
...
+
+      stub_mapit_does_not_have_a_postcode(postcode)
+
+      render_show_page(postcode)
+
+      assert_includes rendered, I18n.t("coronavirus_local_restrictions.errors.postcode_not_found.input_error")
+    end
+  end
+
+  def render_show_page(postcode)
```
kevindew, line 34:
```suggestion
  def render_show_view(postcode)
```

test/integration/coronavirus_local_restrictions_test.rb
```diff
@@ -1,9 +1,16 @@
 require "integration_test_helper"
 require "gds_api/test_helpers/mapit"
+require_relative "../support/coronavirus_local_restrictions_helpers"
 
 class CoronavirusLocalRestrictionsTest < ActionDispatch::IntegrationTest
```
kevindew, line 5: Nice one on the tests removed from here. I think you can remove more too and may want to adjust the phrasing on ones we keep to better capture their purpose.

I think it could suffice with (note: comments intended to communicate intent rather than usage in the code):

```
  it "displays restrictions for an area with restrictions" do
    given_i_am_on_the_local_restrictions_page
    # then_i_can_see_the_postcode_lookup_form # I don't think you need this assertion as entering a postcode will implicitly test it
    # and_there_is_metadata # this doesn't seem a feature concern
    # then_i_enter_a_valid_english_postcode_in_tier_one
    when_i_enter_a_english_postcode_for_a_restriction_area
    then_i_click_on_find
    # then_i_see_the_results_page_for_level_one
    then_i_see_the_restriction_information
  end

  it "displays guidance for a devolved nation" do
    given_i_am_on_the_local_restrictions_page
    # then_i_can_see_the_postcode_lookup_form # tested implicitly
    # then_i_enter_a_valid_welsh_postcode
    when_i_enter_a_postcode_for_a_devolved_nation_area
    then_i_click_on_find
    # then_i_see_the_results_for_wales
    then_i_see_the_devolved_nation_guidance
  end

  it "displays no information for an area without tier information" do
    given_i_am_on_the_local_restrictions_page
    # then_i_can_see_the_postcode_lookup_form # tested implicitly
    then_i_enter_a_valid_english_postcode_with_no_tier_information
    then_i_click_on_find
    then_i_see_the_no_information_page
  end

  describe "with javascript" do
    before { Capybara.current_driver = Capybara.javascript_driver }
    after { Capybara.use_default_driver }

    it "normalises the submitted postcode" do
      given_i_am_on_the_local_restrictions_page
      then_i_can_see_the_postcode_lookup_form
      then_i_enter_an_unusually_formatted_postcode
      then_i_click_on_find
      then_i_see_the_results_with_conventional_postcode_formatting
    end
  end
```

And then all the other feature tests here deleted.

kevindew (commented):
This is looking great Rosa 👍  - there seems a few bits to form conventions on still (the whole fun of introducing a new paradigm) but don't think they'll be too hard to fix.

test/views/coronavirus_local_restrictions/no_information_html_test.rb
```diff
@@ -0,0 +1,41 @@
...
+    end
+  end
+
+  describe "no data returned from Mapit" do
+    test "rendering no information found if the postcode is in a valid format, but there is no data" do
+      postcode = "IM1 1AF"
+      mapit_endpoint = Plek.current.find("mapit")
+
+      stub_request(:get, "#{mapit_endpoint}/postcode/" + postcode.tr(" ", "+") + ".json")
+          .to_return(body: { "postcode" => postcode.to_s, "areas" => {} }.to_json, status: 200)
```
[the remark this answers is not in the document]
Rosa-Fox: Hi, 

They are different in that the first tests for when there is no tier information for an area in the local_restrictions.yml file. The 2nd is for when there is no area data returned from Mapit. I reworded the `describe` of the first test to make it clearer.

test/views/coronavirus_local_restrictions/no_information_html_test.rb
```diff
@@ -0,0 +1,41 @@
+require "test_helper"
+require "gds_api/test_helpers/mapit"
+require_relative "../../support/coronavirus_local_restrictions_helpers"
+
+class NoInformationHtmlTest < ActionView::TestCase
+  include GdsApi::TestHelpers::Mapit
+  include CoronavirusLocalRestrictionsHelpers
+  helper Rails.application.helpers
+
+  describe "current restrictions" do
```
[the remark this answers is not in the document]
Rosa-Fox: I reworded it as a result of the comment below:

"They are different in that the first tests for when there is no tier information for an area in the local_restrictions.yml file. The 2nd is for when there is no area data returned from Mapit. I reworded the describe of the first test to make it clearer."

test/views/coronavirus_local_restrictions/no_information_html_test.rb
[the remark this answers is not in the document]
kevindew: Sure - I don't think they're actually testing anything in the view though. There is no conditionals that are based around this so the same code executes in the same for both tests so at a unit level there is nothing to test. 

Perhaps this is something that should be part of the controller test as I assume it is to test both options on this [OR condition](https://github.com/alphagov/collections/blob/516d34e79ca6c496da9652bcec546bc5804da3c2/app/controllers/coronavirus_local_restrictions_controller.rb#L25-L26) ?

test/views/coronavirus_local_restrictions/no_information_html_test.rb
```diff
@@ -0,0 +1,41 @@
+require "test_helper"
+require "gds_api/test_helpers/mapit"
+require_relative "../../support/coronavirus_local_restrictions_helpers"
+
+class NoInformationHtmlTest < ActionView::TestCase
```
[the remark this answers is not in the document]
Rosa-Fox: Thanks! I have updated.

test/views/coronavirus_local_restrictions/no_information_html_test.rb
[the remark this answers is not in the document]
Rosa-Fox: yeah I see what you mean. I have removed the view test for `no_information` and added a controller test for the scenario that the tier information is missing from the local_restrictions.yml file.

test/views/coronavirus_local_restrictions/no_information_html_test.rb
```diff
@@ -0,0 +1,41 @@
...
+  end
+
+  def render_no_information_page
+    @search = PostcodeLocalRestrictionSearch.new("E1 8QS")
+
+    view.stubs(:out_of_date?).returns(false)
+
+    render template: "coronavirus_local_restrictions/no_information"
+  end
+end
```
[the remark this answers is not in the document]
Rosa-Fox: thanks! I have removed the view test for no_information and added a controller test for the scenario that the tier information is missing from the local_restrictions.yml file.

test/views/coronavirus_local_restrictions/results_html_test.rb
```diff
@@ -0,0 +1,169 @@
+require "test_helper"
+require "gds_api/test_helpers/mapit"
+require_relative "../../support/coronavirus_local_restrictions_helpers"
```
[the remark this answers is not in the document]
Rosa-Fox: Thanks! Have removed.

test/views/coronavirus_local_restrictions/results_html_test.rb
Rosa-Fox: ah yeah it doesn't! Thanks, have removed.

test/views/coronavirus_local_restrictions/results_html_test.rb
Rosa-Fox: ah that is way better!! thank you! I have refactored to use your suggestion.

test/views/coronavirus_local_restrictions/show_html_test.rb
Rosa-Fox: consider it gone!

test/views/coronavirus_local_restrictions/show_html_test.rb
Rosa-Fox: thanks have renamed.

test/integration/coronavirus_local_restrictions_test.rb
Rosa-Fox: Thanks, will do this Monday before stand up!

test/support/coronavirus_local_restrictions_helpers.rb
```diff
@@ -0,0 +1,56 @@
+require "test_helper"
+module CoronavirusLocalRestrictionsHelpers
```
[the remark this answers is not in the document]
Rosa-Fox: Thanks, I have required the mapit test helpers from test_helper.

test/views/coronavirus_local_restrictions/results_html_test.rb
```diff
@@ -0,0 +1,117 @@
...
+
+      test "rendering results for a Northern Irish postcode" do
+        render_results_view(local_restriction: { country_name: "Northern Ireland" })
+
+        assert_includes rendered, I18n.t("coronavirus_local_restrictions.results.devolved_nations.northern_ireland.guidance.label")
+      end
+    end
+
+    describe "future restrictions" do
+      area = "Naboo"
```
kevindew, line 58: It feels a little unusual to define a variable like this.

I think the convention for minitest would be to set an ivar in a before block. For example: `before { @area = "Naboo" }`

test/views/coronavirus_local_restrictions/results_html_test.rb
```diff
@@ -0,0 +1,117 @@
...
+
+      test "rendering restrictions changing from level one to level two" do
+        render_results_view(local_restriction: { current_alert_level: 1, future_alert_level: 2, name: area })
```
kevindew, line 61: Nice that this didn't need the helper function to change.

test/views/coronavirus_local_restrictions/show_html_test.rb
```diff
@@ -0,0 +1,38 @@
+require "test_helper"
+require_relative "../../../test/support/coronavirus_helper"
+
+module CoronavirusLocalRestrictions
+  class ShowHtmlTest < ActionView::TestCase
+    include CoronavirusLocalRestrictionsHelpers
+    helper Rails.application.helpers
```
kevindew, line 7: I couldn't work out how this could happen automatically 😢 . It looks like we'll need this on any view tests that use helpers outside of the default ones that Rails has. 

It's annoying as auto loading of all helpers works fine in RSpec.

test/integration/coronavirus_local_restrictions_test.rb
```diff
@@ -1,207 +1,47 @@
...
   include GdsApi::TestHelpers::Mapit
   include GdsApi::TestHelpers::ContentStore
```
kevindew, line 7: I think you can remove both of these too.

test/integration/coronavirus_local_restrictions_test.rb
```diff
@@ -362,134 +92,17 @@ def then_i_see_the_results_with_conventional_postcode_formatting
...
-        "gss" => gss,
-        "name" => name,
-        "type" => "LBO",
-        "country_name" => country_name,
-      },
-    ]
-    stub_mapit_has_a_postcode_and_areas(postcode, [], areas)
-
-    LocalRestriction.stubs(:find).with(gss).returns(nil)
-  end
```
kevindew, line 494: This is great - it makes the PR a net code reduction of 200 lines and makes this file a much easier to digest ~100 lines from ~500.

kevindew (commented):
Looks good to me, just a couple of small thing to fix.

Perhaps the commits could use a squash too - there's a lot of them for just a few files changed.

test/integration/coronavirus_local_restrictions_test.rb
Rosa-Fox: thanks, have removed.

test/views/coronavirus_local_restrictions/results_html_test.rb
Rosa-Fox: thanks, have updated.

kevindew (approved):
Great stuff 👍

PR #2268, opened 2021-02-08
Include layout header component styles for /transition

danacotoran:
We have made some adjustments to use the layout header component in `static` where the account navigation should be present. This would be on the `transition` page as well as Brexit checker pages.
For more context: [PR to replace hard coded account nav with layout header](https://github.com/alphagov/static/pull/2408)

This is fine, however some of the layout styles are overriding the layout header component styles in static, resulting in extra left/right padding around the layout header content like so: 

Screenshot 2021-02-08 at 15 17 34

Importing the `layout_header` styles into `collections` seems to fix the issue:

Screenshot 2021-02-08 at 15 17 09

https://trello.com/c/9Kjy1hjb/575-use-header-component-in-static

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -26,6 +26,7 @@ $covid-yellow: #fff500;
 @import 'govuk_publishing_components/components/inset-text';
 @import 'govuk_publishing_components/components/inverse-header';
 @import 'govuk_publishing_components/components/label';
+@import 'govuk_publishing_components/components/layout-header';
 @import 'govuk_publishing_components/components/lead-paragraph';
 @import 'govuk_publishing_components/components/metadata';
 @import 'govuk_publishing_components/components/notice';


andysellick (approved):
Nice fix. Do you also need to add the print stylesheet and JS file for the layout header component?

andysellick (commented):
@danacotoran I've suddenly had a thought - the layout header component isn't used in this application, is it? Shouldn't the styles for it go into `static` instead?

danacotoran:
@andysellick it's a weird one because the styles are already in static [here](https://github.com/alphagov/static/blob/355904b5574cc4f3cf72bc7b1b2cbfc14c0d96c7/app/assets/stylesheets/header-footer-only.scss#L11). But the layout styles take precedence which is why there is extra padding on the right/left. 
Adding it here seems to override that.

danacotoran:
@andysellick I've opened https://github.com/alphagov/govuk_publishing_components/pull/1924 with a fix for this problem as well as a few other changes. Therefore, will be closing this PR.

PR #2280, opened 2021-02-12
Only show social media links with the same locale as the current page

MuriloDalRi:
Requires [this PR](https://github.com/alphagov/whitehall/pull/5981) to be merged first
[Trello](https://trello.com/c/RRvGDDmD/2243-5-make-whitehall-social-media-links-translatable)

### DIFF
--- a/app/views/organisations/_what_we_do.html.erb
+++ b/app/views/organisations/_what_we_do.html.erb
@@ -27,7 +27,7 @@
       <% end %>
     </div>
 
-    <% if @what_we_do.has_share_links? %>
+    <% if @what_we_do.has_share_links?(@_request).present? %>
       <div class="govuk-grid-column-one-third">
         <%= render "govuk_publishing_components/components/heading", {
           text: t('organisations.follow_us'),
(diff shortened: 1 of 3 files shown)


app/helpers/organisation_helper.rb
```diff
@@ -42,4 +42,11 @@ def search_results_to_documents(search_results, organisation, include_metadata:
       end,
     }
   end
+
+  def t_social_media_accounts(social, req)
```
brucebolt, line 46: Would this be better in the `translation_helper`?

app/views/organisations/_what_we_do.html.erb
```diff
@@ -27,7 +27,9 @@
       <% end %>
     </div>
 
-    <% if @what_we_do.has_share_links? %>
+    <% social = t_social_media_accounts(@what_we_do.social_media_links, @_request) %>
```
brucebolt, line 30: It feels like this assignment probably doesn't belong in the view.  Would it be possible to remove this line from the view and do the translations in the presenter (a bit like the Contacts Presenter), so the `share_links` method returns the translated links only?

brucebolt (commented):
Thanks for making these changes @MuriloDalRi.  I've added some comments, and I think we could probably move all the translation logic into the presenter.

I'm thinking something like:
```
org.social_media_links.select { |link| link["locale"] == I18n.locale }.each do |link|
...
```

That would remove the need for a helper method and the callback type of logic in the view.

Also, would it be possible to add a test for this?

PR #2281, opened 2021-02-12
Run pact tests with gds-api-adapters as part of deployment pipeline

hannako:
## What

Follow up PR to https://github.com/alphagov/collections/pull/2230

Verify via the deployment pipeline that each new build of collections honours its contracts with gds-api-adaptors.

See commit messages for further details. 

- GDS API Adaptors already has a [rake](https://github.com/alphagov/gds-api-adapters/blob/f838456c65c6ef40b6e85692145351405790c8ec/Rakefile#L26) task to publish its pactfiles to the broker as part of its deployment pipeline by [jenkins](https://github.com/alphagov/gds-api-adapters/blob/f838456c65c6ef40b6e85692145351405790c8ec/Jenkinsfile#L29).
- So the pactfile defining contracts between collections and gds-api-adapters is already [on the broker.](https://pact-broker.cloudapps.digital/pacts/provider/Collections%20Organisation%20API/consumer/GDS%20API%20Adapters/versions/master).
- This means that this PR is safe to be merged. 

- [trello](https://trello.com/c/Qpd7jgOq/628-add-contract-testing-to-the-collections-app)

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -9,7 +9,7 @@ node {
     brakeman: true,
     extraParameters: [
       stringParam(
-        name: "GDS_API_PACT_BRANCH",
+        name: "GDS_API_ADAPTERS_PACT_BRANCH",
         defaultValue: "master",
         description: "The branch of gds-api-adapters pact tests to run against"
       ),
@@ -17,9 +17,9 @@ node {
     beforeTest: {
       govuk.setEnvar("PACT_BROKER_BASE_URL", "https://pact-broker.cloudapps.digital")
     },
-    afterTest : {
+    afterTest: {
       stage("Verify pact with gds-api-adapters") {
-        govuk.runRakeTask("pact:verify:branch[${env.GDS_API_PACT_BRANCH}]")
+        govuk.runRakeTask("pact:verify:branch[${env.GDS_API_ADAPTERS_PACT_BRANCH}]")
       }
     }
   )
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -7,5 +7,20 @@ node {
   govuk.buildProject(
     publishingE2ETests: true,
     brakeman: true,
+    extraParameters: [
+      stringParam(
+        name: "GDS_API_PACT_BRANCH",
+        defaultValue: "master",
+        description: "The branch of gds-api-adapters pact tests to run against"
+      ),
+    ],
+    beforeTest: {
+      govuk.setEnvar("PACT_BROKER_BASE_URL", "https://pact-broker.cloudapps.digital")
+    },
+    afterTest : {
+      stage("Verify pact with gds-api-adapters") {
+        govuk.runRakeTask("pact:verify:branch[${env.GDS_API_PACT_BRANCH}]")
+      }
+    }
   )
 }
--- a/lib/tasks/pact.rake
+++ b/lib/tasks/pact.rake
@@ -1,5 +1,17 @@
 return if Rails.env.production?
 
 require "pact/tasks"
-require "pact_broker/client/tasks"
 require "pact/tasks/task_helper"
+
+desc "Verify that Collections honours all contracts in the pactfile created by a given branch of gds-api-adapters"
+task "pact:verify:branch", [:branch_name] => :environment do |t, args|
+  abort "Please provide a branch name. eg rake #{t.name}[my_feature_branch]" unless args[:branch_name]
+
+  pact_version = args[:branch_name] == "master" ? args[:branch_name] : "branch-#{args[:branch_name]}"
+
+  ClimateControl.modify GDS_API_PACT_VERSION: pact_version do
+    Pact::TaskHelper.handle_verification_failure do
+      Pact::TaskHelper.execute_pact_verify
+    end
+  end
+end
--- a/lib/tasks/pact.rake
+++ b/lib/tasks/pact.rake
@@ -9,7 +9,7 @@ task "pact:verify:branch", [:branch_name] => :environment do |t, args|
 
   pact_version = args[:branch_name] == "master" ? args[:branch_name] : "branch-#{args[:branch_name]}"
 
-  ClimateControl.modify GDS_API_PACT_VERSION: pact_version do
+  ClimateControl.modify GDS_API_ADAPTERS_PACT_VERSION: pact_version do
     Pact::TaskHelper.handle_verification_failure do
       Pact::TaskHelper.execute_pact_verify
     end
--- a/Gemfile
+++ b/Gemfile
@@ -32,6 +32,7 @@ end
 group :development, :test do
   gem "better_errors"
   gem "binding_of_caller"
+  gem "climate_control"
   gem "govuk_test"
   gem "jasmine"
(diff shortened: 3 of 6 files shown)


Jenkinsfile
```diff
@@ -7,5 +7,20 @@ node {
   govuk.buildProject(
     publishingE2ETests: true,
     brakeman: true,
+    extraParameters: [
+      stringParam(
+        name: "GDS_API_PACT_BRANCH",
```
thomasleese, line 12:
```suggestion
        name: "GDS_API_ADAPTERS_PACT_BRANCH",
```

Jenkinsfile
```diff
@@ -7,5 +7,20 @@ node {
...
+        defaultValue: "master",
+        description: "The branch of gds-api-adapters pact tests to run against"
+      ),
+    ],
+    beforeTest: {
+      govuk.setEnvar("PACT_BROKER_BASE_URL", "https://pact-broker.cloudapps.digital")
+    },
+    afterTest : {
```
thomasleese, line 20:
```suggestion
    afterTest: {
```

Jenkinsfile
```diff
@@ -7,5 +7,20 @@ node {
   govuk.buildProject(
     publishingE2ETests: true,
     brakeman: true,
+    extraParameters: [
+      stringParam(
+        name: "GDS_API_PACT_BRANCH",
+        defaultValue: "master",
```
thomasleese, line 13: It's unfortunate we need to hard code this! I'm going to make a note of it: https://trello.com/c/zcuPB4dq/212-update-master-references-in-pact-tests

lib/tasks/pact.rake
```diff
@@ -1,5 +1,17 @@
...
-require "pact_broker/client/tasks"
 require "pact/tasks/task_helper"
+
+desc "Verify that Collections honours all contracts in the pactfile created by a given branch of gds-api-adapters"
+task "pact:verify:branch", [:branch_name] => :environment do |t, args|
+  abort "Please provide a branch name. eg rake #{t.name}[my_feature_branch]" unless args[:branch_name]
+
+  pact_version = args[:branch_name] == "master" ? args[:branch_name] : "branch-#{args[:branch_name]}"
+
+  ClimateControl.modify GDS_API_PACT_VERSION: pact_version do
```
thomasleese, line 12:
```suggestion
  ClimateControl.modify GDS_API_ADAPTERS_PACT_VERSION: pact_version do
```

lib/tasks/pact.rake
```diff
@@ -1,5 +1,17 @@
 return if Rails.env.production?
 
 require "pact/tasks"
-require "pact_broker/client/tasks"
 require "pact/tasks/task_helper"
+
+desc "Verify that Collections honours all contracts in the pactfile created by a given branch of gds-api-adapters"
+task "pact:verify:branch", [:branch_name] => :environment do |t, args|
```
thomasleese, line 7: I wonder if it's time to extract this code out into a common location, because I guess it should always be the same as: https://github.com/alphagov/publishing-api/blob/master/lib/tasks/pact.rake

Perhaps [govuk_test](https://github.com/alphagov/govuk_test) would be a good place for it.

lib/tasks/pact.rake
```diff
@@ -1,5 +1,17 @@
...
+  abort "Please provide a branch name. eg rake #{t.name}[my_feature_branch]" unless args[:branch_name]
+
+  pact_version = args[:branch_name] == "master" ? args[:branch_name] : "branch-#{args[:branch_name]}"
+
+  ClimateControl.modify GDS_API_PACT_VERSION: pact_version do
+    Pact::TaskHelper.handle_verification_failure do
+      Pact::TaskHelper.execute_pact_verify
+    end
+  end
+end
```
thomasleese, line 17: I notice in the Publishing API version of this file we've got a bit at the bottom with credentials, do we need that here?

https://github.com/alphagov/publishing-api/blob/a0c59b9c7b8ec48fa4256274abed626567aea87a/lib/tasks/pact.rake#L20-L30

hannako: I believe the credentials and the definition of the publication task are needed to publish the pactfile to the broker, which is the role of the service consumer. 

In the case of publishing api as it is both a [consumer](https://github.com/alphagov/content-store/blob/c665fa0fb3e513c42187a1f8dd4ff4a5f264fe28/spec/service_consumers/pact_helper.rb#L19) and a [provider](https://github.com/alphagov/publishing-api/blob/a0c59b9c7b8ec48fa4256274abed626567aea87a/spec/service_consumers/pact_helper.rb#L19) it defines the `pact:verify:branch` in its role as a provider,  and it calls `PactBroker::Client::PublicationTask` to publish the contracts it has defined with the content-store.

The publication task for the pact between collections and gds-api-adapters is defined in the [latter](https://github.com/alphagov/gds-api-adapters/blob/3f14140bf8e7170308baf125a161921546d88bef/Rakefile#L19toL29) so it isn't needed here. That's my understanding, I may have got the wrong end of the stick!

hannako (commented):
I kept [this](https://github.com/alphagov/collections/pull/2281#discussion_r576018574) the same as the variable [defined](https://github.com/alphagov/publishing-api/blob/8b9b753866a013d012093f93a51c0809e2a74f96/lib/tasks/pact.rake#L13) in publishing api. But I don't think there's any benefit of them being the same. The variable is only use to fetch the correct branch for the verification task. So will update, thanks 👍

lib/tasks/pact.rake
thomasleese: Yeah I think that's fair!

lib/tasks/pact.rake
thomasleese: Ah yes, I think that makes sense. I guess as we roll this out to more apps we may find that we want a common file which includes both, even if it's unused for some apps, but I think this if fine for now.

thomasleese (approved):

lib/tasks/pact.rake
```diff
@@ -1,5 +1,17 @@
...
-require "pact_broker/client/tasks"
 require "pact/tasks/task_helper"
+
+desc "Verify that Collections honours all contracts in the pactfile created by a given branch of gds-api-adapters"
+task "pact:verify:branch", [:branch_name] => :environment do |t, args|
+  abort "Please provide a branch name. eg rake #{t.name}[my_feature_branch]" unless args[:branch_name]
+
+  pact_version = args[:branch_name] == "master" ? args[:branch_name] : "branch-#{args[:branch_name]}"
+
+  ClimateControl.modify GDS_API_ADAPTERS_PACT_VERSION: pact_version do
```
ChrisBAshton, line 12: Is this the right ENV to use? It looks like you're defining `GDS_API_ADAPTERS_PACT_BRANCH` but not `GDS_API_ADAPTERS_PACT_VERSION`, so I'd imagine this doesn't work unless there's some Jenkins magic happening under the hood 🤔

hannako: https://github.com/alphagov/collections/pull/2297

PR #2283, opened 2021-02-15
Fix whitespace problem for some searches on organisations page

### LINKED ISSUE #1905: Government organisations page search has a big gap
On https://www.gov.uk/government/organisations

If you search for something that happens to be at the bottom of the page, you might be forgiven for thinking your search hasn't found anything, because you might get this result and your browser window might not be tall enough:

Screenshot 2020-09-14 at 08 31 25

The reason for this is that each of the wrappers to each organisation type has a margin bottom of 60px, and the search only hides the individual items within each wrapper, not the wrapper itself:

Screenshot 2020-09-14 at 08 31 40

Surprisingly on mobile, this isn't a problem, even though the margin bottom is still there, each wrapper collapses over the next one.

andysellick:
## What

Amend the JS for the organisations search so that if a department (i.e. a group of organisations, contained in a parent div) contains no matching results, we completely hide that wrapper, instead of just hiding the list items for each non-matching organisation and hiding the result count.

## Why

This change is because previously if you searched for something that occurred at the bottom of the list (e.g. 'the scottish government') we only hid the non-matching contents of the departments, not the element that contained them. This element has bottom margin, so the result was a load of effectively empty divs with bottom margin, creating a gap between the top of the page and the visible result.

Fixes #1905.

## Visual changes

Before | After
------ | ------
Screenshot 2021-02-15 at 13 50 09 | Screenshot 2021-02-15 at 13 50 15

### DIFF
--- a/app/assets/javascripts/organisation-list-filter.js
+++ b/app/assets/javascripts/organisation-list-filter.js
@@ -86,13 +86,14 @@
       var totalMatchingOrgs = 0
 
       listsToFilter.each(function () {
-        var matchingOrgCount = $(this).find('[data-filter="item"]:visible').length
         var departmentSection = $(this).closest('[data-filter="block"]')
-        var departmentCountWrapper = departmentSection.find('[data-filter="count"]')
+        departmentSection.removeClass('js-hidden')
+
+        var matchingOrgCount = $(this).find('[data-filter="item"]:visible').length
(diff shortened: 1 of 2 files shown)


injms (approved):
You beat me to this - I was musing about this layout problem after spotting it when I had to fix the in page search. Super!

Issue #2259, opened 2021-02-02
Convert tests from Minitest to RSpec


hannako:
## Issue
- We are currently implementing Pact tests between Collections and gds-api-adapters in order to [meet the criteria](https://github.com/alphagov/govuk-rfcs/blob/master/rfc-128-continuous-deployment.md#check-api-contract-tests-pass-ie-adapters-work-with-their-apis) for CD of collections. 
- The pact gem brings in RSpec [as a dependency](https://github.com/alphagov/collections/blob/e3bc50f10b62357499053cfb4e214c63207f1c34/Gemfile.lock#L246). This will mean Collections runs two test suites, Minitest and RSpec. 
- This is not necessarily bad, eg Smart Answers also has both suites. However in the case of collections it adds a little bit of complexity. Collections uses cucumber for its feature tests, and they fail without [patching](https://github.com/alphagov/collections/pull/2245/commits/7e046459ea7abfb624495234f54de7a088a69e40) the Minitest assertions methods.
- RSpec is the GDS testing frame work [of choice](https://gds-way.cloudapps.digital/manuals/programming-languages/ruby.html#conventional-tooling), so it is probably worth taking the plunge and converting to Minitest. 

### Regular tests

There is a [rake task](https://github.com/alphagov/collections/blob/e5b305b7e7b3eae0020eed90987ea0abc83bb33d/lib/tasks/converter.rake) that can be used to do some of the syntax conversion.
```
bundle exec rake convert:to_rspec\["test/components","spec/views/components"\]
```
#### Convert the following
- [x] test/components
- [x] test/controllers
- [x] test/integration
- [x] test/lib
- [x] test/models
- [x] test/models/schemas
- [x] test/models/topic
- [x] test/presenters
- [x] test/presenters/organisations
- [x] test/presenters/supergroups
- [x] test/services
- [x] test/services/search
- [x] test/services/supergroup_sections
- [x] test/unit 

#### Once all tests requiring a specific helper have been ported over to RSpec, move the helpers from `test/support` ==> `spec/support`
- [x] analytics_helper
- [x] coronavirus_helper
- [x] coronavirus_landing_page_steps
- [x] custom_assertions
- [x] dit_landing_page_helper
- [x] gds_api_rummager_test_helpers
- [x] helpers
- [x] organisation_feed_helpers
- [x] organisation_helper
- [x] organisations_helpers
- [x] rummager_helpers
- [x] services_and_information_helpers
- [x] supergroups_helpers
[...]

PR #2284, opened 2021-02-16
Convert model tests

hannako:
## What

- As part of issue https://github.com/alphagov/collections/issues/2259 to convert Collections test suite from Minitest to RSpec 

- See commit messages for details.

---
--
[Trello](https://trello.com/c/m7wfWUqc/1354-convert-model-tests-to-rspec)


spec/spec_helper.rb
```diff
@@ -2,6 +2,8 @@
 require File.expand_path("../config/environment", __dir__)
 abort("The Rails environment is running in production mode!") if Rails.env.production?
 require "rspec/rails"
+require "gds_api/test_helpers/search"
```
sihugh, line 5: fair point, search is used a lot in Collections

spec/models/list_set_spec.rb
```diff
@@ -36,50 +37,51 @@
...
+      expect(list_set.map(&:title)).to eq(["Paying HMRC", "Annual PAYE and payroll tasks"])
     end
 
     it "provides the title and base_path for group items" do
-      groups = @list_set.to_a
+      groups = list_set.to_a
 
-      assert_equal "Employee tax codes", groups[1].contents.to_a[2].title
-      assert_equal "/pay-psa", groups[0].contents.to_a[1].base_path
+      expect(groups.[](1).contents.to_a.[](2).title).to eq("Employee tax codes")
```
sihugh, line 49: the array syntax here looks a little weird - it's converted `groups[1]` to `groups.[](1)` which looks like it's doing something clever when it isn't. (It's in a number of places)

spec/spec_helper.rb
```diff
@@ -15,6 +15,7 @@
 # See http://rubydoc.info/gems/rspec-core/RSpec/Core/Configuration
 RSpec.configure do |config|
   config.include GdsApi::TestHelpers::Search
+  config.include ActiveSupport::Testing::TimeHelpers
```
sihugh, line 18: 👍

spec/models/mainstream_browse_page_spec.rb
```diff
@@ -12,97 +10,99 @@
...
-      assert_equal "Information about eligibility, claiming and when Child Benefit stops", @page.description
+      expect(page.description).to eq("Information about eligibility, claiming and when Child Benefit stops")
     end
 
     describe "#second_level_pages_curated?" do
       it "is true when second_level_ordering == curated" do
-        @api_data["details"]["second_level_ordering"] = "curated"
-        assert @page.second_level_pages_curated?
+        api_data["details"]["second_level_ordering"] = "curated"
+        expect(page).to be_second_level_pages_curated
```
sihugh, line 37: `expect(page).to be_second_level_pages_curated` works, but I wonder if it's clearer to say:

`expect(page.second_level_pages_curated?).to be(true)`, not least because we use this pattern to assert false just below

sihugh (commented):
Initial comments... Looks generally good, there's just some slightly clunky bits of the conversion that we might want to address. 

I've only got halfway through so far, but thought this might be a useful start

hannako:
I will squash these commits before merging

sihugh (approved):

PR #2288, opened 2021-02-18
Switch from trackClick to gemTrackClick script

alex-ju:
### What

Switch from using `trackClick` script to `gemTrackClick` script. They are virtually the same, the latter being [copied over from `static` to `govuk_publishing_components`](https://github.com/alphagov/govuk_publishing_components/pull/1751).

### Why

Part of [a bigger piece of work](https://github.com/alphagov/static/pull/2405) to reduce the size of assets served to users and keep our codebase to a minimum.


andysellick (approved):
Approved, but I think you might have missed a few instances in the yml files (which I assume get rendered), specifically

- https://github.com/alphagov/collections/blob/master/config/locales/cy/transition_landing_page.yml#L16
- https://github.com/alphagov/collections/blob/master/config/locales/en/transition_landing_page.yml#L16
- https://github.com/alphagov/collections/blob/master/config/locales/en/transition_landing_page.yml#L47

andysellick (approved):
🚀 🚀

PR #2292, opened 2021-02-24
Add support for the new session header & dev cookie

barrucadu:
The user is treated as logged in if:

- The `GOVUK-Account-Session` request header is set
- The `govuk_account_session` cookie is present and `Rails.env.development?`
- The user is in the `LoggedIn` A/B bucket

The A/B approach will be removed when we fully transition to the new
cookie.

We don't validate that the header or cookie actually contains correct
session data.  This is fine because if an attacker supplies a fake
value, all they'll do is change the "sign in" link in the header to a
"sign out" link.

---

[Trello card](https://trello.com/c/ZhjlysBW/642-preferentially-use-the-new-cookie-on-the-transition-landing-page)

### DIFF
--- a/app/controllers/concerns/account_ab_testable.rb
+++ b/app/controllers/concerns/account_ab_testable.rb
@@ -1,8 +1,14 @@
+# frozen_string_literal: true
+
 module AccountAbTestable
   extend ActiveSupport::Concern
 
   ACCOUNT_AB_CUSTOM_DIMENSION = 42
-  ACCOUNT_AB_TEST_NAME = "AccountExperiment".freeze
+  ACCOUNT_AB_TEST_NAME = "AccountExperiment"
+
+  ACCOUNT_SESSION_REQUEST_HEADER_NAME = "HTTP_GOVUK_ACCOUNT_SESSION"
+  ACCOUNT_SESSION_RESPONSE_HEADER_NAME = "GOVUK-Account-Session"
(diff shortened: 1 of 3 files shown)


app/controllers/concerns/account_ab_testable.rb
```diff
@@ -1,6 +1,19 @@
...
 module AccountAbTestable
+  extend ActiveSupport::Concern
+
   ACCOUNT_AB_CUSTOM_DIMENSION = 42
-  ACCOUNT_AB_TEST_NAME = "AccountExperiment".freeze
+  ACCOUNT_AB_TEST_NAME = "AccountExperiment"
+
+  ACCOUNT_SESSION_REQUEST_HEADER_NAME = "HTTP_GOVUK_ACCOUNT_SESSION"
+  ACCOUNT_SESSION_RESPONSE_HEADER_NAME = "GOVUK-Account-Session"
+  ACCOUNT_SESSION_DEV_COOKIE_NAME = "govuk_account_session"
```
AlanGabbianelli, line 11: Why aren't we freezing these constants?

barrucadu: I added `# frozen_string_literal: true` to the top of the file, which makes it automatic

AlanGabbianelli: Ah, I had missed that!

AlanGabbianelli (approved):

PR #2294, opened 2021-02-25
Increase test coverage of Collections

hannako:
## What
Increase our SimpleCov test coverage score.

## Why
Collections currently has a test coverage of 89.68% which is [below the 95% threshold](https://github.com/alphagov/govuk-rfcs/blob/main/rfc-128-continuous-deployment.md#safety-checks) for CD to be switched back on for this repo. 

Screenshot 2021-02-25 at 21 00 29

## How
This low coverage score was caused by the [addition](https://github.com/alphagov/collections/pull/2245/commits/c303341c7ea3afd121d1428f397824a6e9eb325e) of a temporary parser (lib/parsers/convert_to_rspec.rb) to help convert tests to RSpec from Minitest. 

- Exclude the file
- Add a test for the roles controller as that had no tests at all.

Screenshot 2021-02-25 at 21 08 25

[trello](https://trello.com/c/cKyD8kke/1368-increase-test-coverage-for-collections)


test/controllers/roles_controller_test.rb
```diff
@@ -0,0 +1,29 @@
+require "test_helper"
+
+describe RolesController do
+  include SearchApiHelpers
+
+  def ministerial_role_json
+    json = File.read("test/fixtures/content_store/ministerial_role.json")
```
sihugh, line 7: Do you think this might be better as a [content-schema example](https://github.com/alphagov/govuk-content-schemas/tree/master/examples)? (I appreciate the prior art in this repo...)

test/controllers/roles_controller_test.rb
```diff
@@ -0,0 +1,29 @@
...
+    JSON.parse(json)
+  end
+
+  describe "GET show" do
+    before do
+      stub_content_store_has_item("/government/ministers/prime-minister", ministerial_role_json)
+      stub_content_store_does_not_have_item("/government/ministers/she-ra")
```
sihugh, line 14: And why not!

sihugh (approved):
Looks good barring the decision regarding using an example from the schema repo.

test/controllers/roles_controller_test.rb
hannako: Yep good idea. So I would need to make an example of [this](https://github.com/alphagov/govuk-content-schemas/blob/25cad7a436763199272507b8ac8e4d80da03ac1b/formats/role.jsonnet)?

hannako: https://github.com/alphagov/govuk-content-schemas/pull/1049

sihugh (approved):
🎢

PR #2306, opened 2021-03-08
Add placeholder locale files for supported languages

brucebolt:
We support publishing in a total of 65 languages, but do not have translations available for all of them.

Adding placeholder locale files and required keys here so it is easier to identify which keys will need translating in later work.

Trello card: https://trello.com/c/XbwUsOQr

### DIFF
--- a/config/locales/no/shared.yml
+++ b/config/locales/no/shared.yml
@@ -1,3 +1,3 @@
 "no":
   shared:
-    language_name:
+    language_name: Norsk
--- a/config/locales/no/shared.yml
+++ b/config/locales/no/shared.yml
@@ -1,0 +1,3 @@
+"no":
+  shared:
+    language_name:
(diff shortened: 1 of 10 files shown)


config/locales/no/shared.yml
```diff
@@ -0,0 +1,3 @@
+"no":
```
1pretz1, line 1: Is there a reason this key is in quotes?

1pretz1 (approved):
Looks good Bruce! 👍 Approving as the comment is minor

config/locales/no/shared.yml
brucebolt: `no` is [treated as a boolean rather than a string](https://juliankniephoff.wordpress.com/2012/09/01/reserved-words-in-yaml-and-translating-booleans-in-rails/), so we get an error when the tests runs

1pretz1: Interesting, thanks for explaining!

PR #2315, opened 2021-03-17
Show statistics on Coronavirus landing page

kevindew:
Trello: https://trello.com/c/WVZPCn3t/203-spike-explore-embedding-covid-statistics-into-coronavirus-page

I've marked this PR as draft and do-not-merge as there are a number of steps that are needed before this can be merged:

- [x] Receive approval that statistics can be displayed in this form
- [x] Move the hardcoded copy into https://github.com/alphagov/govuk-coronavirus-content for consistency
- [x]  Publish the hardcoded copy in production

This adds a synchronous call when rendering /coronavirus to load statistic data from https://coronavirus.data.gov.uk. There are some concerns that this data could timeout, so there are a few safeguards. See commits for more details. 

The change looks like this:

Screenshot 2021-03-23 at 17 25 39

### DIFF
--- a/app/services/fetch_coronavirus_statistics_service.rb
+++ b/app/services/fetch_coronavirus_statistics_service.rb
@@ -1,0 +1,88 @@
+class FetchCoronavirusStatisticsService
+  CACHE_KEY = "coronavirus_statistics".freeze
+
+  Statistics = Struct.new(:cumulative_vaccinations,
+                          :cumulative_vaccinations_date,
+                          :hospital_admissions,
+                          :hospital_admissions_date,
+                          :new_positive_tests,
+                          :new_positive_tests_date,
+                          keyword_init: true)
+
+  def self.call
+    new.call
+  end
+
+  def call
+    # This API times out on occassion, so we'll keep a copy of our potentially
+    # stale data to use as a backup
+    backup_statistics = Rails.cache.fetch(CACHE_KEY)
+    statistics = fetch_current_statistics || backup_statistics
+
+    Statistics.new(statistics) if statistics && statistics.any?
+  end
+
+  private_class_method :new
+
+private
+
+  def fetch_current_statistics
+    Rails.cache.fetch(CACHE_KEY, expires_in: 30.minutes, race_condition_ttl: 5.minutes) do
+      request_statistics
+    end
+  rescue StandardError => e
+    Rails.logger.warn("Failed to load coronavirus statistics: #{e}")
+    intermittent_error = case e
+                         when Faraday::TimeoutError, Faraday::ConnectionFailed
+                           true
+                         when Faraday::ServerError
+                           [502, 504].include?(e.response_status)
+                         end
+
+    GovukError.notify(e) unless intermittent_error
+    nil
+  end
+
+  def request_statistics
+    connection = Faraday.new("https://coronavirus.data.gov.uk") do |f|
+      f.response(:raise_error)
+      f.options.timeout = 2
+    end
+
+    response = connection.get do |request|
+      request.url("/api/v1/data", {
+        filters: "areaName=United Kingdom;areaType=overview",
+        structure: {
+          "date" => "date",
+          "cumulativeVaccinations" => "cumPeopleVaccinatedFirstDoseByPublishDate",
+          "hospitalAdmissions" => "newAdmissions",
+          "newPositiveTests" => "newCasesByPublishDate",
+        }.to_json,
+      })
+    end
+
+    parse_response(response)
+  end
+
+  def parse_response(response)
+    data = JSON.parse(response.body).fetch("data")
+    parsed = {}
+
+    if (latest_vaccinations = data.find { |d| d["cumulativeVaccinations"] })
+      parsed[:cumulative_vaccinations] = latest_vaccinations["cumulativeVaccinations"]
+      parsed[:cumulative_vaccinations_date] = Date.parse(latest_vaccinations["date"])
+    end
+
+    if (latest_admissions = data.find { |d| d["hospitalAdmissions"] })
+      parsed[:hospital_admissions] = latest_admissions["hospitalAdmissions"]
+      parsed[:hospital_admissions_date] = Date.parse(latest_admissions["date"])
+    end
+
+    if (latest_tests = data.find { |d| d["newPositiveTests"] })
+      parsed[:new_positive_tests] = latest_tests["newPositiveTests"]
+      parsed[:new_positive_tests_date] = Date.parse(latest_tests["date"])
+    end
+
+    parsed
+  end
+end
--- a/spec/services/fetch_coronavirus_statistics_service_spec.rb
+++ b/spec/services/fetch_coronavirus_statistics_service_spec.rb
@@ -1,0 +1,122 @@
+RSpec.describe FetchCoronavirusStatisticsService do
+  describe ".call" do
+    it "returns a Statistics object" do
+      body = {
+        data: [
+          { "date" => "2021-03-18", "cumulativeVaccinations" => nil, "hospitalAdmissions" => nil, "newPositiveTests" => 6303 },
+          { "date" => "2021-03-17", "cumulativeVaccinations" => 25_735_472, "hospitalAdmissions" => nil, "newPositiveests" => 5758 },
+          { "date" => "2021-03-16", "cumulativeVaccinations" => 25_273_226, "hospitalAdmissions" => nil, "newPositiveests" => 5294 },
+          { "date" => "2021-03-15", "cumulativeVaccinations" => 24_839_906, "hospitalAdmissions" => nil, "newPositiveests" => 5089 },
+          { "date" => "2021-03-14", "cumulativeVaccinations" => 24_453_221, "hospitalAdmissions" => 426, "newPositiveests" => 4618 },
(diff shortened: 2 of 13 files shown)


spec/controllers/taxons_controller_spec.rb
```diff
@@ -1,5 +1,3 @@
-require_relative "../../test/support/coronavirus_helper"
```
JonathanHallam, line 1: just doing an initial glance at the code rather than a full, why is this being removed?

JonathanHallam: because that helper's been removed, I'll show myself out

kevindew: Ah don't worry - I should have left a note in the commit. I think it was just an error that someone left this in the taxons_controller spec as that doesn't have anything to do with the coronavirus helpers.

spec/services/fetch_coronavirus_statistics_service_spec.rb
```diff
@@ -0,0 +1,124 @@
...
+
+    context "when there is missing data" do
+      it "returns nil for no data" do
+        stub_request(:get, /coronavirus.data.gov.uk/)
+          .to_return(status: 200, body: { data: [] }.to_json)
+
+        expect(described_class.call).to be_nil
+      end
+
+      it "sets only the fields that have data" do
```
JonathanHallam, line 38: It's good that it can deal with this situation but are we ever expecting it to happen? IE can some fields on coronavirus.data.gov.uk return nil whilst others have values? Also, can the frontend deal with this? Will it show headings for all fields or just those that have data?

I'm not sure what the user would think if some values showed whilst others didn't.

test/support/coronavirus_helpers.rb
```diff
@@ -0,0 +1,130 @@
+module CoronavirusHelpers
```
JonathanHallam, line 1: There's two lines of change on this file when compared to the helper it's replacing. I'm going to assume that that's just the `module CoronavirusHelpers` line 1 and the `end` that this necessitates rather than going thought line by line 😂

JonathanHallam (commented):
Happy with the approach, this looks good 🎉

test/support/coronavirus_helpers.rb
kevindew: Yup, that's correct no further modifications. I should have done the module change and rename as separate commits then it'd have been easier to see the changes with whitespace off.

spec/services/fetch_coronavirus_statistics_service_spec.rb
kevindew: Yup, the frontend can deal with this - we have conditionals for each option and just won't show a block if there isn't the  data. I don't think a user would be particularly concerned if some showed and not others, we don't tell them what to expect to see.

The API does currently return some fields null as per output actual output from: https://github.com/alphagov/collections/pull/2315/files#diff-583212f27e5e2c931d9ead6be0a3e8b8b55aac9420e3e1dc277d4259004acbb2R5-R13

Question is whether the API could return with all the fields being null, that'd seem very unlikely if it was all operating ok as they'd have to not update the data for a very long time. However given it's an external service and it would be correct to their schema to all be null it feels like it should be handled.

spec/services/fetch_coronavirus_statistics_service_spec.rb
```diff
@@ -0,0 +1,124 @@
...
+      body = {
+        data: [
+          { "date" => "2021-03-18", "cumulativeVaccinations" => nil, "hospitalAdmissions" => nil, "newPositiveTests" => 6303 },
+          { "date" => "2021-03-17", "cumulativeVaccinations" => 25_735_472, "hospitalAdmissions" => nil, "newPositiveests" => 5758 },
+          { "date" => "2021-03-16", "cumulativeVaccinations" => 25_273_226, "hospitalAdmissions" => nil, "newPositiveests" => 5294 },
+          { "date" => "2021-03-15", "cumulativeVaccinations" => 24_839_906, "hospitalAdmissions" => nil, "newPositiveests" => 5089 },
+          { "date" => "2021-03-14", "cumulativeVaccinations" => 24_453_221, "hospitalAdmissions" => 426, "newPositiveests" => 4618 },
+          { "date" => "2021-03-13", "cumulativeVaccinations" => 24_196_211, "hospitalAdmissions" => 460, "newPositiveests" => 5534 },
+        ],
+      }
```
leenagupte, line 13: This is great! Thanks for clarifying what we expect data from coronavirus.data.gov.uk to look like ❤️

app/services/fetch_coronavirus_statistics_service.rb
```diff
@@ -0,0 +1,84 @@
...
+  def fetch_current_statistics
+    Rails.cache.fetch(CACHE_KEY, expires_in: 30.minutes, race_condition_ttl: 5.minutes) do
+      request_statistics
+    end
+  rescue StandardError => e
+    Rails.logger.warn("Failed to load coronavirus statistics: #{e}")
+    intermittent_error = e.is_a?(Faraday::ConnectionFailed) ||
+      e.is_a?(Faraday::ServerError) && [502, 504].include?(e.response_status)
+
+    GovukError.notify(e) if !e.is_a?(Faraday::TimeoutError) && !intermittent_error
```
leenagupte, line 38: The logic for `intermittent_error` is a bit of brain twister. Until I saw the tests, I read it to mean we notify Sentry if the error **is** a 502 or 504.

spec/services/fetch_coronavirus_statistics_service_spec.rb
```diff
@@ -0,0 +1,124 @@
...
+        stub_request(:get, /coronavirus.data.gov.uk/).to_timeout
+        expect(GovukError).not_to receive(:notify)
+
+        described_class.call
+      end
+
+      it "doens't notify GovukError on 502 and 504 errors" do
+        stub_request(:get, /coronavirus.data.gov.uk/)
+          .to_return(status: 502)
+          .then
```
leenagupte, line 76: This is nice! I didn't know it was possible to do this.

spec/services/fetch_coronavirus_statistics_service_spec.rb
```diff
@@ -0,0 +1,124 @@
...
+          .to receive(:fetch)
+          .with(described_class::CACHE_KEY)
+          .and_return(stale_stats)
+
+        allow(Rails.cache)
+          .to receive(:fetch)
+          .with(described_class::CACHE_KEY, expires_in: 30.minutes, race_condition_ttl: 5.minutes)
+          .and_call_original
+
+        expect(described_class.call).to have_attributes(stale_stats)
```
leenagupte, line 110: Sorry, I'm a bit confused, how does this test work? How do we know that cache data is stale?

leenagupte (commented):
I had a quick look at this out of curiosity, I'll leave Alex to do the official review. 
The test data is amazing! It really helps to see what we're expecting the data to look like. I've left a couple of inline questions/comments, but nothing major.

GDSNewt:
Hi @kevindew , much like Leena I felt like just commenting 'nice!' on a lot of this code 😄 . I especially liked how you dealt with the inconsistent API data, and as Leena noted the test data really shows how the information displayed can be staggered. 

The only note I really had was regarding the naming the variable names in the partials, I was reviewing commit by commit, but I see you have removed some of the ambiguity in the final commit. Also content in your explanations to John's earlier review. Happy to approve 👍

GDSNewt (approved):
:shipit:

app/services/fetch_coronavirus_statistics_service.rb
kevindew: Yup, you're right. I've refactored it to be an easier read.

spec/services/fetch_coronavirus_statistics_service_spec.rb
kevindew: I had to Google 😁

spec/services/fetch_coronavirus_statistics_service_spec.rb
kevindew: Yeah, this is a bit nuanced - it was quite hard to document as a test. I've refactored it a little to try make it a better read, though I don't think it's ever going to be great.

So the code works by first making a request to the cache without an expiry to load data whether it's fresh or stale, and then it then makes a second call to the cache which will load fresh data or make the API request.

What this test does is it stubs the call of the first cache call, so the output of that is predictable. It then has the second cache call fail. By the code returning the data from the first cache call we know that it had to have been from the first cache and therefore resilient to failure on the second call.

app/services/fetch_coronavirus_statistics_service.rb
```diff
@@ -0,0 +1,88 @@
...
+  rescue StandardError => e
+    Rails.logger.warn("Failed to load coronavirus statistics: #{e}")
+    intermittent_error = case e
+                         when Faraday::TimeoutError, Faraday::ConnectionFailed
+                           true
+                         when Faraday::ServerError
+                           [502, 504].include?(e.response_status)
+                         end
+
+    GovukError.notify(e) unless intermittent_error
```
leenagupte, line 42: This is much easier to read. Thanks!

leenagupte (approved):
Thanks for the explanations 🏅 👍

PR #2316, opened 2021-03-17
Remove unused pluralisation keys from locale files

brucebolt:
In https://github.com/alphagov/collections/pull/2312, we updated the plural rules for each language that we support that is not included [in the ruby-i18n gem](https://github.com/ruby-i18n/i18n/blob/master/test/test_data/locales/plurals.rb).

This removes the keys that are no longer required in each language from their locale files.

Trello card: https://trello.com/c/lkVB57ri

I have not made any changes to `tr` as I believe the i18n rules for this could be wrong, so will write a card to look into this.

### DIFF
--- a/config/locales/az/people.yml
+++ b/config/locales/az/people.yml
@@ -2,7 +2,6 @@ az:
   people:
     biography:
     heading:
-      one:
       other:
     previous_roles:
     previous_roles_in_government:
--- a/config/locales/az/roles.yml
+++ b/config/locales/az/roles.yml
@@ -1,7 +1,6 @@
 az:
   roles:
     heading:
-      one:
       other:
     headings:
       current_holder:
--- a/config/locales/fa/people.yml
+++ b/config/locales/fa/people.yml
@@ -2,7 +2,6 @@ fa:
   people:
     biography:
     heading:
-      one:
       other:
     previous_roles:
     previous_roles_in_government:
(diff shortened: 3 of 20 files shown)


MuriloDalRi:
Wondering how did you decide which ones to remove and which ones to keep?

For instance `tr` only has an `other` rules according to i18n, did you keep it because we're using that key somewhere?

brucebolt:
> Wondering how did you decide which ones to remove and which ones to keep?
> 
> For instance `tr` only has an `other` rule according to i18n, did you keep the `one` because we're using that key somewhere?

I used the rules from i18n.  Good spot with `tr`.  I _think_ the rules for this are wrong and there should be singular/plural forms, which I why I left that specific one in.  I had meant to add that as a comment to this PR, but forgot to do that.  I'm also writing a card to look into this.

MuriloDalRi:
> > Wondering how did you decide which ones to remove and which ones to keep?
> > For instance `tr` only has an `other` rule according to i18n, did you keep the `one` because we're using that key somewhere?
> 
> I used the rules from i18n. Good spot with `tr`. I _think_ the rules for this are wrong and there should be singular/plural forms, which I why I left that specific one in. I had meant to add that as a comment to this PR, but forgot to do that. I'm also writing a card to look into this.

Yeah, I think you're right, according to [the Unicode rules](https://unicode-org.github.io/cldr-staging/charts/37/supplemental/language_plural_rules.html#tr)

MuriloDalRi (approved):

brucebolt:
I have raised a PR on `rails-i18n` to correct the keys used for Turkish: https://github.com/svenfuchs/rails-i18n/pull/927

PR #2342, opened 2021-03-29
Create NHS Notice partial for April 2021 COVID guidance

kevindew:
Trello: https://trello.com/c/NSprpQLr/295-update-nhs-box

This creates a partial that is tailored to display the NHS guidance that
is expected to be agreed for April 2021.

The copy is so far removed from the previous component (header-notice)
that I've decided it is best to create something dedicated for this and
remove that component.

I've gone with an approach that sets up a modestly flexible section
approach that can be used to set the content for this box mostly through
markdown. I did initially see if the whole block could be markdown but
this causes a bunch of spacing issues with the headers (and creates some
unnecessary anchors) it seemed a pretty messy thing to resolve.

I'm somewhat skeptical that this content designer editable structure
will be flexible enough for future needs but I've tried. I've left the
Content Item data hardcoded into the view for now, until it can be
ported to the YAML file.

---

## Before:

Screenshot 2021-03-29 at 14 38 32

## After:

Screenshot 2021-03-29 at 14 38 47


app/views/coronavirus_landing_page/components/landing_page/_nhs_notice.html.erb
```diff
@@ -0,0 +1,45 @@
...
+        "heading" => "If you have no symptoms:",
+        "markdown" => "Check if you should get [regular rapid lateral flow tests](/getting-tested-for-coronavirus).",
+      },
+      {
+        "heading" => "If you have coronavirus symptoms:",
+        "markdown" => "- [get a PCR test](/get-coronavirus-test)\n- stay at home",
+      }
+    ]
+  }
+%>
```
DilwoarH, line 15: will this be removed before merging?

kevindew: No, it will be removed after deployment when I've updated the YAML file.

app/views/coronavirus_landing_page/components/landing_page/_nhs_notice.html.erb
```diff
@@ -0,0 +1,45 @@
...
+
+  <% nhs_banner["sections"].each do |section| %>
+    <%= render "govuk_publishing_components/components/heading", {
+      text: section["heading"],
+      heading_level: 2,
+      font_size: "m",
+      margin_bottom: 4,
+    } if section["heading"].present? %>
+
+    <%= render_govspeak(section["markdown"]) if section["markdown"].present? %>
```
Rosa-Fox, line 43: What will happen if the markdown is invalid? I tried to break it and found that it can spill over the box quite easily... would some sort of a line length check work?

markdown

spec/views/components/header_notice.html.erb_spec.rb
```diff
@@ -1,146 +0,0 @@
-RSpec.describe "header-notice component" do
```
Rosa-Fox, line 1: What was the reason for not having a `_nhs_notice.html.erb_spec`?

Rosa-Fox (commented):
Looks good locally! I am glad Dilwoar is also reviewing as my input on a styling perspective is limited. I left a couple of questions.

app/views/coronavirus_landing_page/components/landing_page/_nhs_notice.html.erb
kevindew: Thanks for pointing this out. I see this as a concern for someone editing the copy rather than a code concern, it's an existing possibility for any of the currently editable fields of the existing approach, rather than something introduced in this change.

spec/views/components/header_notice.html.erb_spec.rb
kevindew: It didn't strike me as something that would be conventional to test. All of the code paths are executed implicitly as part of other tests.

app/views/coronavirus_landing_page/components/landing_page/_nhs_notice.html.erb
DilwoarH: I think it'll be unlikely that we'd have such a long word but we could add a `word-break: break-word;` css for good measure.

DilwoarH (approved):
Looks good to me 👍

spec/views/components/header_notice.html.erb_spec.rb
Rosa-Fox: I am surprised... I thought you were a view test fan @kevindew! It did seem very rigidly tested before which was maybe overkill but then testing rendering the data from the `coronavirus_landing_page` feels maybe too vague?

app/views/coronavirus_landing_page/components/landing_page/_nhs_notice.html.erb
Rosa-Fox: 👍  I suppose unless we actually put this into the publishing tool then its going to be difficult to validate.

spec/views/components/header_notice.html.erb_spec.rb
kevindew: Haha, I wouldn't really describe myself as a fan of them - but it is preferable, if they contain logic, to test them at a unit level view test rather than using a higher level test to unit test them.

I think essentially all you really want testing wise from views is that you want all code paths exercised (which is often all that is sufficient for most unit tests) which just gives you the basic reassurance that all the code executes and there's not some hidden typo or dependency that isn't exercised until runtime. 

There seems low value in testing that a view outputs particular HTML as it just basically confirms you typed what you did in the code.

PR #2364, opened 2021-04-21
Add helper for hiding global bar

injms:
## What

Adds a helper that checks whether a page should hide the global bar; the view can then add a class to the body element that will hide the global bar.

These changes won't have any effect until Static has been updated as Static needs to contain the CSS for this to have any impact. Static needs to be updated to flip the behaviour from hide by default to show by default.

## Why

Currently the global bar is initially hidden, then shown using JavaScript - this causes a shift in the layout of the page (aka jank aka Cumulative Layout Shift) and means that the global bar is not available for users without JavaScript. 

([More about Cumulative Layout Shift on web.dev](https://web.dev/cls/).)

### CLS

To avoid this, a change will need to be made in Static to the make the global bar be initially shown by default. The bar can then be hidden on pages that don't need it - and doing this using CSS will avoid jank.

The screenshot below shows the layout shift that occurs when the global bar appears:

image

As the search form is present, it's frustrating for users when they select a link only for the layout shift to move the page and a different link to be selected.

This is also present on larger screens, though the effect is dwarfed by the CLS caused by the appearance of the cookie banner:

image

Worth bearing in mind that for users that have made a cookie choice, the CLS will only be due to the global bar. 

### No JavaScript

And for users without JavaScript turned on, the global bar does not appear at all:

image

### But why does this change involve collections?

The global bar should not appear on certain pages - such as the Brexit or Coronavirus landing pages. Previously there was some JavaScript to prevent the global bar from appearing. This has been changed to an helper in the application that can be used to add a class to the body if the global should not appear. 

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -65,4 +65,18 @@ module ApplicationHelper
       raw(Govspeak::Document.new(content, sanitize: false).to_html)
     end
   end
+
+  def hide_global_bar?
+    paths_to_hide_global_bar_on = [
+      "^/coronavirus$",
+      "^/coronavirus/.*$",
+      "^/transition(.cy)?$",
+      "^/transition-check/.*$",
+      "^/eubusiness(\\..*)?$"
+    ]
+
+    concatenated_regexes = Regexp.new(paths_to_hide_global_bar_on.join("|"))
+
+    return request.path.match(concatenated_regexes).present?
+  end
 end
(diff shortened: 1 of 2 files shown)


injms:
I've updated the caching to match that described in the [action/setup-node docs](https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#caching-packages-dependencies). The new setup seems a lot easier to manage and was straightforward change.

PR #2370, opened 2021-04-27
Remove unused i18n keys

ChrisBAshton:
This PR adds the [i18n-coverage gem](https://github.com/hiptest/i18n-coverage) to monitor untested i18n keys. I've then removed used the coverage reports to identify and remove unused keys. See commits for details.

In case it's needed for future reference, I used the i18n-tasks gem to remove properties across multiple locales. It needed a `config/i18n-tasks.yml` file with the following contents, after which I was able to run commands such as `i18n-tasks rm 'roles.policies_responsible_with_person'`:

```yml
base_locale: en
data:
  read:
    - config/locales/%{locale}.yml
    - config/locales/%{locale}/*.yml
  write:
    - config/locales/%{locale}.yml
    - config/locales/%{locale}/*.yml
```

Unfortunately there's no way to tell the gem not to reorder keys, so this looks like quite a sizeable diff, even when no content has changed.

Trello: https://trello.com/c/Z3Pitrfg/2454-identify-and-remove-unused-keys-in-locale-files-5


thomasleese (approved):
I haven't checked in file in detail, but in general it all looks good.

PR #2383, opened 2021-05-10
Adds missing keys

1pretz1:
This PR adds missing keys to the corresponding locale files. Also adds configuration for missing keys in our dev and test environments.

More info in the commits ->

Trello:
https://trello.com/c/kYXvexWo/2455-add-missing-keys-to-locale-files-3 


config/locales/bg/organisations.yml
```diff
@@ -15,62 +15,62 @@ bg:
           one:
           other:
         announcement:
-          one:
-          other:
+          one: Съобщение
+          other: Съобщения
```
ChrisBAshton, line 19: Hmm, we have both `one` and `other` here in commit 640580a62db62b89c7af80af12f38ad8207473f5, but we've removed the `one` key in other places, e.g. https://github.com/alphagov/collections/pull/2383/commits/640580a62db62b89c7af80af12f38ad8207473f5#diff-c8d38f9dfce6230fc33b107b1fb2f14406fac90137b008aeb55bb29d5382bacbL192.

Should we not be keeping the `one` key everywhere, even if we don't have values in all locales?

app/views/dit_landing_page/show.html.erb
```diff
@@ -2,4 +2,4 @@
 
 <%= render partial: 'guidance' %>
 
-<%= render partial: 'training', locals: { table: TablePresenter.new(t("dit_landing_page.training_section_table")) } %>
```
ChrisBAshton, line 5: good digging 👍

ChrisBAshton (commented):
LGTM except for one comment which may need addressing - see below.

app/views/dit_landing_page/show.html.erb
1pretz1: Thanks Chris!

config/locales/bg/organisations.yml
1pretz1: Yes we should, if the language supports that form of pluralisation. The reason `one` has been deleted from certain locale files is because the language doesn't have an equivalent to the English `one`.

Work has been done prior to this to work out which pluralisation keys are actually needed for each language, you can see the decision to remove `one` from certain in locales in [this commit](https://github.com/alphagov/government-frontend/pull/2052/commits/2a81e71175cad844286188c9be5b2e8b8909c6be).

ChrisBAshton (approved):
LGTM - thanks for explaining.

PR #2401, opened 2021-05-26
Redirect taxons to their override url if present

hannako:
## What

We have added an optional field to the details hash of the taxon [content schema](https://github.com/alphagov/govuk-content-schemas/pull/1059). If that field has been populated (in [content tagger](https://github.com/alphagov/content-tagger/pull/1204)) , then Collections should redirect to it.

[trello](https://trello.com/c/V2nmHHA6/1516-make-it-possible-to-add-an-overrideurl-to-taxons-in-content-tagger) 

### DIFF
--- a/app/controllers/taxons_controller.rb
+++ b/app/controllers/taxons_controller.rb
@@ -4,6 +4,8 @@ class TaxonsController < ApplicationController
   def show
     setup_content_item_and_navigation_helpers(taxon)
 
+    redirect_to(url_override, status: :temporary_redirect) and return if url_override
+
     render locals: {
       presented_taxon: presented_taxon,
       presentable_section_items: presentable_section_items,
@@ -35,5 +37,9 @@ private
     end
 
     @presentable_section_items
+  end
+
+  def url_override
+    @content_item.dig("details", "url_override")
   end
 end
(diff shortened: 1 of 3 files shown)


app/controllers/taxons_controller.rb
```diff
@@ -4,6 +4,8 @@ class TaxonsController < ApplicationController
   def show
     setup_content_item_and_navigation_helpers(taxon)
 
+    redirect_to(url_override, status: :temporary_redirect) and return if url_override
```
sihugh, line 7: I reckon we might want to use `.present?` on the end here, as an empty string evaluates to `true`.

sihugh: See:

```
irb(main):001:0> "" ? "bernard" : "jim"
=> "bernard"
```

spec/support/taxon_browsing_helper.rb
```diff
@@ -349,12 +350,6 @@ def tagged_content
     generate_search_results(5)
   end
 
-  def taxon
```
sihugh, line 352: nice deleting :-)

sihugh (approved):

app/controllers/taxons_controller.rb
hannako: Thanks!

PR #2402, opened 2021-05-27
Add route for the Brexit taxon's new base path.

hannako:
## What

We are planning to change the url of the transition landing page from `/transition` to `/brexit`. 

Once merged, a user visiting `/transition` will see a basic taxon page. 

The taxon can then be given a new base path of `/brexit` in content tagger. This will create a redirect from /transition to /brexit, so users visiting either path will see a correctly styled page. 

https://trello.com/c/wxR5o06H/1539-implement-brexit-landing-page-url-change

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -79,6 +79,7 @@ Rails.application.routes.draw do
   end
 
   get "/world/*taxon_base_path", to: "world_wide_taxons#show"
+  get "/brexit(.:locale)", to: "transition_landing_page#show"
   get "/transition(.:locale)", to: "transition_landing_page#show"
 
   # We get requests for URLs like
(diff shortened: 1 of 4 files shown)


sihugh (changes requested):
I think we'll have to leave both in place for a short time. We'll have to add /brexit before we rename, and then remove the /transition path after the rename. Otherwise we'll have a period where just show a standard taxon page for a few

hannako:
Hmm I thought this would have been the least bad option so there's less risk of collections trying to render a redirect. Whichever order we do things in one of those urls will be a redirect at some point in the process, so I thought it would be better to render an unstyled brexit page temporarily whilst collection is being deployed, rather than an error page. But I am probably just confused 😂

sihugh (approved):

PR #2435, opened 2021-06-16
Add reset to lists on Coronavirus landing page

injms:
## What

Add `govuk-list` to override the browser default padding and margin to the lists on the Coronavirus landing page.

## Why

The global CSS reset was been removed with the move to using `gem_layout` and the public layout component - this caused the lists to fall back to the browsers default padding and margin, which meant that the alignment of the lists was not correct.

## Visual differences
| Before | After |
| - | - |
| image | image |

---

### DIFF
--- a/app/views/coronavirus_landing_page/components/hub/_guidance_section.html.erb
+++ b/app/views/coronavirus_landing_page/components/hub/_guidance_section.html.erb
@@ -9,7 +9,7 @@
             font_size: 19,
             margin_bottom: 5
           } %>
-          <ul class="covid__list govuk-!-margin-bottom-8">
+          <ul class="govuk-list covid__list govuk-!-margin-bottom-8">
             <% guidance["list"].each do | item | %>
               <li class="covid__list-item">
                 <%= render 'govuk_publishing_components/components/action_link', {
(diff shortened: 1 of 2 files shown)


app/views/coronavirus_landing_page/components/hub/_guidance_section.html.erb
```diff
@@ -9,7 +9,7 @@
             font_size: 19,
             margin_bottom: 5
           } %>
-          <ul class="covid__list govuk-!-margin-bottom-8">
+          <ul class="govuk-list covid__list govuk-!-margin-bottom-8">
```
chris-gds, line 12: Could be swapped at some point to the [list gem](https://components.publishing.service.gov.uk/component-guide/list)

chris-gds (approved):

app/views/coronavirus_landing_page/components/hub/_guidance_section.html.erb
injms: Unfortunately it can't in this case - the `covid_list` does some extra things and the list component (along with all the other components) doesn't allow custom classes to be added.

PR #2437, opened 2021-06-18
[DO NOT MERGE] Timeline frontend

andysellick:

## What

Add a nation picker to the coronavirus landing page timeline information.

Screenshot 2021-06-18 at 09 13 45

With JS disabled, this works using a standard form GET submission to show/hide the relevant section of the page. If JS is enabled, we hide the form submit button and clicking a radio button automatically switches the visibility of the relevant section.

Includes `aria-live` attributes so that changing the radio button announces the results of the change, and tracking attributes on the radio buttons themselves to give us some stats about usage (if cookies have been consented to).

Things to do:

- backend has to be hooked up (currently just shows and hides some dummy text)
- code needs to be ready to automatically handle the switchover from the current state to data based on specific countries
- text of the elements needs to be confirmed
- text needs adding to translation files
- more tests

## Why

So that users can access specific information relevant to their location.

Trello cards: 

- https://trello.com/c/eE62ythU/447-coronavirus-timeline-frontend
- https://trello.com/c/Bn4ekYnO


kevindew (commented):
This is looking great, thanks @andysellick and @leenagupte - I've tried it out locally and aside from a small no-js snag on Northern Ireland it's working great.

I've added a few suggestions - hope they're helpful, happy to chat through any.

Leena with regards to your question about what to test on this. My thoughts are that there's a feature to test, both with and without javascript:

With & without javascript: you can go to the landing page, see England and UK wide entries visible, not see wales, change country to wales, can see Wales and UK Wide entries but not England.

Then at a controller level there's a few things to test: test old and new timeline switchover, each country selected (via param), when a country has no updates,

leenagupte:
@kevindew Thanks for your comments. I've added the feature tests and controller tests as suggested. Thanks. I've also moved some of the code out of the view and into the helper.

I think I've implemented most of the suggestions in the comments, however, I'm unsure about this [one](https://github.com/alphagov/collections/pull/2437#discussion_r655639832):

> I wonder if there should be some of the view logic in the controller. I feel like whether we show the feature and which country is selected would make sense to set as instance variables. It might also make sense to do the country filtering in the controller.

Is it possible to do the filtering in the controller and have the client-side JS still work?

Things still left to do:

- [ ] Get links to general guidance for devolved nations
- [ ] Move content into translation file
- [ ] Sort out commit messages

app/controllers/coronavirus_landing_page_controller.rb
```diff
@@ -3,14 +3,20 @@
...
     if @statistics
       set_expiry 5.minutes
     else
       logger.warn "Serving /coronavirus without statistics"
       set_expiry 30.seconds
     end
 
-    @content_item = content_item.to_hash
+    @content_item = if params[:timeline_nation] && !Rails.env.production?
+                      timeline_nations_content_item.to_hash
```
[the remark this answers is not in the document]
leenagupte: It appears that brakeman doesn't trust models. Calling `CoronavirusTimelineNationsContentItem.load.to_hash` directly in the `if` caused brakeman to throw a security warning:

== Warning Types ==

Cross-Site Scripting: 1

== Obsolete Ignore Entries ==

344ea763c233c1799950725baf2a9cd55fca8ccb96a0e67af99fbaee080b4487
840e07ce0072ce08b8e6875bde330c91ba40ac3cd6545a95952a567008a3aa83
fcdb8d748855278e7cb8d9639a2f0a4fd446d022d744162a440d2d967d0a2ae3

== Warnings ==

Confidence: Weak
Category: Cross-Site Scripting
Check: CrossSiteScripting
Message: Unescaped model attribute
Code: JSON.pretty_generate(presenter.faq_schema((CoronavirusTimelineNationsContentItem.load.to_hash or content_item.to_hash)))
app/views/coronavirus_landing_page/components/shared/_schema.html.erb
Line: 2

I had a look at some [docs](https://brakemanscanner.org/docs/warning_types/cross_site_scripting/) and rather than ignore the warning, I'm just going to keep it wrapped in the method.

app/views/coronavirus_landing_page/components/shared/_timeline.html.erb
```diff
@@ -2,22 +2,70 @@
...
+        <%= render "govuk_publishing_components/components/button", {
+          text: "View",
+          margin_bottom: 6
+        } %>
+      </div>
+    <% end %>
+
+    <div aria-live="polite">
+      <% CoronavirusTimelineNationsHelper::UK_COUNTRY_LIST.each do |nation| %>
+        <div id="nation-<%= nation %>" class="js-covid-timeline <%= 'covid-timeline__item--hidden' unless display_country(@selected_country) == nation %>">
```
kevindew, line 30: Since this has `covid-timeline__item--hidden` we should have `covid-timeline__item` set on it as well.

spec/controllers/coronavirus_landing_page_controller_spec.rb
```diff
@@ -25,5 +25,54 @@
...
+        get :show, params: { timeline_nation: "wales" }
+
+        expect(response.body).to have_content("There haven't been any recent updates for Wales.")
+      end
+
+      it "shows timeline for England" do
+        stub_content_store_has_item("/coronavirus", coronavirus_content_item_with_timeline_national_applicability)
+        get :show, params: { timeline_nation: "england" }
+
+        expect(response.body).to have_content("In England")
```
kevindew, line 53: This test will pass for any param values of timeline_nation, if I understand right, as we'll always have this content, the determining factor will be what class the selector has. E.g. for England this would be `div#nation-england:not(.covid-timeline__item--hidden)`.

Might be good to have a second assertion that we only have England selected with an assertion of `expect(response.body).not_to have_selector("div.covid-timeline__item:not(.covid-timeline__item--hidden).not(#nation-england)")` though it's a pretty fierce selector (also requires `covid-timeline__item` as per my previous comment)

spec/controllers/coronavirus_landing_page_controller_spec.rb
```diff
@@ -25,5 +25,54 @@
         expect(response.headers["Cache-Control"]).to eq("max-age=#{30.seconds}, public")
       end
     end
+
+    context "when testing national_applicability" do
+      render_views
+
+      before do
+        allow(Rails.env).to receive(:production?).and_return(true)
```
kevindew, line 33: This should only belong in the first test as the other tests should all fail if Rails has a production environment. I'm a bit confused why they're not so not sure if I've missed something.

spec/controllers/coronavirus_landing_page_controller_spec.rb
```diff
@@ -25,5 +25,54 @@
...
+      end
+
+      it "loads the production content item in production environments" do
+        expect(CoronavirusTimelineNationsContentItem).to_not receive(:load)
```
kevindew, line 37: Nitpick: this description doesn't seem to describe what is tested. It feels like it should be asserting on `ContentItem.find!("/coronavirus")` being called - I feel like there should be an inverse test too.

spec/support/coronavirus_landing_page_steps.rb
```diff
@@ -51,6 +55,11 @@ def when_i_visit_the_coronavirus_landing_page
     visit CORONAVIRUS_PATH
   end
 
+  def when_i_visit_a_coronavirus_landing_page_with_timeline_nation_applicability
+    allow(Rails.env).to receive(:production?).and_return(true)
```
kevindew, line 59: I'm a bit confused by this too as I'd have thought this disables the feature?

spec/support/coronavirus_landing_page_steps.rb
```diff
@@ -124,6 +141,16 @@ def then_i_can_see_the_timeline
     expect(page).to have_selector(".covid-timeline__item .gem-c-govspeak", text: "Face coverings are mandatory in shops")
   end
 
+  def then_i_can_see_the_timeline_for_england
+    expect(page).to have_selector(".js-covid-timeline .govuk-heading-m", text: "Changes in England")
+    expect(page).to have_selector(".js-covid-timeline .govuk-heading-m", text: "Changes in Wales", visible: false)
```
kevindew, line 146: Do these need to be selector specific? 

I imagine it'd be a cleaner read with `expect(page).to have_content` and `expect(page).to_not have_content` assertions

app/helpers/coronavirus_timeline_nations_helper.rb
```diff
@@ -0,0 +1,50 @@
...
+        "<strong class='govuk-tag govuk-tag--blue'>#{nation.titleize}</strong>"
+      end
+
+      nation_tags.join(" ")
+    end
+  end
+
+  def uk_wide?(national_applicability)
+    UK_COUNTRY_LIST.sort == national_applicability.uniq.sort
+  end
```
kevindew, line 49: Nitpick: this would be better suited to being a variable than a helper method given this is a method is only used in one place, doesn't need public visibility and would be globally available despite only a niche context.

```
uk_wide = UK_COUNTRY_LIST.sort == national_applicability.uniq.sort
if uk_wide
```

kevindew (commented):
Thanks @leenagupte and great progress 👍 Nice one getting tests working.

Re your question

> Is it possible to do the filtering in the controller and have the client-side JS still work?

The suggestion for that would be to create a hash in the controller (do the logic) so that things were simple in the view. E.g. create `{ "england" => [entry_1, entry_2], "northern_ireland" => [entry_1], ... }`. However now we know about the presenter class it might be be better suited there.

And Slack comment: 

> there’s a bit of weirdness about how the presenter object is passed to the partials

You can pass the presenter through to the view as per here: https://github.com/alphagov/collections/blob/492f8fcf18628f79d77ae74fe4d9eaa7ea8d5bbf/app/views/coronavirus_landing_page/show.html.erb#L33 I'm not sure why it was named details rather than presenter. 

--- 

I've been having a bit of a think presenter. It's unfortunate we learn about it late as it'll be rather confusing to have some stuff in presenter and some in helper with ambiguity about what belongs in either.  We should resolve that, so do add working this out to your list of things still to do.

I imagine we can make use of the presenter to set the passed param as part of initialisation, e.g: `CoronavirusLandingPagePresenter.new(content_item, params[:timeline_nation])`, which can then simplify the questions we ask in the view.

It seems to make sense that, if we're not using the controller, that the presenter decides whether we `show_timeline_nations?`. Since it has the content item available and won't need params.

The options for the form could be a bit simpler with a presenter class as it wouldn't need a parameter if the presenter knows which country is selected. 

Then I wonder if we can do something a bit simpler with nation handling. Something like:

```
  <% details.timelines_for_nation do |nation, timeline| %>
    <div id="nation-<%= nation %>" class="js-covid-timeline <%= 'covid-timeline__item--hidden' unless details.selected_timeline_nation == nation %>">
      <h3 class="govuk-heading-m">Changes in <%= nation.titleize %></h3>
      <% if timeline.empty? %>
        <p class="govuk-body">There haven't been any recent updates for <%= nation.titleize %>.</p>
      <% end %>
      <% timeline.each do | item | %>
        <% text = capture do %>
          <%= item["heading"] %>
          <%= details.timeline_nation_tags(item["national_applicability"]) %>
        <% end %>
        <div class="covid-timeline__item">
            text: text,
            heading_level: 3,
            font_size: "s",
            margin_bottom: 1,
          } %>
          <%= render_govspeak(item["paragraph"]) %>
        </div>
      <% end %>
```

I find myself being persuaded that nearly everything we have in CoronavirusTimelineNationsHelper would be slightly cleaner through the presenter, whilst also cutting out future ambiguity. 

The one thing I find myself on the fence on is the `timeline_nation_tags` method since it wouldn't utilise any internal presenter state and more act as a utility method. However if everything else is the presenter then life would definitely be simpler if it's there too.

spec/support/coronavirus_landing_page_steps.rb
leenagupte: No, setting this to true means that it doesn't try to load the temporary content item from "config/data" and instead loads the fixture file instead. Otherwise we'd be testing the wrong content item. It doesn't really matter in this particular case as both content items ("config/data" and fixture file) have the same contents.

spec/support/coronavirus_landing_page_steps.rb
leenagupte: Strangely enough that doesn't work. I think it's because the content is in the page source, but is only visually hidden.

spec/support/coronavirus_landing_page_steps.rb
kevindew: Oh I see 🤯 

Would we be able to avoid most of the mocking of `Rails.env.production?` by changing the controller condition to be if `@selected_country && Rails.env.development?`? It's the type of method I don't feel very comfortable with mocking given it likely has surprising effects to any unrelated code that queries the environment.

spec/controllers/coronavirus_landing_page_controller_spec.rb
leenagupte: We do need it this, otherwise, it will load the content item from `CoronavirusTimelineNationsContentItem` rather than the live one we're stubbing.

For most of the tests it doesn't matter as the content item loaded for `CoronavirusTimelineNationsContentItem` is the same as the fixture for `coronavirus_content_item_with_timeline_national_applicability`, but that's not true of `coronavirus_content_item_with_timeline_national_applicability_without_wales`, so that test fails.

spec/controllers/coronavirus_landing_page_controller_spec.rb
leenagupte: This works too:

```
expect(response.body).to have_selector(".js-covid-timeline", text: "In England")
expect(response.body).to have_selector(".js-covid-timeline", text: "In Wales", visible: false)
```

app/views/coronavirus_landing_page/components/shared/_timeline.html.erb
leenagupte: @andysellick ☝️ 🥺

app/controllers/coronavirus_landing_page_controller.rb
```diff
@@ -2,6 +2,7 @@
 
 class CoronavirusLandingPageController < ApplicationController
   def show
+    @selected_country = params[:timeline_nation]
```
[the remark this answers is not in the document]
leenagupte: To be honest, there isn't really an "invalid" value. It's either "scotland", "wales" or "northern_ireland" or it defaults to "england".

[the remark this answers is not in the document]
kevindew: It'd be confusing if @selected_country had a value of "foo" for instance, you'd want it to be "england" if that's the selected country.

spec/controllers/coronavirus_landing_page_controller_spec.rb
kevindew: As it's a controller spec it's not rendering the content so it doesn't know if the content is visible or not (it's unfortunate it doesn't throw an error).

For example in this test:

```
(byebug) expect(response.body).to have_content("In England")
true
(byebug) expect(response.body).to have_content("In Wales")
true
(byebug) expect(response.body).to have_selector(".js-covid-timeline", text: "In England", visible: false)
true
(byebug) expect(response.body).to have_selector(".js-covid-timeline", text: "In England", visible: true)
true
(byebug) expect(response.body).to have_selector(".js-covid-timeline", text: "In Wales", visible: true)
true
(byebug) expect(response.body).to have_selector(".js-covid-timeline", text: "In Wales", visible: false)
true
```

app/views/coronavirus_landing_page/components/shared/_timeline.html.erb
andysellick: If you want to follow BEM convention to the letter, then yes, I suppose so. Problem is that there are no other styles required on this element other than to hide it, so the suggested additional class wouldn't provide any styles. I can look at renaming it to something more appropriate.

kevindew: Yes please to take a look. I found myself a bit confused at the existence of a modifier without the thing it's modifying existing. It seems to be a bit doubly confusing at the moment as the class is used on the previous version of this and has a style.

One of the things it'd have been useful for is making assertions about whether this is hidden or not. Outside of a rendered browser testing environment (JS Chrome) this is a bit tricky so we have to assert on selectors (e.g. `expect(response.body).not_to have_selector("div.covid-timeline__item:not(.covid-timeline__item--hidden).not(#nation-england)")` - which seems to look a bit odd if we do `expect(response.body).not_to have_selector("div.js-covid-timeline:not(.covid-timeline__item--hidden).not(#nation-england)")`)

spec/support/coronavirus_landing_page_steps.rb
kevindew: I see. `have_content` does consider visibility (you can actually pass in a [first argument of `:visible` or `:all`](https://www.rubydoc.info/gems/capybara/Capybara/Node/Matchers#assert_text-instance_method))

I imagine we've got a situation where `have_content` works for the Javascript version, as that actually involves a page being rendered. However your failing situation occurs without JS and the page is accessed through rack-test, in that scenario I'd imagine we have the issue where `expect(page).to have_selector(".js-covid-timeline .govuk-heading-m", text: "Changes in Wales", visible: false)` ignores the visible argument and ends up being a bit of a bogus test.

spec/controllers/coronavirus_landing_page_controller_spec.rb
leenagupte: It feels like the controller tests and non-JS feature tests are suffering from the same problems. It would be great if we could figure out a nice, readable, way to test for the absence of content that applies to both.

cc @andysellick

spec/support/coronavirus_landing_page_steps.rb
leenagupte: To clarify, there isn't anything that disables the feature. If the timeline in the content item contains "national_applicability" the radio buttons are displayed. That means that we won't need to "turn it on".

spec/controllers/coronavirus_landing_page_controller_spec.rb
leenagupte: I've added and extra test and re-done this one in a later commit.

leenagupte:
@kevindew I've had a stab at moving all of the helper methods into the presenter. The presenter now returns a megalo-object that has the timeline broken down by nation as well as the tags for each entry. I'm not sure if it's how you envisioned it, but it does work, and has reduced the amount of code in the view considerably.

There's still an issue with how the content on the page is being tested, and it would be really helpful if myself, you and @andysellick could figure out the best way to approach this.

Obviously, I haven't gone near the CSS classes, so these are still as they were.

Other things left to do:

- [ ] Get links to general guidance for devolved nations
- [ ] Move content into translation file
- [ ] Ditch this PR once we've agreed an approach, and cherry pick the good bits into a new PR with good commit messages.

app/assets/javascripts/modules/coronavirus-landing-page.js
```diff
@@ -42,5 +43,21 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
     })
   }
 
+  CoronavirusLandingPage.prototype.addTimelineCountrySelector = function () {
+    var timelineRadios = document.querySelector('.js-change-location')
+    if (timelineRadios) {
```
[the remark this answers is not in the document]
andysellick: It would be cleaner but it would add fractionally more JS for users to download. Given it's not a complicated chunk of code I'm inclined to leave it as is.

app/assets/javascripts/modules/coronavirus-landing-page.js
```diff
@@ -42,5 +43,21 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
+        e.preventDefault()
+      })
+      timelineRadios.addEventListener('change', function (e) {
+        var sections = document.querySelectorAll('.js-covid-timeline')
+        for (var i = 0; i < sections.length; i++) {
+          sections[i].style.display = 'none'
+        }
+        document.getElementById('nation-' + e.target.value).style.display = 'block'
+      })
+    }
```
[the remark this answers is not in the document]
andysellick: I think this line has changed now although we are using `document.querySelector` elsewhere.

It would be more in the spirit of the module system to scope within the module's $element but in this case that is almost the entire page content, which itself is not in keeping with the module system!

I could initialise this particular bit of the code to a specific area of the page but this feels like a bit of a sticking plaster at this point. I think it might be more sensible to revisit this whole JS file to see if we can improve it separately, I'll make a note.

app/assets/javascripts/modules/coronavirus-landing-page.js
```diff
@@ -42,5 +43,21 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
+  CoronavirusLandingPage.prototype.addTimelineCountrySelector = function () {
+    var timelineRadios = document.querySelector('.js-change-location')
+    if (timelineRadios) {
+      timelineRadios.addEventListener('submit', function (e) {
+        e.preventDefault()
+      })
+      timelineRadios.addEventListener('change', function (e) {
+        var sections = document.querySelectorAll('.js-covid-timeline')
+        for (var i = 0; i < sections.length; i++) {
+          sections[i].style.display = 'none'
```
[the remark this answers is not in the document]
andysellick: This seems to have all changed now so closing this comment.

app/views/coronavirus_landing_page/components/shared/_timeline.html.erb
andysellick: I think this has been resolved now.

kevindew:
Superseded by: https://github.com/alphagov/collections/pull/2456

PR #2454, opened 2021-06-30
Remove encoding from locale files

brucebolt:
Removes `%20` (encoding for a space) from the locale files and instead encodes the string at runtime.

This will make it easier for the translators to understand the meaning of this string when we request translations.

[Trello card](https://trello.com/c/ifbyj6pJ)


1pretz1 (approved):
Good idea Bruce! 💯

PR #2456, opened 2021-06-30
Display timeline by nation on coronavirus landing page

leenagupte:
Trello cards:
- https://trello.com/c/Bn4ekYnO
- https://trello.com/c/eE62ythU

Supersedes: #2437

Paired with: @andysellick 

# What's changed?

Adds a nation picker to the coronavirus landing page to show timeline information.

With JS disabled, this works using a standard form GET submission to show/hide the relevant section of the page. If JS is enabled, we hide the form submit button and clicking a radio button automatically switches the visibility of the relevant section.

Includes aria-live attributes so that changing the radio button announces the results of the change, and tracking attributes on the radio buttons themselves to give us some stats about usage (if cookies have been consented to).

# Why?

So that users can access specific information relevant to their location.

# Expected changes

## With timeline entries

Screenshot 2021-07-01 at 17 07 12

## When there are no changes

Screenshot 2021-07-01 at 16 09 38

# How to test locally

The changes can only be seen in a development environment.

To see the changes pass `nation` as a querystring parameter.

To see England as the default: http://localhost:3070/coronavirus?nation=foo
To see another nation selected: http://localhost:3070/coronavirus?nation=wales
To see the current live content item (without nations): http://localhost:3070/coronavirus


spec/support/coronavirus_landing_page_steps.rb
```diff
@@ -75,6 +83,15 @@ def when_i_visit_a_coronavirus_subtaxon_without_a_hub_page
     visit OTHER_SUBTAXON_PATH
   end
 
+  def when_i_click_on_wales
+    find("span", text: "Change to another nation").click
```
kevindew, line 87: Nice one for working this out - it looks like a potentially painful gotcha.

spec/support/coronavirus_landing_page_steps.rb
```diff
@@ -124,6 +141,16 @@ def then_i_can_see_the_timeline
     expect(page).to have_selector(".covid-timeline__item .gem-c-govspeak", text: "Face coverings are mandatory in shops")
   end
 
+  def then_i_can_see_the_timeline_for_england
```
kevindew, line 144: I think it'd be good to assert in here (and Wales) that it has the showing updates text too (e.g `expect(page).to have_text("Showing updates for England")`) since that seems quite a nuanced thing of JS / HTML / Ruby / CSS integrating.

kevindew (commented):
This looks awesome 🌟 

Most of my comments are pretty minor here and could be dealt with post merge if it blocks getting this shipped.

A somewhat minor visual thing is that the spacing inside the details box looks a little uneven where there is perhaps an unhelpful margin bottom on the fieldset:

Screenshot 2021-05-27 at 15 01 39

I wondered if we had an option that sorts it? if not may be just worth getting the ok from Rebecca.

andysellick:
> A somewhat minor visual thing is that the spacing inside the details box looks a little uneven where there is perhaps an unhelpful margin bottom on the fieldset:
> 
> I wondered if we had an option that sorts it? if not may be just worth getting the ok from Rebecca.

@kevindew yep, already raised a PR on the components gem to fix it. If you look at line 29 of `_timeline.html.erb` the option for zero bottom margin is already set, so this should just work once collections gets the next components gem upgrade.

app/views/coronavirus_landing_page/components/shared/_timeline.html.erb
```diff
@@ -1,23 +1,90 @@
...
-    <%= render_govspeak(timeline["summary"]) %>
-  <% end %>
+  <% if details.show_timeline_nations? %>
+    <%= form_tag coronavirus_landing_page_path,
+      method: :get,
+      class: "js-change-location" do |f| %>
+
+      <p class="govuk-body">
+        <%=t("coronavirus_landing_page.show.timeline.controls.status") %>
+        <span class="js-current-nation covid-timeline__nation"><%= @selected_nation.titleize %></span>.
```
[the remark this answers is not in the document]
andysellick: I wanted to give it an `aria-live=polite` so that screen readers would announce the text when the radio button is changed, but it turns out that some of them only announce what has changed, so for example Voiceover would announce a lot of things, including simply the word 'England' (or Scotland, etc.), instead of the full text (which would be more helpful) of 'Showing updates for England'. This would get lost inside the other things it was saying, including the word England as the thing that had been selected - it sounded like a weird repetition.

So I've put the aria-live attribute onto the actual content instead, and there's visually hidden headings so the first things users should hear is the full 'Updates for England' when they change the radio button.

leenagupte:
@kevindew Thanks for your review. I've actioned all of your comments and fixed up. I've also rebased against main to pick up the margin changes to the button component.

config/locales/en/coronavirus_landing_page.yml
```diff
@@ -2,3 +2,23 @@
...
+          heading: Changes in %{nation}
+        no_updates:
+          heading: No updates
+          body: There are no recent updates for %{nation}.
+          northern_ireland:
+            additional_country_guidance_html: "Check the latest guidance on <a href='https://www.nidirect.gov.uk'>nidirect</a>"
+          scotland:
+            additional_country_guidance_html: "Check the latest guidance on <a href='https://www.gov.scot'>gov.scot</a>"
+          wales:
+            additional_country_guidance_html: "Check the latest guidance on <a href='https://gov.wales'>gov.wales</a>"
```
kevindew, line 24: Minor: It'd be good to have a period on the end of each of these.

app/views/coronavirus_landing_page/components/shared/_timeline.html.erb
```diff
@@ -1,23 +1,89 @@
...
-    <%= render_govspeak(timeline["summary"]) %>
-  <% end %>
+  <% if details.show_timeline_nations? %>
+    <%= form_tag coronavirus_landing_page_path,
+      method: :get,
+      class: "js-change-location" do |f| %>
+
+      <p class="govuk-body">
+        <%=t("coronavirus_landing_page.show.timeline.controls.status") %>
+        <span class="js-current-nation covid-timeline__nation"><%= details.selected_nation.titleize %></span>.
```
kevindew, line 15: I think [the comment](https://github.com/alphagov/collections/pull/2456#discussion_r663026960) I made on this last time got lost by the thread going on a tangent.

> Rather than having an incomplete sentence in the translation file it'd be good to move the HTML into there and call it as HTML. e.g:
> 
> ```erb
> <%=t("coronavirus_landing_page.show.timeline.controls.status_html", nation: @selected_nation.titleize) %>
> ```
> 
> ```yaml
>             status_html: Showing updates for <span class="js-current-nation covid-timeline__nation">%{nation}</span>.
> ```

kevindew (commented):
Looks awesome. Just a couple of minor things flagged that slipped through in earlier comments I made.

I've put a note on the card about the back button issue but I don't think we need to wait for that to merge this.

leenagupte:
@kevindew Thanks for your re-review. I've addressed the two comments that were missed the first time around and fixed-up.

kevindew (approved):
Looks great, thanks ⭐ 

You have a typo in the last commit message but I can't see you needing another review.

PR #2465, opened 2021-07-06
Remove the details element from the nation picker

leenagupte:
Trello cards:
- https://trello.com/c/PSKxvsS3
- https://trello.com/c/eE62ythU

# What's changed?

Remove the details component that show/hides the radio button and button components in the nation picker in the timeline.

# Why?

The original plan was to change the "Change to another nation" text to instead list the other nations that the user hadn't picked, e.g. "Change to England, Scotland, or Wales" etc.

However, the nation names in the link would need to be dynamically updated, which could be confusing without testing, and also looks odd when the details tab is opened.

There is a concern that leaving the text as it is will mean that user's won't intuitively understand what is being hidden in the detail. So the decision was made to remove the details component completely.

We can add back the details component later on when we're more certain it won't cause users an issue.

# Things to do
- [x] Sort out margin between radio buttons / button and timeline
 
With Javascript enabled:

Screenshot 2021-07-06 at 13 17 59

With Javascript disabled:

Screenshot 2021-07-06 at 13 17 51

### DIFF
--- a/app/assets/javascripts/modules/coronavirus-landing-page.js
+++ b/app/assets/javascripts/modules/coronavirus-landing-page.js
@@ -51,9 +51,7 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
       })
       timelineRadios.addEventListener('change', function (e) {
         var sections = document.querySelectorAll('.js-covid-timeline')
-        var currentNation = document.querySelector('.js-current-nation')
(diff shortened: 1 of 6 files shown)


kevindew (approved):
Thanks - looks good 👍 , such short lived code 😢

PR #2472, opened 2021-07-09
Handle back button navigation for Coronavirus timeline

kevindew:

Trello: https://trello.com/c/Bn4ekYnO

---

Before:

Jul-09-2021 19-10-12-before

---

After:

Jul-09-2021 19-10-12-after

---

We learnt that there was a UX bug for the Coronavirus timeline if a user
navigated to a back using the browsers back/forward controls.

The experience for a user would be:

- select a constituent country other than England, and see
  the results for that country
- click a link on the page
- click the browser back button
- see that your previously selected country was still selected but that
  the results were for England.

The approach I've taken to resolve this is to make the event handler
no longer understand state from the event and instead read it from the app.
This allows the handler to be called outside of a change event occurring.
I've then updated the code set that the event handler is called during
initialisation which can set the state.

This approach works for Browsers outside the Chrome family, as the form
is set with the restored state before JavaScript initialises. Chrome
however differs and initially renders the form state as per the source
HTML. Once the page reaches a "load" state Chrome then adjusts the form
to restore the previous state. In order to resolve this I've had to put
a listener in for the "pageshow" event that occurs when you navigate to
a page.

A rejected, simpler, approach that I tried was to set the form to
"autocomplete=off". This stops the browser restoring the previous form
state when you navigate by the back button. I found that this had an
undesirable behaviour in Safari where clicking back would set the
country back to England but leave the results as the previous country.
This seems to be because Safari keeps a page existing in memory
initially and doesn't actually reload it on a back button press, however
it will try acknowledge the form autocomplete, while leaving the DOM as
it was previously.

### DIFF
--- a/app/assets/javascripts/modules/coronavirus-landing-page.js
+++ b/app/assets/javascripts/modules/coronavirus-landing-page.js
@@ -45,23 +45,34 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
 
   CoronavirusLandingPage.prototype.addTimelineCountrySelector = function () {
     var timelineRadios = document.querySelector('.js-change-location')
-    if (timelineRadios) {
-      timelineRadios.addEventListener('submit', function (e) {
-        e.preventDefault()
-      })
-      timelineRadios.addEventListener('change', function (e) {
-        var sections = document.querySelectorAll('.js-covid-timeline')
-        var nation = e.target.value
+    if (!timelineRadios) { return }
 
-        for (var i = 0; i < sections.length; i++) {
-          var show = sections[i].id === 'nation-' + nation
-          if (show) {
-            sections[i].classList.remove('covid-timeline__wrapper--hidden')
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/modules/coronavirus-landing-page.js
```diff
@@ -45,23 +45,34 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
-        }
-      })
+    timelineRadios.addEventListener('submit', function (e) {
+      e.preventDefault()
+    })
+
+    timelineRadios.addEventListener('change', this.timelineCountryChangeHandler)
+
+    // Set initial state based of timeline country based on checked inputs.
+    // In Chrome this can set the wrong
```
andysellick, line 57: You might want to tweak this comment?

andysellick (approved):
Seems good 👍

PR #2478, opened 2021-07-14
Allow links in covid landing page header

andysellick:

## What

- the header section of this page has a blue background and white text, so adding links to the govspeak component wasn't working, they were appearing as the default link style of blue, which was unreadable
- this change to use the new govspeak inverse option allows links to be added that will be automatically styled correctly
- so we can remove the styles that previously made it white also

Note that this PR doesn't actually add any links - it only allows links to be added (which is handled through content). If you want to test that it's working you'll need to manipulate the dom to add a link into the page.

## Visual changes

Before | After
------ | -------
Screenshot 2021-07-14 at 10 21 44 | Screenshot 2021-07-14 at 10 21 32

Trello card: https://trello.com/c/o7kekKgI/471-not-to-do-until-we-have-approval-link-styling-within-coronavirus-header-markdown

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -61,8 +61,8 @@ module ApplicationHelper
     end
   end
 
-  def render_govspeak(content)
-    render "govuk_publishing_components/components/govspeak" do
+  def render_govspeak(content, inverse: false)
+    render "govuk_publishing_components/components/govspeak", inverse: inverse do
       raw(Govspeak::Document.new(content, sanitize: false).to_html)
     end
   end
--- a/app/assets/stylesheets/views/_covid.scss
+++ b/app/assets/stylesheets/views/_covid.scss
@@ -75,8 +75,6 @@ $c19-landing-page-header-background: govuk-colour("blue");
   }
 
   .gem-c-govspeak {
-    color: govuk-colour("white");
-
     li {
       margin: 0 0 govuk-spacing(2);
     }
--- a/app/views/coronavirus_landing_page/components/landing_page/_page_header.html.erb
+++ b/app/views/coronavirus_landing_page/components/landing_page/_page_header.html.erb
@@ -40,7 +40,7 @@
           } %>
 
           <div class="covid__page-header-content">
-            <%= render_govspeak(header["intro"]) %>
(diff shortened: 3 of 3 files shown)


app/helpers/application_helper.rb
```diff
@@ -61,8 +61,8 @@ def t_fallback(key, options = {})
     end
   end
 
-  def render_govspeak(content)
-    render "govuk_publishing_components/components/govspeak" do
+  def render_govspeak(content, inverse = nil)
```
kevindew, line 64:
```suggestion
  def render_govspeak(content, inverse: false)
```

Since this is a boolean it's probably better to set this as false, setting it as a keyword argument will make the code that calls it a bit easier to understand too as to what the true means.

kevindew (commented):
Awesome, nice one getting the change in govuk_publishing_components. I hadn't seen that go through.

Just a suggestion on the method signature.

app/helpers/application_helper.rb
andysellick: Maybe you know the answer to this - I tried this syntax originally (`inverse:` instead of `inverse =`) and I got a massive page error of `wrong number of arguments (given 2, expected 1)`, so I went with `=` instead. Any idea what the difference is?

andysellick: I've changed it to `false` though as that made sense.

andysellick: Oh. Wait. Changing it to false causes a linting error. Weird that `inverse = false` causes a lint error but `inverse = nil` doesn't. Or maybe I'm going mad.

kevindew: Sure, the linter doesn't realise it's meant to be a boolean when you set it to nil so it doesn't complain.

I think the issue you've got is that once you change to `def render_govspeak(content, inverse: false)`

Then you'll also need to change `<%= render_govspeak(header["intro"], true) %>` to `<%= render_govspeak(header["intro"], inverse: true) %>` - sorry I should have pointed that out. This makes this render_govspeak call easier to understand since the boolean has an associated key.

andysellick: Ah, of course. Thanks, have changed.

kevindew (approved):
Nice one 👍

PR #2484, opened 2021-07-20
Remove jq from feed

JonathanHallam:
## What
Remove jQuery from the feed file.

## Why

## Trello 

https://trello.com/c/ybtss04a/484-remove-jquery-from-collections-modules-feedsjs


JonathanHallam:
Superseded by https://github.com/alphagov/collections/pull/2487

PR #2485, opened 2021-07-21
Add event listeners to feed

JonathanHallam:
We have a suspicion that the feed is never actually used. This adds an event listener to it so that we can know for sure.

This is part of a wider effort to remove jQuery from our stack

Trello - https://trello.com/c/ybtss04a/484-remove-jquery-from-collections-modules-feedsjs

### DIFF
--- a/app/assets/javascripts/modules/feeds.js
+++ b/app/assets/javascripts/modules/feeds.js
@@ -7,9 +7,15 @@
 
   window.GOVUK.feeds = {
     init: function () {
+      if (window.GOVUK.analytics && window.GOVUK.analytics.trackEvent) {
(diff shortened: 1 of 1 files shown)


andysellick (commented):
Looks good. Unfortunately I think we should have a test file for this, to check that the tracking events get fired properly - at the moment it's reliant on the analytics code already existing, and if it isn't, this isn't going to work.

If you look at the JS test for the details component, you should be able to copy some of the stuff for testing this - basically you create a spy on the analytics code, and then the spy tells you if the code was called or not when certain things happen. Happy to go through it.

JonathanHallam:
Superseded by https://github.com/alphagov/collections/pull/2487

PR #2491, opened 2021-07-26
Remove jQuery from toggle attribute

JonathanHallam:
## What

Remove jQuery from the toggle-attribute file.

## Why

## Trello

https://trello.com/c/RBRTymH1/485-remove-jquery-from-collections-modules-toggle-attributejs

### DIFF
--- a/app/assets/javascripts/modules/toggle-attribute.js
+++ b/app/assets/javascripts/modules/toggle-attribute.js
@@ -2,19 +2,19 @@ window.GOVUK = window.GOVUK || {}
 window.GOVUK.Modules = window.GOVUK.Modules || {};
 
 (function (Modules) {
-  'use strict'
-
-  Modules.ToggleAttribute = function () {
-    this.start = function ($element) {
-      $element.find('[data-toggle-attribute]').click(function (event) {
-        var clicked = event.target
-        var toggleAttribute = clicked.getAttribute('data-toggle-attribute')
-        var current = clicked.getAttribute(toggleAttribute)
-        var closedText = clicked.getAttribute('data-when-closed-text')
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/modules/toggle-attribute.js
```diff
@@ -2,19 +2,19 @@ window.GOVUK = window.GOVUK || {}
...
-
-        clicked.setAttribute(toggleAttribute, current === closedText ? openText : closedText)
-      })
-    }
+  function ToggleAttribute ($module) {
+    this.$module = $module
+  }
+  ToggleAttribute.prototype.init = function ($element) {
+    $element = this.$module.querySelector('[data-toggle-attribute]')
+    $element.addEventListener('click', function (event) {
```
gclssvglx, line 10: Based on @kevindew's explanation [on this PR](https://github.com/alphagov/collections/pull/2493#pullrequestreview-716050444), it might be more idiomatic to do something like in order to keep the setup code in the constructor separate from the `init` code...

```
'use strict'

Modules.ToggleAttribute = function (element) {
  this.toggleAttribute = element.querySelector('[data-toggle-attribute]')
}

Modules.ToggleAttribute.prototype.init = function () {
  this.toggleAttribute.addEventListener('click', function (event) {
    // ...
  })
}
```

gclssvglx (commented):
This is looking nice, just one suggestion/comment...

app/assets/javascripts/modules/toggle-attribute.js
JonathanHallam: Done :) thanks

gclssvglx (approved):
Great job Jon 👍

PR #2493, opened 2021-07-27
Remove jQ from track link file

JonathanHallam:
## What

Remove jQuery from the track-link file.

## Why

## Trello

https://trello.com/c/T3iJDYst/486-remove-jquery-from-collections-modules-track-linksjs

### DIFF
--- a/app/assets/javascripts/modules/track-links.js
+++ b/app/assets/javascripts/modules/track-links.js
@@ -4,27 +4,26 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
 (function (Modules) {
   'use strict'
 
-  Modules.TrackLinks = function () {
-    this.start = function (element) {
-      var category = element[0].getAttribute('data-track-category')
-      var action = element[0].getAttribute('data-track-action')
-      var links = element[0].getElementsByTagName('a')
+  Modules.TrackLinks = function (element) {
+    this.category = element.getAttribute('data-track-category')
+    this.action = element.getAttribute('data-track-action')
+    this.links = element.getElementsByTagName('a')
+  }
 
-      if (!category) return
+  Modules.TrackLinks.prototype.init = function () {
+    if (!this.category) return
 
-      for (var i = 0; i < links.length; i++) {
-        var link = links[i]
-        link.addEventListener('click', function (e) {
-          var options = {
-            transport: 'beacon',
-            label: e.target.getAttribute('href')
-          }
+    for (var i = 0; i < this.links.length; i++) {
+      var link = this.links[i]
+      link.addEventListener('click', function (e) {
+        var options = {
+          transport: 'beacon',
+          label: e.target.getAttribute('href')
+        }
+        if (!this.action) this.action = e.target.innerText
 
-          if (!action) action = e.target.innerText
-
-          GOVUK.analytics.trackEvent(category, action, options)
-        })
-      }
+        GOVUK.analytics.trackEvent(this.category, this.action, options)
+      }.bind(this))
     }
   }
 })(window.GOVUK.Modules)
--- a/spec/javascripts/modules/track-links_spec.js
+++ b/spec/javascripts/modules/track-links_spec.js
@@ -38,8 +38,8 @@ describe('Track links', function () {
       e.preventDefault()
     })
 
-    var tracker = new GOVUK.Modules.TrackLinks()
-    tracker.start([element])
+    var tracker = new GOVUK.Modules.TrackLinks(element)
+    tracker.init()
   })
 
   afterEach(function () {


app/assets/javascripts/modules/track-links.js
```diff
@@ -5,7 +5,7 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
   'use strict'
 
   Modules.TrackLinks = function () {
-    this.start = function (element) {
+    this.init = function (element) {
       var category = element[0].getAttribute('data-track-category')
       var action = element[0].getAttribute('data-track-action')
       var links = element[0].getElementsByTagName('a')
```
gclssvglx, line 11: I think you can change `element[0]` to `element` in here as `init` takes a vanilla JS object and not a JQuery one - in GOV.UK JS-land. Worth checking with a FE dev.

app/assets/javascripts/modules/track-links.js
```diff
@@ -5,7 +5,7 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
   'use strict'
 
   Modules.TrackLinks = function () {
-    this.start = function (element) {
+    this.init = function (element) {
```
andysellick, line 8: I'm not certain, but I think changing this to use `init` instead of `start` involves slightly more than renaming the function. If you take a look at https://github.com/alphagov/govuk_publishing_components/blob/master/app/assets/javascripts/govuk_publishing_components/components/details.js#L8 you'll see that this component receives the element it acts on (`$module`) in the initial function, which is then set as a variable `this.$module`, which is then accessed directly from the `init` function rather than being passed to it. There's also a slight difference at the end of the file in how the whole thing gets added to the `window.GOVUK.Modules` object.

I'm basing this on https://github.com/alphagov/govuk_publishing_components/pull/2095. @alex-ju will be able to provide more information about why it should be like this, but I think he's not in today.

spec/javascripts/modules/track-links_spec.js
```diff
@@ -39,7 +39,7 @@ describe('Track links', function () {
     })
 
     var tracker = new GOVUK.Modules.TrackLinks()
-    tracker.start([element])
+    tracker.init([element])
```
andysellick: There's something odd here, which isn't part of your change but worth raising - this test passes a single element into the module by first wrapping it in an array. The module seems to expect this, acting on the zeroth element of that array (see lines 9 - 11). This is particularly strange as it doesn't do anything if the passes array contains more than one item.

I'm not sure what to suggest here, but it might be worth adjusting the code to see what impact it has on the tests - I can't see a reason why the module is structured like this.

app/assets/javascripts/modules/track-links.js
andysellick: Yeah, this is an odd one - see my other comment about line 42 of the spec file. Ah yes, but that's probably the reason - this file was converted to vanilla JS, but the test was also updated, and confusion. Yes, try what @gclssvglx is suggesting, but modify the test to pass `element` instead of `[element]`.

spec/javascripts/modules/track-links_spec.js
JonathanHallam: no dice I'm afraid. I'll dig into this a bit more and see what I can do!

spec/javascripts/modules/track-links_spec.js
```diff
@@ -39,7 +39,7 @@ describe('Track links', function () {
     })
 
     var tracker = new GOVUK.Modules.TrackLinks()
-    tracker.start([element])
+    tracker.init(element)
```
kevindew, line 42: With the init function this code will look like this:

```js
    var tracker = new GOVUK.Modules.TrackLinks(element)
    tracker.init()
```

app/assets/javascripts/modules/track-links.js
```diff
@@ -5,10 +5,10 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
   'use strict'
 
   Modules.TrackLinks = function () {
```
kevindew, line 7: With the conversion to init this will be broken up into two functions

This is the constructor:

```js
Module.TrackLinks = function (element) {
  // set this variables so probably:
  // this.category = element.getAttribute('data-track-category')
  // this.action = element.getAttribute('data-track-action')
  // this.links = element.getElementsByTagName('a')
}
```

Then you'll have the init function

```js
Modules.TrackLinks.prototype.init = function () { 
  // do the actual work e.g.
  // if (!this.category) return

  // for (var i = 0; i < this.links.length; i++) {
  // etc
}
```

kevindew (commented):
I've just had a quick go at trying to explain the init pattern.

app/assets/javascripts/modules/track-links.js
alex-ju: Kevin covered the difference between the `start` API and the `init` API, but it seems like you've already addressed this.

alex-ju (approved):
Thanks for working on this! I see the comments were addressed, the final code looks great and follows convention.

PR #2495, opened 2021-07-28
Remove jquery from organisation-list-filter

leenagupte:
Trello: https://trello.com/c/nXatVRcM

# What's changed?

Removes jQuery from [organisation-list-filter.js](https://github.com/alphagov/collections/blob/a223e2fccb456b62358d08fd40aeba4e54ab62c5/app/assets/javascripts/organisation-list-filter.js) and make it a conventional GOV.UK module. 

Also converts [start methods into init methods](https://trello.com/c/02ot7gbP/842-convert-finder-frontend-module-initialisers-to-not-use-jquery). 

# Why?

We're removing jQuery from GOV.UK and this file is a little unusual.

# Testing

Used on: https://www.gov.uk/government/organisations

Paired with: @andysellick 


app/assets/javascripts/modules/organisation-list-filter.js
```diff
@@ -0,0 +1,128 @@
...
+      this.filterTimeout = setTimeout(function () {
+        this.$module.filterList(searchTerm)
+      }.bind(this), 200)
+    }.bind(this))
+  }
+
+  FilterOrganisations.prototype.filterList = function (searchTerm) {
+    var itemsToFilter = this.$module.querySelectorAll('[data-filter="item"]')
+    var listsToFilter = this.$module.querySelectorAll('[data-filter="list"]')
+    for (var i = itemsToFilter.length - 1; i >= 0; i--) {
```
owenatgov, line 39: How come we're counting down here? The typical convention I've observed in our JS is to count up when iterating. I can't see anything in the loop itself that would benefit from reverse the count. Is it a bother to swap this around to keep with convention?

```suggestion
    for (var i = 0; i <= itemsToFilter.length; i++) {
```

app/assets/javascripts/modules/organisation-list-filter.js
```diff
@@ -0,0 +1,128 @@
...
+(function (Modules) {
+  function FilterOrganisations ($module) {
+    this.$module = $module
+    this.filterTimeout = null
+    this.form = this.$module.querySelector('[data-filter="form"]')
+    this.orgResults = this.$module.querySelector('#organisations_search_results')
+  }
+
+  FilterOrganisations.prototype.init = function () {
+    this.$module.filterList = this.filterList.bind(this)
```
owenatgov, line 15: This may simply be my lack of knowhow around binding in js so sorry if I'm regurgitating something you've already run into: Why do you need to create a separate call for this function? It's only being used on line 31 so could you not just use something like...

```
this.filterList(searchTerm).bind(this)
```

...on line 31 in place of this variable call? There's already loads of binding going on in that timeout and the event listener itself so the bind might not be needed.

spec/javascripts/modules/organisation-list-filter_spec.js
```diff
@@ -59,28 +59,30 @@ describe('organisation-list-filter.js', function () {
     '</div>'
 
   beforeAll(function () {
-    $(document.body).append(stubStyling)
-    $(document.body).append(form)
-    $(document.body).append(organisations)
-    GOVUK.filter.init()
+    var wrapper = $('<div>').addClass('wrapper')
```
owenatgov, line 62: Hissssssssssssss there's still jQuery here! This isn't as big a problem since Jasmine loads it's own JQuery installation and it doesn't reach users so this isn't a blocker for approval. Tidying up here would make a it a little bit nicer though :)

owenatgov (commented):
Some little nitpicks. Looking strong 💪

spec/javascripts/modules/organisation-list-filter_spec.js
andysellick: There's a lot of jQuery in the tests generally. We've decided to leave it for now as it doesn't cause a problem, generally helps when writing tests, and would likely more than double the effort required to remove jQuery.

Having said that, there are lots of gotchas involved in having JS code and jQuery tests, but that's a separate conversation.

app/assets/javascripts/modules/organisation-list-filter.js
leenagupte: You're right this is odd. We originally did a reverse loop because we thought we were going to be removing things from the array, but in the end all we needed was to add a class.

I'll change the loop and fix up.

leenagupte:
@owenatgov Thanks for your review. I've made the changes to the for loop and fixed up, and responded to the other comments.

app/assets/javascripts/modules/organisation-list-filter.js
owenatgov: How bizarre 🤔  There might be something going on with the function being bound before it can be called. If it works it works, I'm not super precious about this. Feel free to resolve.

owenatgov (approved):
💯

PR #2498, opened 2021-07-30
Hide high_profile_groups heading for organisations without live groups

edwardkerry:
Organisation pages display High Profile Groups that have  a status of `live`, `exempt` or `transitioning`. Other statuses are filtered out, and could leave an empty array. This can result in the heading rendering with no High Profile Groups to display.

This adds a check to make sure there is at least one High Profile Group before rendering the section.

#Before
Screenshot 2021-07-30 at 12 07 52

#After
Screenshot 2021-07-30 at 12 09 16


app/views/organisations/show.html.erb
```diff
@@ -33,7 +33,7 @@
     <% unless @organisation.is_promotional_org? %>
       <%= render partial: 'freedom_of_information' %>
     <% end %>
-    <% if @show.high_profile_groups[:items] && !@organisation.is_promotional_org? %>
+    <% if @show.high_profile_groups[:items]&.any? && !@organisation.is_promotional_org? %>
```
1pretz1, line 36: We could used `present?` here as it takes into account `nil` so you don't need the `&`, which is slightly nicer to read.

1pretz1 (approved):
Nice work Ed! 🐛  I've just left one minor recommendation but nothing blocking.

app/views/organisations/show.html.erb
edwardkerry: ah very good, thank you!

1pretz1 (approved):
🍏

PR #2501, opened 2021-08-06
Revert "Bump govuk_publishing_components from 24.21.1 to 25.1.0"

MuriloDalRi:
Reverts alphagov/collections#2500


alex-ju (approved):
[Reason for reverting it](https://govuk.zendesk.com/agent/tickets/4656102)

PR #2509, opened 2021-08-10
Send AB test header response back for all variants

sihugh:
We need the variant header to be sent for every variant otherwise Fastly doesn't know how to cache them properly.


injms (approved):
Looks grand to me 👍

PR #2513, opened 2021-08-18
Adjust behaviour of ToggleAttribute

kevindew:
Trello: https://trello.com/c/71ehy3NG/562-fix-action-value-for-ga-event-for-accordion-opens-on-coronavirus-hubs

This module wasn't behaving correctly for a couple of reasons:

1) It wasn't finding multiple elements with [data-toggle-attribute]
   because it was using querySelector (as opposed to querySelectorAll)
2) The elements with [data-toggle-attribute] were being removed from the
   DOM and replaced with button elements [1], which meant any elements
   set on them lost their event handlers.

To resolve these issues I have placed a click event handler on the
module itself. Whenever an element within the module is clicked this
event will be triggered, it then resolves whether that element has a
[data-toggle-attribute] attribute or a parent with it to then execute
the code.

This change has been done to fix the problem that our accordions are
sending accordionClosed events each time they are opened or closed -
using this toggleAttributes module will change that attribute when they
are clicked before a later event handler to send the event.

This process to track accordion clicks is rather unusual and fragile and
should really be re-considered, the ideal would probably be to have a
custom event triggered from the accordion that can be listened to, or to
bake in analytic tracking into the accordion itself.

[1]: https://github.com/alphagov/govuk_publishing_components/blob/fe80e8b591bcec1f1855cb34d8601137a36f782c/app/assets/javascripts/govuk_publishing_components/components/accordion.js#L144-L153

### DIFF
--- a/app/assets/javascripts/modules/toggle-attribute.js
+++ b/app/assets/javascripts/modules/toggle-attribute.js
@@ -4,17 +4,27 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
 (function (Modules) {
   function ToggleAttribute ($module) {
     this.$module = $module
-    this.toggleAttribute = $module.querySelector('[data-toggle-attribute]')
   }
-  ToggleAttribute.prototype.init = function ($element) {
-    this.toggleAttribute.addEventListener('click', function (event) {
-      var clicked = event.target
-      var toggleAttribute = clicked.getAttribute('data-toggle-attribute')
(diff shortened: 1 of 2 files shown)


alex-ju (approved):
> This process to track accordion clicks is rather unusual and fragile and
should really be re-considered, the ideal would probably be to have a
custom event triggered from the accordion that can be listened to, or to
bake in analytic tracking into the accordion itself.

Totally agree with this. Great work fixing it for now.

PR #2548, opened 2021-10-01
C19 header change

andysellick:

## What
Update landing page at https://www.gov.uk/coronavirus to include new design for header.

~~**NOTE** the design includes a new blue arrow for the action link at the bottom of the header. This will remain yellow until the [next version of the components gem](https://github.com/alphagov/govuk_publishing_components/pull/2332) is deployed to `static` (which is why it's wrong in the screenshots below).~~ 
This has now been deployed.

## Why
New designs.

## Visual changes

Before | After
------- | ------
Screenshot 2021-10-01 at 11 46 14 | Screenshot 2021-10-04 at 14 58 46

After (mobile)

Screenshot 2021-10-04 at 14 59 31

Trello card: https://trello.com/c/KeJVLjie/649-collections-frontend-implementation-of-the-new-coronavirus-header-design


app/views/coronavirus_landing_page/components/landing_page/_page_header.html.erb
```diff
@@ -16,26 +16,20 @@
...
     </div>
-    <div class="govuk-grid-row">
-      <div class="govuk-grid-column-column-full covid__page-header-title-section-break">
-        <hr class="govuk-section-break govuk-section-break--visible covid__page-header-title-section-break--white">
-      </div>
-    </div>
   </div>
 
-  <div class="covid__page-header-content-section">
+  <div class="covid__page-header-content-section covid__page-guidance">
```
kevindew, line 21: As an aside (as I know this isn't introduced here) what is a covid__page-guidance? Is it just used as a vessel for styles or do we consider this to be a "page-guidance" element of covid block?

I'm wondering if this is perhaps an isolation issue because this should probably be something scoped within `covid__page-header`  but perhaps the horse bolted long ago on this page having BEM isolation?

kevindew (approved):
Looks great and worked great for me. Excited to see it ship 🚢

app/views/coronavirus_landing_page/components/landing_page/_page_header.html.erb
andysellick: `covid__page-guidance` is a class that already existed that was being used on the hub pages. It matched exactly what the design called for so I reused it.

To be honest a lot of this CSS scares me so I tried to do the minimum change possible without disrupting anything. I'm not sure the horse has bolted but it's definitely restless and eyeing the distant fields.

PR #2576, opened 2021-10-19
Iterate stats section

leenagupte:
Trello: https://trello.com/c/EM3HJ1RV

# What's changed?

We want to update the statistics section of /coronavirus to represent the new design captured in https://trello.com/c/5PdpKCiC/639-landing-page-re-jig-updating-the-statistics-section

# Why?

When revealed the wider public will have a much better picture of Coronavirus content.

# Expected changes
## Before
screenshot-www gov uk-2021 11 02-14_04_26

## After
Screenshot 2021-11-02 at 13 43 05


app/views/coronavirus_landing_page/components/landing_page/_statistics_section.html.erb
```diff
@@ -1,80 +1,151 @@
...
+        <p class="govuk-body">
+          <%= topic_section.dig("tests", "daily_heading") %><br/>
+          <span class="covid-statistics__emphasis"><%= number_with_delimiter(statistics.new_positive_tests) %></span>
+        </p>
+        <p class="govuk-body">
+          <%= topic_section.dig("tests", "weekly_heading") %><br/>
+          <span class="covid-statistics__emphasis"><%= number_with_delimiter(statistics.current_week_positive_tests) %></span>
+          <span class="govuk-tag govuk-tag--yellow covid-statistics__tag">
+            <% tests_number = number_with_delimiter(statistics.current_week_positive_tests_change_number).to_i %>
+            <%= render partial: 'coronavirus_landing_page/components/landing_page/arrow', locals: { number: tests_number } %>
```
leenagupte, line 41: `statistics.current_week_positive_tests_change_number` is already a number, so `to_i` doesn't need to be applied to it. The value as is can be passed directly to the "arrow" partial:
```suggestion
            <%= render partial: 'coronavirus_landing_page/components/landing_page/arrow', locals: { number: statistics.current_week_positive_tests_change_number } %>
```

leenagupte: The same applies for "hospital admissions".

andysellick: Thanks, have updated. It's a bit less contorted now.

Interestingly this highlighted a bug - I was using `number_with_delimiter` to convert the number from e.g. `1234` to `1,234` and then applying rails `.to_i` method to convert it back into a number. Unfortunately that doesn't work properly - instead of converting `1,234` to `1234` it converts it to `1` (or `1.0` if using `.to_f`). Bizarre.

leenagupte:
@andysellick If the [statistics partial ](https://github.com/alphagov/collections/blob/0b5611d5b49f87545b3299fcb82fe32bf375d342/app/views/coronavirus_landing_page/components/landing_page/_statistic.html.erb) is no longer being used, it should be removed as well.

app/services/fetch_coronavirus_statistics_service.rb
```diff
@@ -68,21 +80,135 @@ def parse_response(response)
...
-      parsed[:new_positive_tests_date] = Date.parse(latest_tests["date"])
+    total_current_week_admissions = current_week_admissions(weeks_with_admissions)
+    total_previous_week_admissions = previous_week_admissions(weeks_with_admissions)
+
+    return {} unless total_current_week_admissions && total_previous_week_admissions
+
+    {
+      current_week_hospital_admissions: total_current_week_admissions,
+      current_week_hospital_admissions_change_number: total_current_week_admissions - total_previous_week_admissions,
+      current_week_hospital_admissions_change_percentage: percentage_change(total_current_week_admissions, total_previous_week_admissions),
```
andysellick, line 140: How are these percentage changes calculated? Here's the current numbers on my local:

Screenshot 2021-11-03 at 10 13 02

Doing the maths myself, 10.4% of 280,479 is `29169.816`, and 2.5% of 7,197 is `179.925`, which doesn't match the numbers above, although my maths isn't great and maybe this is being calculated in a different way?

leenagupte:
```
((previous weeks total - current week total) / previous weeks total) * 100
```

leenagupte: Thou shalt find that the percentages on coronavirus.data.gov.uk matcheth mine 😎

andysellick: Oh I see. So it's calculating it based on the difference between this weeks total (shown) and last weeks total (not shown). That makes sense, thanks.

app/services/fetch_coronavirus_statistics_service.rb
```diff
@@ -113,12 +114,34 @@ def latest_admissions(data)
...
+  def daily_tests(latest_tests)
     {
       new_positive_tests: latest_tests["newPositiveTests"],
       new_positive_tests_date: Date.parse(latest_tests["date"]),
     }
   end
+
+  def weekly_tests(latest_tests, data)
+    weeks_with_tests = []
+    data.each do |day_cases|
```
JonathanHallam, line 177: I feel like there should be a way to use `select` here to clean it up a little bit but I don't know of a way to do what you want without iterating twice which we obviously don't want! Just doing a first pass through at the moment to understand what's going on :)

app/views/coronavirus_landing_page/components/landing_page/_statistics_section.html.erb
```diff
@@ -13,6 +13,8 @@
         <p><%= topic_section.dig("new_positive_tests", "title") %></p>
         <p>Daily</p>
         <p><%= number_with_delimiter(statistics.new_positive_tests) %></p>
+        <p>Last 7 days</p>
```
JonathanHallam: This means that the heading will show up even if there's data missing and `current_week_positive_tests` doesn't exist right? Is that expected behaviour or should we hide the heading in an `if` statement? (extrapolate comment to other cases where it applies)

leenagupte: You're right. I think I'm missing a guard clause before the merge, so this:

```ruby
daily_tests = daily_tests(latest_tests)
weekly_tests = weekly_tests(latest_tests, data)

daily_tests.merge!(weekly_tests)
```

Should be:

```ruby
daily_tests = daily_tests(latest_tests)
weekly_tests = weekly_tests(latest_tests, data)

return unless daily_tests && weekly_tests

daily_tests.merge!(weekly_tests)
```

leenagupte: I think the same guard is missing for hospital admissions as well.

JonathanHallam: Sounds good :)

leenagupte: Thank you for being so thorough with this review. It's much appreciated.

app/services/fetch_coronavirus_statistics_service.rb
```diff
@@ -135,13 +137,33 @@ def weekly_tests(latest_tests, data)
...
+  end
+
+  def previous_week_positive_tests(weeks_with_tests)
+    previous_week_cases = weeks_with_tests[7..13].compact
+    return unless previous_week_cases.size == 7
+
+    previous_week_cases.inject(:+)
+  end
+
+  def percentage_change(current_total, previous_total)
```
JonathanHallam, line 207: Maths is good, nice one :) 
Amazed there's not a gem or rails library for this though! (I checked)

leenagupte: So did I!

app/services/fetch_coronavirus_statistics_service.rb
```diff
@@ -105,15 +108,53 @@ def latest_vaccinations(data)
...
+  def daily_admissions(latest_admissions)
     {
       hospital_admissions: latest_admissions["hospitalAdmissions"],
       hospital_admissions_date: Date.parse(latest_admissions["date"]),
     }
   end
 
+  def weekly_admissions(latest_admissions, data)
+    weeks_with_admissions = []
+    data.each do |day_admissions|
```
JonathanHallam, line 128: Can you expand the commit message slightly to say that `nil` would be created both by a lack of data or a `false` on the line below and we want one but not the other please (discussed offline (: )

app/services/fetch_coronavirus_statistics_service.rb
```diff
@@ -105,15 +108,53 @@ def latest_vaccinations(data)
...
+
+  def daily_admissions(latest_admissions)
     {
       hospital_admissions: latest_admissions["hospitalAdmissions"],
       hospital_admissions_date: Date.parse(latest_admissions["date"]),
     }
   end
 
+  def weekly_admissions(latest_admissions, data)
+    weeks_with_admissions = []
```
JonathanHallam, line 127: not a blocking comment but I'm not sure `weeks_with_admissions` really sums up what this is doing, it's more like `daily_admissions_list` or `daily_admissions`

JonathanHallam: which makes lines 132 and 133 read a little more clearly, but again, super minor

app/services/fetch_coronavirus_statistics_service.rb
JonathanHallam: Might write a gem as a side project, open source community and all that (love off topic review comments that send emails (sorry for the spam (also love brackets in brackets (i hope I don't embarrass myself by not closing them all))))

app/services/fetch_coronavirus_statistics_service.rb
```diff
@@ -207,4 +207,8 @@ def previous_week_positive_tests(weeks_with_tests)
   def percentage_change(current_total, previous_total)
     (((current_total - previous_total) / previous_total.to_f) * 100).round(1)
   end
+
+  def on_or_before?(day_date, latest_date)
```
JonathanHallam, line 211: nice refactor 🥇

app/views/coronavirus_landing_page/components/landing_page/_arrow.html.erb
```diff
@@ -0,0 +1,9 @@
+<% if number.positive? %>
```
JonathanHallam, line 1: Do we need a 3rd case here for 0? Otherwise I assume we're going to get a down arrow for it which seems wrong

JonathanHallam (commented):
This is looking really nice, well done all! I've made a couple of minor comments but I'm happy that all of the maths works. I haven't given too much attention to the frontend but will have a gander when I get in on Monday, should be able to get this out pretty soon 🎉

app/views/coronavirus_landing_page/components/landing_page/_arrow.html.erb
leenagupte: Perhaps the conditional should be the other way around? i.e. check for `negative?` to show a down arrow and show an up arrow for everything else?  Or change the conditional to also check for `number.zero?`? 

Do we even want to so an arrow for zero?

What do you think @andysellick? Maybe add a note to the commit message about `positive?` and `negative?` both returning `false` for zero.

andysellick: I've changed this to an `if / elsif` so that if it's ever `0` we won't get an arrow shown at all.

leenagupte:
@JonathanHallam Thanks for your review. I've made the changes you suggested (missing guard clauses, misleading variable names) and updated the commit messages. I believe @andysellick has also made the changes to the arrow logic, so this PR is ready for re-review.

app/services/fetch_coronavirus_statistics_service.rb
```diff
@@ -113,12 +114,36 @@ def latest_admissions(data)
...
     }
   end
+
+  def weekly_tests(latest_tests, data)
+    day_tests = []
+    data.each do |day_cases|
+      day_tests << day_cases["newPositiveTests"] if latest_tests["date"] >= day_cases["date"]
+    end
+
+    current_week_tests = day_tests.first(7).compact
```
Rosa-Fox: Probably just me missing something, but I am confused about how the latest values with `nil` records are ignored (as demonstrated in the commit message). 

`a = [nil, 2, 3, 4, 5, 6, 7, 8, 9, 10]` 

` b = a.first(7).compact`  

`=> [2, 3, 4, 5, 6, 7]`

`b.count`

`=> 6` 

which would `return` on the line below

Rosa-Fox (commented):
Looks good to me, I had one question but probably just me missing something.

app/services/fetch_coronavirus_statistics_service.rb
leenagupte: So, in you example, if `a` or `data` is `[nil, 2, 3, 4, 5, 6, 7, 8, 9, 10]`
Then `day_tests` is populated with `[2, 3, 4, 5, 6, 7, 8, 9, 10]`

```
b = day_tests.first(7).compact
=> [2, 3, 4, 5, 6, 7, 8]

leenagupte: `latest_tests["date"]` is the first hash in the data had has a value for number of positive tests, so thats when we start recording the value.

Rosa-Fox: ah ok! Looking at the data in the tests it looks like it is possible for the `date` value to be present but then the other fields in that chunk are `nil`. So I wondered if `day_cases["newPositiveTests"]` would be `nil` which gets stored in `day_tests`? Which could mean the `nil` would get stored and effect the count?

app/views/coronavirus_landing_page/components/landing_page/_arrow.html.erb
JonathanHallam: thanks :)

app/services/fetch_coronavirus_statistics_service.rb
Rosa-Fox: ah sorry just spoke to Jon and realised I misunderstood. If the first value is nil then we would want to `return` as you mentioned in the commit message 'as this proves that we don't have the full 7 days worth of data and should not display the section'.

I incorrectly thought we would want to remove the nils at the start, but keep counting from the place where we had data and display that, whereas actually we don't want to display anything due to missing data.

JonathanHallam (approved):
Am happy with this, good work all!

app/services/fetch_coronavirus_statistics_service.rb
leenagupte: @Rosa-Fox Is there anything I can add to the commit message to make this clearer?

leenagupte: You're right. If we have nils at the start we want to remove them. And start our "7 days" in the past. But if there is nil data in the middle of a week, we want to `return` as we won't have a full week's worth of data.

PR #2586, opened 2021-11-01
Allow links in NHS section

gclssvglx:
## What

This change allows for a section heading in the NHS notice to be (optionally) made a link. 

## Why

This improves the user experience by allowing simple, clean copy to be created that has a single link to the desired action.

The use of links in headers is not typically done on GOV.UK. It may also create some undesirable accessibility issues. To avoid this situation, we style a `P` element to look like a an `H3` in these cases.

Section headers without a URL are rendered as an `H3` element as before.

## Example YAML (see [coronavirus_landing_page.yml](https://github.com/alphagov/govuk-coronavirus-content/blob/12bacb98cdfa9a96a8f6f344143be0be1361babd/content/coronavirus_landing_page.yml#L33))

```yaml
nhs_banner:
  header: Testing for coronavirus (COVID-19)
  sections:
    - heading: "Testing for coronavirus (COVID-19)"
      url: "https://www.nhs.uk/conditions/coronavirus-covid-19/testing/"
      markdown: "Find out how to get tested, what your test result means and how to report your result.\n"
```

### Before

Screenshot 2021-11-01 at 13 40 14

### After

Screenshot 2021-11-01 at 13 41 08

[Trello](https://trello.com/c/MRWKzJg6/299-improve-nhs-section-to-allow-for-links)

### DIFF
--- a/app/views/coronavirus_landing_page/components/landing_page/_nhs_notice.html.erb
+++ b/app/views/coronavirus_landing_page/components/landing_page/_nhs_notice.html.erb
@@ -21,12 +21,18 @@
 
   <div class="govuk-grid-column-two-thirds">
     <% nhs_banner["sections"].each do |section| %>
-      <%= render "govuk_publishing_components/components/heading", {
-        text: section["heading"],
-        heading_level: 2,
-        font_size: "s",
-        margin_bottom: 4,
-      } if section["heading"].present? %>
+      <% if section["heading"].present? && section["url"].present? %>
+        <p class="govuk-heading-s">
+          <a class="govuk-link" href="<%= section["url"] %>"><%= section["heading"] %></a>
+        </p>
+      <% else %>
+        <%= render "govuk_publishing_components/components/heading", {
+          text: section["heading"],
+          heading_level: 3,
+          font_size: "s",
+          margin_bottom: 4,
+        } if section["heading"].present? %>
+      <% end %>
 
       <%= render_govspeak(section["markdown"]) if section["markdown"].present? %>
     <% end %>
--- a/spec/views/coronavirus_landing_page/components/landing_page/nhs_notice.html.erb_spec.rb
+++ b/spec/views/coronavirus_landing_page/components/landing_page/nhs_notice.html.erb_spec.rb
@@ -1,0 +1,103 @@
+require "spec_helper"
+
+RSpec.describe "the nhs notice partial" do
+  before do
+    render(
+      "coronavirus_landing_page/components/landing_page/nhs_notice",
+      {
+        nhs_banner: {
+          header: "Hello world",
+          sections: sections,
+        },
+      }.with_indifferent_access,
+    )
+  end
+
+  describe "header" do
+    context "when there is header text" do
+      let(:sections) { [] }
+
+      it "renders the header as an H2 with the text" do
+        expect(rendered).to have_selector("h2", text: "Hello world")
+      end
+    end
+  end
+
+  describe "sections" do
+    context "when there are no sections" do
+      let(:sections) { [] }
+
+      it "renders an empty DIV" do
+        expect(rendered).to have_selector("div.govuk-grid-column-two-thirds", text: "")
+      end
+    end
+
+    describe "markdown" do
+      context "when the markdown is empty" do
+        let(:sections) { [{ markdown: "" }] }
+
+        it "does not render a DIV with the govuk-govspeak class" do
+          expect(rendered).to have_no_selector("div.govuk-govspeak")
+        end
+      end
+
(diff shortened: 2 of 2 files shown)


app/views/coronavirus_landing_page/components/landing_page/_nhs_notice.html.erb
```diff
@@ -21,12 +21,18 @@
...
   <div class="govuk-grid-column-two-thirds">
     <% nhs_banner["sections"].each do |section| %>
-      <%= render "govuk_publishing_components/components/heading", {
-        text: section["heading"],
-        heading_level: 2,
-        font_size: "s",
-        margin_bottom: 4,
-      } if section["heading"].present? %>
+      <% if section["link_url"].present? %>
+        <p class="gem-c-heading govuk-heading-s govuk-!-margin-bottom-4">
```
andysellick, line 25: The `gem-c-heading` class can probably be removed here as it's a component specific class - `govuk-heading-s` should be sufficient. Also we tend to avoid using the margin override classes if possible unless they're actually overriding something. It might work if you tried `govuk-body` instead?

gclssvglx: Thanks for the suggestions, will try them out and fixup.  Also, fix the breaking test(s)...

gclssvglx: Thanks @andysellick - styles updated leaving only `govuk-heading-s` - visually and margin/padding wise looks the same.

andysellick (commented):
Frontend looks good to me 👍 

Also very nice clear PR description, thanks.

app/views/coronavirus_landing_page/components/landing_page/_nhs_notice.html.erb
```diff
@@ -21,12 +21,18 @@
 
   <div class="govuk-grid-column-two-thirds">
     <% nhs_banner["sections"].each do |section| %>
-      <%= render "govuk_publishing_components/components/heading", {
-        text: section["heading"],
-        heading_level: 2,
-        font_size: "s",
-        margin_bottom: 4,
-      } if section["heading"].present? %>
+      <% if section["url"].present? %>
```
Rosa-Fox, line 24: I am wondering if we should also check the presence of heading? Incase of the unlikely event there is a url but no heading?

spec/views/coronavirus_landing_page/components/landing_page/nhs_notice.html.erb_spec.rb
```diff
@@ -0,0 +1,121 @@
...
+        }.with_indifferent_access,
+      )
+
+      expect(rendered).to have_selector("h2", text: "Hello world")
+    end
+  end
+
+  context "sections" do
+    it "renders an empty DIV when there are no sections" do
+      render(
```
Rosa-Fox, line 22: These tests look good. I wonder if it might be possible to DRY up this `render` part by moving it into a method and passing in the `sections` value as an argument from each test?

Rosa-Fox (commented):
Looks cool! I added a couple of small suggestions, let me know what you think.

app/views/coronavirus_landing_page/components/landing_page/_nhs_notice.html.erb
gclssvglx: Ah yes, my bad - we should absolutely check for that.  Great spot - thanks @Rosa-Fox  👏  Fixed up

spec/views/coronavirus_landing_page/components/landing_page/nhs_notice.html.erb_spec.rb
gclssvglx: Another great spot - thanks 👍 I went with just using `let(:sections)` instead of method as it seems more RSpecy

gclssvglx: Also, I think it read better when you apply the `-f d` flag

app/views/coronavirus_landing_page/components/landing_page/_nhs_notice.html.erb
Rosa-Fox: 👍 thanks!

spec/views/coronavirus_landing_page/components/landing_page/nhs_notice.html.erb_spec.rb
Rosa-Fox: nice one! Agreed the let is better.

Rosa-Fox (approved):

PR #2595, opened 2021-11-05
Remove jQuery from coronavirus landing page JS

andysellick:

## What

Remove jQuery from the JS for the coronavirus landing page, and restructure the code slightly. Also converts it into a GOVUK Frontend 'init' module, rather than a 'start' module.

## Why

We're removing jQuery across GOV.UK, and there was also some tidying needed on this particular file.

## Visual changes

None.

Trello card: https://trello.com/c/GmHGdvoH/72-convert-coronavirus-landing-page-javascript-to-follow-a-more-conventional-module-approach-without-jquery

### DIFF
--- a/app/assets/javascripts/modules/coronavirus-landing-page.js
+++ b/app/assets/javascripts/modules/coronavirus-landing-page.js
@@ -1,3 +1,5 @@
+//= require govuk_publishing_components/vendor/polyfills/closest
+
 window.GOVUK = window.GOVUK || {}
 window.GOVUK.Modules = window.GOVUK.Modules || {};
 
--- a/app/assets/javascripts/modules/coronavirus-landing-page.js
+++ b/app/assets/javascripts/modules/coronavirus-landing-page.js
@@ -2,11 +2,11 @@ window.GOVUK = window.GOVUK || {}
 window.GOVUK.Modules = window.GOVUK.Modules || {};
 
 (function (Modules) {
-  function CoronavirusLandingPage () {}
+  function CoronavirusLandingPage ($module) {
+    this.module = $module
+  }
 
-  CoronavirusLandingPage.prototype.start = function ($element) {
-    this.module = $element[0]
-
+  CoronavirusLandingPage.prototype.init = function () {
     // Confirm the user is on the coronavirus landing page
     if (this.checkOnLandingPage()) {
       this.globarBarSeen()
--- a/app/assets/javascripts/modules/coronavirus-landing-page.js
+++ b/app/assets/javascripts/modules/coronavirus-landing-page.js
@@ -5,17 +5,15 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
   function CoronavirusLandingPage () {}
 
   CoronavirusLandingPage.prototype.start = function ($element) {
-    var $this = this
+    this.module = $element[0]
+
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/modules/coronavirus-landing-page.js
```diff
@@ -32,15 +30,21 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
-    $element.find('.gem-c-accordion__open-all').on('click', function (e) {
-      var expanded = $(e.target).attr('aria-expanded') === 'true'
-      var label = expanded ? 'Expand all' : 'Collapse all'
-      var action = expanded ? 'accordionOpened' : 'accordionClosed'
-      var options = { transport: 'beacon', label: label }
+  CoronavirusLandingPage.prototype.addAccordionOpenAllTracking = function () {
+    var accordions = this.module.querySelectorAll('.gem-c-accordion')
+    for (var i = 0; i < accordions.length; i++) {
+      accordions[i].addEventListener('click', function (e) {
+        var clicked = e.target.closest('button')
```
danacotoran, line 37: `closest()` will return `null` if an element matching the selector is not found. I understand such an element should in theory always exist in the context of the accordion component, but in a scenario where it is not found, `clicked.classList` would throw an `Uncaught TypeError`.
So basically what I'm saying is, should this conditional  be changed to `if (clicked && clicked.classList.contains('gem-c-accordion__open-all')) {}` 😄

danacotoran: Actually I can replicate this hypothetical issue in a browser too (recreate it by clicking inside the accordion but outside the `gem-c-accordion__open-all` link)

https://user-images.githubusercontent.com/7116819/140968420-aa0d4fe3-5015-48a6-8229-970e11c93ea1.mov

danacotoran (changes requested):
Found a bug with the code which needs fixing, but I think it's looking good other than that

app/assets/javascripts/modules/coronavirus-landing-page.js
andysellick: Great spot, thanks. Have updated.

app/assets/javascripts/modules/coronavirus-landing-page.js
```diff
@@ -32,15 +30,21 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
   }
 
-  CoronavirusLandingPage.prototype.addAccordionOpenAllTracking = function ($element) {
-    $element.find('.gem-c-accordion__open-all').on('click', function (e) {
-      var expanded = $(e.target).attr('aria-expanded') === 'true'
-      var label = expanded ? 'Expand all' : 'Collapse all'
-      var action = expanded ? 'accordionOpened' : 'accordionClosed'
-      var options = { transport: 'beacon', label: label }
+  CoronavirusLandingPage.prototype.addAccordionOpenAllTracking = function () {
+    var accordions = this.module.querySelectorAll('.gem-c-accordion')
```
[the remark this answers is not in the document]
andysellick: That's fair. I tried to document it in the commit but I've added a comment to the code as well to make it super clear.

danacotoran (approved):

PR #2615, opened 2021-11-29
Import translations and sync locales

1pretz1:
Imports translations and sync locales.

More info in the commits ->

Trello:
https://trello.com/c/Xz78vNR4/2757-import-newly-sourced-translations-into-locale-files-5


config/i18n-tasks.yml
```diff
@@ -1,4 +1,8 @@
 data:
+  read:
+    - config/locales/%{locale}/*.yml
+  write:
+    - ['{:}.*', 'config/locales/%{locale}/\1.yml']
```
ChrisBAshton, line 5: This syntax is quite tricky to parse - is it copied and pasted from somewhere? Derived from documentation? Could you explain in the commit message where it came from? Perhaps even an inline comment in the file would be useful too.

config/i18n-tasks.yml
```diff
@@ -6,3 +6,8 @@ data:
   yaml:
     write:
       line_width: -1
+ignore_unused:
```
ChrisBAshton, line 9: Why are we ignoring these rather than removing them? Could you explain in the commit? (Especially as they're then not removed in the subsequent "Remove unused keys" commit).

My guess is that they actually _are_ used, but the static analysis doesn't spot that.

config/locales/be/organisations.yml
```diff
@@ -438,16 +438,16 @@ be:
     jobs_contracts: Праца і кантакты
     latest_from: Апошняе ад %{title}
     notices:
-      changed_name: "%{title} зараз завецца <a href='%{link_href}'>%{link_text}</a>"
-      devolved: "%{title} з'яўляецца органам <a href='%{link_href}'>%{link_text}</a>"
+      changed_name_html: "%{title} зараз завецца <a href='%{link_href}'>%{link_text}</a>"
```
ChrisBAshton, line 441: Like with the [previous PR](https://github.com/alphagov/government-frontend/pull/2305#discussion_r759984448), could you explain the benefit of converting these to `_html`?

config/locales/ar/shared.yml
```diff
@@ -6,7 +6,7 @@ ar:
       prefix: 'تصفَّح:'
       title: تصفَّح GOV.UK
     get_emails: احصل على رسائل بريد إلكتروني عن هذا الموضوع
-    language_direction: ltr
+    language_direction: rtl
```
ChrisBAshton, line 9: Nitpick: could your commit message link to the resource you used to determine which languages are RTL?

config/locales/bn/shared.yml
```diff
@@ -7,6 +7,6 @@ bn:
       title: GOV.UK ব্রাউজ করুন
     get_emails: এই বিষয়ের জন্য ইমেইল নিন
     language_direction: ltr
-    language_name: ইংরেজি
```
ChrisBAshton: Well spotted!

ChrisBAshton (commented):
This is looking great! Some very minor suggestions below:

config/i18n-tasks.yml
1pretz1: It is indeed. I've added a commit message explaining where the config was sourced from and what it means.

config/i18n-tasks.yml
1pretz1: We're ignoring them as they are actually used, but i18n-tasks default unused locale scanners can't detect them. Adding a list prevents i18n-tasks from removing those keys when we run destructive commands like: `i18n-task remove-unused`.

config/locales/be/organisations.yml
1pretz1: Good point - this commit was older than the one you originally commented on.

config/locales/ar/shared.yml
1pretz1: Sure thing, have added some more detail to the commit message.

1pretz1:
Cheers for the review @ChrisBAshton. I've addressed your comments and thanks for making them as my commit messages were quite non-descriptive!

ChrisBAshton (approved):
Great stuff - thanks Peter ⭐

PR #2617, opened 2021-11-30
Remove GOV.UK Account sign-in from /brexit header

huwd:
## What

GOV.UK account integration has now been removed from the brexit checker,
so there is no longer a need to have a sign in link on the taxon page.

Here we remove the the option that calls this out of slimmer.

We also replace the Header with the new explore banner that we had
previously been overriding to keep our button in place.

And as no other place in collections uses methods from the
personalisation gem, we can also remove that here to keep things tidy.

The sign in link will likely return, but as part of the site-wide
explore banner. So we can likely clear up any GOV.UK Account specific
slimmer stuff after this too.

## Mobile

image

## Desktop

image


app/controllers/brexit_landing_page_controller.rb
```diff
@@ -1,12 +1,11 @@
 class BrexitLandingPageController < ApplicationController
-  include GovukPersonalisation::ControllerConcern
   include Slimmer::Headers
 
   skip_before_action :set_expiry
   before_action -> { set_expiry(1.minute) }
-  before_action -> { set_slimmer_headers(remove_search: true, show_accounts: logged_in? ? "signed-in" : "signed-out") }
+  before_action -> { set_slimmer_headers(remove_search: true) }
```
danacotoran, line 6: This `remove_search` flag was useful back when the site header had a search field on the left hand side. There would be instances where we would want to hide the search field (e.g on the search page, or to put something else in there such as navigational links). It isn't useful here so I'd remove it.

huwd: Ok cool so the entire L6 can go?

huwd: Just force pushed!

danacotoran (approved):
I think it's good to go!

PR #2619, opened 2021-12-01
Use standard metadata for Brexit page

sihugh:
We don't want custom images or code to worry about if we don't need them.

This page is going to be replaced by a standard format soon, and we want to make that transition ( 😄 ) simple by removing as much unnecessary customisation as possible. Using components makes things more manageable.

I had to inject the translated versions of the title and description into the content item because the [Welsh content item](https://www.gov.uk/api/content/brexit.cy) is barely more than a placeholder - it doesn't have a description.

I tweaked a metadata test for og:title because the metadata component sometimes puts [Withdrawn] in front of the title if the generated test content item is a withdrawn one (which won't be true in practice).


hannako (approved):
✂️

PR #2645, opened 2022-01-12
Display correct translation for "email"

leenagupte:
Zendesk ticket: https://govuk.zendesk.com/agent/tickets/4836551

# What's changed and why?
A change was made to always set the translation for "email" to the english
version for "FOI Contacts" as they come from Whitehall and are unreliable:
https://github.com/alphagov/collections/pull/1857/commits/b321aaeb285b6da0ffd7fafd21b5902621278a01

However this also had the effect of not translating the text in the contacts block.
This was odd as "telephone" would be translated, but "email" is not.
See: https://www.gov.uk/government/organisations/land-registry.cy

# Expected changes
## Before
Screenshot 2022-01-12 at 17 11 53

## After
Screenshot 2022-01-12 at 17 11 35

### DIFF
--- a/app/views/organisations/_contacts.html.erb
+++ b/app/views/organisations/_contacts.html.erb
@@ -21,7 +21,12 @@
         <div class="<%= contact[:post_addresses].any? ? "govuk-grid-column-one-half" : "govuk-grid-column-two-thirds" %>">
           <% if contact[:email_addresses].any? %>
             <div class="organisation__margin-bottom">
-              <p <%= "lang=en" if contact[:locale] != "en" %>><%= t('organisations.contact.email', locale: :en) %></p>
+              <% if I18n.exists?('organisations.contact.email', contact[:locale]) %>
+                <p><%= t('organisations.contact.email') %></p>
+              <% else %>
+                <p lang="en"><%= t('organisations.contact.email') %></p>
+              <% end %>
+
               <% contact[:email_addresses].each do |email| %>
                 <p><%= email.html_safe %></p>
               <% end %>


JonathanHallam (approved):
Brilliant, thanks :)

app/views/organisations/_contacts.html.erb
```diff
@@ -21,7 +21,7 @@
         <div class="<%= contact[:post_addresses].any? ? "govuk-grid-column-one-half" : "govuk-grid-column-two-thirds" %>">
           <% if contact[:email_addresses].any? %>
             <div class="organisation__margin-bottom">
-              <p <%= "lang=en" if contact[:locale] != "en" %>><%= t('organisations.contact.email', locale: :en) %></p>
+              <p <%= "lang=en" if contact[:locale] != "en" %>><%= t('organisations.contact.email') %></p>
```
danacotoran, line 24: When we are viewing a translated version of a page, and a translation in that language exists for a particular string/sentence/term on the page, then that string should not have a `lang="en"` attribute. 

On the [Welsh version of this page](https://govuk-collec-display-co-i94w6w.herokuapp.com/government/organisations/land-registry.cy), the term "E-bost" is now being indicated as being in English which is not accurate:
Screenshot 2022-01-14 at 12 21 20
 

If, however, a translation does not exist for an element, and i18n  is falling back to the English language default, then that element should have `lang="en"`. So let's say a translation did not exist for `t('organisations.contact.email')` so it had to fall back to the English language term "Email" – that is when a `lang` attribute is required to correctly indicate the language that bit of text is in.

Reference: https://www.w3.org/WAI/WCAG21/Understanding/language-of-parts.html

leenagupte: So in this case, the a Welsh language translation does exist, and all of the addresses are in Welsh, and the "div" it's contained in is also Welsh:

Screenshot 2022-01-14 at 13 25 15

Should I remove the `if` statement on this line as well: https://github.com/alphagov/collections/blob/main/app/views/organisations/_contacts.html.erb#L24 ?

leenagupte: Perhaps it would be better to check if the translation for the language exists before replacing it? https://stackoverflow.com/questions/12353416/rails-i18n-check-if-translation-exists

leenagupte:
@danacotoran Thanks for your review and comments. I've added a check to make sure a translation exist for "email" before defaulting to English.

danacotoran (approved):
Thank you 🚀

PR #2651, opened 2022-01-21
Validate locale files during test runs

benjamineskola:
https://trello.com/c/faHEHUr8/2827-apply-locale-file-validation-to-collections


spec/lib/locales_validation_spec.rb
```diff
@@ -1,4 +1,21 @@
 #!/usr/bin/env ruby
+
+class IncompatiblePlurals
+  # monkey-patching to avoid a false positive
```
ChrisBAshton, line 4: You mentioned perhaps Rails Translation Manager needs a feature whereby we can 'ignore' certain translations, and I think you're right - monkey-patching is a smell, and monkey-patching this amount of code makes me quite uncomfortable!

You're closest to the coal face on this, so probably have good ideas about how that could be achieved. E.g. should we expose a way of specifying ignorable translations in [plurals.rb](https://github.com/alphagov/collections/blob/main/config/locales/plurals.rb)? Or have a `config/initializers/rails_translation_manager.rb` file that allows some calls to the API to customise it? Or perhaps a root-level `rails_translation_manager.yml` file where you can specify overrides?

ChrisBAshton (commented):
Thanks @benjamineskola - the changes look good, but I agree that we should avoid monkey-patching if possible. Happy to discuss on Slack what that might look like.

spec/lib/locales_validation_spec.rb
benjamineskola: My initial thought was that if the `validate_locales` method took an `ignore_plurals` parameter then the `IncompatiblePlurals::incompatible_plurals` could just do `next if ignored_plurals.include? plural_group[:parent]`. My only concern there would be having to pass the array around through multiple levels.

ChrisBAshton: Thanks for those thoughts. Makes sense. Just had a quick look at rails_translation_manager, it feels like this kind of customisation would work nicely:

```diff
-     checker = RailsTranslationManager::LocaleChecker.new("config/locales/*/*.yml")
+     checker = RailsTranslationManager::LocaleChecker.new("config/locales/*/*.yml", skip_validation: ["organisations.type"])
```

Passing the array around shouldn't be too much of an inconvenience, as we'd only do it [in one place](https://github.com/alphagov/rails_translation_manager/blob/8724505fa12100962d3297ba89643f1b41b65510/lib/rails_translation_manager/locale_checker.rb#L44), the [BaseChecker class](https://github.com/alphagov/rails_translation_manager/blob/8724505fa12100962d3297ba89643f1b41b65510/lib/rails_translation_manager/locale_checker/base_checker.rb#L6-L8) being quite useful here. Then you could set `@all_locales` in the BaseChecker class to exclude the ignorable keys.

config/locales/gd/organisations.yml
```diff
@@ -484,5 +484,7 @@ gd:
     view_less: féach níos lú
     what_we_do: Cad atá á dhéanamh ag %{title}
     works_with_statement:
+      two:
+      few:
```
ChrisBAshton, line 488: Looks like these crept into the wrong commit. Should they have been in the `Add missing pluralisation keys` instead? There are several examples dotted around this commit.

ChrisBAshton (commented):
LGTM apart from the final two commits look like they've crossed over into each other a bit. Could you take a look?

Thanks for adding the RTM feature, much nicer! ⭐

ChrisBAshton (approved):
🚀

PR #2659, opened 2022-01-27
Honour content store ttl

kentsanggds:
Content Store provides a TTL in the Cache-Control response header which frontend apps should respect.

This ensures that this app honours the cache TTL provided by Content Store. There is a 5 minute TTL on the FindersController which I have left as-is, since we want content in search to be fresh and the current TTL provided by Content Store is 30 minutes.

This partially implements RFC 144.

Links:

Content Store TTL: https://github.com/alphagov/content-store/blob/9661e2047f2167f87066fbb6ef38da71668a1327/app/controllers/content_items_controller.rb#L14
RFC 144: https://github.com/alphagov/govuk-rfcs/blob/main/rfc-144-consistent-cache-expiry.md

### DIFF
--- a/app/controllers/brexit_landing_page_controller.rb
+++ b/app/controllers/brexit_landing_page_controller.rb
@@ -2,7 +2,9 @@ class BrexitLandingPageController < ApplicationController
   include Slimmer::Headers
 
   skip_before_action :set_expiry
-  before_action -> { set_expiry(1.minute) }
+  before_action do
+    set_expiry content_item.max_age, public_cache: content_item.public_cache
+  end
 
   slimmer_template "gem_layout_full_width_explore_header"
 
--- a/app/controllers/coronavirus_landing_page_controller.rb
+++ b/app/controllers/coronavirus_landing_page_controller.rb
@@ -6,13 +6,11 @@ class CoronavirusLandingPageController < ApplicationController
   def show
     @statistics = FetchCoronavirusStatisticsService.call
     if @statistics
-      set_expiry 5.minutes
+      set_expiry content_item.max_age, public_cache: content_item.public_cache
     else
       logger.warn "Serving /coronavirus without statistics"
       set_expiry 30.seconds
     end
-
-    @content_item = content_item.to_hash
 
     breadcrumbs = [{ title: t("shared.breadcrumbs_home"), url: "/", is_page_parent: true }]
     title = {
@@ -39,10 +37,10 @@ private
   end
 
   def presenter
-    @presenter ||= CoronavirusLandingPagePresenter.new(@content_item, params[:nation])
+    @presenter ||= CoronavirusLandingPagePresenter.new(content_item.to_hash, params[:nation])
   end
 
   def special_announcement
-    @special_announcement ||= SpecialAnnouncementPresenter.new(@content_item)
+    @special_announcement ||= SpecialAnnouncementPresenter.new(content_item.to_hash)
   end
 end
--- a/app/controllers/organisations_controller.rb
+++ b/app/controllers/organisations_controller.rb
@@ -1,6 +1,8 @@
 class OrganisationsController < ApplicationController
   skip_before_action :set_expiry
-  before_action -> { set_expiry(5.minutes) }
+  before_action do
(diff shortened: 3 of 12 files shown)


app/models/content_item.rb
```diff
@@ -2,8 +2,10 @@ class ContentItem
   attr_reader :content_item_data
 
   def self.find!(base_path)
-    response = Services.content_store.content_item(base_path)
-    new(response.to_hash)
+    content_item = Services.content_store.content_item(base_path)
+    content_item_hash = content_item.to_h
+    content_item_hash["cache_control"] = { "max_age" => content_item.cache_control["max-age"], "public" => content_item.cache_control.include?("public") }
```
bilbof, line 7: 😄 cheeky, mutating the content item - it's unorthodox (you could keep the provided `content_item` class around and generate the hash as needed later) but it'll work.

here you could do `"public" => !content_item.cache_control.private?`

app/controllers/brexit_landing_page_controller.rb
```diff
@@ -2,7 +2,10 @@ class BrexitLandingPageController < ApplicationController
   include Slimmer::Headers
 
   skip_before_action :set_expiry
-  before_action -> { set_expiry(1.minute) }
+  before_action do
+    content_item = ContentItem.find!(request.path)
```
bilbof, line 6: In the best case this is making a call to the cache (memcached ?) so it's not super expensive, but if we're calling the cache twice for every request I think we can do better. In the worst case and the cache is unavailable this doubles the calls to content store.

app/controllers/coronavirus_landing_page_controller.rb
```diff
@@ -4,16 +4,15 @@ class CoronavirusLandingPageController < ApplicationController
   slimmer_template "gem_layout_full_width"
 
   def show
+    @content_item = content_item.to_hash
```
bilbof, line 7: Is `content_item` already a hash by this point?

app/controllers/organisations_controller.rb
```diff
@@ -1,6 +1,9 @@
 class OrganisationsController < ApplicationController
   skip_before_action :set_expiry
-  before_action -> { set_expiry(5.minutes) }
+  before_action do
+    content_item = ContentItem.find!(request.path)
+    set_expiry(content_item.content_item_data["cache_control"]["max_age"])
```
bilbof, line 5: Same issue here - fetching the content item twice.

bilbof (changes requested):
In general the change looks good! I'd like to avoid calling the content-store / cache twice, and I wonder if you could put a note in to explain why the Yarn change went it.

app/controllers/coronavirus_landing_page_controller.rb
kentsanggds: I just moved the code up from line 15, it doesn't appear to be a hash from my testing of the code

app/models/content_item.rb
kentsanggds: Is that a better way of doing things? it will mean more code changing if I keep the object, I'll see if I can get that working

app/controllers/brexit_landing_page_controller.rb
kentsanggds: Think I've fixed this

app/controllers/brexit_landing_page_controller.rb
```diff
@@ -26,7 +29,8 @@ def show
 
   def taxon
     base_path = request.path
-    @taxon ||= Rails.cache.fetch("collections_content_items#{base_path}", expires_in: 10.minutes) do
+    content_item = ContentItem.find!(base_path)
+    @taxon ||= Rails.cache.fetch("collections_content_items#{base_path}", expires_in: content_item.content_item_data["cache_control"]["max_age"]) do
```
bilbof, line 33: I think this change is not needed, since we only want to set the Cache Control header to cache pages for the correct length of time as declared in the Content Store Cache-Control response header. We don't want to cache the responses from internal API calls differently.

kentsanggds: ok, I'll remove

bilbof (approved):

PR #2661, opened 2022-01-31
Fix browse analytics location

andysellick:

## What / why
Fixes Google analytics pageviews in the browse page sent on category change after the page has loaded.

This problem is due to the browse pages behaving like a Single Page Application. The `Document location URL` (location) and the page title are set in GA when the page loads but not updated for any further pageviews sent manually by the browse JavaScript if the user navigates to other sections within the browse page.

So if you started from `https://www.gov.uk/browse/citizenship` but then browsed to other sections of the browse page, the location would remain the same, even though the `path` value in the pageview might be something quite different. In the example below, `path` accurately reflects the page/URL that is currently being viewed, whereas location does not.

Screenshot 2022-01-31 at 15 23 40

This fixes this by explicitly setting the `location` value in the options object passed to GA. It also does the same for the document title.

## Visual changes
None.

### DIFF
--- a/app/assets/javascripts/browse-columns.js
+++ b/app/assets/javascripts/browse-columns.js
@@ -289,7 +289,8 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
     if (GOVUK.analytics && GOVUK.analytics.trackPageview) {
       var options = {
         dimension1: sectionTitle,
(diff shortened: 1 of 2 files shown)


chao-xian (approved):
Thank you Andy!

PR #2665, opened 2022-02-01
Upgrade Ruby to 2.7.5

ChrisBAshton:
Upgrades Ruby to 2.7.5. See commits for more details.
            This PR should be tested before merge, as per https://docs.publishing.service.gov.uk/manual/ruby.html#update-ruby-version-in-the-relevant-repos.


KludgeKML (approved):
Looks good to me.

PR #2669, opened 2022-02-03
Fix Google Analytics events on browse pages

andysellick:

## What
Google analytics tracking on the browse page was incorrect. There were two main problems:

- some details of tracking were incorrect (e.g. `dimension1`, `dimension26` and `dimension27`)
- event tracking was recording information based on the state of the page after a click, not before (which included wrong details, including the page title and URL)

Events fired when users navigate the browse page were being handled partly by `browse-columns.js` and partly by the `gemTrackClick` module from the gem. Now all events are handled by `browse-columns.js`.

This work has involved the following changes:

Updating `browse-columns.js`:

- to include a concept of the right most column that is currently visible
- to include functions for getting `dimension26` and `dimension27` correctly, which reflect the number of links and sections of links within the right most column that is currently visible
- explicitly firing click events (previously handled by the `gemTrackClick` script from the gem)
- updating the `displayState` attribute more frequently to more accurately reflect the current state
- getting the h2 rather than the h1 when tracking a pageview (there was no h1, so it wasn't working)
- including dimension1 in the pageview (the title of the current section, which is the middle column)

Updating the browse controller and templates:

- to calculate dimension26 and dimension27 so those numbers can be applied to links
- to include the page slug as a `section` meta tag, which is picked up by the general pageview analytics in the gem
- to include the same value for dimension1 on links, picked up by event clicking code in browse-columns.js

And also writing a few extra tests and rewriting some of the fixtures stuff to make it easier to read.

There remains an outstanding problem to do with pageviews that happen when the page loads, but this is resolved separately within the [gem analytics code in a separate PR](https://github.com/alphagov/govuk_publishing_components/pull/2646).

## Why
We need accurate reporting on this page.

## Visual changes
None.

Trello card: https://trello.com/c/uyaekaKj/394-fix-google-analytics-page-path-on-ajax-events-on-mainstream-browse-pages-and-finders


chao-xian (approved):
Code changes LGTM - just waiting for Performance Analyst to confirm that it's what they want.

chao-xian:
> * On pageviews and events on level 1 pages, custom dimension 1 should be just benefits (or whichever section the page is in) rather than /browse/benefits.
> * On events on level 1 pages, the URL, title and custom dimensions should still reflect the level 1 page you were on, not the level 2 page you're clicking through to.
> * On pageviews on level 2 pages, the URL/path and title are correct (apart from the 'Browse - ' prefix issue that I mentioned before). Custom dimension 1 should still be benefits  rather than /browse/benefits/manage-your-benefit.
> * On events on level 2 pages, again the URL, title and custom dimensions should reflect the level 2 page you were on, not the content page you're clicking through to.
> * I'm also confused about 'Third Level Browse' in dimension 32, on the event when you click on a link to a content page. We only use 'First Level Browse' for /benefits for example, and 'Second Level Browse' for /benefits/manage-your-benefit. (And 'Browse Index' for the very top /browse page.)

andysellick:
Thanks @chao-xian I'll take a look at fixing this today.

app/assets/javascripts/browse-columns.js
```diff
@@ -51,6 +61,47 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
+
+    if (window.GOVUK.analytics && window.GOVUK.analytics.trackEvent) {
+      GOVUK.analytics.trackEvent(category, action, options)
+    }
+  }
+
+  // dimension26 should be the number of 'sections' of links in the
+  // right-most visible column. For the first two columns this is
+  // always just 1
+  BrowseColumns.prototype.getDimension26 = function () {
```
chao-xian, line 95: Could we rename the function to be more reflective of what it does and shorten the comment? Something like sectionLinkCount?

app/assets/javascripts/browse-columns.js
```diff
@@ -51,6 +61,47 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
+    return this.rightMostColumn.querySelectorAll('.browse__list').length
+  }
+
+  // dimension27 should be the total number of links in the right
+  // most visible column
+  BrowseColumns.prototype.getDimension27 = function () {
```
chao-xian, line 101: Can we rename this to linkCount?

app/controllers/second_level_browse_page_controller.rb
```diff
@@ -38,4 +40,21 @@ def page
       "/browse/#{params[:top_level_slug]}/#{params[:second_level_slug]}",
     )
   end
+
+  def dimension26(page)
```
chao-xian, line 44: Similar to the JS functions, can we rename this to sectionLinkCount or similar?

app/controllers/second_level_browse_page_controller.rb
```diff
@@ -38,4 +40,21 @@ def page
...
+    page.lists.count + page.related_topics.count
+  end
+
+  def dimension27(page)
```
chao-xian, line 48: Again a rename here too would be useful. link_count as the method name is ok.

chao-xian (dismissed):
A couple of minor things. Worth checking if the new dimensions will come through in mobile view too.

chao-xian:
A couple of minor things. Worth checking if the new dimensions will come through in mobile view too.

chao-xian (approved):

PR #2671, opened 2022-02-04
Replace the "view all" link on departments page with the details component

jon-kirwan:
TBC


maxgds (approved):
This looks sound. 

Before merging, we should run this one by Adam to confirm he's happy and whether we need to do any comms - and check whether the pattern exists anywhere else we should update along with this.

In terms of the document structure, please write another card for that as it's not being introduced with this change and what to do about it requires some thought.

app/views/organisations/_expanded_works_with_organisations.html.erb
```diff
@@ -1,17 +1,13 @@
...
-<div class="organisation-list__works-with js-hidden" id="toggle_<%= current_organisation.parameterize %>">
+<div id="toggle_<%= current_organisation.parameterize %>">
   <% works_with_organisations.each do |type, work_with_organisations| %>
     <%= render "govuk_publishing_components/components/heading", {
       text: t('organisations.type.' + type),
       heading_level: 4,
       font_size: "s",
     } %>
-    <ul class="organisation-list__works-with-list">
+    <ul class="govuk-!-margin-bottom-3">
```
chris-gds, line 8: The bullets on this section are white and are announced by AT weirdly - it sort of goes _"white bullet, link"_

Updating to [`<ul class=" govuk-list govuk-list--bullet">`](https://design-system.service.gov.uk/styles/typography/) should sort.

https://user-images.githubusercontent.com/71266765/152956582-43e3cdef-9ecc-4ff7-b316-88a0ea95c644.mov

app/views/organisations/_organisations_list.html.erb
```diff
@@ -33,31 +33,40 @@
...
+                    brand: organisation["brand"],
+                    crest: organisation["logo"]["crest"]
+                  },
+                  heading_level: 3,
+                } %>
+              </div>
             <% else %>
-              <%= link_to(organisation["title"], organisation["href"], class: "govuk-link organisation-list__item-title") %>
+              <div class="govuk-!-margin-bottom-3">
+                <h3 class="govuk-heading-s govuk-!-font-weight-regular govuk-!-margin-bottom-0 govuk-!-display-inline govuk-!-margin-right-2">
```
chris-gds, line 49: Something's not quite right here, this doesn't look like heading

image

maxgds: Hmm, you make a good point. Initially I thought it made sense as it preserves the hierarchy of the siblings that are sort of parents, but I'm now considering whether any of these h3s should actually be headings. The relationship is that the listed org works with the ones listed under it, but it's not a parent to them. I think we should probably be removing all the heading levels throughout.

app/views/organisations/_organisations_list.html.erb
```diff
@@ -33,31 +33,40 @@
...
+                  <%= link_to(organisation["title"], organisation["href"], class: "govuk-link organisation-list__item-title") %>
+                </h3>
```
chris-gds, line 51: _"Separate website"_ gets announced after the event for the user.  It ideally needs to be tied 
together to the user hears _"National Archives, separate website"_ 

image

maxgds (changes requested):
As per Chris' comments this needs some more thought around the headings.

app/views/organisations/_organisations_list.html.erb
jon-kirwan: Ah that is a good point. I'd changed the markup to use a heading because the document outline is currently broken for entries that don't use the organisation logo component but do contain 'works with' content.

validator w3 org_nu__doc=https%3A%2F%2Fwww gov uk%2Fgovernment%2Forganisations showoutline=yes (1)

As Max mentioned above, this should probably be tackled separately so will create a new card and revert this back to the previous markup.

jon-kirwan:
Placeholder card added for the headings: https://trello.com/c/jhnpLsS4/1160-document-outline-issues-on-collection-pages

app/views/organisations/_organisations_list.html.erb
chris-gds: I've just seen this - we should remove this entirely inline with guidance.

_[External links](https://design-system.service.gov.uk/styles/typography/#links)
If it’s an external link to a non-government website, make that clear in the link text. For example, ‘read advice on writing link text from [name of organisation]’. There’s no need to say explicitly that you’re linking to an external site. [Do not use an external link icon](https://designnotes.blog.gov.uk/2016/11/28/removing-the-external-link-icon-from-gov-uk/)._

maxgds: They're not even actual links to the external site... I think we should bring this to Adam's attention before going ahead though - might need to do comms around it first.

maxgds: Let's bundle this into the headings card to address separately from this PR.

jon-kirwan: Okay, I won't make the changes now but have captured here: https://trello.com/c/jhnpLsS4/1160-document-outline-issues-on-collection-pages

jon-kirwan:
> ... and check whether the pattern exists anywhere else we should update along with this.

I've checked for other occurrences in `collections` and the only similar use I can find is here: https://www.gov.uk/government/technology-guidance-technology-guidance-open-standards
The same Gem Toggle script is being used but with a different pattern so I think no need to make any further updates.

config/locales/gu/organisations.yml
```diff
@@ -331,8 +331,6 @@ gu:
       special_health_authority:
       sub_organisation:
       tribunal: ટ્રિબ્યુનલ
-    view_all: બધા જુઓ
```
chris-gds, line 334: @maxgds / @jon-kirwan I'm thinking - this content translation work has already been done for this phrase. Might it make sense to note to update the details component with this?  Or even just capture this for when localisation is prioritised?

maxgds: I'm not sure I follow, Chris? At a glance the details component looks like it requires a string input and doesn't have a default. Am I missing something?

chris-gds: 🤔 Maybe I am, let me explain in more detail -  so the details element here on `collections` before this change is bespoke, however the phrase "view all" has already had Content translation work done to pull in the different text should the language be changed

As we're swapping to the gem version what happens with all these translations. How does a user see `બધા જુઓ`?

maxgds: So any app that has copy in it should have had the copy put into a translation file and the translations should have been sourced and updated for all relevant languages. I'd be surprised if the phrase "view all" doesn't appear anywhere else across GOV.UK but it's possible. If that's the case, yes, we're throwing away some translation work with this deletion but that's better than leaving it to clutter up the yaml files when it's not needed. I guess we could capture this deletion by linking to this PR in a doc somewhere in case it ever needs to be reimplemented, but I'm not sure we'd be able to communicate that to everyone who might ever remove a translation key and we don't know for sure that we'll ever need it again. And when we look to add "view all" in future we *could* look for it in a linked doc, but it's likely people will forget to... I think we'd just be adding overhead if we were to try to capture this so favour just letting it go, to be honest.

chris-gds: Fair - also it's in git so if someone does go looking (they might not find it) but it's there.

jon-kirwan: Thanks, i'll make sure to capture this in a separate commit too which might it a bit easier to find if needed.

jon-kirwan:
Commit history and screenshots both now updated. @maxgds - could you please merge this in if otherwise all looks okay?

PR #2672, opened 2022-02-07
Spike to visualise browse powered by topics

hannako:
Things to note:

- mainstream_browse_page content items contain a list of all top level browse categories in the attribute
`top_level_browse_pages`. But topic content_items do not contain a list of all the topic categories. I've
added an additional page fetch for the "/topic" content item which contains all the topic categories in
it's "children" attribute but obviously that's just a hack, we should add the list to all topic content items
when they're published.

- mainstram_browse_page content items contain a representation of their parent, children, and siblings.
Topics only contain their parent and child categories. To get this prototype functioning, I've added an addiontal
page fetch. ie when we are on browsetopics/topic/subtopic, the subtopic content item doesn't know who it's siblings are.
So I fetch the parent in order to get the list of siblings. We should add the siblings for subtopics to all topic
content items when they're published.

- There's a concept of links having a curated order, or an alphabetical order. This prototype does not respect the curated
order, because the second_level_ordering attribute is not present on the topic content item.

- Not all topics have a populated details["groups"] attribute.
eg topic/benefits-credits/child-benefit does
but topic/business-enterprise/business-auditing-accounting-reporting

If the groups aren't set, we just show anything tagged to the topic via
https://github.com/alphagov/collections/blob/ac1003904ef6770778ec606e80da14a370e42d7c/app/models/list_set.rb#L43

This is the same behaviour as on /browse at the moment. But just FYI.

- Trello https://trello.com/c/uux5bshi/756-prototype-a-representation-of-a-browse-topic-from-within-specialist-topic

## Review App
https://govuk-collec-topic-spik-hy1lrn.herokuapp.com/browsetopics/benefits-credits

### DIFF
--- a/app/controllers/browsetopics_controller.rb
+++ b/app/controllers/browsetopics_controller.rb
@@ -1,0 +1,51 @@
+class BrowsetopicsController < ApplicationController
+  enable_request_formats show: [:json]
+
+  def index
+    page = MainstreamBrowseTopicPage.find("/topic")
+    setup_content_item_and_navigation_helpers(page)
+
+    render :index,
+           locals: {
+             page: page,
+           }
+  end
+
+  def show
+    page =
+      MainstreamBrowseTopicPage.find("/topic/#{params[:topic_slug]}")
+    setup_content_item_and_navigation_helpers(page)
+
+    respond_to do |f|
+      f.html do
+        show_html(page)
+      end
+      f.json do
+        render json: {
+          content_id: page.content_id,
+          navigation_page_type: "First Level Browse",
+          breadcrumbs: breadcrumb_content,
+          html: second_level_browse_pages_partial(page),
+        }
+      end
+    end
+  end
+
+private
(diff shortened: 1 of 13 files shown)


chao-xian:
Thank you so much Jess! Are the first two hacks you listed only for supporting the Miller Column functionality on `/browse`?

hannako:
> Thank you so much Jess! Are the first two hacks you listed only for supporting the Miller Column style functionality on `/browse`? As in showing everything as part of the page?
> 
> As in this: [5980208#diff-009eff9981e11379157ca9d5d331759042ea3dd36ae470694588b52278ffd803R39-R45](https://github.com/alphagov/collections/commit/598020825fb1880f15305dbfbfc6f1ba34de9d17#diff-009eff9981e11379157ca9d5d331759042ea3dd36ae470694588b52278ffd803R39-R45)

@chao-xian Exactly

PR #2630, opened 2021-12-15
Add intervention banner on specific mainstream browse and specialist topics

chao-xian:
⚠️ This repo is Continuously Deployed: make sure you [follow the guidance](https://docs.publishing.service.gov.uk/manual/development-pipeline.html#merge-your-own-pull-request) ⚠️

## Note 

Depends on https://github.com/alphagov/govuk_publishing_components/pull/2516 (tests will fail until that is released)

## What

Adds the intervention banner component onto select topic pages and the Miller Columns.

## Why

We want to lead users of these specific topics to an online tree testing exercise.

## Visual changes

https://user-images.githubusercontent.com/424772/146251828-0fa2d84a-c7ac-4cb3-afdb-f5a7696f4b36.mov

https://user-images.githubusercontent.com/424772/146252361-4c782b3a-0346-4d2e-bb2c-5be4b1b4f09e.mov

localhost_3070_topic_dealing-with-hmrc(Moto G4)
localhost_3070_browse_tax_capital-gains(Moto G4)
localhost_3070_browse_benefits_help-for-carers(Moto G4)

PR #2683, opened 2022-02-25
Add recruitment banner on specific mainstream browse pages

hannako:

This PR is entirely based on the excellent work of @chao-xian https://github.com/alphagov/collections/pull/2630

Final commit wip.

https://trello.com/c/qPhlRiUx/779-launch-and-monitor-the-recruitment-banner-for-money-tax

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -26,6 +26,7 @@ $govuk-new-link-styles: true;
 @import 'govuk_publishing_components/components/image-card';
 @import 'govuk_publishing_components/components/input';
 @import 'govuk_publishing_components/components/inset-text';
+@import 'govuk_publishing_components/components/intervention';
 @import 'govuk_publishing_components/components/inverse-header';
 @import 'govuk_publishing_components/components/label';
 @import 'govuk_publishing_components/components/lead-paragraph';
--- a/app/views/second_level_browse_page/_links.html.erb
+++ b/app/views/second_level_browse_page/_links.html.erb
@@ -1,6 +1,15 @@
 <div class="browse__inner <%= page.lists.curated? ? 'browse__inner--curated' : 'browse__inner--alphabetical' %>">
   <h2 tabindex="-1" class="browse__heading js-heading"><%= t("shared.browse.prefix") %> <%= page.title %></h2>
 
+  <% if @show_recruitment_banner %>
+    <%= render "govuk_publishing_components/components/intervention", {
+        suggestion_text: "Help improve GOV.UK",
+        suggestion_link_text: "Take part in user research",
+        suggestion_link_url: "https://gdsuserresearch.optimalworkshop.com/treejack/lb5eu75l",
+        new_tab: true,
+    } %>
+  <% end %>
+
   <% page.lists.each_with_index do |list, section_index| %>
     <% if page.lists.curated? %>
       <h3 class="browse__list-header"><%= list.title %></h3>
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -6,6 +6,7 @@
 //= require govuk_publishing_components/components/details
 //= require govuk_publishing_components/components/feedback
 //= require govuk_publishing_components/components/govspeak
+//= require govuk_publishing_components/components/intervention
 //= require govuk_publishing_components/components/radio
 //= require govuk_publishing_components/components/step-by-step-nav
 //= require govuk_publishing_components/components/error-summary
--- a/app/controllers/second_level_browse_page_controller.rb
+++ b/app/controllers/second_level_browse_page_controller.rb
@@ -1,8 +1,11 @@
 class SecondLevelBrowsePageController < ApplicationController
+  include RecruitmentBannerHelper
   enable_request_formats show: [:json]
 
   def show
     setup_content_item_and_navigation_helpers(page)
+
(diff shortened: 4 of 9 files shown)


app/assets/stylesheets/application.scss
```diff
@@ -26,6 +26,7 @@ $govuk-new-link-styles: true;
 @import 'govuk_publishing_components/components/image-card';
 @import 'govuk_publishing_components/components/input';
 @import 'govuk_publishing_components/components/inset-text';
+@import 'govuk_publishing_components/components/intervention';
```
andysellick, line 29: The intervention component also has a JavaScript file, I'd recommend including that as well in this commit.

app/views/second_level_browse_page/_links.html.erb
```diff
@@ -1,6 +1,14 @@
 <div class="browse__inner <%= page.lists.curated? ? 'browse__inner--curated' : 'browse__inner--alphabetical' %>">
   <h2 tabindex="-1" class="browse__heading js-heading"><%= t("shared.browse.prefix") %> <%= page.title %></h2>
 
+  <%= render "govuk_publishing_components/components/intervention", {
+    suggestion_text: "Help improve GOV.UK",
+    suggestion_link_text: "Take part in user research",
```
andysellick, line 6: I'm guessing there aren't translations yet for the `suggestion_text` or `suggestion_link_text` but could this text come from a translations file please?

app/assets/javascripts/browse-columns.js
```diff
@@ -190,6 +191,7 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
     this.updateBreadcrumbs(state)
 
     this.changeColumnVisibility(3)
+    this.showBanner(state.slug)
```
andysellick, line 194: This call is in the `showSubsection` function but not the `showSection` function. I don't have a complete picture yet of the intended functionality here but just wanted to check that this was only needed here.

app/views/second_level_browse_page/_links.html.erb
```diff
@@ -1,6 +1,14 @@
...
+    suggestion_link_url: "https://gdsuserresearch.optimalworkshop.com/treejack/35gap0ij",
```
hannako, line 7: This needs to be updated to https://GDSUserResearch.optimalworkshop.com/treejack/lb5eu75l

app/assets/javascripts/browse-columns.js
```diff
@@ -303,5 +305,16 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
     }
   }
 
+  BrowseColumns.prototype.showBanner = function (slug) {
```
andysellick, line 308: I'm confused why this needs JavaScript at all, but perhaps I've missed something so happy to discuss.

The logic to show/hide the intervention component is already present in the backend, and the JS gets its data from that same place (via JSON in an AJAX call, which includes a chunk of markup for the column to be rendered, which includes/doesn't include the intervention component as appropriate).

As an experiment, I checked out this branch locally, removed this JS and modified `_links.html.erb` as follows:

```
  <% if !@hide_recruitment_banner %>
    <%= render "govuk_publishing_components/components/intervention", {
      suggestion_text: "Help improve GOV.UK",
      suggestion_link_text: "Take part in user research",
      suggestion_link_url: "https://gdsuserresearch.optimalworkshop.com/treejack/35gap0ij",
      new_tab: true,
    } %>
  <% end %>
```

...and it worked as expected, showing the intervention banner only where it was required (and persisting through the JS cache that's part of the browse JS also as required).

(might need a bit of refactoring in this PR to change `if !@hide_recruitment_banner` to something clearer like `if @show_recruitment_banner`)

app/assets/javascripts/browse-columns.js
chao-xian: Removed the JS - thanks Andy

app/assets/javascripts/browse-columns.js
chao-xian: Yup trying this out now. Will refactor the name in a bit.

app/assets/stylesheets/application.scss
chao-xian: Done

app/views/second_level_browse_page/_links.html.erb
chao-xian: I'd normally agree to do this, but we then have to add keys with missing translations to _all_ of the non-en files as [this test expects](https://github.com/alphagov/collections/blob/main/spec/lib/locales_validation_spec.rb). As the banner is temporary, can we get away with not?

app/assets/javascripts/browse-columns.js
chao-xian: Done

app/views/second_level_browse_page/_links.html.erb
andysellick: Ah, yeah that's a bit of a pain, I hadn't realised that. In that case yes, that's fine.

chao-xian: Thank you!

chao-xian:
Works on integration and locally - something weird with Heroku

andysellick (approved):
We can't reproduce the error I found anywhere except in the Heroku preview, so I'm going to approve this for now to avoid blocking.

PR #2690, opened 2022-03-04
Add employing people topic to tree study

chao-xian:
Follows on from #2683. Adding in one more page to display intervention banner to direct users to a tree study for User Research.

[Trello](https://trello.com/c/i6PqMLdp/842-add-banner-to-guidance-pages)

### DIFF
--- a/app/helpers/recruitment_banner_helper.rb
+++ b/app/helpers/recruitment_banner_helper.rb
@@ -1,7 +1,16 @@
 module RecruitmentBannerHelper
-  TOPICS = ["/browse/business", "/browse/tax"].freeze
+  STUDY_URLS_FOR_TOPICS = {
+    "/browse/business" => "https://GDSUserResearch.optimalworkshop.com/treejack/lb5eu75l",
+    "/browse/tax" => "https://GDSUserResearch.optimalworkshop.com/treejack/lb5eu75l",
+    "/browse/employing-people" => "https://GDSUserResearch.optimalworkshop.com/treejack/724268fr",
+  }.freeze
 
   def show_banner?(path)
-    path.starts_with?(TOPICS.first) || path.starts_with?(TOPICS.last)
+    STUDY_URLS_FOR_TOPICS.keys.any? { |topic| path.starts_with?(topic) }
+  end
+
+  def study_url_for(topic)
+    parent_topic = topic.rpartition("/").first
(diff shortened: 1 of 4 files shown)


app/helpers/recruitment_banner_helper.rb
```diff
@@ -1,7 +1,17 @@
...
-  TOPICS = ["/browse/business", "/browse/tax"].freeze
+  TOPICS = ["/browse/business", "/browse/tax", "/browse/employing-people"].freeze
 
   def show_banner?(path)
-    path.starts_with?(TOPICS.first) || path.starts_with?(TOPICS.last)
+    TOPICS.each do |topic|
+      return true if path.starts_with?(topic)
+    end
+
+    false
```
AgaDufrat, line 9:
```suggestion
TOPICS.any? { |topic| path.starts_with?(topic) }
```
I think this logic could be captured in a slightly shorter manner using `.any?` if it's easily understandable for other dev.

AgaDufrat (approved):
Looks good to me! :shipit: 

What do you think about refactoring the Helper into something like this:
```
module RecruitmentBannerHelper
  STUDY_URLS_FOR_TOPICS = {
    "/browse/business" => "https://GDSUserResearch.optimalworkshop.com/treejack/lb5eu75l",
    "/browse/tax" => "https://GDSUserResearch.optimalworkshop.com/treejack/lb5eu75l",
    "/browse/employing-people" => "https://GDSUserResearch.optimalworkshop.com/treejack/724268fr",
  }.freeze

  def show_banner?(path)
    STUDY_URLS_FOR_TOPICS.keys.any? { |topic| path.starts_with?(topic) }
  end

  def study_url_for(topic)
    STUDY_URLS_FOR_TOPICS[topic]
  end
end
```

I think the constant (name tbc) documents it a bit better so we can see the topic with corresponding study URL. 
It means we don’t have to rely on the order of elements in the array (TOPICS.last). 
I think it’ll make it easier to delete and add study topics in the future. When adding a new one we won't need to work out the logic for getting the correct url.

app/helpers/recruitment_banner_helper.rb
chao-xian: I thought exactly this in the middle of the night! Thanks Aga

chao-xian:
Thanks Aga I did think a hash might be better but was in a rush.

PR #2696, opened 2022-03-09
Add new topic level templates and tracking

hannalaakso:
🚧 Work in progress 🚧

**This PR now contains all the work from #2673,  #2709, #2740, #2737, #2782, #2778, #2774, and #2773**

## What

Adds the new topic page templates for the 2nd (eg `/browse/benefits/[topic]`) level of the topics.

This is hidden behind a flag to fake the A/B test mechanism - append `?b=true` to the URL to view the new topic pages.

[Trello card](https://trello.com/c/FY0o23pp/762-build-the-level-0-1-and-2-browse-templates-grid-accordion-a-z)

## Why

The topic pages are undergoing a set of A/B tests, and the templates need to be present and usable for this test to take place.

## Visual differences

[Review app while Heroku + Github incident is ongoing](https://explore-collections-browse.herokuapp.com/browse?b=true)

...

---

⚠️ This repo is Continuously Deployed: make sure you [follow the guidance](https://docs.publishing.service.gov.uk/manual/development-pipeline.html#merge-your-own-pull-request) ⚠️

PR #2709, opened 2022-03-22
Add accordion tracking

chris-gds:
## What

Switch on feature flags for click event tracking for the Accordion on the new [Level 2 Browse Templates ](https://github.com/alphagov/collections/pull/2696)

## Why 

When cookies are accepted and "Show all sections" or a section on the Accordion is clicked then send this event to Google Analytics (GA).

## Visuals 

NA

## Done when

- [x] [Accordion](https://github.com/alphagov/govuk_publishing_components/pull/2693) updated
- [ ] Dependancy in `collections`

### Capture for future update

- [ ] Update other instances
- [ ] Remove bespoke tracking

### DIFF
--- a/app/views/second_level_browse_page/_new_links.html.erb
+++ b/app/views/second_level_browse_page/_new_links.html.erb
@@ -1,10 +1,25 @@
-<ul class="govuk-list">
+<%
+  list_tracking_values = []
+%>
+<ul class="govuk-list" data-module="gem-track-click">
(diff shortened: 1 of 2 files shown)


hannalaakso (approved):
This is a good WIP solution the problem. There are still some things that need to be done like the link tracking and splitting out some of the attributes, but we can address this https://github.com/alphagov/collections/pull/2696 which this will be merged into.

PR #2711, opened 2022-03-23
Backend logic for the Browse AB test

hannako:
## Add AB test logic for the new browse templates

In https://github.com/alphagov/collections/pull/2696 temporary logic was added to the browse controllers so that the new templates are only shown if the url contains a query string with the b param. 

This PR replaces that temp logic with working AB test code.

Currently the CDN config has not been updated to add the header that the logic here is using to switch between variants. As it's absent, all users will be seeing the A (control) variant. We are doing this so that we can validate all is well before we activate the AB test and put it in front of real users. 

### How to view the B variant before we turn the AB test on
To toggle between the variants install [modheader](https://chrome.google.com/webstore/detail/modheader/idgpnmonknjnojddfkpgkljpfnnfcklj?hl=en) and then add a new **Request Header** by clicking the **+**. Enter a `Name` of **govuk-abtest-newbrowse** and a `Value` of **B**

---

- Trello card: https://trello.com/c/ACFzuepy/850-prep-prs-for-ab-test-of-browse-pages

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,11 +1,16 @@
 class ApplicationController < ActionController::Base
   include Slimmer::Template
+  include AbTests::NewBrowseAbTestable
+
+  helper_method :new_browse_variant
+  helper_method :page_under_test?
   helper_method :content_item_h
 
   protect_from_forgery with: :exception
 
   before_action :set_expiry
   before_action :restrict_request_formats
+  before_action :set_new_browse_response_header_if_browse_page
 
   slimmer_template "gem_layout"
 
@@ -67,6 +72,10 @@ private
     return true if format == Mime[:html] || format == Mime::ALL
 
     format && self.class.acceptable_formats.fetch(params[:action].to_sym, []).include?(format.to_sym)
+  end
+
+  def set_new_browse_response_header_if_browse_page
+    set_new_browse_response_header if page_under_test?
   end
 
   def error_403
--- a/app/controllers/browse_controller.rb
+++ b/app/controllers/browse_controller.rb
@@ -6,19 +6,12 @@ class BrowseController < ApplicationController
     @dimension26 = 1
     @dimension27 = page.top_level_browse_pages.count || 0
     setup_content_item_and_navigation_helpers(page)
-    slimmer_template "gem_layout_full_width" if is_variant_b?
-
-    template = is_variant_b? ? :new_index : :index
-
-    render(template, locals: {
-      page: page,
-    })
+    index_html(page)
   end
 
   def show
     page = MainstreamBrowsePage.find("/browse/#{params[:top_level_slug]}")
     setup_content_item_and_navigation_helpers(page)
-    slimmer_template "gem_layout_full_width" if is_variant_b?
 
     respond_to do |f|
       f.html do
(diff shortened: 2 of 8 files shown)


app/controllers/browse_controller.rb
```diff
@@ -36,10 +32,21 @@ def show
...
+    render template, locals: { page: page }
+  end
+
+  def index_html(page)
+    template = :index
+    if new_browse_variant_b?
+      slimmer_template "gem_layout_full_width"
+      template = :new_index
+    end
+    render template, locals: { page: page }
```
hannako, line 49: These are itching to be refactored to remove the duplication. And the slimmer logic could also be abstracted into a module and included into both controllers. I have not done that because:
- It's easy to read and understand.
- It was a bit buggy to move the slimmer code out, and only include it in the browse controllers. It lead to some weird state persistence between requests so the relevant layout wasn't consistently picked up. Including the module in the application controller avoided this issue, but then the module is available to ALL the controllers unnecessarily.

spec/features/mainstream_browsing_spec.rb
```diff
@@ -29,4 +30,46 @@
       expect(page).to have_selector(".gem-c-breadcrumbs")
     end
   end
+
+  context "for AB tests" do
+    scenario "AB testing of new browse templates" do
+      with_variant NewBrowse: "B" do
+        schemas = GovukSchemas::Example.find_all("mainstream_browse_page").shuffle
+
+        schemas.each do |content_item|
```
chao-xian: Nice!

app/controllers/browse_controller.rb
chao-xian: A small amount of duplication is ok and these methods are small and simple. In this case being explicit is good.

app/views/browse/new_index.html.erb
```diff
@@ -0,0 +1,2 @@
+# Replace with Ian/Chris's level 0 template
```
chao-xian, line 1: Can we prefix these with `TODO:`?

spec/support/mainstream_browse_helpers.rb
```diff
@@ -0,0 +1,37 @@
+module MainstreamBrowseHelpers
```
chao-xian, line 1: Lovely!

chao-xian (approved):
Lovely work. Is it worth (even feasible?) to add a test for Z variant?

spec/features/mainstream_browsing_spec.rb
hannako: I can't claim the glory for the shuffle, pinched it from a test above! 😂

chao-xian (approved):
One minor comment but otherwise looks good!

app/controllers/application_controller.rb
```diff
@@ -69,6 +74,14 @@ def can_handle_format?(format)
     format && self.class.acceptable_formats.fetch(params[:action].to_sym, []).include?(format.to_sym)
   end
 
+  def browse_page?
+    request.path.starts_with?("/browse")
```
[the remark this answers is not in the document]
hannako: https://github.com/alphagov/collections/pull/2711/commits/07c0f6d257ebf7c20c498c1a0dcd8b38ef83b38d

PR #2717, opened 2022-03-28
Stop using the custom page for Brexit

sihugh:
This will now use an override URL that we'll put on the Brexit taxon, so it will redirect to the new collections page.

We no longer need any of the custom bits for Brexit in this app.


alex-ju (approved):
Lovely spring clean-up!

hannako (approved):
Well this is VERY nice. I just pulled the branch and grep for Brexit.... ZERO RESULTS! WOOOO

PR #2736, opened 2022-04-05
New index new show use partials

patrickpatrickpatrick:
### What

Create partials for several sections of the new pages that were being repeated

### Why

Tidy up the code somewhat and make it more maintainable 

### Visual Differences 

None, should be the same


chao-xian:
Superseded by https://github.com/alphagov/collections/pull/2737

PR #2737, opened 2022-04-05
Use partials for topic pages

patrickpatrickpatrick:
### What

Create partials for several sections of the new pages that were being repeated

### Why

Tidy up the code somewhat and make it more maintainable 

### Visual Differences 

None, should be the same

### DIFF
--- a/app/views/application/_breadcrumbs.html.erb
+++ b/app/views/application/_breadcrumbs.html.erb
@@ -1,7 +1,4 @@
-<% inverse ||= false %>
-
 <%= render 'govuk_publishing_components/components/breadcrumbs', {
   breadcrumbs: Breadcrumbs.new(@content_item).breadcrumbs,
-  collapse_on_mobile: true,
-  inverse: inverse
+  collapse_on_mobile: true
 } %>
--- a/app/views/application/_breadcrumbs.html.erb
+++ b/app/views/application/_breadcrumbs.html.erb
@@ -1,4 +1,7 @@
+<% inverse ||= false %>
+
 <%= render 'govuk_publishing_components/components/breadcrumbs', {
   breadcrumbs: Breadcrumbs.new(@content_item).breadcrumbs,
-  collapse_on_mobile: true
+  collapse_on_mobile: true,
+  inverse: inverse
 } %>
--- a/app/views/browse/new_show.html.erb
+++ b/app/views/browse/new_show.html.erb
@@ -18,20 +18,14 @@
   } %>
 <% end %>
 
-<div class="topic-page__header-wrapper">
-  <div class="govuk-width-container">
-    <div class="govuk-grid-row">
-      <div class="govuk-grid-column-full">
-        <h1 class="topic-page__heading govuk-heading-xl">
-          <%= page.title %>
-        </h1>
-        <p class="gem-c-lead-paragraph gem-c-lead-paragraph--inverse govuk-!-margin-bottom-2">
-          <%= page.description %>
-        </p>
-      </div>
-    </div>
-  </div>
-</div>
+<%= render "shared/topic_page_header" do %>
+  <h1 class="topic-page__heading govuk-heading-xl">
+    <%= page.title %>
+  </h1>
+  <p class="gem-c-lead-paragraph gem-c-lead-paragraph--inverse govuk-!-margin-bottom-2">
+    <%= page.description %>
+  </p>
+<% end %>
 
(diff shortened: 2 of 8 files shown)


app/views/application/_breadcrumbs.html.erb
```diff
@@ -1,4 +1,7 @@
+<% inverse ||= false %>
```
chao-xian, line 1: With a caveat that I'm missing a ton of context coming into this fairly cold, but I'm not clear why you've set this default, and it also doesn't seem to be relevant to the intent of the commit of creating new partials - could you split this into a separate commit with more explanation around the change?

app/views/browse/new_show.html.erb
```diff
@@ -13,24 +13,9 @@
...
-            ],
-            collapse_on_mobile: true,
-            inverse: true,
-          } %>
-        </div>
-      </div>
-    </div>
-  </div>
+  <%= render "shared/topic_page_breadcrumbs", {
+    two_thirds: true
```
chao-xian, line 17: Should this be `column_two_thirds`?

app/views/second_level_browse_page/new_show.html.erb
```diff
@@ -11,48 +11,24 @@
...
-      <div class="govuk-grid-row">
-        <div class="govuk-grid-column-full">
-          <%= render "govuk_publishing_components/components/breadcrumbs", {
-            breadcrumbs: [
-              {
-                title: "Home",
-                url: "/"
-              },
-              {
-                title: page.active_top_level_browse_page.title,
```
chao-xian, line 25: I may have missed a step but I can't see how this data gets passed to the new partials now?

app/views/shared/_topic_page_breadcrumbs.html.erb
```diff
@@ -0,0 +1,18 @@
...
+  column_two_thirds ||= false
+
+  column_classes = %w()
+  column_classes << column_two_thirds ? "govuk-grid-column-full" : "govuk-grid-column-two-thirds"
+%>
+
+<div class="topic-page__breadcrumb-wrapper">
+  <div class="govuk-width-container">
+    <div class="govuk-grid-row">
+      <div class="<%= column_classes.join(' ') %>">
```
hannalaakso, line 11: This outputs `false` instead of the class name.

app/views/shared/_topic_page_header.html.erb
```diff
@@ -0,0 +1,20 @@
...
+   header_wrapper_classes << "govuk-!-margin-bottom-#{margin_bottom}" if !margin_bottom.zero?
+
+   column_classes = %w()
+   column_classes << two_thirds ? "govuk-grid-column-two-thirds" : "govuk-grid-column-full"
+%>
+
+<div class="<%= header_wrapper_classes.join(' ') %>">
+  <div class="govuk-width-container">
+    <div class="govuk-grid-row">
+      <div class="<%= column_classes.join(' ') %>">
```
hannalaakso, line 15: This outputs `false` instead of the class name.

hannalaakso (commented):
This is a really nice tidy up of the duplicated markup, thanks @patrickpatrickpatrick 🙌  Just noticed a small issue with the class name rendering.

app/views/browse/new_show.html.erb
patrickpatrickpatrick: Yes, thanks for catching that!

app/views/shared/_topic_page_breadcrumbs.html.erb
patrickpatrickpatrick: was missing some parenthesis, should be working now!

app/views/shared/_topic_page_header.html.erb
patrickpatrickpatrick: was missing some parenthesis, should be working now!

app/views/second_level_browse_page/new_show.html.erb
chao-xian: Huh so it does. @hannalaakso are you able to shed any light on that? Why we were using `page.active_top_level_browse_page.title`?

hannalaakso: I think this was on me, I was mirroring the hardcoded breadcrumbs on the new level 1 templates and assumed that we needed to explicitly pass through the links for the items. @patrickpatrickpatrick's solution is definitely more robust.

PR #2740, opened 2022-04-08
split analytics dimension for browse templates

patrickpatrickpatrick:
## What

This PR is for splitting the navigation_page_type dimension into navigation_page_type and navigation_list_type.

## Why 
These changes were requested to make analytics better.

## Visual Changes
There should be none.

### DIFF
--- a/app/views/browse/new_show.html.erb
+++ b/app/views/browse/new_show.html.erb
@@ -5,8 +5,8 @@
 <% content_for :meta_tags do %>
   <%= render "govuk_publishing_components/components/meta_tags", {
     content_item: {
-      # TODO align this naming "level 1" vs "second_level" mismatch
-      navigation_page_type: "Browse level 1 - #{page.second_level_pages_curated? ? "curated" : "default"}",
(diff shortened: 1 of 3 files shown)


patrickpatrickpatrick:
https://github.com/alphagov/govuk_publishing_components/pull/2722 needs to be merged also for changes to work

hannalaakso (approved):
I've tested this together with https://github.com/alphagov/govuk_publishing_components/pull/2722 and the correct dimensions are being set 🎉

app/views/browse/new_show.html.erb
```diff
@@ -6,7 +6,8 @@
   <%= render "govuk_publishing_components/components/meta_tags", {
     content_item: {
       # TODO align this naming "level 1" vs "second_level" mismatch
```
hannalaakso, line 8: Nit: I'm pretty sure we can remove this comment, I think the fix here is that this refers to.

patrickpatrickpatrick: ah yes good spot

PR #2750, opened 2022-04-13
Add a helper to generate accordion content markup

hannalaakso:
Add a helper to generate the accordion shaped markup (following documentation [here](https://components.publishing.service.gov.uk/component-guide/accordion)) instead of duplicating the code for doing it multiple times inside the new browse templates. 

### DIFF
--- a/app/views/second_level_browse_page/_new_second_level_browse_pages.html.erb
+++ b/app/views/second_level_browse_page/_new_second_level_browse_pages.html.erb
@@ -22,17 +22,7 @@
 
   <%# Generate accordion shaped markup %>
   <% if curated_order %>
-    <%
-      additional_contents = {
-        heading: {
-          text: list["title"],
-        },
-        content: {
-          html: contents,
-        },
-      }
-      accordion_contents << additional_contents
-      %>
+    <% accordion_items(section_title: list["title"], section_contents: contents, accordion_contents: accordion_contents) %>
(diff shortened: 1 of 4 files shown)


app/views/second_level_browse_page/_new_second_level_browse_pages.html.erb
```diff
@@ -22,17 +22,7 @@
...
-        heading: {
-          text: list["title"],
-        },
-        content: {
-          html: contents,
-        },
-      }
-      accordion_contents << additional_contents
-      %>
+    <% accordion_contents = accordion_items(list["title"], accordion_contents, contents) %>
```
patrickpatrickpatrick, line 25: If assignment is taking place here still, maybe the concatenation could just happen here instead of it being in the helper function.

hannalaakso: I've simplified the code a bit to remove the need to do the assignment in the template, have a look.

patrickpatrickpatrick: yeah that's looking better!

patrickpatrickpatrick: yeah that's looking better!

patrickpatrickpatrick (approved):

PR #2765, opened 2022-04-28
Enable CI in GitHub Actions

theseanything:
This enables a GitHub Action workflow to run CI using a re-usableworkflow template in govuk-infrastructure repo. This is the initial set up to test the reusable workflow.

This PR also prevents the deploy workflow from being run before the application tests have been completed. This is to ensure that the commit passes application tests before being deployed.


fredericfran-gds (approved):
ok

PR #2772, opened 2022-05-04
Add tracking fixes

hannalaakso:
This PR:
- corrects `dimension1` for Level 1 pageview to be `<level 1 topic name>` eg `births, deaths, marriages and care`
- removes a `gem-track-click` attribute from the accordion wrapper as it's already set on the lists and was making the link click events fire twice; keeping it on the lists (instead of the wrapper) to be more explicit about what it impacts since the 'Show all' button also sets its (button click) tracking with a separate attribute inside the accordion
- removes explicitly setting `location` or `title` on the accordion link clicks; this makes `Document location URL` and `Document Title` flow from the pageview directly so that they are the URL and title of the page the accordion is on (so similar to the current homepage implementation)

### DIFF
--- a/app/views/browse/new_show.html.erb
+++ b/app/views/browse/new_show.html.erb
@@ -7,7 +7,7 @@
     content_item: {
       navigation_page_type: "Browse level 1",
       navigation_list_type: page.second_level_pages_curated? ? "curated" : "default",
-      section: yield(:title).downcase.sub!('browse: ', ''),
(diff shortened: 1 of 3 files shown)


patrickpatrickpatrick:
looks good to me

patrickpatrickpatrick (approved):

chao-xian (approved):

PR #2773, opened 2022-05-04
Split 2nd level topic into two templates

injms:
## What

Adds separate templates for the two types of second level browse page - one for alphabetically listed topics, one for curated topics.

## Why

The code needed to bend a single template was causing complexity which could be removed by having two separate templates - there were enough conditionals inside the template to lead to the conclusion that it was simpler to have two separate templates for each type of second level topic page.

The logic for deciding which template to use now sits in the controller, which means that the number of branching `if` statements used in the template has drastically reduced.

## Visual changes

None.

---


chao-xian (approved):
LGTM thanks for doing this Ian!

PR #2778, opened 2022-05-09
Add blue bar for users with JavaScript available

injms:
## What
Switches the blue bar back on for people who have JavaScript available.

## Why

When the site-wide coronavirus banner was enabled, there was a bit of jigger-pokery needed to ensure that the blue bar that's attached to just below the navigation header was not visible when the coronavirus banner was visible. The coronavirus banner has been removed - so this isn't needed any longer.

Sidenote - this does raise an issue that the blue bar isn't implemented consistently across different applications - for example, the new topics pages have a full width dark header which require a constrained width blue bar; a taxon (eg [transport]) has a full width dark header that doesn't require a blue bar at all; the [coronavirus page] has full width yellow bar (the colour is different, but the placement and dimensions are the same as the blue coloured bar). Mentioning this problem here as it's worth noting but out of scope for this task.

## Visual changes

Before:
image

After:
image

---

[transport]: https://www.gov.uk/transport
[coronavirus page]: https://www.gov.uk/coronavirus


hannalaakso (approved):
Looks good, thanks @injms 👍  I'm assuming this also fixes https://trello.com/c/BhTAmYwG/351-fix-blue-bar-missing-from-below-top-nav-on-new-templates ?

injms:
> Looks good, thanks @injms 👍 I'm assuming this also fixes https://trello.com/c/BhTAmYwG/351-fix-blue-bar-missing-from-below-top-nav-on-new-templates ?

Yes, it fixes that layout bug as well.

PR #2781, opened 2022-05-10
Use new_links partial in the refactored 2nd level browse templates

patrickpatrickpatrick:
This PR addresses the suggested changes made in https://github.com/alphagov/collections/pull/2696#discussion_r864672508 and https://github.com/alphagov/collections/pull/2696#discussion_r830989865. 

Square brackets are now used in `_topic_page_breadcrumbs` partial. Where new_links could be used in the new templates, it has been used (with a small change made to accept a different value for `track_action` which is the only difference between the partial and the markup it is replacing.


hannalaakso (approved):
Looks good to me, thanks @patrickpatrickpatrick 👍

PR #2792, opened 2022-05-16
Update component asset imports

andysellick:

## What
Update the individual component asset imports.

- removing assets for components not needed as components have been removed (e.g. radios on the covid landing page)
- removing assets for components that are already included in static and were therefore duplicated

## Why
We've not done this for a while and a lot of components have been removed since then, which means that we're importing these assets unnecessarily, and inflating the page weight when we don't need to.

## Visual changes
None.

However, asset size has decreased. This is dev before (production is more compressed, so the change is smaller, but it's still noticeable) - 1.8MB.

Screenshot 2022-05-16 at 07 08 35

This is dev after - 1.5MB.

Screenshot 2022-05-16 at 07 34 40


injms (approved):
Good to go! (after a tiny rebase)

PR #2800, opened 2022-05-24
Update accordion GTM analytics tracking

andysellick:

## What
Adds data attributes to the accordion component on the /coronavirus page, which will be used to track clicks on the 'show/hide all' link.

Depends upon https://github.com/alphagov/govuk_publishing_components/pull/2786 but will fail silently without it.

## Why
We're testing the technical approach of implementing GTM. This is only going to be used for testing.

## Visual changes
None.

Trello card: https://trello.com/c/V0RFrSMK/43-test-technical-approach-to-tracking-on-tabs-and-accordions


gclssvglx (approved):
Looks great - nice one @andysellick 👍

PR #2804, opened 2022-05-25
Remove data-track-count from accordion component parameters

patrickpatrickpatrick:
## What

Remove data-track-count from the parameters of the `govuk_publishing_components` accordion.

## Why

Passing data-attributes to the accordion adds those attributes to the container of the accordions that the accordion component renders. This means that analytics will count the container itself and the accordions within the container instead of only counting the accordions. Therefore we need to remove it from the parameters of the accordion component but keep it on each of the accordion_contents to get the correct count of accordions on the page.

### DIFF
--- a/app/views/second_level_browse_page/new_show_curated.html.erb
+++ b/app/views/second_level_browse_page/new_show_curated.html.erb
@@ -85,7 +85,6 @@
           heading_level: 2,
           data_attributes: {
             module: "govuk-accordion",
-            track_count: "accordionSection"
           },
           items: accordion_contents,
           margin_bottom: 4,


hannako (commented):
I'm not sure I follow. Does it need removing from line 51 too then? Maybe you could you link to the code in the other partials?

patrickpatrickpatrick:
> I'm not sure I follow. Does it need removing from line 51 too then? Maybe you could you link to the code in the other partials?

When I added the data-attribute to [`govuk_publishing_components/components/accordion`](url) I misunderstood what it renders and where the data-attributes are appended to. I thought that passing it as a parameter would add it to each of the accordions it renders but it actually adds the attribute to a container element which the accordions reside in. This meant that the analytics would count the container itself as well as the accordions (which line 51 correctly adds the data-attribute to). I do remember removing it but it might have gotten lost at some point during refactoring of the partials in the large `add-topic-level-2-template` branch.

chao-xian (approved):

hannako:
> > I'm not sure I follow. Does it need removing from line 51 too then? Maybe you could you link to the code in the other partials?
> 
> When I added the data-attribute to [`govuk_publishing_components/components/accordion`](url) I misunderstood what it renders and where the data-attributes are appended to. I thought that passing it as a parameter would add it to each of the accordions it renders but it actually adds the attribute to a container element which the accordions reside in. This meant that the analytics would count the container itself as well as the accordions (which line 51 correctly adds the data-attribute to). I do remember removing it but it might have gotten lost at some point during refactoring of the partials in the large `add-topic-level-2-template` branch.

Ahhhh gotcha! Thank you for explaining :-)

PR #2815, opened 2022-06-01
Render document lists on Topical Event page

brucebolt:
[Trello card](https://trello.com/c/huPXst3c)

This PR is part of work to move rendering of Topical Events from Whitehall to Collections. It adds the ability to render the lists of related documents under a topical event, under the categories of 'publications', 'consultations' and 'announcements'. The logic for retrieving these is copied from Whitehall, with the rendering implementation simplified.

This screenshot shows these sections as rendered by collections:

Screenshot 2022-06-15 at 14 34 25

If you want to spin the service up locally via `govuk-docker-up app-live`, this is a good example of a topical event with lots of related documents that you can compare with the live site:

http://collections.dev.gov.uk/government/topical-events/budget-2021

vs 

http://gov.uk/government/topical-events/budget-2021

This also adds translations for topical event documents, with these defaulted to blank for all locales except English, as we do not currently support translations for topical events.

### DIFF
--- a/app/helpers/topical_events_helper.rb
+++ b/app/helpers/topical_events_helper.rb
@@ -1,0 +1,31 @@
+module TopicalEventsHelper
+  def search_url(content_type, slug)
+    base_path(content_type) + query_params(content_type, slug).to_param
+  end
+
+private
+
+  def base_path(content_type)
+    { publication: "/search/all?",
+      consultation: "/search/policy-papers-and-consultations?",
+      announcement: "/search/news-and-communications?" }[content_type]
+  end
+
+  def query_params(content_type, slug)
+    case content_type
+    when :publication
+      {
+        topical_events: [slug],
+      }
+    when :consultation
+      {
+        content_store_document_type: %w[open_consultations closed_consultations],
+        topical_events: [slug],
+      }
+    when :announcement
+      {
+        topical_events: [slug],
+      }
+    end
+  end
+end
--- a/app/models/topical_event.rb
+++ b/app/models/topical_event.rb
@@ -3,6 +3,7 @@ class TopicalEvent
 
   def initialize(content_item)
     @content_item = content_item
+    @documents_service = TopicalEventsDocuments.new(slug)
   end
 
   def self.find!(base_path)
@@ -81,5 +82,28 @@ class TopicalEvent
 
     advice = ContentItem.find!("/foreign-travel-advice/afghanistan").to_hash
     [advice.slice("base_path", "title")]
+  end
+
+  def publications
+    @publications ||= @documents_service.fetch_related_documents_with_format({ filter_format: "publication" })
+  end
+
+  def consultations
+    @consultations ||= @documents_service.fetch_related_documents_with_format({ filter_format: "consultation" })
+  end
+
+  def announcements
+    announcement_document_types = %w[
+      press_release
+      news_article
+      news_story
+      fatality_notice
+      speech
+      written_statement
+      oral_statement
+      authored_article
+      government_response
+    ]
+    @announcements ||= @documents_service.fetch_related_documents_with_format({ filter_content_store_document_type: announcement_document_types })
   end
 end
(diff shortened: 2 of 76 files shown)


app/helpers/topical_events_helper.rb
```diff
@@ -0,0 +1,11 @@
+module TopicalEventsHelper
+  def query_string_from_content_type(content_type, slug)
+    string_hash = {
+      publication: "/search/all?topical_events%5B%5D=",
+      consultation: "/search/policy-papers-and-consultations?content_store_document_type%5B%5D=open_consultations&content_store_document_type%5B%5D=closed_consultations&topical_events=",
+      announcement: "/search/news-and-communications?topical_events=",
+    }
+
+    string_hash[content_type] + slug
+  end
```
brucebolt, line 10: Generating the query string will probably be clearer if you create an array of parameters then use the `to_query` method to turn this into a query string, e.g.

```ruby
def search_url(content_type, slug)
  base_path(content_type) + query_params(content_type, slug).to_query
end

private

def base_path(content_type)
  case content_type
  when :publication
    "/search/all"
  when :consultations
    "/search/policy-papers-and-consultations"
  when :announcement
    "/search/news-and-communications"
  end
end

def query_params(content_type, slug)
  case content_type
  when :publication
    [
      topical_events: [slug],
    ]
  when :consultations
    [
      content_store_document_type: ["open_consultations", "closed_consultations"],
      topical_events: [slug],
    ]
  when :announcement
    [
      topical_events: [slug],
    ]
  end
end
```

app/models/topical_event.rb
```diff
@@ -63,4 +71,19 @@ def ordered_featured_documents
...
+  def fetch_related_publications
+    @documents_service.fetch_related_documents_with_format({ filter_format: "publication" })
+  end
+
+  def fetch_related_consultations
+    @documents_service.fetch_related_documents_with_format({ filter_format: "consultation" })
+  end
+
+  def fetch_related_announcements
+    announcement_document_types = %w[press_release news_article news_story fatality_notice speech written_statement oral_statement authored_article government_response]
```
brucebolt, line 86: A really small nitpick, but I think these would be more aesthetically pleasing if each item were on a separate line (and it'd make the Git diff clearer if we update this list in the future).

app/models/topical_event.rb
```diff
@@ -1,15 +1,23 @@
 class TopicalEvent
-  attr_reader :content_item
+  attr_reader :content_item, :publications, :consultations, :announcements
```
brucebolt, line 2: Is there a reason we need to use `attr_reader` for these three things, but for the other methods we access them directly?

brucebolt (commented):
This is looking really great.

I've done a very quick review and added a few comments, which are mostly stylistic.

app/helpers/topical_events_helper.rb
BeckaL: Yup really like this, have copied almost verbatim in what I'm about to push up :-)

app/models/topical_event.rb
BeckaL: So one reason for having them as currently is to cache the results of getting each document type since the method actually does the database call, as we need to access them twice in the views (once for determining whether we even display a documents section, once for rendering the partial). So didn't want to access the method directly. 

Does that answer your question or am I misunderstanding?

brucebolt: Cool, thanks for clarifying.

It's also possible to replicate the same behaviour by memoizing within the method itself, e.g.
```ruby
  def publications
    @publications ||= format_results(fetch_related_documents_with_format("publication"))
  end
```

brucebolt (commented):
Thanks for making these changes @BeckaL.  This all looks great now.  I've left a non-blocking comment about memoization within the method, but not fussed either way really.

The approve option is disabled as I opened the PR, so maybe someone from Find & View could approve this when they look?

app/models/topical_event.rb
brucebolt: It looks like the memoization within the method is used quite widely within this app, so I'd probably go with that to retain consistency.

chao-xian (approved):
Thanks folks for doing this! Looks good

PR #2822, opened 2022-06-09
Render topical event organisation names logos

callumknights:
Trello: https://trello.com/c/iIfXqVKi/71-add-rendering-of-topical-event-organisation-names-and-logos-to-collections


app/helpers/organisation_helper.rb
```diff
@@ -24,4 +24,10 @@ def search_results_to_documents(search_results, organisation, include_metadata:
       end,
     }
   end
+
+  def array_of_links_to_organisations(organisations)
```
JonathanHallam, line 28: I think this is probably the wrong place for this method. When collections thinks about Organisations in this context it means pages listed here https://www.gov.uk/government/organisations

I think this method is in the organisation helper in whitehall because it's actually used by the [organisation frontend here](https://github.com/alphagov/whitehall/blob/5cba22ae73efabd6e93f8da75b28a2ae88ffdec7/app/views/organisations/_organisations_name_list.html.erb#L4-L6). Because we're not using this method in collections organisation view and it would be limited to topical events this is probably better placed in the topical events controller

app/models/topical_event.rb
```diff
@@ -64,12 +64,13 @@ def ordered_featured_documents
     end
   end
 
-  def organisations
+  def organisations(*args)
```
JonathanHallam, line 67: why do you need the splat operator here?

config/locales/az/topical_events.yml
```diff
@@ -3,3 +3,16 @@ az:
   topical_events:
     archived:
     featured:
+    lead_organisations:
+      arts_council_england:
```
BeckaL, line 7: Just a thought: As far as I understand it, these are hacks for special cases (first world war centenary etc) that we want to eventually remove. In this case I wonder if it'd be better to not use translations for them and just hardcode their titles, then we don't have to make any changes to the translation files which we'll later go back and need to remove.

spec/features/topical_event_spec.rb
```diff
@@ -56,12 +51,15 @@
     expect(page).to have_link(content_item.dig("details", "about_page_link_text"), href: "#{base_path}/about")
   end
 
-  context "when there are social media links" do
```
BeckaL, line 59: Have you intentionally got rid of this context?

config/locales/ar/browse.yml
```diff
@@ -2,5 +2,5 @@
 ar:
   browse:
     all_categories: جميع الفئات
-    description: 
-    title: 
+    description:
```
BeckaL, line 5: Minor nitpick, this looks unintentional. Would be good for this not to appear in the change history for this PR

app/views/topical_events/show.html.erb
```diff
@@ -16,8 +16,14 @@
...
-    <% if @topical_event.image_url %>
-      <%= image_tag(@topical_event.image_url, alt: @topical_event.image_alt_text) %>
+    <hr class="govuk-section-break govuk-section-break--m govuk-section-break--visible">
+
+    <% if @topical_event.organisations %>
+      <%= render "govuk_publishing_components/components/metadata", {
+        other: {
+          "Organisations": sanitize(array_of_links_to_organisations(@topical_event.organisations).to_sentence)
+        }
+      } %>
```
brucebolt, line 27: I notice there is [a partial](https://github.com/alphagov/collections/blob/ae77f463a80aaec1ee27f9b43e9d4d6f64422670/app/views/roles/_organisations.html.erb) (used solely by roles at the moment) which does exactly this.

Maybe we could make that partial shared so the same code isn't duplicated across multiple document types?

spec/features/topical_event_spec.rb
```diff
@@ -56,12 +51,15 @@
...
-    it "includes links to the social media accounts" do
-      visit base_path
-      expect(page).to have_link("Facebook", href: "https://www.facebook.com/a-topical-event")
-      expect(page).to have_link("Twitter", href: "https://www.twitter.com/a-topical-event")
-    end
+  it "includes a link to the organisation" do
```
brucebolt, line 54: Would it be possible to have a test that covers the scenario when there are no organisations (and therefore tests that the header isn't included)?

Also a test for the organisation logos section would be good too.

app/views/topical_events/show.html.erb
```diff
@@ -59,4 +65,33 @@
...
+              text: t("topical_events.organisations"),
+              font_size: "l",
+              margin_bottom: 4,
+              padding: true,
+              border_top: 2,
+            } %>
+
+            <div class="govuk-grid-row">
+              <% @topical_event.organisations.each do |organisation| %>
+                <div class="govuk-grid-column-one-quarter govuk-!-padding-bottom-7">
```
brucebolt: The layout seems to go a bit wobbly when there are more than 4 images (e.g. for [this topical event](http://collections.dev.gov.uk/government/topical-events/russian-invasion-of-ukraine-uk-government-response)).

Screenshot of the "Who's involved" section showing five organisation logos.  The first four are on a single line. The fifth is on the following line but aligned to the right.

config/locales/az/topical_events.yml
brucebolt: It might be a good idea to keep the two hacky logo sections in their own partials (i.e. one for each of the topical events that has these logos).  Everything in those could then be hardcoded, as @BeckaL suggests.

config/locales/ar/browse.yml
callumknights: Thanks for pointing this out 👍

config/locales/en/topical_events.yml
```diff
@@ -3,3 +3,16 @@ en:
...
+      bbc: BBC
+      birmingham_2022_commonwealth_games: Birmingham 2022 Commonwealth Games
+      british_library: British Library
+      commonwealth_war_graves_commission: Commonwealth War Graves Commission
+      heritage_lottery_fund: Heritage Lottery Fund
+      historic_england: Historic England
+      iwm: Imperial War Museum
+      queens_platinum_jubilee: Queen's Platinum Jubilee
+      unboxed: 'Unboxed: Creativity in the UK'
+      war_memorials_trust: War Memorials Trust
```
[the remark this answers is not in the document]
callumknights: Thanks for pointing this out, now resolved 👍

spec/features/topical_event_spec.rb
callumknights: 👍

config/locales/az/topical_events.yml
callumknights: Moved into partials as suggested! 👍

app/controllers/topical_events_controller.rb
```diff
@@ -18,4 +18,12 @@ def show
       end
     end
   end
+
+  def array_of_links_to_organisations(organisations)
+    organisations.map do |organisation|
+      helpers.link_to organisation[:title], organisation[:base_path], class: "organisation-link govuk-link"
```
brucebolt, line 24: It should be possible to just use `link_to` instead of `helpers.link_to`.

config/locales/en/topical_events.yml
```diff
@@ -10,3 +10,6 @@ en:
       latest: Latest
       publications: Publications
       travel_advice: Travel Advice
+      travel_advice: "Travel Advice"
+    publications: "Publications"
```
brucebolt, line 14: It looks like the `publications` key has crept into this PR, probably because of rebasing.

app/views/topical_events/show.html.erb
```diff
@@ -123,30 +124,39 @@
...
+          <% end %>
+        </div>
+
+        <% if @topical_event.slug == 'first-world-war-centenary' %>
+          <%= render "first_world_war_centenary"%>
+        <% end %>
+
+        <% if @topical_event.slug == '2022-events-platinum-jubilee-commonwealth-games-unboxed' %>
+          <%= render "platinum_jubilee_unboxed" %>
         <% end %>
```
brucebolt, line 158: Nice!  The partials make this much clearer.

app/views/topical_events/show.html.erb
```diff
@@ -123,30 +124,39 @@
...
+          <% @topical_event.organisations.each do |organisation| %>
+            <div class="govuk-grid-column-one-quarter govuk-!-padding-bottom-7">
+              <%= render "govuk_publishing_components/components/organisation_logo", {
+                organisation: {
+                  name: organisation[:title],
+                  url: organisation[:base_path],
+                  brand: organisation[:brand],
+                  crest: organisation[:crest],
+                }
+              } %>
```
brucebolt: It looks like the only change here is the indentation.  Could this be corrected in the first commit, so there's no diff here?  That'll make it easier to see the changes.

app/models/topical_event.rb
```diff
@@ -110,11 +110,18 @@ def announcements
...
         brand: organisation.dig("details", "brand"),
       }
     end
   end
+
+  def emphasised_organisations
+    @content_item.content_item_data.dig("details", "emphasised_organisations")&.map do |organisation_id|
+      [] << organisation_id
+    end
+  end
```
brucebolt, line 126: As far as I can see, this method is taking an array of content IDs and returning the same array.  Is there something I've missed?

spec/fixtures/content_store/topical_event.json
```diff
@@ -24,6 +25,7 @@
       "alt_text": "Text describing the image"
     },
     "end_date": "2016-04-28T00:00:00+00:00",
+    "emphasised_organisations": [],
```
brucebolt, line 28: Would it be possible to add a test for the presence of the organisation logos?  And also check that a non-emphasised organisation doesn't have a logo.

app/views/topical_events/show.html.erb
```diff
@@ -123,7 +123,7 @@
   </div>
 
   <div class="govuk-grid-column-two-thirds">
-    <% if @topical_event.organisations %>
+    <% if @topical_event.emphasised_organisations && @topical_event.emphasised_organisations.any? %>
```
brucebolt, line 126: I wonder if it might be better to just add this section in the final commit, rather than introducing it earlier?  I got slightly confused looking at the diff as to what changed each time.

app/models/topical_event.rb
brucebolt: Oh, looks like it has been updated in the next commit.  Maybe it could just be introduced in the latter commit only?

brucebolt (commented):
This is looking great @callumknights.  The functionality all looks to be complete :star:.

I've left a few comments about a couple more tests that could be useful and also some changes to what goes in each commit, just so it's clearer for others to see what changed if they look back at this PR in the future.

app/controllers/topical_events_controller.rb
callumknights: This doesn't work unfortunately, normally helpers methods are only expected to be used in the views, but the [Helpers Proxy](https://api.rubyonrails.org/v5.1/classes/ActionController/Helpers.html) is a way around that.

app/views/topical_events/show.html.erb
callumknights: Resolved this by just squashing the last two commits together, they weren't really distinct from each other

app/models/topical_event.rb
callumknights: Also resolved through squashing

spec/fixtures/content_store/topical_event.json
callumknights: I've pushed a new test for this in a separate commit, re-review would be fantastic!

PR #2826, opened 2022-06-14
Use safe navigation operator for topical event details

brucebolt:
The social media links and featured documents are optional for topical events.

This ensures we cover a scenario of `dig` returning `nil`, which would otherwise result in a `undefined method 'map' for nil:NilClass` error when we attempt to `map`.


JonathanHallam (approved):
Nice one, good test coverage, thanks :)

PR #2838, opened 2022-06-29
Add dimension 114 to click event on show/hide all sections

patrickpatrickpatrick:
### What

Add dimension114 (seciton index) to click event on Show/Hide All Sections on the accordion component.

### Why

The section index has been added to all other accordion sections apart from the Show/Hide All Sections. This meant that when filtering by section index the Show/Hide All Sections clicks would not appear in the data. This fix adds an index of '0' (as it is the first in the list of clickable buttons).


hannalaakso (approved):
Reviewed by running locally, I can see the dimension 114 being set to '0' for the Show all / Hide all button clicks.

PR #2843, opened 2022-07-01
Add detailed guidance

BeckaL:
Add detailed guidance section to topical events.

There are some differences from the Whitehall rendering for this section. We have decided to render as a sub group under the `documents` header, rather than separately as Whitehall does. We do not think there is a good product reason to treat these as logically separate from other document types. We are also using the shared document lists component which makes the styling consistent with the other document types.

|Collections rendering | Whitehall rendering |
|-------------|-----------|
| Screenshot 2022-07-01 at 15 59 46 |Screenshot 2022-07-01 at 16 00 16 |

The non-visual / less immediately obvious differences between the Whitehall rendered content and this are: 

* Detailed guides are retrieved from search api, rather than from the local Whitehall database
* The number of documents displayed has changed, from a maximum of 5 in Whitehall, to 3 here for consistency with other document types. However, we have added a 'see all' link, which was previously not given in the Whitehall rendering, meaning that there was no obvious way to see documents beyond the first 5 returned.
* The ordering has changed, and is now based on the 'public timestamp' from search api. There was some other ordering going on in Whitehall that I'm not sure it makes sense to replicate, timestamp seems like a sensible ordering.
* Whitehall was also doing some [eager loading](https://github.com/alphagov/whitehall/blob/f57384eb5f87c95038b68b7037032d9e0ee9a4b3/app/controllers/topical_events_controller.rb#L22) of the detailed guidance it displayed. I'm not sure whether we need to do anything equivalent here.

Note that the 'see all' link actually links to the 'guidance and regulations' topics. I'm not aware that regulations are a separate document type, but if we think that this will return some results we don't want, I'm not sure there is a way of doing this without creating a separate finder. But is at least better than the current Whitehall situation where you can't see anything beyond the 5 links presented to you 🤷 

### DIFF
--- a/spec/features/topical_event_spec.rb
+++ b/spec/features/topical_event_spec.rb
@@ -77,11 +77,13 @@ RSpec.feature "Topical Event pages" do
     let(:related_publications) { { "Policy on Topicals" => "/foo/policy_paper", "PM attends summit on topical events" => "/foo/news_story" } }
     let(:related_consultations) { { "A consultation on Topicals" => "/foo/consultation_one", "Another consultation" => "/foo/consultation_two" } }
     let(:related_announcements) { { "An announcement on Topicals" => "/foo/announcement_one", "Another announcement" => "/foo/announcement_two" } }
+    let(:related_guidance) { { "Some guidance" => "/foo/detailed_guidance_one", "Another bit of guidance" => "/foo/detailed_guidance_two" } }
 
     it "displays links to all related documents" do
       stub_search(body: search_api_response(related_announcements))
       stub_search(body: search_api_response(related_publications), params: { "filter_format" => "publication" })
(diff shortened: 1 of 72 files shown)


spec/features/topical_event_spec.rb
```diff
@@ -110,6 +110,7 @@
 
       within("#detailed-guides") do
         related_guidance.each { |title, link| expect(page).to have_link(title, href: link) }
+        expect(page).to have_link("See all detailed guides", href: "/search/guidance-and-regulation?topical_events%5B%5D=something-very-topical")
```
brucebolt, line 113: Could this change be fixed up to the previous commit, as that's when the test was first introduced

brucebolt (approved):
Looks good to me.

I've left a few really small comments.  Plus this will need rebasing once https://github.com/alphagov/collections/pull/2849 is merged.

PR #2849, opened 2022-07-05
Add topical event latest documents section

BeckaL:
This adds a 'latest' documents section to the topical events page, which displays the latest documents (of any type) linked to the topical collection.

| Collections Rendering | Whitehall Rendering |
|--------|--------|
|Screenshot 2022-07-05 at 15 19 28 | Screenshot 2022-07-05 at 15 19 45 |

[Trello card](https://trello.com/c/5BG65SGT/145-add-topical-event-latest-documents-to-collections)

### DIFF
--- a/app/helpers/topical_events_helper.rb
+++ b/app/helpers/topical_events_helper.rb
@@ -6,13 +6,19 @@ module TopicalEventsHelper
 private
 
   def base_path(content_type)
-    { publication: "/search/all?",
+    { latest: "/search/all?",
+      publication: "/search/all?",
       consultation: "/search/policy-papers-and-consultations?",
       announcement: "/search/news-and-communications?" }[content_type]
   end
 
   def query_params(content_type, slug)
     case content_type
+    when :latest
+      {
+        order: "updated-newest",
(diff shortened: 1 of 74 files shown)


callumknights (approved):
LGTM 👍

chao-xian:
Looks ok but just a small suggestion on whether the 74 file commit could be broken down into smaller commits at all?

BeckaL:
> Looks ok but just a small suggestion on whether the 74 file commit could be broken down into smaller commits at all?

Hey @chao-xian thanks for having a look!

The vast majority of the 74 files are locale files that needed updating with default values. We could do this as part of a separate commit, but they're logically part of the changes from the other commit. I'm not sure the commits would make sense individually, and I think the actual logic changes in this PR are relatively small, just that we need to add default values in ~70 locale files in order to add a translation for one bit of the PR.

But happy to break this up if you think it makes sense 😄

chao-xian:
Ah ok - might be worth adding that as part of the commit message to explain why. That's just a small suggestion but otherwise yeah that makes sense! 👍

PR #2853, opened 2022-07-07
Some styling tweaks for topical events

BeckaL:
This PR gathers a number of styling tweaks to the topical events page in collections, either to: a) make the layout more logical, or b) make it more consistent with the Whitehall rendering, following review.

Fixes in this PR:
* Headings for different document types are h2s rather than h3s to make more sense with the format of the page
* Styling of 'latest' heading is the same as that of the 'Documents' heading, since they occupy the same place in the layout hierarchy of the page.
* Adds padding at appropriate points on the page
* Corrects ordering of layout of elements to be consistent with Whitehall
* Long summaries on featured documents are truncated to be consistent with Whitehall.

To see a rendered page:

`govuk-docker-up app-live`
Visit http://collections.dev.gov.uk/government/topical-events/nepal-earthquake-april-2015 and compare to the live site.
This topical event has most of the elements possible to standard topical events.

Slugs of some other topical events (for a locally spun up version of collections pointing to the live site:
* [budget-2021](http://collections.dev.gov.uk/government/topical-events/budget-2021)
* [afghanistan-uk-government-response](http://collections.dev.gov.uk/government/topical-events/afghanistan-uk-government-response) - this has a custom 'travel advice' section not seen on other topical events.
* [first-world-war-centenary](http://collections.dev.gov.uk/government/topical-events/first-world-war-centenary) - this has a custom external organisations set of logos at the bottom not seen on other topical events

### DIFF
--- a/app/assets/stylesheets/views/_topical_events.scss
+++ b/app/assets/stylesheets/views/_topical_events.scss
@@ -91,4 +91,10 @@
     }
     /* stylelint-enable */
   }
+
+  .topical-events__section-break,
+  .topical-events__social-media-links,
+  .topical-events__image {
+    @include govuk-responsive-margin(8, "bottom");
+  }
 }
--- a/app/assets/stylesheets/views/_topical_events.scss
+++ b/app/assets/stylesheets/views/_topical_events.scss
@@ -94,7 +94,8 @@
 
   .topical-events__section-break,
   .topical-events__social-media-links,
-  .topical-events__image {
+  .topical-events__image,
+  .topical-event__body-wrapper {
     @include govuk-responsive-margin(8, "bottom");
   }
 }
--- a/app/assets/stylesheets/views/_topical_events.scss
+++ b/app/assets/stylesheets/views/_topical_events.scss
@@ -8,6 +8,8 @@
 
   .organisations-list {
     list-style: none;
+    padding: 0;
+
     /* stylelint-disable */
     .organisations-list__item {
       float: left;
--- a/app/models/topical_event.rb
+++ b/app/models/topical_event.rb
@@ -72,7 +72,7 @@ class TopicalEvent
         image_src: document.dig("image", "url"),
         image_alt: document.dig("image", "alt_text"),
         heading_text: document["title"],
-        description: document["summary"],
+        description: document["summary"].truncate(160, separator: " "),
       }
     end
   end
--- a/app/views/topical_events/_document_list_from_search_api.html.erb
+++ b/app/views/topical_events/_document_list_from_search_api.html.erb
@@ -7,6 +7,7 @@
     padding: true,
     font_size: 27,
     margin_bottom: 3,
+    heading_level: 3
   } %>
 <% end %>
 
--- a/app/views/topical_events/_document_list_from_search_api.html.erb
+++ b/app/views/topical_events/_document_list_from_search_api.html.erb
@@ -1,7 +1,7 @@
 <% heading ||= nil %>
 
 <% if heading %>
-  <section id="<%= heading.downcase.parameterize %>" class="document-block">
+  <section id="<%= heading.downcase.parameterize %>" class="govuk-!-margin-bottom-7">
   <%= render "govuk_publishing_components/components/heading", {
     text: heading,
     padding: true,
@@ -13,7 +13,6 @@
 
   <%= render "govuk_publishing_components/components/document_list", {
     items: documents,
-    margin_bottom: 6,
   } %>
 
   <p class="govuk-body">
--- a/app/views/topical_events/show.html.erb
+++ b/app/views/topical_events/show.html.erb
@@ -32,16 +32,18 @@
       </div>
     <% end %>
 
-    <%= render "govuk_publishing_components/components/govspeak", {
-      } do %>
-        <%= @topical_event.body&.html_safe %>
-    <% end %>
+    <div class="topical-event__body-wrapper">
+      <%= render "govuk_publishing_components/components/govspeak", {
+        } do %>
+          <%= @topical_event.body&.html_safe %>
+      <% end %>
 
-    <% if @topical_event.about_page_link_text %>
-      <p class="govuk-body">
-        <%= link_to(@topical_event.about_page_link_text, @topical_event.about_page_url, html_options: { class: "govuk-link" }) %>
-      </p>
-    <% end %>
+      <% if @topical_event.about_page_link_text %>
+        <p class="govuk-body">
+          <%= link_to(@topical_event.about_page_link_text, @topical_event.about_page_url, html_options: { class: "govuk-link" }) %>
+        </p>
+      <% end %>
+    </div>
 
     <% if @topical_event.social_media_links %>
       <section id="social-media-links" class="topical-events__section-break">
--- a/app/views/topical_events/show.html.erb
+++ b/app/views/topical_events/show.html.erb
@@ -17,10 +17,6 @@
       margin_bottom: 8,
     } %>
 
-    <% if @topical_event.image_url %>
-      <%= image_tag(@topical_event.image_url, alt: @topical_event.image_alt_text) %>
-    <% end %>
-
     <hr class="govuk-section-break govuk-section-break--m govuk-section-break--visible">
 
     <% if @topical_event.organisations %>
@@ -29,6 +25,10 @@
           "Organisations": sanitize(array_of_links_to_organisations(@topical_event.organisations).to_sentence)
         }
       } %>
+    <% end %>
+
+    <% if @topical_event.image_url %>
+      <%= image_tag(@topical_event.image_url, alt: @topical_event.image_alt_text) %>
     <% end %>
 
     <%= render "govuk_publishing_components/components/govspeak", {
--- a/app/views/topical_events/show.html.erb
+++ b/app/views/topical_events/show.html.erb
@@ -40,7 +40,7 @@
 
       <% if @topical_event.about_page_link_text %>
         <p class="govuk-body">
-          <%= link_to(@topical_event.about_page_link_text, @topical_event.about_page_url, html_options: { class: "govuk-link" }) %>
+          <%= link_to(@topical_event.about_page_link_text, @topical_event.about_page_url, class: "govuk-link" ) %>
         </p>
       <% end %>
     </div>
--- a/app/views/topical_events/show.html.erb
+++ b/app/views/topical_events/show.html.erb
@@ -50,7 +50,7 @@
     <% end %>
   </div>
 
-  <div class="govuk-grid-column-full govuk-!-margin-bottom-7">
+  <div class="govuk-grid-column-full">
     <% if @topical_event.ordered_featured_documents %>
       <section id="featured" class="govuk-!-margin-bottom-6">
         <%= render "govuk_publishing_components/components/heading", {
@@ -72,7 +72,9 @@
         <% end %>
       </section>
     <% end %>
+  </div>
 
+  <div class="govuk-grid-column-two-thirds govuk-!-margin-bottom-7">
     <section id="latest">
       <%= render "govuk_publishing_components/components/heading", {
         text: I18n.t("topical_events.headings.latest"),
@@ -95,7 +97,7 @@
     </section>
   </div>
 
-  <section id="documents" class="govuk-grid-column-full">
+  <section id="documents" class="govuk-grid-column-two-thirds">
     <% if @topical_event.publications.any? || @topical_event.consultations.any? || @topical_event.announcements.any? || @topical_event.detailed_guidance.any? %>
       <%= render "govuk_publishing_components/components/heading", {
         text: I18n.t("topical_events.headings.documents"),
--- a/app/views/topical_events/show.html.erb
+++ b/app/views/topical_events/show.html.erb
@@ -14,10 +14,9 @@
 
     <%= render "govuk_publishing_components/components/lead_paragraph", {
       text: @topical_event.description,
-      margin_bottom: 8,
     } %>
 
-    <hr class="govuk-section-break govuk-section-break--m govuk-section-break--visible">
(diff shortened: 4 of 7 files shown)


app/models/topical_event.rb
```diff
@@ -72,7 +72,7 @@ def ordered_featured_documents
         image_src: document.dig("image", "url"),
         image_alt: document.dig("image", "alt_text"),
         heading_text: document["title"],
-        description: document["summary"],
+        description: document["summary"].truncate(160, separator: " "),
```
JonathanHallam, line 75: Oooh, this truncate method is a new one for me, love that you can pass the separator as well 🤩

app/views/topical_events/show.html.erb
```diff
@@ -31,6 +27,10 @@
       } %>
     <% end %>
 
+    <% if @topical_event.image_url %>
+      <%= image_tag(@topical_event.image_url, alt: @topical_event.image_alt_text) %>
+    <% end %>
```
kevindew, line 35: Just a quick check. Are we happy the page renders ok when there isn't an image?

app/models/topical_event.rb
kevindew: I couldn't find an existing use of this approach in this app. What do we do for other document types - like organisations that have this same issue?

Just concerned about introducing an inconsistency.

app/assets/stylesheets/views/_topical_events.scss
```diff
@@ -9,6 +9,7 @@
   .organisations-list {
     list-style: none;
     /* stylelint-disable */
+    padding-inline-start: 0px;
```
kevindew, line 12: I don't think this should be within the stylelint-disable block

This is also newer CSS that IE doesn't support so feels a bit funky. I wonder if there's a more basic CSS approach to solve the problem.

Also for quite a nuanced fix there isn't much information in the commit message.

app/views/topical_events/_document_list_from_search_api.html.erb
```diff
@@ -1,7 +1,7 @@
 <% heading ||= nil %>
 
 <% if heading %>
-  <section id="<%= heading.downcase.parameterize %>" class="document-block">
+  <section id="<%= heading.downcase.parameterize %>" class="govuk-!-margin-bottom-7">
```
kevindew, line 4: Is there anything we can add to the commit message as to why these are the appropriate spacing? 

There seems a bit of a mental leap from the commit message saying it's appropriate to understanding why - so could be something like "in orgs we have a 30px padding [link to rule] and margin-bottom-7 replicates that"

app/views/topical_events/show.html.erb
```diff
@@ -12,12 +12,13 @@
...
-      text: @topical_event.description,
-      margin_bottom: 8,
-    } %>
+    <div class="govuk-!-margin-bottom-8">
+      <%= render "govuk_publishing_components/components/lead_paragraph", {
+        text: @topical_event.description,
+      } %>
 
-    <hr class="govuk-section-break govuk-section-break--m govuk-section-break--visible">
+      <hr class="govuk-section-break govuk-section-break--m govuk-section-break--visible">
```
kevindew, line 20: I wonder if rather than needing an extra div for the margin, whether we can just apply it to this hr?

I imagine it's worth explaining in the commit that we want the paragraph close to the HR and then apply the vertical margin.

app/assets/stylesheets/views/_topical_events.scss
```diff
@@ -90,4 +90,8 @@
     }
     /* stylelint-enable */
   }
+
+  .topical-events__margin-bottom {
+    margin-bottom: govuk-spacing(6);
+  }
```
kevindew, line 96: This seems a somewhat unusual class. By [BEM](http://getbem.com/naming/) this would be describing the margin-bottom element of a topical-events object. But we neither have a margin-bottom element or a topical-events object.

I think it'd be better to just the govuk-!-margin modifiers directly to be honest, as a simple suggestion. As I guess we want just a utility class and we have that in the govuk- option but without the risks of this utility class which will couple any changes to affect multiple things (which makes the components not isolated)

app/views/topical_events/show.html.erb
```diff
@@ -34,16 +34,18 @@
...
-      <p class="govuk-body">
-        <%= link_to(@topical_event.about_page_link_text, @topical_event.about_page_url, html_options: { class: "govuk-link" }) %>
-      </p>
-    <% end %>
+      <% if @topical_event.about_page_link_text %>
+        <p class="govuk-body">
+          <%= link_to(@topical_event.about_page_link_text, @topical_event.about_page_url, html_options: { class: "govuk-link" }) %>
+        </p>
+      <% end %>
+    </div>
```
kevindew, line 48: Hmm we do seem to be adding this extra margin a lot - how does similar pages like organisation resolve it? I'm just wondering if we've started digging ourselves a hole where we're fighting the basic layout - but could be that the basic layout just isn't great for a hub style page.

kevindew (commented):
Looks great - I've added a few comments and questions. Only thing I'm a bit worried about is the frequent need for margin bottom and not sure if there's already conventions for that or if we should use the `govuk-!-margin` a lot (perhaps that is the convention).

It'd be good to see it in action too if it's available on integration?

andysellick:
> Only thing I'm a bit worried about is the frequent need for margin bottom and not sure if there's already conventions for that or if we should use the `govuk-!-margin` a lot (perhaps that is the convention).

Some hopefully helpful insight into the margin question. The margin classes in `govuk-frontend` (`govuk-!-margin` etc.) are override classes that use `!important`, which means they are intended for use only where another option is not available. 

For example, if you were working on a legacy codebase, a new margin might be overridden by the existing CSS that cannot for some reason be changed (because the effort required to do so might exceed the time allowed), resulting in an incorrect margin. The override classes would be used in that situation to overcome that problem e.g.

```
// existing, unchangeable CSS
.sectionClass .elementClass h1 {
  margin-bottom: 0px;
}

// new CSS
.heading1Class {
  margin-bottom: 10px; // we want this, but it'll be overridden by the existing CSS
}
```

```html
<!-- solution -->
<h1 class="heading1Class govuk-!-margin-bottom-10">Heading</h1>
```

Where we're writing new CSS and/or have full control over our CSS we should try to use custom classes for margin as a general rule. Otherwise we risk having multiple `!important` declarations fighting with each other.

kevindew:
> > Only thing I'm a bit worried about is the frequent need for margin bottom and not sure if there's already conventions for that or if we should use the `govuk-!-margin` a lot (perhaps that is the convention).
> 
> Some hopefully helpful insight into the margin question. The margin classes in `govuk-frontend` (`govuk-!-margin` etc.) are override classes that use `!important`, which means they are intended for use only where another option is not available.
> 
> For example, if you were working on a legacy codebase, a new margin might be overridden by the existing CSS that cannot for some reason be changed (because the effort required to do so might exceed the time allowed), resulting in an incorrect margin. The override classes would be used in that situation to overcome that problem e.g.
> 
> ```
> // existing, unchangeable CSS
> .sectionClass .elementClass h1 {
>   margin-bottom: 0px;
> }
> 
> // new CSS
> .heading1Class {
>   margin-bottom: 10px; // we want this, but it'll be overridden by the existing CSS
> }
> ```
> 
> ```
> 
> <h1 class="heading1Class govuk-!-margin-bottom-10">Heading</h1>
> ```
> 
> Where we're writing new CSS and/or have full control over our CSS (as in this PR) we should try to use custom classes for margin as a general rule. Otherwise we risk having multiple `!important` declarations fighting with each other.

Thanks @andysellick great to have your input ⭐ 

Ah I didn't realise that used important, is there a utility class that doesn't use it? I guess I might be missing the point.

Do you have a suggestion for what should be done to apply these frequent uses of margin here and/or what is most conventional. I figure for some things that are very similar (the section elements) we could have them be a particular block with their own class and then may have to something different on the adhoc places, but interested in your view - particularly if the utility class that I've suggested should be removed is something is something that should be considered (to avoid conflicting reviews).

app/assets/stylesheets/views/_topical_events.scss
hannalaakso: Agree that it would be good to know more about the issue we're trying to solve, and also to get a URL for a page that this affects.

app/views/topical_events/_document_list_from_search_api.html.erb
hannalaakso: Agree with Kevin. Also it would be useful to have a URL for a page that this impacts.

hannalaakso (commented):
Below is what I've managed to successfully review so far. Some commits didn't have enough information for me to review them, have left comments for those. 

### Make headings of individual document lists h3s 

Checked http://collections.dev.gov.uk/government/topical-events/budget-2021 against live, heading structure deffo makes more sense now 👍 
 
### Move image to be below organisation names 

Compared with https://www.gov.uk/government/topical-events/cop26, the image now appears in the same location as live.
 
### Use stylised apostrophe 

Matches https://www.gov.uk/guidance/style-guide/a-to-z-of-gov-uk-style
 
### Change topical event sections to be 2/3rds width

Document list width now uses `govuk-grid-column-two-thirds` both on http://collections.dev.gov.uk/government/topical-events/budget-2021 and live.

With the last four commits it would be good to hear the outcome of the conversation between Kevin and @andy.sellick 👀

andysellick:
> Ah I didn't realise that used important, is there a utility class that doesn't use it? I guess I might be missing the point.

I don't believe there is. I think the Design System encourages you to write your own, using the `govuk-spacing()` mixin.

> Do you have a suggestion for what should be done to apply these frequent uses of margin here and/or what is most conventional. I figure for some things that are very similar (the section elements) we could have them be a particular block with their own class and then may have to something different on the adhoc places, but interested in your view - particularly if the utility class that I've suggested should be removed is something is something that should be considered (to avoid conflicting reviews).

Sounds about right - where something happens repeatedly create a generic class for that, and apply specifics where needed. It depends on context. This PR doesn't look too complex, so a `section` class and the `topical-events__margin-bottom` class should be fine.

You could also look for existing classes within this application that could be reused, but that might be overthinking it.

app/assets/stylesheets/views/_topical_events.scss
BeckaL: I was taking inspiration from [this existing css for organisations](https://github.com/alphagov/collections/blob/main/app/assets/stylesheets/views/_organisation.scss#L83-L85) in the repo, and trying to fix the issue where we were using a lot of different values for margin-bottom.

But I also don't really know much about frontend practice so maybe this doens't make sense here, happy to change

app/assets/stylesheets/views/_topical_events.scss
BeckaL: Thanks both, this now updated to just be padding: 0 and be outside the stylelint disable block. I've also beefed up the commit description, so considering this set of comments resolved 🙂

app/views/topical_events/show.html.erb
BeckaL: The organisations page uses a lot of the class `organisation__margin-bottom` to add spacing, so I don't think we're being inconsistent here. But again, I will defer to anybody with better frontend knowledge than me on the best thing to do.

app/views/topical_events/show.html.erb
kevindew: Nah - just wondering if we'd tested the lack of this condiditon.

app/assets/stylesheets/views/_topical_events.scss
kevindew: Ah sure I didn't realise we had that for organisations. It'd be good to reference that in the commit message.

From Andy's comment it seems like he is happy with it https://github.com/alphagov/collections/pull/2853#issuecomment-1187475459 so by that logic and precedence it seems ok to keep.

It makes me personally a bit sad as it's not using BEM conventions correctly but not a hill I'm willing to die on, particularly as there is precedence.

app/views/topical_events/show.html.erb
DilwoarH: I am inclined to agree with @BeckaL - I think it's good to be consistent with the rest of the application. I am also in the process of doing a write-up to suggest that we discontinue using the [override classes](https://design-system.service.gov.uk/styles/spacing/#using-the-override-classes) as they use `!important` as part of their css. It is better to use the [static spacing mixins](https://design-system.service.gov.uk/styles/spacing/#using-the-static-spacing-function) than to use these classes.

DilwoarH: Depending on what you need, you might want to use the responsive helpers instead:

```
@include govuk-responsive-padding(6);
@include govuk-responsive-margin(6, "bottom");
```

DilwoarH: Just to add additional context, the usage of these helper classes was raised in a GOVUK Frontend community meeting (in 2019) where it was decided that we would avoid using override classes and use a single class instead.

Some things to consider might be that rather than having a class specifically for margin (`organisation__margin-bottom`) it should be describing what the element is, ie. `organisation__content-body`.

app/views/topical_events/_document_list_from_search_api.html.erb
injms: Worth noting that `govuk-!-margin-bottom-7` has a margin of 40px of desktop, but 25px on mobile. All of the override classes are responsive - so will have a different margin depending on screen size.

app/models/topical_event.rb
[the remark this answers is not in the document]
hannalaakso: That makes sense @BeckaL 👍

brucebolt (approved):
This looks good to me as we're reproducing the behaviour of the Whitehall rendered page.  Any further improvements can be done once the rendering is switched over to Collections.

JonathanHallam:
I'm happy that we've addressed the comments on this PR. Am merging now :) 🎉

PR #2865, opened 2022-07-15
Iterate browse topics templates

hannalaakso:
### Context
We ran an A/B test to compare the current "miller column" GOV.UK Browse topics templates (variant A) against the new more accessible and mobile friendly GOV.UK topics templates (variant B). The results of the test have given us confidence that we can retire the old templates and make the new templates the default ones. 

We are further iterating the second level template by running an AB test with curated lists (instead of curated accordions). The curated lists show users all the links but break them into sections which could help users scan potentially quite long topic pages since we're not using accordions to organise the content. 

In the AB test the curated lists become the A (default) variant, and the curated accordions continue to be the B variant. 

This PR:
- Deletes the templates, tests, the controller logic, the sass and the JS for the old miller column templates (thanks @chao-xian for helping with this)
- Adds new accordion & list templates for curated level 2 pages to allow us to switch between them based on the variant
- Makes the new curated lists the A (default variant) and switches between the variants using the query string (this is only for testing purposes, we will implement the proper logic in a later PR)
- Makes the new list topic level 2 template the default one
- Divides the topic 2 lists into sections divided by borders
- Adjusts spacing on the topics pages to match new design

### Notes for design review
- It's probably best to quickly review the spacing in the other topic pages too since the spacing is now shared programmatically between some of the templates and I could have missed something.
- I might have not solved [this card](https://trello.com/c/odXz3tiq/1151-improve-spacing-of-links-within-topic-page-accordions), please let me know if further adjustments are needed

### Notes for developer review
- It's easiest to review the commits in order as the PR deletes and renames files

### Preview URLs 
- [Level 2 curated](https://collections-pr-2865.herokuapp.com/browse/benefits/manage-your-benefit) - NEW
- [Level 2 curated with More on this topic](https://collections-pr-2865.herokuapp.com/browse/business/setting-up) - NEW
- [Level 0](https://collections-pr-2865.herokuapp.com/browse)
- [Level 1](https://collections-pr-2865.herokuapp.com/browse/births-deaths-marriages)
- [Level 2 curated](https://collections-pr-2865.herokuapp.com/browse/benefits/manage-your-benefit?b=true)
- [Level 2 curated with More on this topic](https://collections-pr-2865.herokuapp.com/browse/business/setting-up?b=true)
- [Level 2 A-Z](https://collections-pr-2865.herokuapp.com/browse/births-deaths-marriages/register-offices)
- [Level 2 A-Z with More on this topic](https://collections-pr-2865.herokuapp.com/browse/business/limited-company)

### Depends on
- https://github.com/alphagov/govuk-cdn-config/pull/401

https://trello.com/c/RDZA1vJ7/1109-build-iteration-of-level-2-browse-curated-topic-template-b-c-test


app/views/second_level_browse_page/_show_curated_accordion.html.erb
```diff
@@ -0,0 +1,66 @@
+<%
```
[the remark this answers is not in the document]
hannalaakso: @chao-xian Sure, have a look now.

app/controllers/browse_controller.rb
```diff
@@ -1,4 +1,8 @@
 class BrowseController < ApplicationController
+  def new_browse_variant_c?
```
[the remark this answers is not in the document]
hannalaakso: Had a conversation with @chao-xian IRL. For reference the switching in the second browse level templates will be used for doing the design review. But I've removed the level 1 switching mechanism and also flipped the logic so the default variant is now the list version.

chao-xian:
This is all looking good to my eyes, but would also like one of the FE devs to review this too if possible.

chao-xian (approved):

app/controllers/browse_controller.rb
```diff
@@ -12,49 +10,20 @@ def index
...
-          html: second_level_browse_pages_partial(page),
-        })
-      end
-    end
+    show_html(page)
   end
 
 private
 
   def show_html(page)
```
injms, line 18: As both the `show_html` and `index_html` methods have had the logic stripped out, can they be re-integrated back into the `show` and `index` methods?

app/assets/javascripts/application.js
```diff
@@ -8,7 +8,6 @@
 //= require govuk_publishing_components/components/step-by-step-nav
 
 //= require support
-//= require browse-columns
```
injms, line 11: It'd be really interesting to see what effect the removal of the CSS and JavaScript has on page weight - can you give me a shout when this gets deployed to production so I can add a note to Speedcurve?

features/step_definitions/viewing_browse_steps.rb
```diff
@@ -52,8 +52,8 @@
 end
 
 Then(/^I see the list of top level browse pages alphabetically ordered$/) do
-  expect(page).to have_selector("div#root li:nth-child(1)", text: "Benefits"), "Benefits should appear first"
-  expect(page).to have_selector("div#root li:nth-child(2)", text: "Crime and justice"), "Crime and justice should appear second"
+  expect(page).to have_selector("ul.gem-c-cards__list li:nth-child(1)", text: /Benefits/), "Benefits should appear first"
```
injms, line 55: Using the `gem-c-` class in a test can be a bit of a hidden footgun - if the component gets renamed, the tests in the component gem will pass, the gem will be released, and then this test in collections will fail. 

As this is not testing for the use of the card component, could another non-gem selector be used here?

hannalaakso: That's a good point. We no longer have an attribute to hook onto that works without reference to something in the gem. The component currently doesn't allow you to set any custom classes or id's so one option is making that possible although our use case feels quite specific to this context.

Alternatively, I'm wondering if we actually need this test anymore? It seems to have been useful for when the component was custom to collections but now it's just literally testing a component that has its own tests in the gem.

app/assets/javascripts/application.js
hannalaakso: Yes, definitely, will do 👍

features/step_definitions/viewing_browse_steps.rb
injms: Agreed this test isn't helpful any more. I think that if you give the component a list of items I'd expect the order of the list to be respected - and to be tested at a component level if needed.

app/controllers/browse_controller.rb
```diff
@@ -1,60 +1,19 @@
...
-
   def index
     page = MainstreamBrowsePage.find("/browse")
     @dimension26 = 1
     @dimension27 = page.top_level_browse_pages.count || 0
+    template = :index
     setup_content_item_and_navigation_helpers(page)
-    index_html(page)
+    slimmer_template "gem_layout_full_width"
+    render template, locals: { page: page }
```
injms, line 9: We can get ride of setting the `template` variable as it's only used once and just set it directly:

```erb
render :index, locals: { page: page }
```

app/controllers/browse_controller.rb
```diff
@@ -1,60 +1,19 @@
...
-  enable_request_formats show: [:json]
-
   def index
     page = MainstreamBrowsePage.find("/browse")
     @dimension26 = 1
     @dimension27 = page.top_level_browse_pages.count || 0
+    template = :index
     setup_content_item_and_navigation_helpers(page)
-    index_html(page)
+    slimmer_template "gem_layout_full_width"
```
injms, line 8: As both the `index` and `show` methods use the same slimmer template, this can be set once straight after the class is set up - for example:

```ruby
class BrowseController < ApplicationController
  slimmer_template "gem_layout_full_width"

  def index
    ...
```

features/step_definitions/viewing_browse_steps.rb
hannalaakso: I've decoupled the tests from the gem classes and they now trust the gem component to handle the alphabetical ordering.  We're now testing for the presence of links instead, similar to what we do [here](https://github.com/alphagov/collections/blob/c1e5ed91084bc354876c5ac9de29b7a02fb78ee9/features/step_definitions/latest_changes_steps.rb#L17). Although that example actually references the gem classes, I guess we have some inconsistencies in these tests that should be tidied up.

hannalaakso:
@injms Thanks for the review 👍 I think I've addressed all your comments.

app/assets/stylesheets/views/_browse.scss
```diff
@@ -67,24 +67,19 @@ $topic-section-border: 1px solid $govuk-border-colour;
 .browse__curated-section {
   border-top: $topic-section-border;
   padding-top: govuk-spacing(3);
-  margin-bottom: govuk-spacing(9);
+  margin-bottom: 70px;
```
injms, line 70: If we absolutely have to stray completely outside of the Design System spacing, then this either needs a comment explaining why this is a magic number (hopefully preventing it from being undone in the future) or should be achieved by adding together two of the static spacing functions.

```suggestion
  margin-bottom: 70px; // This needs to be 70px because....
```

or



```suggestion
  margin-bottom: govuk-spacing(6) + govuk-spacing(7);
```


If the spacing scale doesn't suit, then we also need to note it down so we can let the Design System team know - feedback like this really helps them ensure that the Design System is useful and relevant.

features/step_definitions/viewing_browse_steps.rb
injms: Yep, there are definitely inconsistencies. It's good to aim for is not introducing _more_ problems, even if the old problems haven't been removed yet.

app/assets/stylesheets/views/_browse.scss
hannalaakso: Good shout. This particular value was actually a bit of a test for now for the designer to see what it would look like if we increased the spacing here (we've got a separate conversation going on in the Trello card about design fixes, my bad for not making it clear here). But I've implemented your suggested way of setting the value and have added a comment in case we decide to keep this spacing.

injms (approved):
Thanks for temporarily keeping the JSON endpoints - means a better user experience during the upgrade. 

This works and looks good - so a thumbs up from me 👍 

Thanks for all your hard work on this 🚀

PR #2867, opened 2022-07-18
Remove Miller Column Browse template

chao-xian:
We have completed the AB Test for the New Browse template and want to switch from the Miller Column based template as the default.

[Preview on Heroku should show you only see the New Browse designs](https://collections-pr-2867.herokuapp.com/browse).

[Trello](https://trello.com/c/glBoTClk/1158-switch-off-new-browse-ab-test-and-show-new-browse-instead-of-miller-columns)


app/controllers/browse_controller.rb
```diff
@@ -1,6 +1,4 @@
 class BrowseController < ApplicationController
-  enable_request_formats show: [:json]
```
hannako, line 2: Why were we supporting json formats before, but not now. Not sure I understand how that relates to miller columns.

chao-xian: The JSON was consumed by the Javascript to build up miller columns.

hannako: Ah gotcha, thank you

chao-xian:
Work from this PR is being incorporated into https://github.com/alphagov/collections/pull/2865

PR #2874, opened 2022-07-21
Only show topical event featured documents section when present

brucebolt:
We are currently showing the featured documents section of topical event pages when there are no topical events present, as the `ordered_featured_documents` method returns an empty array.

Fixing this to check to see if there are any items in the array.

Example of this issue:
Screenshot showing a topical event page with no featured documents.  There is a heading titled Featured but no documents are listed under it.


JonathanHallam (approved):
oops, good spot!

PR #2875, opened 2022-07-21
Update GTM analytics tracking attributes

andysellick:

## What
Updates the data attributes used on accordion on the `/coronavirus` page for GTM analytics. Note that this change depends upon [this change in govuk_publishing_components](https://github.com/alphagov/govuk_publishing_components/pull/2864) being present in the currently installed gem version, but this can be safely merged ahead of that as it doesn't affect live and we're still just testing it anyway.

## Why
We updated the schema for how click tracking like this works, which meant that the tracking would no longer work.

## Visual changes
None.

Trello card: https://trello.com/c/obytEbqo/153-update-existing-code-with-schema-changes


AshGDS (approved):
All good :+1:

PR #2852, opened 2022-07-06
Only use `better_errors` in development environment

brucebolt:
This application uses `better_errors` to produce pretty error messages when an error occurs.

This is great for development in a browser, but as the error is returned as a HTML response, we don't ever see the error when running feature tests (as Capybara thinks the page loaded ok). All we see is that the test expectation fails.

Example of a failing test before this change (you've no idea whether the failure is because the element isn't there or because there is an error in the code):
```
1) Topical Event pages afghanistan response topical event includes travel advice for Afghanistan
   Failure/Error: expect(page).to have_link(titleize_base_path(afghanistan_travel_advice_base_path), href: afghanistan_travel_advice_base_path)
     expected to find link "Foreign travel advice afghanistan" with href "/foreign-travel-advice/afghanistan" but there were no matches
   # ./spec/features/topical_event_spec.rb:72:in `block (3 levels) in <top (required)>'
```

Example of a failing test after this change (you can now see there's an error with the code):
```
 1) Topical Event pages afghanistan response topical event includes travel advice for Afghanistan
    Failure/Error:
      @content_item.content_item_data.dig("links", "organisations").map do |organisation|
        {
          content_id: organisation["content_id"],
          base_path: organisation["base_path"],
          title: organisation["title"],
          crest: organisation.dig("details", "logo", "crest"),
          brand: organisation.dig("details", "brand"),
        }
      end

    ActionView::Template::Error:
      undefined method `map' for nil:NilClass
    # ./app/models/topical_event.rb:111:in `organisations'
    # ./app/views/topical_events/show.html.erb:26:in `_app_views_topical_events_show_html_erb___4491315266640037643_67180'
    # ./spec/features/topical_event_spec.rb:71:in `block (3 levels) in <top (required)>'
    # ------------------
    # --- Caused by: ---
    # NoMethodError:
    #   undefined method `map' for nil:NilClass
    #   ./app/models/topical_event.rb:111:in `organisations'
```

By moving `better_errors` to only be included in the development environment, we will see actual errors when running tests locally.

⚠️ This repo is Continuously Deployed: make sure you [follow the guidance](https://docs.publishing.service.gov.uk/manual/development-pipeline.html#merge-your-own-pull-request) ⚠️

PR #2876, opened 2022-07-21
Fix errors in topical event tests

brucebolt:
We only noticed these problems after changing the usage of `better_errors` in https://github.com/alphagov/collections/pull/2852 as the test previous passed based on matching against the rendered error message.  The errors were introduced into the `main` branch after the `better_errors` PR was branch off `main`, so they weren't spotted in the pre-merge checks.

There are two corrections:

- Use safe navigation operator for emphasised organisations: a content item may not contain any links to organisations for a topical event, so we get a nil object which we cannot map. By adding the safe navigation operator, we overcome this.
- Add a missing `>` to a HTML tag.

### DIFF
--- a/app/views/topical_events/show.html.erb
+++ b/app/views/topical_events/show.html.erb
@@ -120,7 +120,7 @@
 
   <div class="govuk-grid-column-two-thirds">
     <% if @topical_event.travel_advice.any? %>
-      <section id="travel-advice"
+      <section id="travel-advice">
       <%= render "govuk_publishing_components/components/heading", {
         text: t("topical_events.headings.travel_advice"),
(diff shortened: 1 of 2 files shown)


app/views/topical_events/show.html.erb
```diff
@@ -120,7 +120,7 @@
 
   <div class="govuk-grid-column-two-thirds">
     <% if @topical_event.travel_advice.any? %>
-      <section id="travel-advice"
+      <section id="travel-advice">
```
BeckaL, line 123: 🤦‍♀️ good spot

BeckaL (approved):
LGTM

PR #2895, opened 2022-08-10
Add featured documents to world location news page

brucebolt:

## Screenshots
| Collections  | Whitehall  |
|---|---|
| Screenshot showing a world location news page rendered by Collections.  It has a heading 1 labelled "France and the UK" and a heading 2 labelled "Featured".  Under this, there are 3 image cards showing featured documents.  |  Screenshot showing a world location news page rendered by Whitehall.  It has a heading 1 labelled "France and the UK" and a heading 2 labelled "Featured".  Under this, there are 3 image cards showing featured documents in exactly the same way as in Collections.  There is additional page furniture (translation links and featured links) that are not yet present in the Collections rendered page. |

[Trello card](https://trello.com/c/WNZ1IZAe)

### DIFF
--- a/app/models/world_location_news.rb
+++ b/app/models/world_location_news.rb
@@ -17,4 +17,20 @@ class WorldLocationNews
   def description
     @content_item.content_item_data["description"]
   end
+
+  def ordered_featured_documents
+    @content_item.content_item_data.dig("details", "ordered_featured_documents")&.map do |document|
+      {
+        href: document["href"],
+        image_src: document.dig("image", "url"),
+        image_alt: document.dig("image", "alt_text"),
+        heading_text: document["title"],
+        description: ActionView::Base.full_sanitizer.sanitize(document["summary"]).truncate(160, separator: " "),
+        context: {
+          date: document["public_updated_at"]&.to_date,
+          text: document["document_type"],
+        },
+      }
+    end
+  end
 end
--- a/app/views/world_location_news/show.html.erb
+++ b/app/views/world_location_news/show.html.erb
@@ -10,4 +10,28 @@
       title: @world_location_news.title,
     } %>
   </div>
+
+  <div class="govuk-grid-column-full">
+    <% if @world_location_news.ordered_featured_documents&.any? %>
+      <section id="featured">
+        <%= render "govuk_publishing_components/components/heading", {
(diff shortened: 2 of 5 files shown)


app/models/world_location_news.rb
```diff
@@ -17,4 +17,20 @@ def title
   def description
     @content_item.content_item_data["description"]
   end
+
+  def ordered_featured_documents
+    @content_item.content_item_data.dig("details", "ordered_featured_documents")&.map do |document|
```
BeckaL, line 22: Here, I don't think we technically need the '&' since I made ordered_featured_documents mandatory on the schema. Not necessarily anti keeping this in though

app/models/world_location_news.rb
```diff
@@ -17,4 +17,20 @@ def title
...
+      {
+        href: document["href"],
+        image_src: document.dig("image", "url"),
+        image_alt: document.dig("image", "alt_text"),
+        heading_text: document["title"],
+        description: ActionView::Base.full_sanitizer.sanitize(document["summary"]).truncate(160, separator: " "),
```
BeckaL, line 28: Just out of interest, why are we sanitising here? Partly wondering if we need to do anything similar for e.g. topical events

BeckaL (approved):
Just a couple of questions/ minor comments, but looks good, and styling looks good to me.

app/models/world_location_news.rb
brucebolt: The description text contains HTML in the content item.  If we don't sanitize, then truncate will cut off the closing HTML tags.

BeckaL: Ah, do we need the same for topical events [here](https://github.com/alphagov/collections/blob/3aaaf6a662f6b4db60d06fb725af56d906850045/app/models/topical_event.rb#L75) in that case?

brucebolt: For some reason, we aren't converting topical events to HTML in Whitehall.  Maybe we should, just to be consistent?

BeckaL: Oh yeah - actually I think this consistency will come as part of this [spike ticket](https://trello.com/c/6duIhmSQ/192-spike-using-shared-featurable-code-for-topical-events-instead-of-custom-topicaleventfeaturingscontroller)

PR #2896, opened 2022-08-10
Add featured links to world location news page

brucebolt:

There are two differences:
- We are now using the `topic-list` component, which will make this view consistent with organisation pages.  This varies slightly from the Whitehall rendered version of the page.
- There are currently 6 featured links in the screenshot below.  This is because the code in Whitehall to publish these links does not currently limit the number of items.  This will be corrected in https://github.com/alphagov/whitehall/pull/6724.

## Screenshots
| Collections  | Whitehall  |
|---|---|
| Screenshot showing a world location news page rendered by Collections.  It has 6 links in the top right hand corner, which link to featured content.  |  Screenshot showing a world location news page rendered by Collections.  It has 5 links in the top right hand corner, which link to featured content. |

[Trello card](https://trello.com/c/i577upcN)

### DIFF
--- a/app/models/world_location_news.rb
+++ b/app/models/world_location_news.rb
@@ -18,6 +18,15 @@ class WorldLocationNews
     @content_item.content_item_data["description"]
   end
 
+  def ordered_featured_links
+    @content_item.content_item_data.dig("details", "ordered_featured_links")&.map do |link|
+      {
+        path: link["href"],
+        text: link["title"],
+      }
+    end
+  end
+
   def ordered_featured_documents
     @content_item.content_item_data.dig("details", "ordered_featured_documents")&.map do |document|
       {
--- a/app/views/world_location_news/show.html.erb
+++ b/app/views/world_location_news/show.html.erb
@@ -8,9 +8,18 @@
   <div class="govuk-grid-column-two-thirds">
     <%= render "govuk_publishing_components/components/title", {
       title: @world_location_news.title,
+      margin_top: 0,
     } %>
   </div>
 
+  <div class="govuk-grid-column-one-third">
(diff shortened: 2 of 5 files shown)


app/models/world_location_news.rb
```diff
@@ -17,4 +17,29 @@ def title
   def description
     @content_item.content_item_data["description"]
   end
+
+  def ordered_featured_links
+    @content_item.content_item_data.dig("details", "ordered_featured_links")&.map do |link|
```
BeckaL, line 22: same here, we don't technically need the &, but not too fussed

BeckaL (commented):
I think my main question here would be around alignment of the links - links currently look a bit weird to me not being vertically aligned with the title on the page.

brucebolt:
> Code all looks great. I think my main question here would be around alignment of the links - links currently look a bit weird to me not being vertically aligned with the title on the page. Do we have an example of another page where it renders like this? The organisations page is a bit different given it doesn't have the same kind of title.

I think I'll need a FE dev to help out with the alignment.  I'm unfamiliar with the best way to do this here.

BeckaL (approved):
> I think I'll need a FE dev to help out with the alignment. I'm unfamiliar with the best way to do this here.

The alignment actually looks better to me in [this PR](https://github.com/alphagov/collections/pull/2897), so maybe this is all good. Just trying to avoid another big frontend tidy up PR at the end.

brucebolt:
> The alignment actually looks better to me in [this PR](https://github.com/alphagov/collections/pull/2897), so maybe this is all good.

The translation links won't always be displayed (if the page is only available in English), so we should probably try and resolve this.

PR #2899, opened 2022-08-11
Add mission statement to world location news page

brucebolt:

## Screenshots
| Collections | Whitehall |
|---|---|
| Screenshot of mission statement rendered by Collections.  There is a heading 2 titled 'Our mission' then a paragraph of text. | Screenshot of mission statement rendered by Whitehall.  There is a heading 2 titled 'Our mission' then a paragraph of text. |

[Trello card](https://trello.com/c/wvMip7tS/176-add-mission-statement-to-world-location-news-page-in-collections)

### DIFF
--- a/app/views/world_location_news/show.html.erb
+++ b/app/views/world_location_news/show.html.erb
@@ -35,3 +35,21 @@
     <% end %>
   </div>
 </div>
+
+<% if @world_location_news.mission_statement.present? %>
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds govuk-!-margin-bottom-6">
+      <%= render "govuk_publishing_components/components/heading", {
+        text: I18n.t("world_location_news.headings.mission"),
+        padding: true,
+        font_size: "l",
+        border_top: 2,
+        margin_bottom: 2,
+      } %>
+
+      <p class="govuk-body">
+        <%= @world_location_news.mission_statement %>
+      </p>
+    </div>
+  </div>
+<% end %>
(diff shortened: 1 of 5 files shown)


app/views/world_location_news/show.html.erb
```diff
@@ -49,3 +49,21 @@
     <% end %>
   </div>
 </div>
+
+<% if @world_location_news.mission_statement %>
```
BeckaL, line 53: This is always present, but can be an empty string, so I don't think this if statement will currently work the way we want it to

spec/features/world_location_news_spec.rb
```diff
@@ -77,6 +77,29 @@
...
+      expect(page).to have_text(I18n.t("world_location_news.headings.mission"))
+    end
+
+    it "includes the mission statement" do
+      visit base_path
+      expect(page).to have_text("Our mission is to test world location news.")
+    end
+  end
+
+  context "when there is no mission statement" do
```
BeckaL, line 92: As per above, think this test needs to be when it's an empty string

config/locales/ar/world_location_news.yml
```diff
@@ -1,5 +1,12 @@
----
```
BeckaL, line 1: what's up with these lines / lines getting removed? Can we just take them out of the original PR if we don't need them?

app/views/world_location_news/show.html.erb
brucebolt: Ah, I'll update this to check for the empty string.

config/locales/ar/world_location_news.yml
brucebolt: This is an artefact of me branching off an old version of another branch.  I've rebased, so the lines will remain now (they are put there by `rake translation:normalize`).

BeckaL (approved):
Great! LGTM

PR #2900, opened 2022-08-11
Build the Cost of living hub frontend

patrickpatrickpatrick:
This PR adds the frontend for the Cost of Living hub. It is based on the already existing Browse pages (including some of the classes used for those pages) and uses the accordion component. As there is no content in the content store for the pages, this page content is stored in a YML file.

Relevant Trello Cards
- [Building Header for Cost of Living page ](https://trello.com/c/tZXyaXV6/1203-create-header-of-cost-of-living-hub-m)
- [Building Body for Cost of Living page](https://trello.com/c/Ny93RuWT/1206-create-body-of-cost-of-living-hub-m)

### DIFF
--- a/app/views/cost_of_living_landing_page/_links.html.erb
+++ b/app/views/cost_of_living_landing_page/_links.html.erb
@@ -1,0 +1,55 @@
+<% track_action ||= "contentLink" %>
+
+<ul class="govuk-list browse__list" data-module="gem-track-click">
+  <% list.each_with_index do |content_item, index| %>
+    <% if content_item[:links] %>
+      <li class="govuk-list browse__list-item govuk-!-padding-bottom-1">
+         <% if content_item[:subheading] %>
+           <%= render "govuk_publishing_components/components/heading", {
+            text: content_item[:subheading],
+            heading_level: 3,
+            font_size: "s",
+            padding: index > 0 ? true : false,
+            margin_bottom: index === 0 ? 3 : 0,
+          } %>
+         <% end %>
+         <% if content_item[:links] %>
+          <ul class="govuk-list browse__list">
+             <% content_item[:links].each_with_index do |link_item, index| %>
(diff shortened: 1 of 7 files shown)


app/views/browse/show.html.erb
```diff
@@ -22,6 +22,11 @@
   <h1 class="browse__heading govuk-heading-xl">
     <%= page.title %>
   </h1>
+  <%= render "govuk_publishing_components/components/lead_paragraph", {
```
hannalaakso, line 25: This is currently adding a lead second paragraph to the browse pages. Making both the CoL and browse pages use the same header partial feels like a sensible change but one we could do when we do the refactoring later. So it might be safer to delete the changes to this file for now to avoid changing the live pages.

app/views/cost_of_living_landing_page/_links.html.erb
```diff
@@ -0,0 +1,55 @@
+<% track_action ||= "contentLink" %>
+
+<ul class="govuk-list browse__list" data-module="gem-track-click">
+  <% list.each_with_index do |content_item, index| %>
+    <% if content_item[:subheading] || content_item[:links] %>
+      <li class="govuk-list browse__list govuk-!-padding-bottom-1">
```
hannalaakso, line 6: I think this should be `.browse_list-item`.

hannalaakso (commented):
Looks good, just left a couple of small comments. 

(Just for anyone else reading this, we haven't yet refactored the code to consolidate it with the browser and covid templates as we're still waiting for the final version of the content. )

app/views/browse/show.html.erb
patrickpatrickpatrick: oh i forgot to delete the p element below it, should i do that or should i remove this lead paragraph component

hannalaakso (approved):
Thanks for addressing my comments so quickly, LGTM now 👍

PR #2905, opened 2022-08-15
Add document sections to world location news page

brucebolt:

We have tried to reuse as much code as possible from topical events in rendering this section.

The horizontal line between each section has been updated to match the style of Topical Event pages.  This differs from the rendering in Whitehall.

The number of documents in each section has been updated to match Topical Event pages, meaning we can re-use code without having inconsistencies between the content types.

## Screenshots
| Collections | Whitehall |
|---|---|
| Screenshot of part of the world location news page for Cyprus rendered by Collections.  There is a header labelled Documents.  Under this there are three sections: announcements, publications and statistics.  They all contain a maximum of 3 documents.  There is no horizontal line between each document type. | Screenshot of part of the world location news page for Cyprus rendered by Whitehall.  There is a header labelled Documents.  Under this there are three sections: announcements, publications and statistics.  They all contain a maximum of 2 documents.  There is a horizontal line between each document type. |

[Trello card](https://trello.com/c/O7wicoig)

### DIFF
--- a/app/helpers/document_list_helper.rb
+++ b/app/helpers/document_list_helper.rb
@@ -15,7 +15,8 @@ private
       consultation: "/search/policy-papers-and-consultations?",
       detailed_guidance: "/search/all?",
       latest: "/search/all?",
-      publication: "/search/all?" }[content_type]
+      publication: "/search/all?",
+      statistic: "/search/research-and-statistics?" }[content_type]
   end
 
   def topical_event_query_params(content_type, slug)
@@ -47,6 +48,16 @@ private
     when :latest
       {
         order: "updated-newest",
+        world_locations: [slug],
+      }
+    when :publication
+      {
+        content_purpose_supergroup: %w[guidance_and_regulation policy_and_engagement transparency],
+        order: "updated-newest",
+        world_locations: [slug],
+      }
+    else
+      {
         world_locations: [slug],
       }
     end
--- a/app/models/world_location_news.rb
+++ b/app/models/world_location_news.rb
@@ -66,4 +66,16 @@ class WorldLocationNews
   def latest
     @latest ||= @documents_service.fetch_related_documents_with_format
   end
+
+  def announcements
+    @announcements ||= @documents_service.fetch_related_documents_with_format({ filter_content_purpose_supergroup: "news_and_communications" })
+  end
+
+  def publications
+    @publications ||= @documents_service.fetch_related_documents_with_format({ filter_content_purpose_supergroup: %w[guidance_and_regulation policy_and_engagement transparency] })
+  end
+
+  def statistics
+    @statistics ||= @documents_service.fetch_related_documents_with_format({ filter_content_purpose_subgroup: "statistics" })
+  end
 end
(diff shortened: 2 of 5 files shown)


app/helpers/document_list_helper.rb
```diff
@@ -49,6 +50,16 @@ def world_location_news_query_params(content_type, slug)
         order: "updated-newest",
         world_locations: [slug],
       }
+    when :publication
+      {
+        content_purpose_supergroup: %w[guidance_and_regulation policy_and_engagement transparency],
+        order: "updated-newest",
```
BeckaL, line 56: Do we want definitely want this ordering? Latest obvs needs it, but I haven't been sure about the others - I think to me it'd make more sense if they were in order of most viewed (or some other relevance metric). Not sure what the default is

brucebolt: I added this as it's the order we're retrieving documents on the World Location News page, so it'll ensure the preview always matches the first two results in the finder.

BeckaL (commented):
Nice one, rocketing through this rendering work! All looks sensible, just a question about ordering. 

Bit annoying that we've got an inconsistency between the number of documents displayed between topical events and these ones, but the inconsistency already exists between organisations and topical events, so 🤷‍♀️

app/helpers/document_list_helper.rb
BeckaL: I think we have this issue with other document types - was something I was pondering earlier, as topical event doc types have the same issue. I was thinking the nicer way around it would be for everything everywhere except latest to be ordered by relevance

BeckaL: So I guess this is a question that goes beyond just this PR, but wondered what your thoughts were

brucebolt: To me, I'd rather the documents were sorted by most recently published, otherwise there's a chance that new documents won't be seen by users of the finder (and the 'most viewed' stats could be skewed towards the older documents, rather than those that are most popular).

If there's something super important, it could be added as a featured document.

Maybe this is something that a PM could make a decision on?

BeckaL: OK that makes sense. Will approve this and maybe we can raise the general question with a PM

brucebolt:
> Bit annoying that we've got an inconsistency between the number of documents displayed between topical events and these ones, but the inconsistency already exists between organisations and topical events, so 🤷‍♀️

Yes, it is.  I'm tempted to set both to 3.

I can't see any reason why we're limiting to 2 for World Location News, but 3 for Topical Events.

BeckaL (approved):

BeckaL:
> Yes, it is. I'm tempted to set both to 3.

> I can't see any reason why we're limiting to 2 for World Location News, but 3 for Topical Events.

Not anti this at all, totally agree

brucebolt:
> > Yes, it is. I'm tempted to set both to 3.
> 
> > I can't see any reason why we're limiting to 2 for World Location News, but 3 for Topical Events.
> 
> Not anti this at all, totally agree

Done.  I've updated the screenshot in the PR.

PR #2907, opened 2022-08-15
Show guidance and regulation rather than detailed guides on topical events

BeckaL:
We have switched to using the content_purpose_supergroup of guidance_and_regulation rather than the filter_format of detailed_guidance for topical events. This aligns more with other content types in collections, e.g. organisations, which prefer content_purpose_supergroups instead of specific document types. This also means that the link to search is now more relevant, as it links to all guidance and regulation for the topical event, rather than just to detailed guides.

This also updates the announcements section to use the existing content purpose supergroup for this type - no changes expected to content on page.

| Rendering of section currently | Rendering of section as per this PR |
| --------- | ----------- |
| Screenshot 2022-08-15 at 16 26 43 | Screenshot 2022-08-15 at 16 24 45|

[Trello](https://trello.com/c/SkGhAxxl/181-update-topical-events-finder-link-to-just-show-detailed-gudiance)

### DIFF
--- a/app/models/topical_event.rb
+++ b/app/models/topical_event.rb
@@ -111,8 +111,8 @@ class TopicalEvent
     @announcements ||= @documents_service.fetch_related_documents_with_format({ filter_content_store_document_type: announcement_document_types })
   end
 
-  def detailed_guidance
-    @detailed_guidance ||= @documents_service.fetch_related_documents_with_format({ filter_format: "detailed_guidance" })
+  def guidance_and_regulation
+    @guidance_and_regulation ||= @documents_service.fetch_related_documents_with_format({ filter_content_purpose_supergroup: "guidance_and_regulation" })
   end
 
   def organisations
--- a/app/models/topical_event.rb
+++ b/app/models/topical_event.rb
@@ -97,18 +97,7 @@ class TopicalEvent
   end
 
   def announcements
-    announcement_document_types = %w[
-      press_release
-      news_article
-      news_story
-      fatality_notice
-      speech
-      written_statement
-      oral_statement
-      authored_article
-      government_response
-    ]
-    @announcements ||= @documents_service.fetch_related_documents_with_format({ filter_content_store_document_type: announcement_document_types })
+    @announcements ||= @documents_service.fetch_related_documents_with_format({ filter_content_purpose_supergroup: "news_and_communications", reject_content_purpose_subgroup: %w[decisions updates_and_alerts] })
   end
 
   def guidance_and_regulation
--- a/config/locales/en/topical_events.yml
+++ b/config/locales/en/topical_events.yml
@@ -5,9 +5,9 @@ en:
     headings:
       announcements: Announcements
       consultations: Consultations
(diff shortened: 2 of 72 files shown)


brucebolt (approved):

brucebolt (approved):

app/models/topical_event.rb
```diff
@@ -97,18 +97,7 @@ def consultations
...
-      news_story
-      fatality_notice
-      speech
-      written_statement
-      oral_statement
-      authored_article
-      government_response
-    ]
-    @announcements ||= @documents_service.fetch_related_documents_with_format({ filter_content_store_document_type: announcement_document_types })
+    @announcements ||= @documents_service.fetch_related_documents_with_format({ filter_content_purpose_supergroup: "news_and_communications" })
```
AgaDufrat, line 100: It looks like ["news_and_communications"](https://github.com/alphagov/govuk_document_types/blob/main/data/supergroups.yml#L6-L11) will also include [other document types](https://github.com/alphagov/govuk_document_types/blob/main/data/supertypes.yml#L158-L188), not listed in the array that's being replaced. Just wanted to check that it is expected. 

You can use `reject_content_purpose_subgroup` to filter stuff out.

AgaDufrat (approved):
LGTM. Left one comment.

config/locales/en/topical_events.yml
```diff
@@ -5,9 +5,9 @@ en:
     headings:
       announcements: Announcements
       consultations: Consultations
-      detailed_guides: Detailed Guides
       documents: Documents
       featured: Featured
+      guidance_and_regulation: Guidance and Regulation
```
AgaDufrat, line 10:
```suggestion
      guidance_and_regulation: Guidance and regulation
```
We use sentence-case capitalisation for heading e.g. https://github.com/alphagov/collections/blob/main/config/locales/en/organisations.yml#L249

Would you mind also updating `Travel Advice` in a separate commit while you are there ;)

BeckaL: Ah, good spot. Done and added travel advice separately as a commit as you suggested 🙂

app/models/topical_event.rb
BeckaL: Ahh, good spot. I'm not too worried about the original list as it was just copied from Whitehall and I'm inclined not to read too much into the exact list we had there, but have now updated this query to exclude the same content_purpose_supergroups as we're [excluding elsewhere in this app for news_and_communications](https://github.com/alphagov/collections/blob/fc9db4ef6d87ad52d0ac795d1dd05d5fa116dde1/app/services/search/supergroups.rb#L17). We should also switch to using this same supergroups code at some point (don't think we saw it at the time) but I'm considering this outside of the scope of this PR.

BeckaL:
Thanks very much for your review @AgaDufrat! Super helpful.

I'm off tomorrow so won't be merging it until Monday (don't want to do it at this point in the day), sorry to leave it hanging around, but if anything else occurs to you or you want to give it another look tomorrow you're very welcome to 🙂

PR #2908, opened 2022-08-16
Add world location news atom feed and email signup link

brucebolt:
This adds the following for World Location News, which are currently rendered by Whitehall:
- the rendering of the atom feed
- a link to the atom feed (using the same component as topical events, to make this consistent)
- a link to signup for email alerts

### Whitehall rendered links
Screenshot showing the latest document section, with two buttons: Get emails and Subscribe to feed.

### Collections rendered link
Before clicking the "Subscribe to feed" link:
Screenshot showing the latest document section, with two buttons: Get emails and Subscribe to feed.

After clicking the "Subscribe to feed" link:
Screenshot showing the latest document section, with two buttons: Get emails and Subscribe to feed.  The subscribe to feed button is highlighted in yellow and a section has appeared underneath, containing an explanation of the feed and a link to the feed.

### Whitehall rendered atom feed
Screenshot showing part of an atom feed rendered by Whitehall.  It is identical to collections, except the update times are only accurate to the day, not the exact time.

### Collections rendered atom feed
Screenshot showing part of an atom feed rendered by Collections.  It is identical to Whitehall, except the update times are accurate.

[Trello card](https://trello.com/c/3IV2aDkp)

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -83,6 +83,10 @@ Rails.application.routes.draw do
     get "/:slug", to: "step_nav#show"
   end
 
+  get "/world/:name/news(.:locale).atom",
+      to: "world_location_news#show",
+      as: :world_location_news_feed
+
   get "/world/:name/news(.:locale)",
       to: "world_location_news#show",
       as: :world_location_news
--- a/app/controllers/world_location_news_controller.rb
+++ b/app/controllers/world_location_news_controller.rb
@@ -1,8 +1,22 @@
 class WorldLocationNewsController < ApplicationController
   around_action :switch_locale
+  enable_request_formats show: :atom
 
   def show
-    @world_location_news = WorldLocationNews.find!(request.path)
-    setup_content_item_and_navigation_helpers(@world_location_news)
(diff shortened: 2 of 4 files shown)


config/routes.rb
```diff
@@ -81,6 +81,10 @@
     get "/:slug", to: "step_nav#show"
   end
 
+  get "/world/:name/news(.:locale).atom",
```
BeckaL, line 84: Probably more for my understanding than anything since I'm not super familiar with all the routes stuff - why do we have to define this atom route explicitly for world location news but not for topical events? Organisations also does things differently to both but I'm assuming that's because it supports non English feeds

brucebolt: I tried the same approach as topical events, but the `.atom` got captured as `:locale`.

I can't quite understand what's happening with organisations, as it seems rather more complicated than it needs to be.

BeckaL: Ah, I see, that makes sense re topical events vs these

BeckaL (approved):
Looks good to me, have had a look at UI too, all 👍

PR #2920, opened 2022-08-23
Add additional tracking for Cost of living hub

AgaDufrat:
Adds in additional tracking required from [Performance analyst docs](https://docs.google.com/presentation/d/1ERC0U2vI76t2Nm4Y4gLJWNoHgPl6iy2YvVr4stmduJA/edit#slide=id.g14486b99c14_0_23).
[Trello](https://trello.com/c/BrDDL2pi/1244-update-hub-page-with-additional-analytics-m)

Tracking for "Check what support..." is already in place: https://github.com/alphagov/collections/blob/main/app/views/cost_of_living_landing_page/show.html.erb#L67-L72

Tested EventListeners locally.

_Note:_ This could benefit from a presenter so we can move some logic. Given the frontend devs are still working on the views we may want to delay this refactor.

### DIFF
--- a/app/views/cost_of_living_landing_page/_links.html.erb
+++ b/app/views/cost_of_living_landing_page/_links.html.erb
@@ -1,5 +1,3 @@
-<% track_action ||= "contentLink" %>
-
 <ul class="govuk-list browse__list" data-module="gem-track-click">
   <% list.each_with_index do |content_item, index| %>
     <% if content_item[:links] %>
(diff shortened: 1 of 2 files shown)


app/controllers/cost_of_living_landing_page_controller.rb
```diff
@@ -4,11 +4,9 @@ class CostOfLivingLandingPageController < ApplicationController
   rescue_from GdsApi::ContentStore::ItemNotFound, with: :show
 
   def show
-    @content_item = { "locale" => "en" }
-
```
chao-xian, line 8: You can drop this commit as I made this change and it's in `main` now.

injms (approved):
Looks good to me - all set up correctly 👍

PR #2924, opened 2022-08-25
Fix the announcement of unordered lists in Safari+Voiceover

matthillco:
## What
Unordered lists aren't announced by Voiceover in Safari when the bullet styling is removed. This is a known issue, for example see here: https://www.scottohara.me/blog/2019/01/12/lists-and-safari.html

Adding `role=“list”` to the UL element forces Voiceover to make the correct announcement in Safari.


injms (approved):
Thanks for fixing this 👍

PR #2926, opened 2022-08-26
Cost of living page header should match topics page header

matthillco:
## What

The page header for the Cost of Living hub should match the page header for topics pages, both structurally and visually. Specific areas to change are:

* The blue bar should be the width of the content, not the full width of the page
* Matching the vertical margins and padding on header elements
* Removing redundant rows and columns

## Why
For visual consistency with the header on mainstream topics pages.

## Before
image

## After
image

### DIFF
--- a/app/views/cost_of_living_landing_page/show.html.erb
+++ b/app/views/cost_of_living_landing_page/show.html.erb
@@ -29,54 +29,42 @@
   end
 %>
 
-<header class="cost-of-living-hub__header">
-  <%= render "shared/browse_header", { margin_bottom: 7, padding_top: 0 } do %>
-    <%= render 'govuk_publishing_components/components/breadcrumbs', {
-      breadcrumbs: breadcrumbs,
-      collapse_on_mobile: true,
-      inverse: true
-    } %>
+<%= render "shared/browse_breadcrumbs", {
+  breadcrumbs: breadcrumbs,
+  collapse_on_mobile: true,
+  inverse: true
+} %>
 
-    <div class="govuk-grid-row">
-      <div class="govuk-grid-column-two-thirds">
-        <%= render "govuk_publishing_components/components/title", {
-          title: content.header[:heading],
-          inverse: true,
-          margin_top: 4,
-          margin_bottom: 6,
-        } %>
-      </div>
-    </div>
+<%= render "shared/browse_header", { margin_bottom: 7, padding_top: 0 } do %>
 
-    <div class="govuk-grid-row">
-      <div class="govuk-grid-column-two-thirds">
-        <%= render "govuk_publishing_components/components/lead_paragraph", {
-          text: content.header[:lede],
-          inverse: true,
-          margin_bottom: 2,
-        } %>
-      </div>
-    </div>
+  <%= render "govuk_publishing_components/components/title", {
+    title: content.header[:heading],
+    inverse: true,
+    margin_top: 6,
+    margin_bottom: 6,
+  } %>
 
-    <div class="govuk-grid-row covid__action-link-wrapper">
-      <div class="govuk-grid-column-two-thirds">
-        <%= render 'govuk_publishing_components/components/action_link', {
-          white_arrow: true,
-          href: content.header[:href],
-          text: content.header[:link_text],
-          mobile_subtext: true,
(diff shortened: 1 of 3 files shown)


app/views/cost_of_living_landing_page/show.html.erb
```diff
@@ -28,20 +28,25 @@
...
 %>
 
+<%= render "shared/browse_breadcrumbs", {
+  breadcrumbs: breadcrumbs,
+  column_two_thirds: true,
+  collapse_on_mobile: true,
+  inverse: true
+} %>
+
 <header class="cost-of-living-hub__header">
```
hannalaakso, line 38: I think we can now delete the `cost-of-living-hub__header` class from this wrapper, seeing that we removed the related style for it. 

(In fact, I think we don't need the `<header>` tag here, we don't have it on the browse pages. That's unrelated to the changes in this PR but we could just remove the wrapper here too.)

matthillco: Good spot, thank you. I've now removed not only the header but also all the redundant row/column classes, and adjusted the margins to match the browse pages. This is now much cleaner and more robust.

matthillco: Good spot, thank you. I've now removed the header element and also various redundant rows/columns, and adjusted margins accordingly to match the treatment used on Browse pages.

app/views/cost_of_living_landing_page/show.html.erb
```diff
@@ -28,54 +28,42 @@
...
-      <div class="govuk-grid-column-two-thirds">
-        <%= render "govuk_publishing_components/components/title", {
-          title: content[:header][:heading],
-          inverse: true,
-          margin_top: 4,
-          margin_bottom: 6,
-        } %>
-      </div>
-    </div>
+<%= render "shared/browse_header", { margin_bottom: 7, padding_top: 0 } do %>
```
hannalaakso, line 37: Since we removed the two-thirds column, I think we need to pass `two_thirds: true` to the component here to match the width of the browse header markup.

app/views/cost_of_living_landing_page/show.html.erb
```diff
@@ -28,54 +28,42 @@
...
-          text: content[:header][:lede],
-          inverse: true,
-          margin_bottom: 2,
-        } %>
-      </div>
-    </div>
+  <%= render "govuk_publishing_components/components/title", {
+    title: content[:header][:heading],
+    inverse: true,
+    margin_top: 6,
```
hannalaakso, line 42: This adds quite a lot of margin on top of the component on mobile. 

Screenshot 2022-08-31 at 18 37 35

The browse header actually uses the heading component here, not the title component. I'm wondering if switching to it here would fix this.

hannalaakso (commented):
Left a couple more small comments. Also created a [tech debt card](https://trello.com/c/Dm9KrOBT/692-make-browse-header-into-a-component) to look at turning the header into an actual component to make it easier to keep these different versions in sync.

app/views/cost_of_living_landing_page/show.html.erb
matthillco: Good spot, thanks. This structure for the title and lead paragraph has now been changed to match with browse/topics pages. This fixes the too-big top margin on mobile and improves overall consistency.

app/views/cost_of_living_landing_page/show.html.erb
```diff
@@ -28,54 +28,42 @@
...
 %>
 
-<header class="cost-of-living-hub__header">
-  <%= render "shared/browse_header", { margin_bottom: 7, padding_top: 0 } do %>
-    <%= render 'govuk_publishing_components/components/breadcrumbs', {
-      breadcrumbs: breadcrumbs,
-      collapse_on_mobile: true,
-      inverse: true
-    } %>
+<%= render "shared/browse_breadcrumbs", {
```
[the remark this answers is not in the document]
matthillco: This is now working with the shared breadcrumbs, the full-width parameter was removed to enable it.

hannalaakso (approved):
Thanks for addressing my comments @matthillco  , this LGTM 👍

PR #2927, opened 2022-08-30
Add analytics that power page views custom dimensions to Cost of living hub

AgaDufrat:
# Context

Additional meta tags that power the custom dimensions we normally see with events or page views aren't firing on the CoL page due to the content being hardcoded rather than coming from content store.

[Trello](https://trello.com/c/0pjp3DgK/1288-add-meta-tags-that-power-custom-dimensions-m)

Custom dimensions are populated via [govuk_publishing_components](https://github.com/alphagov/govuk_publishing_components/blob/465040b/app/assets/javascripts/govuk_publishing_components/analytics/custom-dimensions.js) gem.

## Depends on

-  https://github.com/alphagov/govuk_publishing_components/pull/2921

# What

This adds missing tracking attributes which populate the following custom dimensions:

- 26 - number of sections 
- 27 - number of links 
- 31 - navigation-list-type - "curated”
- 32 - navigation-page-type - “Cost of living hub”

It also includes a refactor to add a presenter so we can avoid having so much logic in the view and to be able to add unit test.

---

### DIFF
--- a/app/presenters/cost_of_living_landing_page_presenter.rb
+++ b/app/presenters/cost_of_living_landing_page_presenter.rb
@@ -1,0 +1,16 @@
+class CostOfLivingLandingPagePresenter
+  COMPONENTS = %i[
+    page_title
+    metadata
+    header
+    body
+  ].freeze
+
+  def initialize(content_item)
+    COMPONENTS.each do |component|
+      define_singleton_method component do
+        content_item[component]
+      end
+    end
+  end
+end
--- a/app/presenters/cost_of_living_landing_page_presenter.rb
+++ b/app/presenters/cost_of_living_landing_page_presenter.rb
@@ -13,4 +13,24 @@ class CostOfLivingLandingPagePresenter
       end
     end
   end
+
+  def link_clicked_track_data(track_action:, href:)
+    return {} unless internal_link?(href)
+
+    {
+      track_category: "contentsClicked",
+      track_action: track_action,
+      track_label: slug_for_href(href),
+    }
+  end
+
+private
+
+  def internal_link?(link)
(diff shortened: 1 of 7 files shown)


app/presenters/cost_of_living_landing_page_presenter.rb
```diff
@@ -13,4 +13,24 @@ def initialize(content_item)
...
+  end
+
+private
+
+  def internal_link?(link)
+    link.starts_with?("https://www.gov.uk/")
+  end
+
+  def slug_for_href(href)
+    href.gsub("https://www.gov.uk/", "")
```
KludgeKML, line 35: Is this going to cause problems in staging/integration? (I guess we don't really have analytics in those environments, but is it going to cause an unexpected difference in how link attributes are rendered between those and live?)

AgaDufrat: Agree this is sub-optimal. We have a [card](https://trello.com/c/waEkLAut/1279-generate-correct-base-urls-for-internal-links-based-on-environment-m-post-live) in the backlog to generate correct base URLs for internal links based on environment. We'll do that after the page is live.

KludgeKML: Okay, cool. :)

KludgeKML (approved):
LGTM

hannalaakso (approved):
Frontend portion looks good too 👍 I tested this locally together with https://github.com/alphagov/govuk_publishing_components/pull/2921 and the dimensions were being set (and after this is merged there'll be a performance analyst review to double check the tracking is correct.)

PR #2934, opened 2022-08-31
Correct order of translations on World Location news pages

callumknights:
Screenshot 2022-08-31 at 13 04 31

Trello: https://trello.com/c/6mCz9HL2/225-world-location-news-rendering-in-collections-fix-order-of-translated-tabs

### DIFF
--- a/app/models/world_location_news.rb
+++ b/app/models/world_location_news.rb
@@ -52,15 +52,15 @@ class WorldLocationNews
     @content_item.content_item_data.dig("details", "mission_statement")
   end
 
-  def translations
-    @translations ||= @content_item.content_item_data.dig("links", "available_translations")&.map do |translation|
+  def ordered_translations
+    @ordered_translations ||= @content_item.content_item_data.dig("links", "available_translations")&.map { |translation|
       {
         locale: translation["locale"],
         base_path: translation["base_path"],
         text: I18n.t("shared.language_name", locale: translation["locale"]),
         active: I18n.locale.to_s == translation["locale"],
       }
-    end
+    }&.sort_by { |t| t[:locale] == I18n.default_locale.to_s ? "" : t[:locale] }
   end
 
   def latest
--- a/app/views/world_location_news/show.html.erb
+++ b/app/views/world_location_news/show.html.erb
@@ -14,9 +14,9 @@
   </div>
 
   <div class="govuk-grid-column-one-third">
-    <% if @world_location_news.translations.count > 1 %>
+    <% if @world_location_news.ordered_translations.count > 1 %>
       <%= render "govuk_publishing_components/components/translation_nav", {
-        translations: @world_location_news.translations,
+        translations: @world_location_news.ordered_translations,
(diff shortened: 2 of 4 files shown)


app/models/world_location_news.rb
```diff
@@ -63,6 +63,10 @@ def translations
     end
   end
 
+  def ordered_translations
+    translations.sort_by { |t| t[:locale] == I18n.default_locale.to_s ? "" : t[:locale] }
```
BeckaL, line 67: I'm not sure this is worth having as a separate method - think we could just in line this in the translations method, and maybe rename that to ordered translations? Think we have a bit of a tendency to do tiny methods but I don't think it makes it more readable.

So something like 

```
  def ordered_translations
    @ordered_translations ||= @content_item.content_item_data.dig("links", "available_translations")&.map do |translation|
      {
        locale: translation["locale"],
        base_path: translation["base_path"],
        text: I18n.t("shared.language_name", locale: translation["locale"]),
        active: I18n.locale.to_s == translation["locale"],
      }
    end.sort_by { |t| t[:locale] == I18n.default_locale.to_s ? "" : t[:locale] }
  end
```

If not, maybe we should make translations private to make it clear we should only be accessing the ordered version, but my preference would be to inline it

BeckaL:
Looks good! Just a comment on in-lining a method. I also think we could add a bit more detail to the commit message to explain what the issue was and what the new behaviour is.

app/models/world_location_news.rb
```diff
@@ -52,15 +52,16 @@ def mission_statement
...
+  def ordered_translations
+    @ordered_translations ||= @content_item.content_item_data.dig("links", "available_translations")&.map do |translation|
       {
         locale: translation["locale"],
         base_path: translation["base_path"],
         text: I18n.t("shared.language_name", locale: translation["locale"]),
         active: I18n.locale.to_s == translation["locale"],
       }
     end
+    @ordered_translations.sort_by { |t| t[:locale] == I18n.default_locale.to_s ? "" : t[:locale] }
```
BeckaL, line 64: I think there might be an issue here where the cached version of ordered translations is not sorted, so only when we rely on the non-cached version is it sorted. So, we need to cache the sorted stuff as well. 

Suggestion (I think rubocop didn't like my original):

```    
@ordered_translations ||= @content_item.content_item_data.dig("links", "available_translations")&.map { |translation|
      {
        locale: translation["locale"],
        base_path: translation["base_path"],
        text: I18n.t("shared.language_name", locale: translation["locale"]),
        active: I18n.locale.to_s == translation["locale"],
      }
    }&.sort_by { |t| t[:locale] == I18n.default_locale.to_s ? "" : t[:locale] }
```

callumknights: Ended up going with the suggestion above. I didn't realise the caching issue at all.

BeckaL (approved):
Nice one, looks good to me!

hannako (approved):
This looks good Callum!

PR #2935, opened 2022-09-01
Enable Static proxy for Heroku deployments

theseanything:
This sets the GOVUK_PROXY_STATIC_URL env var to enable the proxy to Static in production. This is so the application continues, when Static changes to use relative URLs for assets. See https://github.com/alphagov/govuk_app_config/pull/261 and https://github.com/alphagov/govuk-puppet/pull/11801


theseanything:
Superseded by https://github.com/alphagov/collections/pull/2936

PR #2937, opened 2022-09-02
Document type translation changes

JonathanHallam:
This was originally done to fix some issues we were having with the naming of content types. It morphed into moving a lot of translations to make sure that we kept the functionality. 

## How to review
- Probably best to ignore the I18n changes in the first commit, they're just moving translations from organisations to shared
- There's a gist linked in the 2nd commit message that shows how I moved the I18n stuff, that's way easier to look at than several thousand lines of change

Trello - https://trello.com/c/fPWiIEXI/226-fix-mismatch-in-featured-document-types-on-world-location-news-page-in-collections


callumknights (approved):
LGTM 👍

config/locales/en/shared.yml
```diff
@@ -165,36 +234,54 @@ en:
...
         one: Rail Accident Investigation Branch report
         other: Rail Accident Investigation Branch reports
+      recruitment:
+        one: Recruitment
+        other: Recruitment
       regulation:
         one: Regulation
         other: Regulations
       research:
-        one: Research and analysis
```
AgaDufrat: Are we sure about this content change? I'd suggest leaving it unchanged for the moment.

config/locales/en/shared.yml
```diff
@@ -225,12 +318,15 @@ en:
...
         one: Administrative appeals tribunal decision
         other: Administrative appeals tribunal decisions
+      welsh_language_scheme:
+        one: Welsh language scheme
+        other: Welsh language schemes
       world_location_news_article:
         one: News article
         other: News articles
       world_news_story:
-        one: World news story
```
AgaDufrat: Similar to my previous comment.

AgaDufrat (commented):
It makes sense to bring those translations over to ease the work changing the rendering from Whitehall to Collections.

Very reasonable to move document_type translations to `shared` since they are used by organisations and roles. I like how it simplifies the calls.

I noticed that adding those translations introduced some content changes. 
I’d suggest we keep them as they for now. We can decide whether to change the copy for all use cases or add a new key when we get to the point of using those translations.

It’d be nice if the PR description included info what the translation changes actually are ;)

config/locales/en/shared.yml
JonathanHallam: Heya, have changed this so it stores the existing Collections translations rather than overwriting them :) excellent spot!

AgaDufrat (approved):
Thanks for updating the PR and collating them so nicely. 
Translations are so mind-boggling 😵

PR #2945, opened 2022-09-08
Remove JSON endpoint from BrowseController

patrickpatrickpatrick:
The JSON endpoint was used to generate the miller columns on browse pages. We no longer use miller columns for browse pages and enough time has passed now that cached versions of browse pages using miller columns should have been superseded by the new browse pages. The test for the JSON endpoint has also been removed.

[Relevant Trello Card](https://trello.com/c/brX2fuXL/1193-remove-miller-column-json-api-endpoint-s)


AgaDufrat (approved):
Can we also delete `app/views/second_level_browse_page/_second_level_browse_pages.html.erb` as it doesn't appear to be used?

I'm not entirely sure about the scope of this card. Do we also need to delete it for second level browse? https://github.com/alphagov/collections/blob/527757e/app/controllers/second_level_browse_page_controller.rb#L10-L22

On a separate note, we should probably also delete stuff around the `app/views/second_level_browse_page/old_show.html.erb` (not in this PR)

PR #2953, opened 2022-09-12
Revert "Bump govuk_publishing_components from 30.5.0 to 30.5.1"

chao-xian:
Reverts alphagov/collections#2950


chao-xian:
Opened this to check if changes was causing a build failure (which I didn't think it would be).

PR #2958, opened 2022-09-13
Use request path for World Location News

brucebolt:
We are currently using a path that doesn't contain the locale when requesting World Location News from the content store.

This means we can't currently show anything other than English content.

Changing to `request.path` means we will include the locale when retrieving the content.

This is consistent with the usage on organisation pages.

[Trello card](https://trello.com/c/APn84yVb/223-spike-into-whats-involved-in-rendering-non-english-world-location-news-in-collections)

### DIFF
--- a/app/controllers/world_location_news_controller.rb
+++ b/app/controllers/world_location_news_controller.rb
@@ -3,7 +3,7 @@ class WorldLocationNewsController < ApplicationController
   enable_request_formats show: :atom
 
   def show
-    path = world_location_news_path(params[:name])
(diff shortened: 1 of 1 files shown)


kevindew (approved):
Change looks good - should a test change at all?

brucebolt:
> Change looks good - should a test change at all?

We don't currently have any tests for non-English content, so no tests need to change in that regard, as we'll add the test when we get round to dealing with this. The behaviour here for the atom feed matches as before, so no tests need updating.

PR #2964, opened 2022-09-20
Add featured translation

brucebolt:
The word 'Featured' currently appears on Organisation pages, however we also include it (untranslated) in the Topical Event and World Location News pages.

Therefore moving the translations to the shared locale file so it can be translated the same in all locations.

This PR also normalizes the locale files according to the [pre-defined rules](https://github.com/alphagov/collections/blob/main/config/i18n-tasks.yml), which appears to have not been done for some time.

[Trello card](https://trello.com/c/c6j6LkYg)


BeckaL (approved):
All looks good to me, nice to move some more stuff to shared 👍

AgaDufrat (approved):

PR #2965, opened 2022-09-20
Render right to left languages properly on world location news translations

BeckaL:
[Trello](https://trello.com/c/RxH1E3eG/275-set-up-world-location-pages-to-deal-with-right-to-left-languages)

In order to switch rendering of translated world location news pages to collections, we need to ensure that we are rendering their text right-to-left. This PR fixes this.

| Currently rendered version (Whitehall) | Version rendered by collections after this PR |
|-----|-----|
|Screenshot 2022-09-20 at 16 33 34|Screenshot 2022-09-20 at 16 33 19|

Note here that we are missing some sections in the collections-rendered version of this content, vs the currently rendered version. This is because the currently rendered versions have links to documents that we are not porting over to collections, as we do not have non-English content in search.

Some examples of some RTL non-English world location news that can be used as comparisons if you spin collections up locally:

[https://www.gov.uk/world/iran/news.fa](https://www.gov.uk/world/iran/news.fa)
[https://www.gov.uk/world/israel/news.he](https://www.gov.uk/world/israel/news.he)
[https://www.gov.uk/world/pakistan/news.ur](https://www.gov.uk/world/pakistan/news.ur)
[https://www.gov.uk/world/algeria/news.ar](https://www.gov.uk/world/algeria/news.ar)
[https://www.gov.uk/world/egypt/news.ar](https://www.gov.uk/world/egypt/news.ar) (Prior to the second commit in this PR, this page was erroring - see commit for details)

**Request for frontend help**
I've chucked the lang attributes into all existing divs where we need it. However, should I create one wrapper div and only specify the RTL code once?

**Fix in this PR for some non-English news pages**

I have implemented a fix in the second commit which was causing the following page not to render: http://collections.dev.gov.uk/world/egypt/news.ar (and also any page where a translated document type contains no latin-script characters, so e.g. http://collections.dev.gov.uk/world/china/news.zg)

The alternative would be to do something on the lines of `&.downcase.gsub(/\s+/, '_')}", default: document['document_type']),` but really the only reason why this would work for translated document types was that it would be falling back to the default every time since the content item contains the translated document type (which is not a key in the locales file) so it is simpler to just use the document type from the content item. 

I note we have the same code for organisations, which will also cause issues when trying to render document types with entirely non [a-zA-Z] and [0-9] characters, so I think the same fix should go in there, whatever we decide to do here.

### DIFF
--- a/app/models/world_location_news.rb
+++ b/app/models/world_location_news.rb
@@ -42,7 +42,7 @@ class WorldLocationNews
         description: ActionView::Base.full_sanitizer.sanitize(document["summary"]).truncate(160, separator: " "),
         context: {
           date: document["public_updated_at"]&.to_date,
-          text: I18n.t("shared.schema_name.#{document['document_type']&.parameterize(separator: '_')}", count: 1, default: document["document_type"]),
+          text: document["document_type"],
         },
       }
     end
--- a/app/views/world_location_news/show.html.erb
+++ b/app/views/world_location_news/show.html.erb
@@ -4,7 +4,7 @@
   <%= tag("meta", name: "description", content: @world_location_news.description) if @world_location_news.description %>
 <% end %>
 
-<div class="govuk-grid-row">
+<div class="govuk-grid-row" <%= dir_attribute %> <%= lang_attribute %>>
   <div class="govuk-grid-column-two-thirds">
     <%= render "govuk_publishing_components/components/title", {
       title: @world_location_news.title,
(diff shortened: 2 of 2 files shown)


callumknights (approved):

BeckaL:
Thanks for pointing that out @chao-xian, I hadn't spotted that vim had stripped out the example I'd given because it's funny about backticks sometimes. I've beefed up the explanation a bit now, let me know if it makes sense

chao-xian (approved):

PR #2967, opened 2022-09-21
Remove gov.uk domain prefix

chao-xian:
These are hardcoded and should be relative so that we can check on each environment correctly.

[Trello](https://trello.com/c/waEkLAut/1279-generate-correct-base-urls-for-internal-links-based-on-environment-s)

### DIFF
--- a/app/presenters/cost_of_living_landing_page_presenter.rb
+++ b/app/presenters/cost_of_living_landing_page_presenter.rb
@@ -28,10 +28,10 @@ class CostOfLivingLandingPagePresenter
 private
 
   def internal_link?(link)
-    link.starts_with?("https://www.gov.uk/")
+    link.starts_with?("/")
   end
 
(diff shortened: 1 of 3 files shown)


app/presenters/cost_of_living_landing_page_presenter.rb
```diff
@@ -28,10 +28,10 @@ def link_clicked_track_data(track_action:, href:)
 private
 
   def internal_link?(link)
-    link.starts_with?("https://www.gov.uk/")
+    link.starts_with?("/")
   end
 
   def slug_for_href(href)
```
AgaDufrat, line 34: I think we can now delete this method entirely as it's not making any changes and on line 23 have simply `track_label: href,`

AgaDufrat (approved):
Looks good and turned out simpler than I expected! Tested locally and in integration.

PR #2972, opened 2022-09-23
Add announcement to Cost of living landing page and update other content

chao-xian:
Hard coded for now to get it shipped asap for Monday 26 Sept 2022.

[Trello](https://trello.com/c/YAhDFaNj/1291-iterate-col-navigation-page-content-l)

## Screenshot

localhost_3070_cost-of-living

### DIFF
--- a/app/assets/stylesheets/views/_colhub.scss
+++ b/app/assets/stylesheets/views/_colhub.scss
@@ -1,0 +1,9 @@
+.colhub__announcement {
+  background-color: govuk-colour("light-grey");
+  padding: govuk-spacing(3) 0;
+  margin-bottom: govuk-spacing(4);
+  @include govuk-media-query($from: desktop) {
+    padding: govuk-spacing(6) 0;
+    margin-bottom: govuk-spacing(7);
+  }
+}
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -43,3 +43,4 @@ $govuk-include-default-font-face: false;
 @import "views/organisations";
 @import "views/organisation";
 @import "views/covid";
+@import "views/colhub";
--- a/app/views/cost_of_living_landing_page/show.html.erb
+++ b/app/views/cost_of_living_landing_page/show.html.erb
@@ -78,7 +78,7 @@
           } %>
 
           <p class="govuk-body">There will be an automatic limit on how much you will be charged for each unit of gas or electricity from 1 October.
-            <a class="govuk-link" href="/government/publications/energy-bills-support">Find out how the Energy Price Guarantee limits your energy prices.</a>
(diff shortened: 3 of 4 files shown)


app/assets/stylesheets/views/_colhub.scss
```diff
@@ -0,0 +1,9 @@
+.colhub__announcement {
+  background-color: govuk-colour("light-grey");
+  padding: govuk-spacing(3) 0;
+  margin-bottom: govuk-spacing(4);
+  @include govuk-media-query($from: desktop) {
+    padding: govuk-spacing(6) 0;
```
hannalaakso, line 6: Doesn't have to be blocking as this needs to go out urgently but we might be able to use the Design System responsive spacing mixins here instead of setting the padding and margin separately for mobile and desktop.

app/assets/stylesheets/views/_colhub.scss
```diff
@@ -0,0 +1,9 @@
+.colhub__announcement {
+  background-color: govuk-colour("light-grey");
```
hannalaakso, line 2: Do we have any other examples of using this background colour with the standard link styles? I notice that the blue against the grey fails AAA (passes AA). It's not a WCAG fail but something to look at later, maybe we should follow the footer link styles for this announcement banner.  The footer uses this grey background colour but the links there are black.  (Just for completeness, the focus style yellow fails the colour contrast check here the against the grey background although it's also not a WCAG fail since the underline is still there to indicate focus.)

hannalaakso (approved):
I left some comments but we could circle back to those after publication.

app/assets/stylesheets/views/_colhub.scss
matthillco: Is AA compliance acceptable for our link colours or do they need to be AAA?

app/assets/stylesheets/views/_colhub.scss
matthillco: OK, this is something I wasn't aware existed. Happy to revise it later, if it's fine for this quick release.

hannalaakso: Cool. It's now captured [here](https://trello.com/c/MlXyrLwz/746-remove-fix-inline-changes-for-cost-of-living-hub-page-s) so resolving this comment.

app/assets/stylesheets/views/_colhub.scss
hannalaakso: AA is acceptable but it might be advisable to revise the link colours later to follow something similar to what the footer does. But this requirement has now been captured [here](https://trello.com/c/MlXyrLwz/746-remove-fix-inline-changes-for-cost-of-living-hub-page-s) so resolving this comment as well.

PR #2973, opened 2022-09-23
Small Dockerfile fixes for Replatforming.

sengi:
Remove an unnecessary `mkdir /app` from the Dockerfile, since this will be incompatible with an upcoming version of govuk-ruby-base.

Also remove some unneeded paths from the image.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,6 +1,10 @@
 .git
 .gitignore
 CONTRIBUTING.md
-docs
 Jenkinsfile
 README.md
+docs
+log
+spec
+test
+tmp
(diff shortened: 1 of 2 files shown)


nsabri1 (approved):
You can also move line 24 up to 19 if you like in the dockerfile

sengi:
> You can also move line 24 up to 19 if you like in the dockerfile

Oh that's a good idea. As discussed out-of-band let's do that in another pass later — that way I can get the base image change in and unblock a bunch of other stuff. Happy to do that cleanup though!

PR #2988, opened 2022-10-04
Translate to_sentence for connecting titles

1pretz1:
Previously, to_sentence was not using translations and so injecting English connecting words like 'and' into foreign translations.

## Before

Screenshot 2022-10-04 at 17 05 36

## After

Screenshot 2022-10-04 at 17 05 27


spec/models/person_spec.rb
```diff
@@ -54,6 +54,13 @@
     it "combines the titles into a sentence" do
       expect(person.current_roles_title).to eq("Prime Minister and First Lord of the Treasury")
     end
+
+    it "combines the titles into a sentence for translated content" do
```
JonathanHallam, line 58: odd question, would it be worth putting the text in english to make it clearer what's actually being tested something like `one و two`

1pretz1: I see your point for readability, however the current setup is more realistic, as we shouldn't have a translated page with English words. I have added a comment above the test which should make the test more obvious.

JonathanHallam: thanks :)

JonathanHallam (approved):
👍

gclssvglx (approved):
Nice work @1pretz1 👍

PR #2991, opened 2022-10-06
Use blue bar from gem instead of pseudoelement

patrickpatrickpatrick:
## What

Remove the blue bar pseudo element on browse pages (and Cost of Living hub). Add the blue bar element from the govuk_publishing_components gem instead.

## Why

On gov.uk, other apps/pages that have the blue bar beneath the navbar use the gem_layout template. On that template, the blue bar is part of the template partial itself and individual apps are not responsible for rendering it. However the full width header background of Browse pages means that we have to use a gem_layout_full_width template. This template does not have the blue bar and so the Browse page partials render this blue bar themselves. 

I thought that perhaps the blue bar could be added to the full width gem layout as an option but this resulted in a gap between the blue bar and where the background colour for the banner starts.

Screenshot 2022-10-06 at 12 58 55

So the best solution I could see was to use the styling of the blue bar from the gem itself. This at least will ensure consistency between apps going forward and fixes the issue that was occurring on tablet screen sizes (see below).

### Before

Screenshot 2022-10-06 at 15 16 11

### After

Screenshot 2022-10-06 at 15 16 50

### DIFF
--- a/app/assets/stylesheets/views/_browse.scss
+++ b/app/assets/stylesheets/views/_browse.scss
@@ -22,18 +22,6 @@ $browse-header-background-colour: #263135;
   // Setting the overflow to auto is required to prevent the breadcrumb
   // component's margin from collapsing.
   overflow: auto;
-
-  // We want to show a constrained width blue bar _within_ the grey header.
-  // But the full width layout template doesn't include the blue bar - so we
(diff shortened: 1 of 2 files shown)


MartinJJones (approved):
Changes look good to me, nice work.

PR #3053, opened 2022-11-25
Revert "Upgrade Ruby to 3.1.2"

chao-xian:
Reverts alphagov/collections#2949

As it rolled to staging a ton of errors started


ollietreend:
We no longer need to revert this change. Errors started appearing due to us needing to perform a 'hard restart' of the app on deployment.

PR #3055, opened 2022-11-28
Add GitHub Action workflow for verifying pact tests

kevindew:
Trello: https://trello.com/c/QrxjshEm/270-move-ruby-gems-from-jenkins-to-github-actions

This moves the pact verification to a GitHub Action and will no longer be performed by a Jenkins build. This workflow will also be called by GDS API Adapters as part of it's CI build to assert that newly generated pacts are valid against this app. To avoid running this same task twice `pact:verify` is no longer part of the Jenkins test actions and instead the other rake default steps are included.

For more details see https://github.com/alphagov/gds-api-adapters/pull/1175 where the wider change is being co-ordinated.

### DIFF
--- a/.github/workflows/pact-verify.yml
+++ b/.github/workflows/pact-verify.yml
@@ -1,0 +1,36 @@
+# Pact verify workflow
+#
+# This workflow asserts that Pact contract tests are valid against this
+# codebase. It is trigged when changes are made to this project and it
+# is explicitly called by GDS API Adapters when changes are made there.
+on:
+  pull_request:
+  push:
+  workflow_call:
+    inputs:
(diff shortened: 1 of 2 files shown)


hannako (approved):
Just for my understanding:

I see we still have `pact:verify` as part of the [default rake test](https://github.com/alphagov/collections/blob/5dff34e533cc1090ff0aea353314eaa72bf69a2b/Rakefile#L14) and then we are over-riding that in the jenkinsfile, so jenkins doesn't repeat what github actions has already done. Is the reason we've chosen to do it like that to make it easier for local development? ie so that we don't have to remember to specifically run `pact:verify` we can just run `bundle exec rake`? Or is there something else I'm missing?

PR #3056, opened 2022-11-28
Use content_ids for curated lists

AgaDufrat:
## Context

When a content item is tagged to a curated Mainstream browse or Specialist topic and this content item’s slug changes, the link disappears from the topic page. This is because the curated lists/groups are based on base paths. The “contents” slugs/base paths don’t get updated as a part of dependency resolution in Publishing API.

## What
Use content_ids to present curated lists instead of base_paths, as they are less volatile.

### Example steps to replicate the issue

In Publisher:
1. Create a new edition of a document tagged to [a curated Mainstream browse page](https://www.integration.publishing.service.gov.uk/browse/childcare-parenting/childcare) (you can only edit slug of a draft edition)
2. Metatags -> change slug -> Save
3. Submit for 2i -> Skip review -> Publish 

In the Frontend: 
- the tag is removed from the Mainstream browse curated list
- content Item: still has the old slug in [“details”][“groups”]

In Collections Publisher:
- Content appears under Unused links
- The content (list item) appears in the List but link goes to https://content-tagger.integration.publishing.service.gov.uk/taggings/lookup

### What should happen
There should be no change in the rendering app and Collections Publisher i.e. the item should remain in the list.

## Depends on
- https://github.com/alphagov/publishing-api/pull/2201
- https://github.com/alphagov/collections-publisher/pull/1749
- https://github.com/alphagov/collections-publisher/pull/1772
- [x] All Mainstream browse pages and Specialist topics have been republished

### DIFF
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -14,5 +14,6 @@ class Document
     :content_store_document_type,
     :organisations,
     :image_url,
+    :content_id,
   )
 end
--- a/app/models/list_set.rb
+++ b/app/models/list_set.rb
@@ -48,8 +48,8 @@ private
 
   def curated_list
     curated_data = @group_data.map do |group|
-      contents = group["contents"].map do |base_path|
-        content_tagged_to_tag.find { |content| content.base_path == base_path }
+      contents = group["content_ids"].map do |content_id|
+        content_tagged_to_tag.find { |content| content.content_id == content_id }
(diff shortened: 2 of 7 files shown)


hannako (approved):
This looks good Aga. Wondering if you should set it to draft until all of the republishing has completed though!

PR #3087, opened 2022-12-15
Move 'Follow Us' out of 'What We Do' Row for No. 10 Org Page

patrickpatrickpatrick:
## What

Move 'Follow Us' column out of the same row as 'What We Do' for No. 10 Organisation page. Move 'Follow Us' to same row as 'Contact No. 10' in place of 'About Us'.

## Why

As part of the new design for No. 10 pages, there are some changes in layout. This is the most trivial to do with the least changes needing to be made to content of the page, so therefore it has been done first.

### Before

Screenshot 2022-12-16 at 11 06 34

Screenshot 2022-12-16 at 11 06 39

### After

Screenshot 2022-12-16 at 14 14 02

Screenshot 2022-12-16 at 14 14 06

### DIFF
--- a/app/views/organisations/_corporate_information.html.erb
+++ b/app/views/organisations/_corporate_information.html.erb
@@ -8,7 +8,7 @@
         lang: t_fallback('organisations.corporate_information')
       } %>
     <% end %>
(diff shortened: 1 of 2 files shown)


matthillco (approved):
Looks good to me, no changes required.

PR #3091, opened 2022-12-16
Changes to promotional items on No. 10 page

patrickpatrickpatrick:
## What 

New partial for promotional features for No. 10 pages that only uses the first item of a promotional feature and renders it using the large variant of the image card. Update the `documents_presenter `to set `large` to true for an item if it is part of a promotional feature on the No. 10 page. Update the `documents_presenter `to set `extra_details_no_indent` to true for an item if it is part of a promotional feature on the No. 10 page.  This item gets passed to the image card component.

## Why

In the new redesign for the No. 10 organisation page, the promotional features are no longer in columns but take up entire rows. Each promotional item also no longer will show any other items than the first item. The extra details of each promotional item is no longer indented in the design, so this has been included in this PR.

### Before

Screenshot 2022-12-16 at 12 14 22

### After

Screenshot 2022-12-16 at 15 05 18

### DIFF
--- a/app/presenters/organisations/documents_presenter.rb
+++ b/app/presenters/organisations/documents_presenter.rb
@@ -77,6 +77,7 @@ module Organisations
           brand: org.brand,
           heading_level: 3,
           large: @org.is_no_10?,
+          extra_details_no_indent: @org.is_no_10?,
         }
 
         if item["title"].present?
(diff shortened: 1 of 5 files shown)


hannako:
It was decided that we'd make this change for all org pages. So closing this PR as it's been superseded by https://github.com/alphagov/collections/pull/3096

PR #3096, opened 2022-12-19
Promotional features on organisation pages

hannako:
When a promotional feature contains only one item, expand the column to fill the whole row.

When it contains many items, no change in current behaviour. ie split into columns.

This change was requested as part of the number 10 page redesign, but is being rolled out to all organisation pages.

Trello: https://trello.com/c/nLPbdLc9/36-use-featured-section-for-meet-the-prime-minister-speech-history-m-l

### Review app
- https://collections-pr-3096.herokuapp.com/government/organisations/prime-ministers-office-10-downing-street
- https://collections-pr-3096.herokuapp.com/government/organisations/civil-service

### Before
Screenshot 2022-12-19 at 18 14 14

---
### After
Screenshot 2022-12-20 at 14 57 48

### DIFF
--- a/app/views/organisations/_promotional_features.html.erb
+++ b/app/views/organisations/_promotional_features.html.erb
@@ -1,7 +1,7 @@
 <% if @documents.has_promotional_features? %>
 <section class="organisation__margin-bottom organisation__section-wrap">
   <% @documents.promotional_features.each do |feature| %>
-    <div>
+    <div class="organisation__promotion">
       <%= render "govuk_publishing_components/components/heading", {
         text: feature[:title],
         heading_level: 2,
--- a/app/views/organisations/_promotional_features.html.erb
+++ b/app/views/organisations/_promotional_features.html.erb
@@ -1,7 +1,7 @@
 <% if @documents.has_promotional_features? %>
-<section class="organisation__margin-bottom organisation__promotions">
+<section class="organisation__margin-bottom organisation__section-wrap">
   <% @documents.promotional_features.each do |feature| %>
-    <div class="<%= feature[:parent_column_class] %>">
+    <div>
       <%= render "govuk_publishing_components/components/heading", {
         text: feature[:title],
         heading_level: 2,
--- a/app/views/organisations/_promotional_features.html.erb
+++ b/app/views/organisations/_promotional_features.html.erb
@@ -9,14 +9,19 @@
         border_top: 5,
         brand: @organisation.brand
       } %>
-
-      <div class="<%= "govuk-grid-row" if feature[:number_of_items] > 1 %>">
+      <% if feature[:number_of_items] > 1 %>
+        <div class="govuk-grid-row">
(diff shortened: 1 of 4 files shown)


app/views/organisations/_promotional_features.html.erb
```diff
@@ -1,7 +1,7 @@
 <% if @documents.has_promotional_features? %>
-<section class="organisation__margin-bottom organisation__promotions">
+<section class="organisation__margin-bottom organisation__section-wrap">
   <% @documents.promotional_features.each do |feature| %>
-    <div class="<%= feature[:parent_column_class] %>">
+    <div class="<%= feature[:parent_column_class]%>">
```
matthillco, line 4: Is it possible to remove this wrapping div for single item promos? It's now redundant in that context. 

There are also similar redundant <divs> at lines 13(19) and 15(17) which aren't part of this PR but could also be removed for clarity of output.

However, I appreciate that this would incur extra logic for the sake of slightly cleaner HTML and may not be worthwhile. They're not doing any harm by being left in the output HTML.

matthillco (commented):
See my comment on line 4 of `app/views/organisations/_promotional_features.html.erb`

app/views/organisations/_promotional_features.html.erb
hannako: Thanks Matt, I'll make that change

hannako: Removing the outer div at line 4 breaks some styling - the black dividing line is truncated. So I've left that as is. But have removed the inner divs when only one item is present see https://github.com/alphagov/collections/pull/3096/commits/4818e9fa70b36b58f1121fc0470ac6a6683fd84d

matthillco (approved):
Looks good to me, thank you.

app/views/organisations/_promotional_features.html.erb
```diff
@@ -1,22 +1,27 @@
 <% if @documents.has_promotional_features? %>
-<section class="organisation__margin-bottom organisation__promotions">
```
AgaDufrat, line 2: It would be good if the commit message included a bit more info as to why we are changing the class and want the `word-wrap: break-word` here and we are moving away from the previous styles.

Also since `organisation__promotions` class this is no longer used anywhere could we remove the definition? https://github.com/alphagov/collections/blob/0473b65512dc0e3f57189cce87e32ce74b90fc91/app/assets/stylesheets/views/_organisation.scss#L34-L59

AgaDufrat (approved):
Backend looks good to me. I'm not entirely clear about some of the css changes and design decisions behind them (previews look fine). Maybe a frontend dev could have a look as well.

app/views/organisations/_promotional_features.html.erb
hannako: Thanks Aga, I've removed the legacy code and improved the commit message.

PR #3103, opened 2022-12-29
Update GitHub Action workflow for CI

theseanything:
This updates the initial workflow to enable CI using GitHub Actions. It makes use of
reusable workflows and composite actions defined in govuk-infrastructure. The
CI workflow is initially only used to support deployments to the new EKS
infrastructure and not to immediately replace the CI workflow in Jenkins.

The CI workflow makes use of individual jobs to run each testing and linting
tool. This allows the processes to run concurrently and improves the visibility
of status in the GitHub UI. The workflow runs a precursor job to cache
dependencies in order to speed up other jobs.

Most jobs make use of reusable workflows as they are consistent across the
majority of our repositories - this allows us to easily maintain them in a
single place. However, the test-ruby job (and/or integration-tests) is defined
per repo as backing services and other dependencies vary greatly between repos.
Backing services (such as databases) are created using a composite action
instead of a job service. This is because they contain a lot of shared config
and makes it easier for us to maintain.


.github/workflows/ci.yaml
```diff
@@ -11,8 +11,99 @@ on:
...
+        uses: alphagov/govuk-infrastructure/.github/actions/setup-node@main
+
+      - name: Precompile assets
+        uses: alphagov/govuk-infrastructure/.github/actions/precompile-rails-assets@main
+
+      - name: Run Rspec
+        env:
+          RAILS_ENV: test
+          GOVUK_CONTENT_SCHEMAS_PATH: vendor/publishing-api/content_schemas
+        run: bundle exec rake spec
```
kevindew, line 84: This looks great 👍

.github/workflows/ci.yaml
```diff
@@ -11,8 +11,99 @@ on:
...
-    name: Test Rails
-    uses: alphagov/govuk-infrastructure/.github/workflows/test-rails.yaml@main
+  cache-dependencies:
+    name: Cache dependencies
+    uses: alphagov/govuk-infrastructure/.github/workflows/cache-dependencies.yaml@main
+
+  precompile-assets:
+    name: Precompile Rails assets
+    needs: cache-dependencies
+    uses: alphagov/govuk-infrastructure/.github/workflows/precompile-rails-assets.yaml@main
```
kevindew, line 21: If we're testing the ability to pre-compile assets, we'll want to run it against the production environment so that it tests compression - which is where people get caught out. 

May be good to clarify if this is an optimisation step or a test. For what it's worth I'd imagine it might not add a huge amount of value as an optimisation - maybe only needed for an app with exceptionally slow asset compilation (Whitehall was once in that way but I think it's ok now)

.github/workflows/ci.yaml
```diff
@@ -11,8 +11,99 @@ on:
...
+    with:
+      files: "app/assets/stylesheets/"
+
+  lint-javascript:
+    name: Lint JavaScript
+    needs: cache-dependencies
+    uses: alphagov/govuk-infrastructure/.github/workflows/standardx.yaml@main
     with:
-      requiresJavaScript: true
+      files: "'app/assets/javascripts/**/*.js' 'spec/javascripts/**/*.js'"
```
kevindew, line 40: I'd wonder if you could avoid repetition by having these call the yarn task directly. e.g. `yarn lint:js`. I imagine doing that could allow you to have just one set-up yarn action and would mean you don't need to change both package.json and this if you change the JS files linted.

These shared actions have the issue that they're coupled to yarn which we don't tend to use on non-Rails projects, so might be clearer if we used a set-up yarn action (as per: https://github.com/alphagov/govuk-infrastructure/blob/main/.github/actions/setup-node/action.yaml)

I also like to dream that at some point we'll port tools to ES6 and then be able to use standard again for linting (which we do in node projects), so it wouldn't surprise me if standardx wasn't the nor

kevindew (commented):
Thanks for your work on this, I know this is still draft so just commented a few things that crossed my mind as I saw it.

marcpomfret (approved):

.github/workflows/ci.yaml
theseanything: Yeah, this job was meant to serve as a test for precompiling, but mainly was to cache the assets if they're re-used in multiple subsequent steps. (However, turns out that doesn't save any time.) I've removed this as an individual job and kept it as a step in the test run jobs (if the app has assets). At the moment hat will be compiling them with `RAILS_ENV=test`, but that could be changed [in the composite action.](https://github.com/alphagov/govuk-infrastructure/blob/add91b7629162fb3ff27b6aaebf558260b0d8fe0/.github/actions/precompile-rails-assets/action.yml#L18). 

We might want to move this to separate job, as it'll be easier to identify which bit of CI is failing and to not block the tests running if they take a long time to compile. Will look at that outside the initial PRs.

.github/workflows/ci.yaml
theseanything: Yah - I was trying minimise coupling this reusable workflows to dependencies within the individual repo itself. I.e. we'd need to assume that the calling repo has the "lint:js" yarn script defined. I do realise it's still coupled and assumes yarn is being used, but the reusable workflow could be iterated to support npm as well.

If we wrote this job to use the setup-node composite action as a precursor step, then we couldn't call a re-usable workflow and end up having to define the full job in every repo.

Yup - would be great to go back to standard!

kevindew: Sure, I'd have thought we could have a workflow that just allowed the task to be specified such as:

```
  lint-javascript:
    needs: cache-dependencies
    uses: alphagov/govuk-infrastructure/.github/workflows/node-task.yaml@main
    with:
      task: yarn lint:js
```

PR #3117, opened 2023-01-06
Simplify organisation view template

hannalaakso:
## What
Move the non-organisational content into its own template files. Remove a few now redundant conditional checks for whether an organisation is promotional or not. 

## Why
The `show` template was becoming hard to follow partly because of the length of the file and also because of a large number of conditional blocks of logic depending on whether the organisation is promotional not. 

We're currently working on adding a new organisation header to collections so this work is to prep adding some more logic to `organisations/show`. 

### DIFF
--- a/app/views/organisations/_standard_org_docs_people.html.erb
+++ b/app/views/organisations/_standard_org_docs_people.html.erb
@@ -1,0 +1,15 @@
+<% if @supergroups.has_groups? %>
+    <%= render partial: 'latest_documents_by_supergroup' %>
+<% end %>
+<% if @people.has_people? %>
+    <div id="people">
+        <% @people.all_people.each do |people_data| %>
+        <%= render partial: 'related_people', locals: {
+            people: people_data[:people],
+            title: people_data[:title],
+            lang: people_data[:lang],
+            brand: @organisation.brand
+        } %>
+        <% end %>
+    </div>
+<% end %>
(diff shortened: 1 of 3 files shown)


chao-xian (changes requested):
Nice! Great to see us tidying up the place - thank you @hannalaakso!

Overall this is fine but the naming of files and partials are confusing, long and non-descriptive of what they contain.

I'd suggest naming the partials as:

* `_standard_org_docs_people.html.erb` instead of `_not_promotional_org.html.erb`
* `_standard_org_foi_profiles.html.erb` instead of `_not_promotional_org_2.html.erb`

I also think the `app/views/organisations/_show_content.html.erb` partial is unnecessary and one abstraction level too much, though I can see that it was a good path to follow as part of the process of working out what to split. The name of the partial itself being so similar to `_show.html.erb` suggests it's not necessary. However it would benefit from a rename as well if we kept the partial.

Hope that's ok. Great work so far!

hannalaakso:
Thanks for the review @chao-xian, have a look now. I've renamed the files as suggested. I've now also removed the extra `show_content` partial - I initially added it keeping in mind that we'll be adding more logic to the main page content at a later point but we can always split the main page content out at that point if it feels necessary.

chao-xian (approved):
Thank you for making those suggested changes! Looks good.

PR #3118, opened 2023-01-06
Add 'Past Chancellors of the Exchequer' page

brucebolt:
This code has mostly been ported over from Whitehall with few changes.

It depends on a content item being published for this route, which is covered by https://github.com/alphagov/special-route-publisher/pull/240.  Once this code is merged, we'll then need to change the rendering app to Collections, which is included in https://github.com/alphagov/special-route-publisher/pull/241.

I've split the `past_chancellors_person` partial into two parts to make it more intuitive, as this was rendering a table rather than an individual person as the name would suggest.

Screenshots showing rendering:
| Whitehall | Collections |
| --- | --- |
| Screenshot showing the past chancellors page rendered by Whitehall. | Screenshot showing the past chancellors page rendered by Collections.  It is visually identical to Whitehall.  |

[Trello card](https://trello.com/c/PgJxJILW)

### DIFF
--- a/app/views/past_chancellors/_person.html.erb
+++ b/app/views/past_chancellors/_person.html.erb
@@ -1,0 +1,10 @@
+<div class="govuk-grid-column-one-quarter" role="listitem">
+  <%= render "govuk_publishing_components/components/heading", {
+    text: name,
+    heading_level: 3,
+    font_size: "s"
+  } %>
+  <% info[:service].each do | service | %>
+    <p class="govuk-body"><%= service %></p>
+  <% end %>
+</div>
--- a/app/views/past_chancellors/index.html.erb
+++ b/app/views/past_chancellors/index.html.erb
@@ -1,0 +1,21 @@
+<% content_for :title, @content_item["title"] %>
+
+<% page_class "historic-appointments-index historical-past-chancellors govuk-width-container" %>
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-three-quarters">
+    <%= render "govuk_publishing_components/components/title", {
+      context: "History",
+      title: @content_item["title"],
+    } %>
+  </div>
+</div>
+
+<%= render partial: "century", locals: { heading: "21st century", people: @twentyfirst_century_chancellors } %>
+
+<%= render partial: "century", locals: { heading: "20th century", people: @twentieth_century_chancellors } %>
+
+<%= render partial: "century", locals: { heading: "19th century", people: @nineteenth_century_chancellors } %>
+
+<%= render partial: "century", locals: { heading: "18th century", people: @eighteenth_century_chancellors } %>
+
+<%= render partial: "century", locals: { heading: "16th & 17th centuries", people: @sixteenth_and_seventeenth_century_chancellors } %>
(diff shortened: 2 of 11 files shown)


app/views/past_chancellors/_person.html.erb
```diff
@@ -0,0 +1,14 @@
+<div class="govuk-grid-column-one-quarter" role="listitem">
+  <%= render "govuk_publishing_components/components/heading", {
+    text: name,
+    heading_level: 3,
+    font_size: "s"
+  } %>
+  <% if info[:service].kind_of?(Array) %>
```
BeckaL, line 7: In my past foreign secretaries page, I've removed this logic by just making them all arrays. I'd imagine we'd make it an array in the content item too, plus I'm not a fan of this kind of mixing types. What do you reckon about changing them all to be arrays?

brucebolt: That sounds like a good idea.  I've made the same change here too.

BeckaL (commented):
Looks great! Just a suggestion about cleaning up the index file, otherwise happy to approve.

app/views/past_chancellors/index.html.erb
```diff
@@ -0,0 +1,101 @@
...
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-three-quarters">
+    <%= render "govuk_publishing_components/components/title", {
+      context: "History",
+      title: @content_item["title"],
+    } %>
+  </div>
+</div>
+
+<div class="govuk-grid-row historic-people-index__section-row-header">
```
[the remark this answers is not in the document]
BeckaL: Actually presumably we can just use the same partial here too (the `past_foreign_secretaries_by_century` one renamed to be more general)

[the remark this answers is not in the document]
BeckaL: Cool, I'm expecting to merge all my stuff after you so I can do that tidy up.

spec/features/past_chancellors_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+    }
+  end
+
+  before do
+    stub_content_store_has_item(base_path, content_item)
+  end
+
+  it "sets the page title" do
+    visit base_path
+    expect(page).to have_title("#{content_item['title']} - GOV.UK")
```
BeckaL, line 18: Just noticed this as I was copying these tests for mine - title here (and a few lines below) should be a symbol

brucebolt: Thanks.  Test passed anyway, so the hash must be using indifferent access.  I'll update to a symbol for consistency.

BeckaL: In case of interest, that's what I had thought but actually I noticed this I copied it but the test didn't fail when I expected it to. Turns out it's because of the string interpolation - interpolating a nil gets you an empty string, so: 
`puts "#{content_item['title']} - GOV.UK"`
prints 
` - GOV.UK` 
so I had a passing test because e.g. the title always has `- GOV.UK` at the end

brucebolt: Oh, that makes sense.  I hadn't realised that was happening!

chao-xian (approved):
LGTM - thanks for refactoring it as well

PR #3131, opened 2023-01-11
🚧  Add new header for No 10 organisation page

hannalaakso:
This PR:

- makes No 10 organisation page full width
- adds a No 10 organisation header with an image with gradient overlay to improve colour contrast with heading text, and a three card image card grid without images
- fixes the image card margins and width on mobile - we should investigate the use cases without image for the component and potentially move this fix to the gem
- adds and amends tests to make sure the header is rendering correctly

## Still to do
- Add tests for `app/presenters/organisations_no_10_page_presenter.rb`
- Add confirmed assets

Fixes https://trello.com/c/pdyIMeno/38-implement-the-redesigned-no10-page-header-l


app/views/organisations/_header.html.erb
```diff
@@ -1,5 +1,62 @@
...
+
+        <div class="organisation__no10-header__banner__content">
+          <h1 class="organisation__no10-header__banner__content__heading govuk-heading-l govuk-!-margin-0">
+            The Prime Minister's Office,<br>
+            10 Downing Street
+          </h1>
+        </div>
+      </div>
+      <div class="organisation__no10-header__cards">
+        <h2 class="gem-c-heading gem-c-heading--font-size-27 govuk-!-margin-bottom-5">
```
patrickpatrickpatrick, line 25: I think whenever possible, it's best to use the heading component from govuk_publishing_components. This was the headings across templates will be consistent, especially if there are changes to classes made in govuk-frontend.

app/views/organisations/_header.html.erb
```diff
@@ -1,5 +1,62 @@
...
+    <div class="organisation__no10-header organisation__margin-bottom">
+      <%# To do: add asset here and style and name this container as needed %>
+      <span class="govuk-heading-xl govuk-!-text-align-centre">
+      10
+      </span>
+      
+      <div class="organisation__no10-header__banner govuk-!-margin-bottom-6">
+
+        <%= content_tag(:picture) do %>
+          <%= content_tag(:source, nil, { media: "(min-width: 641px)", srcset: "#{image_path('no10header-desktop.png')}" }) %>
```
patrickpatrickpatrick, line 13: I don't think `nil` is needed here in theory. I would try removing `nil` but if it doesn't work then we can keep it how it is now.

app/views/organisations/_header.html.erb
```diff
@@ -1,5 +1,62 @@
...
+          <%= content_tag(:img, nil, { srcset: "#{image_path('no10header.png')}"}) %>
+        <% end %>
+
+        <div class="organisation__no10-header__banner__content">
+          <h1 class="organisation__no10-header__banner__content__heading govuk-heading-l govuk-!-margin-0">
```
patrickpatrickpatrick, line 18: You can also use the heading component here. The heading component has an 'inverse' option so you can make the text white.

patrickpatrickpatrick:
Looks good, just a couple of places where the [heading component from govuk_publishing_components](https://components.publishing.service.gov.uk/component-guide/heading) could be used.

app/views/organisations/_header.html.erb
hannalaakso: Thank you, this is now done.

app/views/organisations/_header.html.erb
hannalaakso: Done, using the tag element instead of content_tag.

app/views/organisations/_header.html.erb
```diff
@@ -1,5 +1,68 @@
...
+          margin_bottom: 5
+        } %>
+        <div class="govuk-grid-row">
+          <div class="govuk-grid-column-one-third">
+            <%= render "govuk_publishing_components/components/image_card", {
+              brand: "prime-ministers-office-10-downing-street",
+              href: "/not-a-page",
+              heading_text: "Lorem ipsum",
+              heading_level: 3,
+              description: "Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua."
```
patrickpatrickpatrick, line 41: Should this text be in a .yml file? That's probably better than having the content in the view itself.

hannalaakso: @patrickpatrickpatrick We didn't discuss doing it as part of this piece of work but I've added it in anyway now. I also added tests for the yml for the controller but didn't have the time to add tests for the presenter, I've made a note of that in the PR description.

PR #3180, opened 2023-02-15
Make collections show default main language in analytics

Tetrino:
Previously we would filter out the language from analytics if it was our default `en` entry. While clean, it's inconsistent with other areas of GovUK, so this has been changed to return the current language regardless of whether it's the default. Nothing relied on the value being returned being `nil`, so this change should be enough.

The test was updated to ensure that `en` is returned as the default locale, rather than looking for a nil entry.

https://trello.com/c/5D0FH2V5/1644-make-collections-show-default-main-language-in-analytics-m

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -38,7 +38,7 @@ module ApplicationHelper
   end
 
   def lang_attribute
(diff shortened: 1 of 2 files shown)


Tetrino:
Closing for now, needs more investigation.

PR #3186, opened 2023-02-21
Render Past Prime Ministers

callumknights:
This PR adds the Past Prime Ministers index page into collections. Not to be confused with the individual past PM pages, already added in https://github.com/alphagov/collections/pull/3191

Once the PR is merged, the rendering app in the content item will need to be updated PR - [here](https://github.com/alphagov/whitehall/pull/7400)

Trello: https://trello.com/c/LSd8y4zg/421-render-past-prime-ministers-index-page-in-collections

⚠️ This repo is Continuously Deployed: make sure you [follow the guidance](https://docs.publishing.service.gov.uk/manual/development-pipeline.html#merge-your-own-pull-request) ⚠️

Follow [these steps](https://guides.rubyonrails.org/upgrading_ruby_on_rails.html) if you are doing a Rails upgrade.

PR #3191, opened 2023-02-27
Render individual past pm

BeckaL:
This adds the code to render the individual past prime ministers pages in collections (such as [Teresa May](https://www.gov.uk/government/history/past-prime-ministers/theresa-may) and [William Pitt the Younger](https://www.gov.uk/government/history/past-prime-ministers/william-pitt)). This forms part of the overall work to move rendering out of whitehall.

Screenshots:

| Whitehall | Collections |
| ---- | --- |
|Screenshot 2023-02-27 at 16 20 55|Screenshot 2023-02-27 at 16 20 21|
|Screenshot 2023-02-27 at 16 29 58|Screenshot 2023-02-27 at 16 30 54|

Note that there is a slight difference in that we no longer render links to all past prime minister pages on the individual pages - just to the 5 closest PMs who have historical accounts. This decision was made on the Whitehall side, to simplify the content item and presenting code, and also simplify what appears on the page. We've also fixed an issue with older PMs not displaying the prime ministers adjacent to them in term (see William Pitt the Younger above).

[Trello](https://trello.com/c/YJ4RzyVl/424-render-individual-past-prime-ministers-pages-in-collections)

### DIFF
--- a/app/views/past_prime_ministers/show.html.erb
+++ b/app/views/past_prime_ministers/show.html.erb
@@ -4,7 +4,7 @@
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
     <%= render "govuk_publishing_components/components/title", {
-      context: "Past Prime Ministers",
+      context: I18n.t("past_prime_ministers.context"),
       title: @past_prime_minister.title,
     } %>
 
@@ -43,7 +43,7 @@
     <% end %>
 
     <%=  render "govuk_publishing_components/components/heading", {
-      text: "Biography",
+      text: I18n.t("past_prime_ministers.headings.biography"),
       font_size: "m",
       margin_bottom: 1,
     } %>
@@ -61,7 +61,7 @@
       ga4_tracking: true
     } %>
     <p class="govuk-body">
-      <a class="govuk-link" href="/government/history/past-prime-ministers">View all past prime ministers</a>
+      <a class="govuk-link" href="/government/history/past-prime-ministers"><%= I18n.t("past_prime_ministers.view_all")  %></a>
     </p>
   </div>
 </div>
--- a/app/views/past_prime_ministers/show.html.erb
+++ b/app/views/past_prime_ministers/show.html.erb
@@ -1,0 +1,67 @@
+<% content_for :title, @past_prime_minister.page_title %>
+<% page_class "govuk-width-container" %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= render "govuk_publishing_components/components/title", {
+      context: "Past Prime Ministers",
+      title: @past_prime_minister.title,
+    } %>
+
+    <%= render "govuk_publishing_components/components/lead_paragraph", {
+      text: "#{@past_prime_minister.political_party} #{@past_prime_minister.dates_in_office}",
+    } %>
+    <%= render "govuk_publishing_components/components/lead_paragraph", {
+      text: @past_prime_minister.description,
+    } %>
+  </div>
+</div>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <hr class="govuk-section-break govuk-section-break--s govuk-section-break--visible govuk-!-margin-bottom-4">
+
+    <div class="govuk-grid-row">
+      <div class="govuk-grid-column-one-third">
+        <%= image_tag(
+              @past_prime_minister.image_data["url"],
+              alt: @past_prime_minister.image_data["alt_text"],
+              class: "govuk-!-margin-bottom-4 govuk-!-width-full"
+        ) %>
+      </div>
+    </div>
+
+    <% @past_prime_minister.appointment_info_array.each do |info| %>
+      <% if info[:text].present? %>
+        <%= render "govuk_publishing_components/components/heading", {
+          text: info[:title],
+          font_size: "m",
+          margin_bottom: 1,
+        } %>
+        <p class="govuk-body"><%= info[:text] %></p>
+      <% end %>
+    <% end %>
+
+    <%=  render "govuk_publishing_components/components/heading", {
+      text: "Biography",
+      font_size: "m",
+      margin_bottom: 1,
+    } %>
+
+    <%=  render "govuk_publishing_components/components/govspeak" do %>
+      <%= @past_prime_minister.biography %>
+    <%  end %>
+
+  </div>
+
+  <div class="govuk-grid-column-one-third">
+    <p class="govuk-body">
+      <%=  render "govuk_publishing_components/components/related_navigation", {
+      content_item: @past_prime_minister.related_prime_ministers_nav,
+      ga4_tracking: true
+    } %>
+    <p class="govuk-body">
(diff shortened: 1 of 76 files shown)


app/models/historic_appointment.rb
```diff
@@ -0,0 +1,59 @@
...
+  end
+
+  def image_data
+    @content_item.content_item_data.dig("links", "person", 0, "details", "image")
+  end
+
+  def appointment_info_array
+    %w[born died dates_in_office political_party major_acts interesting_facts].map do |field|
+      text = field == "dates_in_office" ? dates_in_office : @content_item.content_item_data.dig("details", field)
+      { title: field.gsub("_", " ").capitalize, text: }
```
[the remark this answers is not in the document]
BeckaL: Good shout, this is a lot nicer

app/views/historic_appointments/show.html.erb
```diff
@@ -0,0 +1,67 @@
...
+        <%= render "govuk_publishing_components/components/image_card", {
+          href: "#",
+          image_src: @historic_appointment.image_data["url"],
+          image_alt: @historic_appointment.image_data["alt_text"]
+        } %>
+      </div>
+    </div>
+
+    <% @historic_appointment.appointment_info_array.each do |info| %>
+      <% unless info[:text].blank? %>
```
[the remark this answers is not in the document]
JonathanHallam: Have fixed this :)

app/models/historic_appointment.rb
```diff
@@ -0,0 +1,59 @@
...
+  end
+
+  def political_party
+    @content_item.content_item_data.dig("details", "political_party")
+  end
+
+  def dates_in_office
+    @content_item.content_item_data.dig("details", "dates_in_office")
+                 .map { |d| "#{d['start_year']} to #{d['end_year']}" }
+                 .join(", ")
```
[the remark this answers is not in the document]
JonathanHallam: We've looked through a few of the PM's with multiple appointments and it's been this way round every time. Whitehall is [actively descending ](https://github.com/alphagov/whitehall/blob/main/app/controllers/historic_appointments_controller.rb#L40) their order on starting date to get the previous behaviour.

Our current approach just takes them in the order they come out of the DB which seems to be ascending on start date but we haven't been able to find any guarantees that this is the case.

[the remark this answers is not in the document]
brucebolt: Great, thanks for looking into this 👍.

app/views/historic_appointments/show.html.erb
```diff
@@ -0,0 +1,67 @@
...
+  </div>
+
+  <div class="govuk-grid-column-one-third">
+    <p class="govuk-body">
+      <%=  render "govuk_publishing_components/components/related_navigation", {
+      content_item: @historic_appointment.related_historic_appointments_nav,
+      ga4_tracking: true
+    } %>
+    <p class="govuk-body">
+      <a class="govuk-link" href="/government/history/past-prime-ministers">View all past prime ministers</a>
```
brucebolt, line 64: Looks like there's another bit of hardcoded text in this link and also in the title on line 1.

brucebolt (approved):
Thanks for making the changes.  There's a couple more bits of hardcoded text in the view, but otherwise all looks goof to me.

app/views/historic_appointments/show.html.erb
BeckaL: Should all be gone now, did this and also moved the page title into a method on the model.

app/controllers/historic_appointments_controller.rb
```diff
@@ -0,0 +1,6 @@
+class HistoricAppointmentsController < ApplicationController
```
hannako, line 1: Following on the pattern from the previous two Whitehall imports ([Past Foreign Secretaries](https://github.com/alphagov/collections/pull/3113) and [Past Chancellors of the Exchequer](https://github.com/alphagov/collections/pull/3118)), could we rename controller/models/views to PastPrimeMinisters?

Or, I'm wondering if this generic name symbolises the first step in a plan to do some refactoring? ie there's currently a lot of duplication, [PastChancellors](https://github.com/alphagov/collections/blob/main/app/models/past_chancellors.rb) and [PastForeignSecretaries](https://github.com/alphagov/collections/blob/main/app/models/past_foreign_secretaries.rb) are identical. And they contain code duplicated again in this new [HistoricAppointment](https://github.com/alphagov/collections/blob/e238bf13e3bffb8554cf0a763c0a9a7c122f9d58/app/models/historic_appointment.rb) model. 

I'm wondering why we don't have a generic historic appointment (as seems to be the case in Whitehall where this code is coming from), with PastForeignSecretaries, PastChancellors and PastPrimeMinisters all inheriting from the parent class (and possibly parent controller too). 

Or is there some reason they all need to be separate that I'm missing?

Sorry to be a pain - but now this is a third import it does feel like we need to refactor a bit.

chao-xian:
@BeckaL can we have the README and the development controller updated with these as live examples please.

app/controllers/historic_appointments_controller.rb
chao-xian: or a module with the same methods to include instead

hannako:
Becka, Jon and I had a chat about this PR offline. Decided to:

1. Rename the models/views/controllers to follow pattern of previous imports, ie `PastPrimeMinister` instead of `HistoricAppointment`
2. Make the same amendments to https://github.com/alphagov/collections/pull/3186
3. Follow up work to move:
 -  the hardcoded stuff out of the [past foreign secretaries controller](https://github.com/alphagov/collections/blob/main/app/controllers/past_foreign_secretaries_controller.rb#L5-L288) and [past chancellors controller](https://github.com/alphagov/collections/blob/main/app/controllers/past_chancellors_controller.rb#L5-L345) 
 - And then if it's not too much work, creating a single HistoricAppointmentsController and associated refactoring that will allow.
 
This is not essential - the code works - but it would be great to get it refactored before it falls out of everyone's head.

app/views/past_prime_ministers/show.html.erb
```diff
@@ -0,0 +1,67 @@
...
+  </div>
+
+  <div class="govuk-grid-column-one-third">
+    <p class="govuk-body">
+      <%=  render "govuk_publishing_components/components/related_navigation", {
+      content_item: @past_prime_minister.related_prime_ministers_nav,
+      ga4_tracking: true
+    } %>
+    <p class="govuk-body">
+      <a class="govuk-link" href="/government/history/past-prime-ministers"><%= I18n.t("past_prime_ministers.view_all")  %></a>
```
hannako, line 64: There's no route for this until Callum's PR is in? Would you mind updating the pr description with next steps? Ie merge them both in before republishing the historic appointments 🙇‍♀️

hannako (approved):
Looks great!

app/views/past_prime_ministers/show.html.erb
BeckaL: Oh this should be OK - I think even with the relative route it still hits router and so is routed through to Whitehall. (I just tested it out by modifying a live collections-rendered page's html to be this relative link, and it worked)

hannako: Ah yep good point, thanks for explaining!

PR #3205, opened 2023-03-16
Refactor historic appointments

BeckaL:
There are currently three pages that are index pages displaying all historic past role holders of major roles in government, which have all recently been moved from being rendered in whitehall to being rendered in this application:

* Past prime ministers
* Past foreign secretaries
* Past chancellors

This refactors these three types of historic appointment to re-use as much code as possible. So specifically, this PR goes:

* from 3 individual controllers to a single historic appointments controller
* from 3 separate view templates to a single one

Other improvements in this PR:

* Remove the hard-coding of past foreign secretaries and past chancellors from the controller, instead creating yaml content items here as per the coronavirus landing page.
* Add more tests for these pages
* Introduce a historic appointments presenter, responsible for presenting some information for past chancellors and past foreign secretaries in a similar way to the past prime ministers presenter
* Move all hard-coded text on the index pages to locale files
* A bit of rationalisation of the partials to render these pages

### A note on the differences between past prime ministers and the other two pages

There's a slight complexity in these pages in that only one (prime ministers) is rendered from a content item, the other two have always been hard-coded. The ideal state is that all of these pages are rendered from a content item that is published by whitehall as the publishing application that has access to this data.  This means that we often have to treat these separately in code.

[Trello](https://trello.com/c/47S36ZI7/506-refactor-past-foreign-secretaries-past-pms-and-past-chancellors-in-collections-to-share-as-much-as-possible)


app/controllers/historic_appointments_controller.rb
```diff
@@ -0,0 +1,44 @@
...
+      @presenter = PastPrimeMinistersIndexPresenter.new(@index_page.all_prime_ministers)
+    else
+      @index_page = HistoricAppointmentsIndex.find!(request.path)
+      @presenter = HistoricAppointmentsIndexPresenter.new(@index_page)
+    end
+    setup_content_item_and_navigation_helpers(@index_page)
+  end
+
+  def show_past_foreign_secretary
+    setup_content_item_and_navigation_helpers(HistoricAppointmentsIndex.find!(request.path.split("/")[0...-1].join("/")))
```
hannako, line 19: I know this is originally imported code, but I wonder if we should improve it whilst you're in the area! We could just pass in`request.path` rather than `request.path.split("/")[0...-1].join("/")` and add a bit more logic to the HistoricAppointmentsIndex find method to figure out which file to read. Eg:

```
def self.find!(base_path)
    directory =
      base_path.include?("past-foreign-secretaries") ? "past_foreign_secretaries" : "past_chancellors"
    yaml_content = YAML.load_file(Rails.root.join("config/#{directory}/content_item.yml"))
    new(ContentItem.new(yaml_content))
  end
```
Not a blocker, just a style thing.

app/views/shared/_historical_figures_by_century.html.erb
```diff
@@ -9,14 +9,15 @@
...
-  <div class="govuk-grid-row historic-people-index__section-row">
-    <% row.each do | name, info | %>
-      <%= render partial: "shared/historical_person", locals: {
-        name: name,
-        info: info,
-      } %>
-    <%end%>
-  </div>
-<% end %>
+<div id=<%= "historical-people-#{heading.downcase.gsub(' ', '-')}" %>>
```
hannako: This is a good idea!!

app/controllers/past_chancellors_controller.rb
```diff
@@ -1,348 +1,7 @@
 class PastChancellorsController < ApplicationController
   def index
-    @index_page = HistoricAppointmentsIndex.find!(request.path)
+    content_item = YAML.load_file(Rails.root.join("config/past_chancellors/content_item.yml"))
+    @index_page = HistoricAppointmentsIndex.new(ContentItem.new(content_item))
```
hannako: Satisfying refactor 🥳

hannako (commented):
This is a great PR Becka! The commits are so easy to follow, amazing job.
There are some really brilliant improvements here. My only thought is that perhaps we're a bit premature to refactor all the way down to a single controller. That feels like the end game when everything is read from a live content item. But whilst we are still forced to source content from different paths, it leaves us with logic that's not super easy to follow.
What do you think about having an `HistoricAppointmentsController` and a `PastPrimeMinistersController`. The former handling the hardcoded yaml content, and the latter reading from the content item. We could still use the refactored templates where possible. E.g.
```
class PastPrimeMinistersController < ApplicationController
  def index
    @index_page = PastPrimeMinistersIndex.find!(request.path)
    @presenter = PastPrimeMinistersIndexPresenter.new(@index_page.all_prime_ministers)
    setup_content_item_and_navigation_helpers(@index_page)
    render "historic_appointments/index"
  end
  ...
end
```
This would mean on the whole, we only refer to PrimeMinistery named things in the PastPrimeMinistersController, and HistoricAppointmenty named things in the HistoricAppointmentsController.

BeckaL:
Thanks for the review @hannako! I did wonder how far to go with the refactoring, so agree with your suggestion of keeping the controllers separate. Hopefully I can just revert the last few commits and do it not too painfully, I think the controllers bit was the last bit. All your other suggestions look sensible too, will have a rework and let you know when it's ready to have another look.

hannako (approved):
Looks great, thank you for doing this work Becka!

PR #3227, opened 2023-04-14
Render /world/embassies

chrislo:
## Todo

- [x] Investigate whether we need to escape `locality` in our organisation partial. See Becka's comment below.
- [x] Move hardcoded strings into locale yaml file. See Becka's comment below.

---

Trello: https://trello.com/c/MMUcRbIT

This commit adds rendering of /world/embassies, a page currently
rendered by Whitehall. When the embassies_index document is published
to the publishing API in production, we intend to switch over the
rendering of this route to this application (and remove the relevant
code from Whitehall).

We've designed the interface of EmbassiesIndexPresenter so that we can
reuse as much of the [existing Whitehall view template][1] as
possible. Hopefully this makes the correspondence between the two as
clear as possible.

We have added tests for the two methods in EmbassiesIndexPresenter
that contain some logic, but we're relying on the higher level feature
spec to exercise the wrapping of json into objects in that
presenter. The fixture file we're using in the feature test is taken
from the [example frontend embassies_index document in the publishing
api repo][2].

Although there are translations for the title of this embassies index
page in Whitehall (e.g. in [en.yml][3]), there's no way to access any of
those translations because the [/world/embassies route doesn't accept a
locale param][4]. Because of this we've chosen to hardcode the title of
the page in `EmbassiesIndexPresenter#title` to preserve the behaviour in
whitehall.

[1]: https://github.com/alphagov/whitehall/blob/a8a0634e6bf15596b43648e1f4cbcd302b73494d/app/views/embassies/index.html.erb
[2]: https://github.com/alphagov/publishing-api/blob/067cd760d76e91d97b47d5e8edecff2ac0fe5ea7/content_schemas/examples/embassies_index/frontend/embassies_index.json
[3]: https://github.com/alphagov/whitehall/blob/8ad388e7c9ad40283928df4baac0b5da8f172243/config/locales/en.yml#L446
[4]: https://github.com/alphagov/whitehall/blob/8ad388e7c9ad40283928df4baac0b5da8f172243/config/routes.rb#L48

## Screenshot before change (rendered by whitehall)

Screenshot 2023-05-11 at 12 06 52

## Screenshot after change (rendered by collections)

Screenshot 2023-05-11 at 12 06 30

### DIFF
--- a/app/presenters/embassies_index_presenter.rb
+++ b/app/presenters/embassies_index_presenter.rb
@@ -4,7 +4,7 @@ class EmbassiesIndexPresenter
   end
 
   def title
-    "Find a British embassy, high commission or consulate"
+    I18n.t("embassies.title")
   end
 
   def details
--- a/app/presenters/embassies_index_presenter.rb
+++ b/app/presenters/embassies_index_presenter.rb
@@ -1,0 +1,54 @@
+class EmbassiesIndexPresenter
+  def initialize(embassies_index)
+    @embassies_index = embassies_index
+  end
+
+  def title
+    "Find a British embassy, high commission or consulate"
+  end
+
+  def details
+    @embassies_index.content_item.details
+  end
+
+  def embassies_by_location
+    details["world_locations"].map do |data|
+      Embassy.new(data)
+    end
+  end
+
+  class Embassy
+    def initialize(data)
+      @data = data
+    end
+
+    def name
+      @data["name"]
+    end
+
+    def can_assist_british_nationals?
+      %w[local remote].include?(@data["assistance_available"])
+    end
+
+    def can_assist_in_location?
+      @data["assistance_available"] == "local"
+    end
+
+    def remote_office
+      RemoteOffice.new(name: @data["remote_office"]["name"],
+                       location: @data["remote_office"]["country"],
+                       path: @data["remote_office"]["path"])
+    end
+
+    def organisations_with_embassy_offices
+      @data["organisations_with_embassy_offices"].map do |o|
+        Organisation.new(locality: o["locality"],
+                         name: o["name"],
+                         path: o["path"])
+      end
+    end
+
+    Organisation = Struct.new(:name, :locality, :path, keyword_init: true)
+    RemoteOffice = Struct.new(:name, :location, :path, keyword_init: true)
+  end
+end
--- a/app/views/embassies/_organisation.html.erb
+++ b/app/views/embassies/_organisation.html.erb
@@ -1,6 +1,6 @@
 <li>
   <%= render "govuk_publishing_components/components/heading", {
-    text: ERB::Util.html_escape(organisation.locality),
+    text: organisation.locality,
     heading_level: 3,
     margin_bottom: 1
   } %>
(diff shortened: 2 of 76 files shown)


app/models/embassies_index.json
```diff
@@ -0,0 +1,3129 @@
...
+            {
+              "locality": "Algiers",
+              "name": "British Embassy Algiers",
+              "path": "/world/organisations/british-embassy-algiers"
+            }
+          ]
+        },
+        {
+          "name": "American Samoa",
+          "has_consular_services": "remote",
```
JonathanHallam, line 53: I think we're going to have to think about this quite carefully. I totally see what you were talking about at stand up. I'd like to see if there's a way that we can model this within World Locations. I realise that's a bit of a pain with places like American Samoa that don't have their own page and exist only as a content item, but I'd like to see if we can work out a way to do this without having to hardcode a field for each of these.

chrislo: Thanks @JonathanHallam! The discussion on the proposed schema was handled in https://github.com/alphagov/publishing-api/pull/2354, so I'll resolve this comment now.

app/views/embassies/_organisation.html.erb
```diff
@@ -0,0 +1,8 @@
+<li>
+  <%= render "govuk_publishing_components/components/heading", {
+    text: ERB::Util.html_escape(organisation.locality),
```
BeckaL, line 3: Why do we need the escaping here? I had a quick scan through the content item and couldn't immediately see

BeckaL (commented):
Looks great! I spun it up locally and all looking good too, as per your screenshot. A question in there about html escaping, and a request for something to be put in translations for consistency with other content types we've migrated across.

app/views/embassies/_organisation.html.erb
chrisroos: I'm afraid we didn't investigate why that's there. We just copied [the partial from Whitehall](https://github.com/alphagov/whitehall/blob/f3c8ace884f8d27c91a130cd83cde96bb174023a/app/views/embassies/_organisation.html.erb) as-is in the hope that it's easier for someone to see where it came from. My guess would be to ensure that we're escaping things like `&` if they're entered in the locality field. We'll investigate further to see whether it's still necessary.

chrisroos:
Thanks, @BeckaL! We've added checklist items for the points you've raised so will investigate those soon.

app/views/embassies/_organisation.html.erb
chrislo: You're right, rails does this by default in ERB templates (and has since Rails 3 😊). We've removed this call to `html_escape` in [66098ce](https://github.com/alphagov/collections/pull/3227/commits/66098ce0768d0947121f55b71ab7d16c0de2c49c)

app/presenters/embassies_index_presenter.rb
```diff
@@ -0,0 +1,54 @@
+class EmbassiesIndexPresenter
+  def initialize(embassies_index)
+    @embassies_index = embassies_index
+  end
+
+  def title
+    "Find a British embassy, high commission or consulate"
```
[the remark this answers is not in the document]
chrislo: We've done this in [b475685](https://github.com/alphagov/collections/pull/3227/commits/b475685c73d50a1431bc5cc59576996cb56def05).

PR #3229, opened 2023-04-17
[DO NOT MERGE!!!] Practice setting up an AB test

Rosa-Fox:
[Trello](https://trello.com/c/gADp9GVM/445-define-configure-and-write-ab-test)

- Displays different versions of the /coronavirus landing page depending on the test variant.
- The different versions will not be noticeably different to the user which minimises the risk if this branch reaches production.
- Please note - the plan is just to push this this branch to integration. The lack of visual change is cautionary!

You can test this out locally by:

1. Run the app using `./startup.sh --live`
2.  Use Postman to send a GET request to `http://localhost:3070/coronavirus` and add a header `GOVUK_ABTEST_UXMPRACTICETEST` with a value of `A` or `B`. 
3. You will see that the `<meta name="govuk:ab-test" content="UxmPracticeTest:B"` content changes depending on the header value. The HTML body will also change to have either one or two <br> tags. 

### DIFF
--- a/app/controllers/coronavirus_landing_page_controller.rb
+++ b/app/controllers/coronavirus_landing_page_controller.rb
@@ -4,6 +4,16 @@ class CoronavirusLandingPageController < ApplicationController
   def show
     breadcrumbs = [{ title: t("shared.breadcrumbs_home"), url: "/", is_page_parent: true }]
 
+    ab_test = GovukAbTesting::AbTest.new(
+      "UxmPracticeTest",
+      dimension: 68,
+      allowed_variants: %w[A B Z],
+      control_variant: "Z",
+    )
+
+    @requested_variant = ab_test.requested_variant(request.headers)
+    @requested_variant.configure_response(response)
+
     render "show",
            locals: {
              breadcrumbs:,
--- a/spec/controllers/coronavirus_landing_page_controller_spec.rb
+++ b/spec/controllers/coronavirus_landing_page_controller_spec.rb
@@ -1,5 +1,6 @@
 RSpec.describe CoronavirusLandingPageController do
   include CoronavirusContentItemHelper
+  include GovukAbTesting::RspecHelpers
 
   describe "GET show" do
     before do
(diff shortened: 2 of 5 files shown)


spec/controllers/coronavirus_landing_page_controller_spec.rb
```diff
@@ -12,7 +13,32 @@
...
+        with_variant UxmPracticeTest: "B" do
+          get :show
+          expect(response).to have_http_status(:success)
+          expect(response.header["Vary"]).to eq("GOVUK-ABTest-UxmPracticeTest")
+          assert_equal("GOVUK-ABTest-UxmPracticeTest", response.headers["Vary"])
+          expect(response.body).to include(
+            "<meta name=\"govuk:ab-test\"  content=\"UxmPracticeTest:B\"  data-analytics-dimension=\"68\"  data-allowed-variants=\"A,B,Z\">",
+          )
+        end
+      end
```
sihugh, line 41: I guess the `Z` variant is covered by the normal tests that just result in the `show` view being rendered. It might be nice to add it in here for explicitness though?

app/controllers/coronavirus_landing_page_controller.rb
```diff
@@ -4,12 +4,38 @@ class CoronavirusLandingPageController < ApplicationController
...
+      dimension: 68,
+      allowed_variants: %w[A B Z],
+      control_variant: "Z",
+    )
+
+    @requested_variant = ab_test.requested_variant(request.headers)
+    @requested_variant.configure_response(response)
+
+    if @requested_variant.variant?("A")
+      render "one_line_break",
```
sihugh, line 18: you don't necessarily have to render an entirely different view if you don't want to, though it's a bit subjective. (Can just have a tiny bit of switching logic in a view if it's small enough) I guess your real test will be a more significant change than a couple of line breaks though!
One risk about entirely different views is that someone else tweaks styles or something for the default view and don't know enough to update the others too. Hopefully that's not an issue here though.

spec/controllers/coronavirus_landing_page_controller_spec.rb
sihugh: Also possibly worth testing what happens if the supplied header is not A,B or Z. Just in case something goes wonky. You know I hate wonky

spec/features/ab_testing_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+    )
+  end
+
+  scenario "viewing the A version of the page" do
+    with_variant UxmPracticeTest: "A" do
+      visit "/coronavirus"
+
+      expect(page.driver.options[:headers]).to eql({ "HTTP_GOVUK_ABTEST_UXMPRACTICETEST" => "A" })
+
+      expect(page.body).to include("<br>").once
```
sihugh, line 21: I tested this locally, and it works 👍

app/controllers/coronavirus_landing_page_controller.rb
```diff
@@ -4,12 +4,38 @@ class CoronavirusLandingPageController < ApplicationController
...
-           }
+    ab_test = GovukAbTesting::AbTest.new(
+      "UxmPracticeTest",
+      dimension: 68,
+      allowed_variants: %w[A B Z],
+      control_variant: "Z",
+    )
+
+    @requested_variant = ab_test.requested_variant(request.headers)
+    @requested_variant.configure_response(response)
```
sihugh, line 15: people often seem to forget this bit 👏

app/views/coronavirus_landing_page/one_line_break.html.erb
```diff
@@ -0,0 +1,79 @@
...
+      heading: "Stay up to date with GOV.UK",
+      data: {
+        "module": "gem-track-click",
+        "track-category": "emailAlertLinkClicked",
+        "track-action": "/coronavirus",
+      },
+      background: true
+    }
+  end %>
+  <br>
```
sihugh, line 78: yep, as I mentioned above - if this is the only change, it might be easier to put this all in the show view with something like:
```
<% if @requested_variant.variant?("A") %>
<br>
<% end %>
```
and likewise for the `B`. It might make this PR a fair bit smaller as you can ditch the two new views and the switching logic from the controller. 

The meta tag section is the same across all three, so that should be fine.

(I haven't exhaustively checked that the views otherwise look the same, but I think they are)

sihugh (commented):
Looks good! I've suggested a couple of things, but they don't affect the user facing output of this (which is the important thing at this stage)

spec/controllers/coronavirus_landing_page_controller_spec.rb
Rosa-Fox: Thanks Simon! I have added these

PR #3264, opened 2023-05-18
Localise hard coded strings

chrislo:


app/views/layouts/application.html.erb
```diff
@@ -34,7 +34,7 @@
 <!DOCTYPE html>
 <html>
 <head>
-  <title><%= yield :title %> - GOV.UK</title>
```
chao-xian, line 37: Not convinced this needs to be ported as GOV.UK _can't_ be translated?

chao-xian (approved):
Super minor question but LGTM - thanks Chris!

app/views/layouts/application.html.erb
chrislo: Yes, I think that's probably true! I don't think I'll leave it as a hard coded string though, just because that would involve adding this file to the exclude list for the linter rules, or adding a single line exception. The former would mean other strings in this file would be missed and the latter makes the file more complicated/ugly. There's several examples of strings that are not currently translated (their corresponding keys in non-`en` locale files have nil values).

chrislo:
Rebasing on `main` and force-pushing before merging.

chrislo:
It's just occurred to me that it's not sufficient to just add the new linter to the default rake tasks, I should also add it to the github CI process so that it runs in the CI action. Looking at that now.

chrislo:
I've added a new task to CI to run the erb linter, so getting ready to merge this now.

PR #3302, opened 2023-06-21
Add ga4 link tracking to topic lists

JamesCGDS:

## What
This PR adds GA4 link tracking to topic lists.

Examples:

- [Topic page](https://www.gov.uk/topic)
- [Subtopic page level 1](https://www.gov.uk/topic/personal-tax)
- [Subtopic page level 2](https://www.gov.uk/topic/personal-tax/national-insurance)
- [Services and information page](https://www.gov.uk/government/organisations/hm-revenue-customs/services-information)

Approach:

All topic and services and information pages make use of the `_topic_list.html.erb` component in their respective templates. When `_topic_list.html.erb` is called, the templates can pass `list_index`, `list_count`, and `list_title` (along with other parameters) to it via `topic_list_helper.rb`.

`topic_list_helper.rb` uses `list_index`, `list_count`, and `list_title` to populate a `ga4_data` hash, which is ultimately what `_topic_list.html.erb` receives. Sometimes, `list_index` (which ultimately sets `index_section`) and `list_count` (which ultimately sets `index_section_count`) will not be available if there are no sections on the page since it is just a list of links [(example)](https://www.gov.uk/topic/personal-tax). This is why the `index_section` and `index_section_count` keys are conditionally rendered. But we would always expect the `list_title` (and hence `section` parameter) to be present.

## Why
Part of the GA4 migration.

[Trello card](https://trello.com/c/1TFIjPLq/595-add-tracking-navigation-topic-pages)

## Visual changes
N/A

### DIFF
--- a/app/helpers/topic_list_helper.rb
+++ b/app/helpers/topic_list_helper.rb
@@ -17,11 +17,17 @@ module TopicListHelper
     }
   end
 
-  def topic_list_params(list, tracking_attributes: nil, list_index: nil, category: nil)
+  def topic_list_params(list, tracking_attributes: nil, list_index: nil, category: nil, list_count: nil, list_title: nil)
     tracking_attributes ||= topic_list_tracking_attributes(list.count, list_index, category)
+    ga4_data = {}
+    ga4_data[:index] = {}
+    ga4_data[:index][:index_section] = list_index + 1 if list_index
+    ga4_data[:index][:index_section_count] = list_count if list_count
+    ga4_data[:section] = list_title if list_title
 
     {
       items: topic_list_items(list, tracking_attributes),
+      ga4_data:,
     }
   end
 
--- a/app/views/components/_topic_list.html.erb
+++ b/app/views/components/_topic_list.html.erb
@@ -4,6 +4,7 @@
   see_more_link ||= false
   small ||= false
   margin_bottom ||= false
+  ga4_data ||= {}
 
   brand ||= false
   brand_helper = GovukPublishingComponents::AppHelpers::BrandHelper.new(brand)
@@ -15,13 +16,29 @@
 %>
 <% if items.any? %>
   <%= tag.ul(class: ul_classes, lang: "en") do %>
-    <% items.each do |item| %>
+    <% items.each_with_index do |item, index| %>
       <li class="app-c-topic-list__item">
+        <% 
+          ga4_link_data = ga4_data.empty? ? {} : {
+            ga4_link: {
+              event_name: "navigation",
+              type: "document list",
+              index: {
+                index_link: index + 1,
+                **ga4_data[:index] || {},
+              },
+              # As see_more_link is not included in the items array, we need to account for it here by adding 1 to items.count
+              # if see_more_link has been passed
+              index_total: (see_more_link ? items.count + 1 : items.count),
+              section: ga4_data[:section],
+            }
+          }
+        %>
         <%=
           link_to(
             item[:text],
             item[:path],
-            data: item[:data_attributes],
+            data: {**item[:data_attributes] || {}}.merge(ga4_link_data),
             "aria-label": item[:aria_label],
             class: "govuk-link govuk-link--no-underline app-c-topic-list__link #{brand_helper.color_class}"
           )
@@ -30,11 +47,25 @@
     <% end %>
     <% if see_more_link %>
       <li class="app-c-topic-list__item">
+        <% 
+          ga4_link_data = ga4_data.empty? ? {} : {
+            ga4_link: {
+              event_name: "navigation",
+              type: "document list",
+              index: {
+                index_link: items.count + 1,
+                **ga4_data[:index] || {},
+              },
+              index_total: items.count + 1,
+              section: ga4_data[:section],
+            }
+          }
+        %>
         <%=
           link_to(
             see_more_link[:text],
             see_more_link[:path],
-            data: see_more_link[:data_attributes],
+            data: {**see_more_link[:data_attributes] || {}}.merge(ga4_link_data),
             class: "govuk-link app-c-topic-list__link #{brand_helper.color_class}"
           )
         %>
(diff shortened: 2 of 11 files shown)


app/views/components/_topic_list.html.erb
```diff
@@ -15,13 +20,30 @@
...
+              **item[:data_attributes] || {},
+              ga4_link: {
+                event_name: "navigation",
+                type: "document list",
+                index: {
+                  index_link: index + 1,
+                  # If ga4_data[:section_index] and ga4_data[:section_count] have not been passed to this component, we don't want to 
+                  # include null values in the push to the dataLayer. Therefore, we conditionally include these keys.
+                  **(ga4_data[:section_index] ? { index_section: ga4_data[:section_index] } : {}),
+                  **(ga4_data[:section_count] ? { index_section_count: ga4_data[:section_count] } : {}),
```
andysellick, line 39: This logic is somewhat complex for the view. I'm wondering if there's an easier approach?

```Ruby
# topic_list_helper.rb
ga4_data = {}
ga4_data[:index_section_count] = list_count if list_count
ga4_data[:index_section] = list_index + 1 if list_index
ga4_data[:section] = list_title if list_title
```

```Ruby
# _topic_list.html.erb
<li class="app-c-topic-list__item">
  <% 
     index_attributes = ga4_data
     index_attributes[:index_link] = index + 1
  %>
  <%= link_to(
    # etc.
    data: {
       ga4_link: {
            event_name: "navigation",
            index: index_attributes,
```

You'd have to figure something out for `section` as that isn't inside `index`, but it should be doable.

andysellick (commented):
Looks like the tracking is all working correctly, just a small refactoring suggestion.

app/helpers/topic_list_helper.rb
```diff
@@ -17,11 +17,17 @@ def topic_list_tracking_attributes(list_count, list_index, category)
...
     tracking_attributes ||= topic_list_tracking_attributes(list.count, list_index, category)
+    ga4_data = {
+      section_count: list_count,
+      section_index: (list_index + 1 if list_index),
+      section_title: list_title,
+    }
 
     {
       items: topic_list_items(list, tracking_attributes),
+      ga4_data:,
```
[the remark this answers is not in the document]
JamesCGDS: Passing `ga4_data` as `ga4_data: ga4_data` results in the following rubocop warning:

image

So I changed it to `ga4_data:`.

app/views/components/_topic_list.html.erb
JamesCGDS: This was along the lines of my initial approach but I felt it was a bit clunky. I'll revert it to how it was before.

app/helpers/topic_list_helper.rb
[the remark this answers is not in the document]
andysellick: Ah right this must be some clever Ruby/Rails thing I wasn't aware of. Makes sense, thanks.

app/views/components/_topic_list.html.erb
andysellick: Great, thanks. I think it's better to try to keep the logic out of the view as much as possible.

app/views/services_and_information/_grouped_links.html.erb
```diff
@@ -13,7 +13,7 @@
       <%
         list_item_count = group.see_more_link ? group.examples.count + 1 : group.examples.count
         tracking_attributes = topic_list_tracking_attributes(list_item_count, group_index, 'navServicesInformationLinkClicked')
-        topic_list_component_params = topic_list_params(group.examples, tracking_attributes: tracking_attributes)
+        topic_list_component_params = topic_list_params(group.examples, list_index: group_index, tracking_attributes: tracking_attributes, list_count: grouped_links.count, list_title: group.title)
```
gclssvglx, line 16: Getting quite a number of params here, not necessarily this time - but maybe a small refactor next time we need to add any.

gclssvglx (changes requested):
Looking great - thanks @JamesCGDS 👍  Just a couple of things to clean up in the tests 
to make it more readable please.

app/views/components/_topic_list.html.erb
```diff
@@ -15,13 +16,27 @@
...
+    <% items.each_with_index do |item, index| %>
       <li class="app-c-topic-list__item">
         <%=
           link_to(
             item[:text],
             item[:path],
-            data: item[:data_attributes],
+            data: {
+              **item[:data_attributes] || {},
+              ga4_link: {
```
andysellick, line 27: Some of these data attributes are here regardless of whether any GA4 data attributes are passed. Is there a situation where they're not passed? Should we make them only appear if `ga4_data` exists? If this data is always here, why not hard code the link tracker module to the parent?

spec/components/topic_list_spec.rb
```diff
@@ -63,3 +142,31 @@ def component_name
     expect(rendered).to have_selector(".govuk-\\!-margin-bottom-6")
   end
 end
+
+# RSpec.describe "topic_list component", type: :view do
```
AshGDS, line 146: This needs removing?

AshGDS (changes requested):
Looks good :+1: requested a couple changes

app/helpers/topic_list_helper.rb
```diff
@@ -17,11 +17,17 @@ def topic_list_tracking_attributes(list_count, list_index, category)
...
     tracking_attributes ||= topic_list_tracking_attributes(list.count, list_index, category)
+    ga4_data = {}
+    ga4_data[:index] = {}
+    ga4_data[:index][:index_section] = list_index + 1 if list_index
+    ga4_data[:index][:index_section_count] = list_count if list_count
+    ga4_data[:section] = list_title if list_title
 
     {
       items: topic_list_items(list, tracking_attributes),
+      ga4_data:,
```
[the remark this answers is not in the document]
JamesCGDS: It is, yeah. It's how the data is passed to the template. It's a similar thing in JS when an object's property and value are the same name so you can just put the property name (the `:` is a rails thing)

spec/components/topic_list_spec.rb
JamesCGDS: Whoops, thanks for spotting, will remove 👍

AshGDS (approved):

PR #3319, opened 2023-07-06
Remove taxons_redirection_controller

chao-xian:
This controller was only for a couple of Coronavirus related taxons, and only for doing an HTTP redirect which the routes file can handle.


BeckaL (approved):
As discussed - this looks good to me, I'm not sure if we want to preserve the "307" (temporary redirect) vs "302" (redirect) behaviour. I'll leave that to you however, this preserves existing behaviour

leenagupte (approved):
The redirect was originally added because the data that powers the coronavirus landing page is from the coronavirus-taxon. 

It was done this way so that it could be styled differently to a regular taxon page. 

I guess the reason for the separate controller was so that all of the child taxon could be redirect appropriately to the correct child coronavirus page. 

Once the child pages were removed we should have removed the redirect controller. This solution is much nicer.

PR #3320, opened 2023-07-06
Add dev page

chao-xian:
Help devs to find pages that are rendered by the app. This is inspired by the same in government-frontend.

This PR is a continuation of work that was started in the previous GIFT week.

This page is only viewable when running collections locally or or Heroku and visting the app directly.

## Screenshot

collections-pr-3320 herokuapp com_(iPad Air)

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -3,7 +3,7 @@ Rails.application.routes.draw do
   # content-store (by collections-publisher) - whenever the routes below
   # change, also change the routes claimed by collections-publisher.
 
-  get "/", to: redirect(path: :browse)
+  root to: "development#index"
 
   mount GovukPublishingComponents::Engine, at: "/component-guide"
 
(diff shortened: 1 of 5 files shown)


config/routes.rb
```diff
@@ -3,7 +3,7 @@
   # content-store (by collections-publisher) - whenever the routes below
   # change, also change the routes claimed by collections-publisher.
 
-  get "/", to: redirect(path: :browse)
+  root to: "development#index"
```
BeckaL, line 6: Not sure what the correct pattern is here, but is there any value of retaining "/" as redirecting to browse (I'm not sure why this redirect was here in the first place given "/" will always be routed to frontend), and having the development controller at /development? Although I know frontend is different because it does actually render the homepage, I like the pattern it has of this plus only showing it in development: https://github.com/alphagov/frontend/blob/main/config/routes.rb#L18-L20

BeckaL (approved):
Approving but with a question about what the appropriate pattern is for what the route is, and whether it's just scoped to the dev environment or not - I'll leave this up to you since you have more context

PR #3397, opened 2023-09-25
Add user research banner

catalinailie:
The banner will be applied on the following pages:
- [/browse/tax/self-assessment](https://www.gov.uk/browse/tax/self-assessment)
- [/browse/tax](https://www.gov.uk/browse/tax)
- [/topic/personal-tax/self-assessment](https://www.gov.uk/topic/personal-tax/self-assessment)

Before:
Screenshot 2023-09-26 at 10 07 02

After:
Screenshot 2023-09-26 at 10 06 20

### DIFF
--- a/app/views/browse/show.html.erb
+++ b/app/views/browse/show.html.erb
@@ -20,7 +20,7 @@
   } %>
 <% end %>
 
-<%= render "shared/browse_header", { margin_bottom: page.slug == "benefits" ? 7 : 9 } do %>
+<%= render "shared/browse_header", { margin_bottom: page.slug == "benefits" || page.recruitment_survey_url ? 7 : 9 } do %>
   <h1 class="browse__heading govuk-heading-xl">
     <%= page.title %>
   </h1>
@@ -29,6 +29,17 @@
     margin_bottom: 2,
     inverse: true
   } %>
+<% end %>
+
+<% if page.recruitment_survey_url %>
+  <div class="govuk-width-container">
+    <%= render "govuk_publishing_components/components/intervention", {
+      suggestion_text: "Help improve a new GOV.UK tool",
+      suggestion_link_text: "Sign up to take part in user research",
+      suggestion_link_url: page.recruitment_survey_url,
+      new_tab: true,
+    } %>
+  </div>
 <% end %>
 
 <% if page.slug == "benefits" %>
--- a/app/views/subtopics/show.html.erb
+++ b/app/views/subtopics/show.html.erb
@@ -18,6 +18,17 @@
   <%= render "govuk_publishing_components/components/title", title_params %>
 <% end %>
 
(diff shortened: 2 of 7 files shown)


app/views/browse/show.html.erb
```diff
@@ -31,6 +31,17 @@
   } %>
 <% end %>
 
+<% if page.recruitment_survey_url %>
```
leenagupte, line 34: I forgot to add this to my original review, but I think you might need to include the view stylesheet at the top of this view:

```ruby
<% add_view_stylesheet("intervention") %>
```

catalinailie: We don't use that stylesheet in this file because we don't want to change the `margin-top` for this page. We are using that only for the subtopic page.

app/assets/config/manifest.js
```diff
@@ -4,6 +4,7 @@
 
 //= link components/_taxon-list.css
 //= link components/_topic-list.css
+//= link components/_intervention.css
```
MartinJJones, line 7: We can remove this line if the approach mentioned in the show.html.erb view template is used

app/assets/stylesheets/components/_intervention.scss
```diff
@@ -0,0 +1,5 @@
+@import "govuk_publishing_components/individual_component_support";
+
+.gem-c-intervention {
+  margin-top: govuk-spacing(4);
+}
```
MartinJJones, line 5: We can remove this file if the approach mentioned in the show.html.erb view template is used

app/views/subtopics/show.html.erb
```diff
@@ -18,6 +19,17 @@
   <%= render "govuk_publishing_components/components/title", title_params %>
 <% end %>
 
+<% if subtopic.recruitment_survey_url %>
+  <div class="govuk-width-container">
```
MartinJJones, line 23: We should be able to update this div element from `<div class="govuk-width-container">` to `<div class="govuk-!-static-margin-top-4">`

Using the spacing class from the design system will allow us to remove the `intervention.scss` component stylesheet and the related line in the `manifest.js` file.

This will also resolve a layout issue caused by the presence of the "govuk-width-container" class, when the screen width is less than 1020px, please see the screenshot below showing additional margin on the left and right:
Screenshot 2023-09-26 at 17 12 23

app/views/subtopics/show.html.erb
```diff
@@ -1,3 +1,4 @@
+<% add_app_component_stylesheet("intervention") %>
```
MartinJJones, line 1: We can remove this line if the approach mentioned below is used

MartinJJones (commented):
Changes look great, just a couple of suggestions on how we can remove the custom intervention component stylesheet.

MartinJJones (approved):

PR #3404, opened 2023-10-03
Prevent "latest changes" link from being tracked as an email subscription link

AshGDS:
## What

- On these types of pages https://www.gov.uk/topic/personal-tax/income-tax there is a "Get emails for this topic" link, and a  "See latest changes to this content" link at the top right of the page.
- Our GA4 tracking was added on a parent `<div>` which gave the tracking to the email link and the "See latest changes" link.
- We don't want the tracking on the "See latest changes" link, so I have wrapped the email link in a `<span>` and put the tracking on that instead.

## Why

https://trello.com/c/tbT92mSB/680-subscribe-link-click-firing-on-non-subscribe-events-on-topic-pages

### DIFF
--- a/app/views/subtopics/_subtopic.html.erb
+++ b/app/views/subtopics/_subtopic.html.erb
@@ -11,17 +11,17 @@
       <% end %>
     </div>
 
-    <div 
-      class="govuk-grid-column-one-third govuk-!-margin-top-8 govuk-!-margin-bottom-3"
-      data-module="ga4-link-tracker"
-      data-ga4-link='{ "event_name": "navigation", "type": "subscribe", "index": { "index_link": 1 }, "index_total": 1, "section": "Top" }'
-      data-ga4-track-links-only
-    >
-      <%= render "govuk_publishing_components/components/subscription_links", {
-        email_signup_link: "/email-signup?link=/topic/#{params[:topic_slug]}/#{params[:subtopic_slug]}",
-        email_signup_link_text: t("shared.get_emails")
-      } %>
-
+    <div class="govuk-grid-column-one-third govuk-!-margin-top-8 govuk-!-margin-bottom-3">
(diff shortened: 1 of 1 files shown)


app/views/subtopics/_subtopic.html.erb
```diff
@@ -11,17 +11,17 @@
...
-      <%= render "govuk_publishing_components/components/subscription_links", {
-        email_signup_link: "/email-signup?link=/topic/#{params[:topic_slug]}/#{params[:subtopic_slug]}",
-        email_signup_link_text: t("shared.get_emails")
-      } %>
-
+    <div class="govuk-grid-column-one-third govuk-!-margin-top-8 govuk-!-margin-bottom-3">
+      <span
+        data-module="ga4-link-tracker"
+        data-ga4-link='{ "event_name": "navigation", "type": "subscribe", "index": { "index_link": 1 }, "index_total": 1, "section": "Top" }'
+        data-ga4-track-links-only
```
andysellick, line 18: I think you could also pass the attributes directly to the component? https://components.publishing.service.gov.uk/component-guide/subscription_links#with_tracking

AshGDS: @andysellick Thanks. I spoke to James and I think there's a reason the tracking is added like this. We couldn't remember fully, but think it was something to do with multiple types of links in the component, and the need to track them separately/with different values? 

Here are the original PRs https://github.com/alphagov/finder-frontend/pull/3081 and https://github.com/alphagov/finder-frontend/pull/3081

Would you recommend I still try and pass it through the component? Thanks

andysellick: No if you have a reason for doing it this way, that's fine.

andysellick (approved):

PR #3426, opened 2023-10-30
Stop guessing urls for Topical Event image srcset attribute

tunylund:
Whitehall has changed to start using modern Asset Manager urls for each asset. This means that the image urls are no longer guessable. Each image variant has its own unique id. In order to have a srcset for images.

This pr gets the correct image urls from the topical event content.

[Trello](https://trello.com/c/l7AHikFh/233-story-implement-non-legacy-urls-for-topical-event-logo)

Specifically this change is about the following acceptance criteria:
```
GIVEN a Topical Event with a logo
WHEN I open the Topical Event on the live stack (View on website btn)
THEN I should see the Topical event logo on the page
AND the logo should be loaded without using legacy_url_path
AND the page should have the image srcset-attribute using asset-id urls instead of legacy_url_path
```

The background context for this work is that the wayhow Whitehall references assets in Asset Manager - including images is changing. Previously Whitehall would have a hardcoded configuration for urls such that  would make it possible for the various sizes of the same image urls to be guessable. 

The new behaviour is that each individual image will have its own url provided by Asset Manager. The urls are therefore no longer guessable as they all have a unique id in them.

Please see these 2 related prs that show how the urls will be carried from Whitehall to Collections.
https://github.com/alphagov/whitehall/pull/8432
https://github.com/alphagov/publishing-api/pull/2534

### DIFF
--- a/app/views/topical_events/show.html.erb
+++ b/app/views/topical_events/show.html.erb
@@ -58,7 +58,13 @@
 
     <% if @topical_event.image_url %>
       <div class="topical-events__image">
-      <%= image_tag(@topical_event.image_url, alt: @topical_event.image_alt_text, srcset: { @topical_event.image_srcset[2][:url] => "2x", @topical_event.image_srcset[0][:url] => "3x",}) %>
+      <%=
+        if @topical_event.image_srcset.any?
+          image_tag(@topical_event.image_url, alt: @topical_event.image_alt_text, srcset: @topical_event.image_srcset)
+        else
+          image_tag(@topical_event.image_url, alt: @topical_event.image_alt_text)
+        end
+      %>
       </div>
     <% end %>
 
--- a/spec/features/topical_event_spec.rb
+++ b/spec/features/topical_event_spec.rb
@@ -30,6 +30,23 @@ RSpec.feature "Topical Event pages" do
     expect(page).to have_css("img[src='https://www.gov.uk/some-image.png'][alt='Text describing the image']")
   end
 
+  it "includes image srcset for various image sizes" do
+    visit base_path
+    expect(page).to have_css("img[srcset='https://www.gov.uk/some-medium-image.png 2x, https://www.gov.uk/some-high-image.png 3x']")
+  end
+
+  context "when the image has no variants" do
+    before do
+      content_item["details"]["image"].delete("medium_resolution_url")
+      content_item["details"]["image"].delete("high_resolution_url")
+      stub_content_store_has_item(base_path, content_item)
+    end
+    it "omits srcset if there are no higher res image sizes" do
+      visit base_path
+      expect(page).not_to have_css("img[srcset]")
+    end
+  end
+
   it "sets the body text" do
     visit base_path
     expect(page).to have_text(content_item.dig("details", "body"))
--- a/app/models/topical_event.rb
+++ b/app/models/topical_event.rb
@@ -23,23 +23,23 @@ class TopicalEvent
     @content_item.content_item_data.dig("details", "image", "url")
   end
 
-  def image_alt_text
-    @content_item.content_item_data.dig("details", "image", "alt_text")
+  def image_medium_resolution_url
+    @content_item.content_item_data.dig("details", "image", "medium_resolution_url")
+  end
+
+  def image_high_resolution_url
+    @content_item.content_item_data.dig("details", "image", "high_resolution_url")
   end
 
   def image_srcset
-    [
-      { url: sized_image_url(960), size: "960w" },
-      { url: sized_image_url(712), size: "712w" },
-      { url: sized_image_url(630), size: "630w" },
-      { url: sized_image_url(465), size: "465w" },
-      { url: sized_image_url(300), size: "300w" },
-      { url: sized_image_url(216), size: "216w" },
-    ]
+    set = {}
+    set[image_medium_resolution_url] = "2x" if image_medium_resolution_url
+    set[image_high_resolution_url] = "3x" if image_high_resolution_url
+    set
   end
 
-  def sized_image_url(width)
-    image_url.gsub("/s300_", "/s#{width}_")
+  def image_alt_text
+    @content_item.content_item_data.dig("details", "image", "alt_text")
   end
 
   def body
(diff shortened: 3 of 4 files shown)


spec/features/topical_event_spec.rb
```diff
@@ -30,6 +30,11 @@
     expect(page).to have_css("img[src='https://www.gov.uk/some-image.png'][alt='Text describing the image']")
   end
 
+  it "includes image srcset for various image sizes" do
+    visit base_path
+    expect(page).to have_css("img[srcset='https://www.gov.uk/some-medium-image.png 2x, https://www.gov.uk/some-high-image.png 3x']")
```
leenagupte, line 35: It's great that you've added a test for this 🎉

app/views/topical_events/show.html.erb
```diff
@@ -58,7 +58,7 @@
 
     <% if @topical_event.image_url %>
       <div class="topical-events__image">
-      <%= image_tag(@topical_event.image_url, alt: @topical_event.image_alt_text, srcset: { @topical_event.image_srcset[2][:url] => "2x", @topical_event.image_srcset[0][:url] => "3x",}) %>
+      <%= image_tag(@topical_event.image_url, alt: @topical_event.image_alt_text, srcset: @topical_event.image_srcset) %>
```
leenagupte, line 61: What happens if `image_srcset` is empty?

leenagupte (commented):
Thanks for adding this. And it's great that the tests have been updated. 🎉 

Are there any expected changes for the user?
Do you have an example of pages with medium and high resolution images so we can see it in action? I tried the top two from the [topical events page on GOV.UK](https://www.gov.uk/government/topical-events) but they don't have the attribute set.

app/views/topical_events/show.html.erb
tunylund: browsers ignore the empty srcset attribute and show the image defined in the src-attribute

tunylund: I made it so that srcset is not rendered if there are no image variants to avoid any possible issues with it

app/views/topical_events/show.html.erb
```diff
@@ -58,7 +58,9 @@
 
     <% if @topical_event.image_url %>
       <div class="topical-events__image">
-      <%= image_tag(@topical_event.image_url, alt: @topical_event.image_alt_text, srcset: { @topical_event.image_srcset[2][:url] => "2x", @topical_event.image_srcset[0][:url] => "3x",}) %>
+      <%= @topical_event.image_srcset.any? ?
```
leenagupte, line 61: Ah. I missed that you added the conditional as a ternary operator, hence my previous question about what happens if `image_srcset` is `nil`. I've never seen a ternary operator written over multiple lines like this before. 

Given that logic in views is generally more difficult to read, I'd recommend using a standard if/else here instead.

leenagupte (commented):
I'd missed the conditional in the view in the first review, and have added an inline comment about it.

The values being pulled out of the content item and used to determine the size of the image look ok and match the content schema. I guess these values aren't in use yet?

spec/features/topical_event_spec.rb
```diff
@@ -30,6 +30,19 @@
...
+  end
+
+  context "when the image has no variants" do
+    let(:content_item) { fetch_fixture("topical_event_with_no_image_variants") }
```
hannako, line 39: I didn't realise collections had this `fetch_fixture` method. We usually something like [this](https://github.com/alphagov/collections/blob/a30feec0625d190724a5e765c400161ff646bfa2/spec/features/atom_feeds_spec.rb#L117)

hannako: It isn't a blocker for this PR, but maybe you could update your spec at some point!

tunylund: oh interesting, the `GovukSchemas::Example.find(schema_name, example_name: schema_name)` seems to fetch test fixtures from the examples written in publishing_api. It basically makes these tests become sort of contract tests between publishing api schema and collections app. however that then means that the test data has to be written in the examples folder of publishing api and therefore makes management of these testes a bit strange.

I can prepare a separate pr for that as that would require a change in publishing api to provide specific exampel files there.

app/views/topical_events/show.html.erb
tunylund: fixed this

spec/features/topical_event_spec.rb
hannako: You can either create a new example file in publishing components (which might not be a bad idea, as there's only one example for topical events which is quite basic [ie](https://github.com/alphagov/publishing-api/tree/main/content_schemas/examples/topical_event/frontend))

Or you can just use the existing schema, and mutate the JSON response like in that link I posted. In that way you don't need to make a change in publishing api.

hannako: Step 7 on here! https://github.com/alphagov/publishing-api/blob/main/docs/content_schemas/changing-an-existing-content-schema.md#adding-a-new-field-to-a-content-schema

tunylund: hmm, i'm not exited to set up test data in another repo for this repo to use. If the example files were desired to be used as contract tests - then publishing_api CI workflow should include an action that executes any tests in Collections app that depend on those example files. Otherwise a dev who changes the schema or examples, would not be aware of collections application breaking.

I did however do as you suggested, i removed the extra fixture file i had added and just modified the data in place.

leenagupte: This was a good catch @hannako. I have to admit I didn't delve into what `fetch_fixture` was doing too closely. Using the examples content items is a pattern that is reflected in other applications, so I agree that it would make sense to use it here too. 

As a bit of historical context, the content used to exist in a separate gem, but I believe they were merged into publishing-api to remove our app overhead.

tunylund:
This fixes this situation in production in production on this page

https://www.gov.uk/government/topical-events/ai-safety-summit-2023

image

richardTowers (approved):
I think this looks good enough to release now (not sure if there are any outstanding stylistic comments, but the code looks fine to me).

Approving so we can fix the issue with the production image

PR #3438, opened 2023-11-13
Past Foreign Secretaries List: Add James Cleverly

Aurorum:
James Cleverly has moved to the Home Office, so should be added to this list here: https://www.gov.uk/government/history/past-foreign-secretaries


hannako:
Thank you for raising this PR -looks like the file was updated as part of our normal internal processes, so I'll close this.

PR #3450, opened 2023-11-24
Revert "Revert "Add user research banner""

patrickpatrickpatrick:
Reverts alphagov/collections#3424


spec/features/research_panel_banner_spec.rb
```diff
@@ -0,0 +1,29 @@
+require "integration_spec_helper"
+
+RSpec.feature "Research panel banner" do
+  include SearchApiHelpers
+  include RecruitmentBannerHelper
+
+  scenario "browse pages where we want to display User Research banner" do
+    schema = GovukSchemas::Example.find("mainstream_browse_page", example_name: "level_2_page")
```
hannako, line 8: If we had a test that iterated through all the paths here we'd have caught that the banner wasn't showing on the A to Z page. So I wonder if it would be good practice to do that. ie 

```
[array_of_paths].each do |path|
  schema["base_path"] = path
  stub_content_store_has_item(path, schema)
  search_api_has_documents_for_browse_page(schema["content_id"], [path], page_size: SearchApiSearch::PAGE_SIZE_TO_GET_EVERYTHING)
  visit path
  expect(page.status_code).to eq(200)
  expect(page).to have_selector(".gem-c-intervention")
end
 ```

hannako (approved):
Thanks Patrick 👍

spec/features/research_panel_banner_spec.rb
patrickpatrickpatrick: Ah yep, that's in the other PRs I reverted but not this one. I'll add it.

PR #3453, opened 2023-11-27
Use medium_resolution_url and high_resolution_url instead of constructing image urls based on size

lauraghiorghisor-tw:
These changes refer to urls of featured documents for `Organisation` and `WorldLocationNews`. This is due to changes in the way `whitehall` references assets in `asset-manager`, including feature images. Previously `whitehall` would have hardcoded urls in the legacy format of `/government/uploads/system/uploads/featured_image_data/file/:id/:filename` that would make it possible for the various sizes of the same image urls to be guessable by simply prefixing the filename in the url `/s465_filename`. The new behaviour is that each individual image will have its own url containing an ID from `asset-manager`, in the format `media/:id/:filename`.

`WorldLocationNews` only uses the original url so no code changes are required. We have nonetheless updated test data to better reflect the desired shape of the urls. `Organisation` uses `:s712` and `:s465` urls which we are now providing from `whitehall` rather than guessing them.

The `publishing-api` schema allows for 3 fields to be sent: `url`, `medium_resolution_url` and `high_resolution_url`. These changes ensure that the `url` is used where the original image asset is required, the `medium_resolution_url` is used when the `:s465` version is required and the `high_resolution_url` when the `:s712` is required. If additional sized are required, the `whitehall` presenters and `publishing-api` schemas would need to be updated. For now, the 3 allowed fields satisfy the user needs.

Since `url` is the only required field in `publishing-api`, it is used as fallback when the other 2 are not available.

These changes have no impact on how the page looks like for the user.

We intend to release these changes first, followed by the whitehall changes:
1. Deploy collections -> organisation and world location news will expect medium and high res url. No such urls are provided so it falls back to original url. This is a temporary solution while we deploy whitehall. 
2. Migrate all Features in whitehall to use assets (just populate DB with assets, no user-facing changes at this point). 
3. Release whitehall changes that use asset-id based urls and send url, medium-res url, and high-res urls to publishing api -> all new organisations and world location news features will now have the desired image version + modern asset-ID based urls ✅
4. Republish all Organisation and World Location News -> because we have migrated all to use assets, republishing them will ensure they use ID-based urls. All organisations and world location news now have modern urls in the size they expect (original, s712, s465)

We tested this branch in integration. Screenshots of a selection of organisations with original URLs instead of :s712 and :s465, rendering fine ✅

New organisations with features
New features using temporary original urls

Existing organisation with s465
Existing features using temporary original urls 

No10 org features
Existing organisation (No 10) using temporary original urls  

Existing World location news with original urls
Existing world location news using temporary original urls 

new world location news feature
New world location news using temporary original urls 

[Trello card](https://trello.com/c/JYIbvFb0/248-feature-to-use-assetids) containing the context of the changes and the release plan
[Whitehall PR](https://github.com/alphagov/whitehall/pull/8570)
[Specific Whitehall changes that provide the medium and high resolution urls](https://github.com/alphagov/whitehall/pull/8570/commits/ea6761676fd5bbb171e90a6a0e0b452bb85e4005)

### DIFF
--- a/app/helpers/organisation_helper.rb
+++ b/app/helpers/organisation_helper.rb
@@ -8,14 +8,6 @@ module OrganisationHelper
     I18n.t("organisations.type.#{organisation_type}", default: I18n.t("organisations.type.other"))
   end
 
-  def image_url_by_size(image_url, size)
-    image_url_array = image_url.split("/")
-    image_by_size_name = "s#{size}_" + image_url_array[-1]
-    image_url_array[-1] = image_by_size_name
-
-    image_url_array.join("/")
-  end
-
   def search_results_to_documents(search_results, organisation, include_metadata: true)
     {
       brand: (organisation.brand if organisation.is_live?),
--- a/app/presenters/organisations/documents_presenter.rb
+++ b/app/presenters/organisations/documents_presenter.rb
@@ -98,15 +98,16 @@ module Organisations
 
     def featured_news(featured, first_featured: false)
       news_stories = []
-      image_size = first_featured ? 712 : 465
 
       featured.each do |news|
         date = Date.parse(news["public_updated_at"]) if news["public_updated_at"]
         text = I18n.t("shared.schema_name.#{news['document_type']&.parameterize(separator: '_')}", count: 1, default: news["document_type"]) if news["document_type"]
+        image_src = first_featured ? news["image"]["high_resolution_url"] : news["image"]["medium_resolution_url"]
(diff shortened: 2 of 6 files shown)


lauraghiorghisor-tw:
> Thanks for providing such a thorough PR description and roll out plan 👍 Hope the roll out goes smoothly! 🍀

Thank you so much! Had a quick look at smokey (which got us before on some csv urls), seems fine. Will merge this in first thing tomorrow, followed up by whitehall release 😎

PR #3459, opened 2023-12-05
Add GA4 tracking to browse action link

andysellick:

## What
- the action link appears on the [/browse/benefits](https://www.gov.uk/browse/benefits) page to signpost to a highlighted link
- adds GA4 tracking to this link

## Why
Part of the GA4 migration.

## Visual changes
None.

Trello card: https://trello.com/c/lstNprcf/717-add-tracking-to-action-link

### DIFF
--- a/app/views/browse/show.html.erb
+++ b/app/views/browse/show.html.erb
@@ -40,6 +40,16 @@
           href: "/check-benefits-financial-support",
           dark_large_icon: true,
           margin_bottom: 7,
+          data_attributes: {
+            module: "ga4-link-tracker",
+            ga4_track_links_only: "",
+            ga4_link: {
+              event_name: "navigation",
+              type: "action",
+              index_link: "1",
+              index_total: "1",
+            }
+          }
         } %>
       </div>
     </div>


AshGDS (approved):
@andysellick Will approve since the attributes aren't incorrect.

I've noticed that there's a small problem we have with wrapping non-links in the `ga4-link-tracker` attributes that we need to address - I've created a PR to fix it here: https://github.com/alphagov/govuk_publishing_components/pull/3748

You can see the issue if you click the SVG icon in the action link in this branch. 

I think you're right that changing it to `data-ga4-track-links-only` will fix it, so feel free to change it to that as well.

andysellick:
Thanks @AshGDS, great job noticing that. I've added the `data-ga4-track-links-only` attribute as well, which has fixed the problem here.

AshGDS (approved):
Thanks @andysellick 👍  think track-links-only just needs an empty string instead of true, but that's not a massive issue 👍

andysellick:
Ah yes good point, have fixed thanks @AshGDS

PR #3573, opened 2024-03-21
Add GA4 focus loss tracker to /world and /government/organisations

AshGDS:
## What
- Adds the `ga4-focus-loss-tracker` to `/world` and `/government/organisations` filter inputs

## Why
- https://trello.com/c/EXuAg9qR/801-add-tracking-filter-search-input

### DIFF
--- a/app/views/world/index.html.erb
+++ b/app/views/world/index.html.erb
@@ -18,7 +18,16 @@
           },
           name: "search-box",
           id: "filter-list",
-          controls: "search_results"
+          controls: "search_results",
+          data: {
+            module: "ga4-focus-loss-tracker",
+            ga4_focus_loss: {
+              event_name: "filter",
+              type: "filter",
+              action: "filter",
+              section: @presented_index.title
+            }
+          }
         } %>
       </form>
     </div>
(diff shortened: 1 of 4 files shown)


app/views/world/index.html.erb
```diff
@@ -18,7 +18,16 @@
...
           id: "filter-list",
-          controls: "search_results"
+          controls: "search_results",
+          data: {
+            module: "ga4-focus-loss-tracker",
+            ga4_focus_loss: {
+              event_name: "filter",
+              type: "filter",
+              action: "filter",
+              section: @presented_index.title
```
andysellick, line 28: Can we also force this to use the English translation string?

AshGDS: @andysellick I don't think so as it seems to come from the content item:

`app/controllers/world_controller.rb`

```ruby
class WorldController < ApplicationController
  def index
    index = WorldIndex.find!("/world")
    @presented_index = WorldIndexPresenter.new(index)
    setup_content_item_and_navigation_helpers(index)
  end
end
```

andysellick: Okay, thanks for checking 👍

andysellick (approved):

PR #3574, opened 2024-03-21
Add use_v1 to world_locations_news search links

richardTowers:
The v2 search api isn't finding any documents for queries like

https://www.gov.uk/search/all?order=updated-newest&world_locations%5B%5D=uk-delegation-to-council-of-europe

This is presumably because it doesn't have the same logic for indexing world locations as v1 did. We should probably fix that and remove the use_v1 flag, but for the short term this is a quick fix for the immediate issue.

This was raised by FCDO on Zendesk[1]

- [1] https://govuk.zendesk.com/agent/tickets/5664229

### DIFF
--- a/app/helpers/document_list_helper.rb
+++ b/app/helpers/document_list_helper.rb
@@ -44,22 +44,9 @@ private
   end
 
   def world_location_news_query_params(content_type, slug)
-    case content_type
-    when :latest
-      {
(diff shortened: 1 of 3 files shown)


richardTowers:
Better to put this logic in Finder Frontend, rather than exposing a debugging parameter in a link. We'll pick it up there.

PR #3577, opened 2024-03-22
Add GA4 tracking to 'organisation__supergroup' see all links

AshGDS:
## What
- Adds GA4 tracking to some "See all" links on organisation pages such as https://www.gov.uk/government/organisations/attorney-generals-office

## Why
- Was missed when the tracking was first added
- https://trello.com/c/53PguFQb/808-add-tracking-on-see-all-links-on-organisation-pages


app/presenters/organisations/supergroups_presenter.rb
AshGDS, line 1: It seems I need `ga4_english_title = ""` otherwise the app crashes, not sure if you have a better solution, things like `ga4_english_title: ga4_english_title || "",` crashed as well - my limited Ruby knowledge is probably showing here

andysellick: Have you tried?

```Ruby
I18n.with_locale(:en) do
  ga4_english_title = I18n.t(:title, scope: [:organisations, :document_types, supergroup.content_purpose_supergroup], lang: :en) || ""
end
```

If that doesn't work maybe a ternary operator would work, something like

```Ruby
ga4_english_title = ga4_english_title ? ga4_english_title : ""
```

AshGDS: @andysellick Thanks - I tried the first one and it didn't work, and I can't add a ternary into the return object so going with the second option would achieve the same thing with the same amount of lines we already have

I'm guessing it's because the variable is block scoped within `I18n.with_locale(:en) do`, so I have to have a definition of it outside the block to prevent it crashing

app/views/organisations/_latest_documents_by_supergroup.html.erb
```diff
@@ -19,7 +19,11 @@
             lang: supergroup[:lang]
           } %>
           <%= render "govuk_publishing_components/components/document_list", supergroup[:documents].merge({ ga4_extra_data: { section: t('organisations.document_types.documents', locale: :en)}}) %>
-          <a href="<%= supergroup[:finder_link][:path] %>" class="govuk-link brand__color" <%= supergroup[:finder_link][:lang] %>>
+          <a
+            href="<%= supergroup[:finder_link][:path] %>"
+            class="govuk-link brand__color" <%= supergroup[:finder_link][:lang] %>
```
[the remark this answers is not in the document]
AshGDS: Good spot 👍

spec/support/organisation_helpers.rb
```diff
@@ -940,4 +940,16 @@ def test_ga4_email_links(selector, section_heading = nil)
       expect(ga4_link_data["section"]).to eq section_heading
     end
   end
+
+  def test_ga4_organisation_supergroup_links(section_headings)
+    supergroups = page.all(".organisation__supergroup--margin-bottom")
```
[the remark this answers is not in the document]
AshGDS: Will update 👍

andysellick (approved):

PR #3584, opened 2024-03-28
Prevent organisation latest documents from including the org page

KludgeKML:
- Organisations with very few documents tend to end up with their own organisation page in their latest documents, which obviously just leads to circular links. Solve this by always asking for 4 latest documents, rejecting documents where the link matches the document base path, then taking the first three.

Fixes report in Zendesk ticket: https://govuk.zendesk.com/agent/tickets/5652462

##Before
Screenshot 2024-03-28 at 11 35 29

##After
Screenshot 2024-03-28 at 12 03 13

### DIFF
--- a/app/presenters/organisations/documents_presenter.rb
+++ b/app/presenters/organisations/documents_presenter.rb
@@ -46,9 +46,16 @@ module Organisations
     end
 
     def latest_documents
-      @latest_documents ||= Services.cached_search(
+      @latest_documents ||= latest_documents_excluding_org_page
+      search_results_to_documents(@latest_documents, @org)
+    end
+
+  private
+
+    def latest_documents_excluding_org_page
+      documents = Services.cached_search(
         {
-          count: 3,
+          count: 4,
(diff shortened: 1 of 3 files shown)


app/presenters/organisations/documents_presenter.rb
```diff
@@ -46,21 +46,26 @@ def has_latest_documents?
...
+          count: 4,
           order: "-public_timestamp",
           filter_organisations: @org.slug,
           fields: %w[title link content_store_document_type public_timestamp],
         },
         metric_key: "organisations.search.request_time",
       )["results"]
 
-      search_results_to_documents(@latest_documents, @org)
+      documents.reject { |doc| doc["link"] == @org.base_path }.take(3)
```
hannako, line 66: Do we need to add an additional step to make sure we're selecting the 3 most recent? eg if 4 documents are returned, and 0 of those documents are the organisation page itself.

hannako (approved):

PR #3605, opened 2024-05-01
Redesign No. 10 organisation page

AshGDS:
## What
- Redesign No. 10 organisation page

## Why
- https://trello.com/c/uXnyGiIa/76-frontend-changes-to-no10-page?filter=label:No%2010%20Org%20page

## Things for frontend to look at:

- [x] Latest news page
- [x] ~~The image card is always `large: true` when a YouTube video is present, causing a formatting issue, so component change is needed~~

## Things for backend to look at:
- [x] Latest news doesn't come with the right data for the image card component so ive hacked it for now.
    - Also I can't figure out how to remove the duplicate news item from the list to the right of it (doing .shift(), .drop() etc on the array doesn't work)
- [x] I can't change the colour of the contacts links as they're coming through from the backend, at `app/presenters/organisations/contacts_presenter.rb` line 68 (basically you need to remove brand__color if its the no 10 page) and I can't figure out how to do it as the `@org.is_no_10?` boolean doesn't work in that presenter
- [ ] The follow links rendering relies on `@what_we_do.has_share_links?`, so maybe they can't remove the 'what we do' section from Whitehall?
- [x] Question around tests - they seem to be complaining that I haven't added loads of translations for the new strings I've added. Is there a way to make the tests ignore them? Or should we be getting translations for these new strings?


app/views/organisations/_videos.html.erb
```diff
@@ -0,0 +1,27 @@
...
+      } %>
+    </div>
+  </div>
+
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+      <%= render "govuk_publishing_components/components/govspeak" do %>
+        <a href="<%= @documents.promotional_video_feature_youtube_url %>" class="govuk-link brand__color"><%= @documents.promotional_video_feature_youtube_url %></a>
+      <% end %>
+      <p class="govuk-heading-s govuk-!-margin-top-1"><%= @documents.promotional_video_feature_title %></p>
```
andysellick, line 17: Ideally the govuk margin override classes should only be used where overrides are necessary, but I appreciate that this work is time sensitive. Might be good to come back to these use of these classes later and write some CSS to support the layout here.

app/views/organisations/_videos.html.erb
```diff
@@ -0,0 +1,27 @@
...
+    </div>
+    <div class="govuk-grid-column-one-third govuk-!-margin-top-1">
+     <%= render "govuk_publishing_components/components/list",
+      extra_spacing: true,
+      items: @documents.promotional_video_feature_item_links(@organisation).map { |item| sanitize("<a href=\"#{item[:href]}\" rel=\"noopener\" class=\"govuk-link brand__color\">#{item[:text]}</a>") }
+      %>
+      <p class="govuk-body govuk-!-margin-top-7"><%= link_to(t("organisations.more_videos_link"), @organisation.more_videos_link, { class: "govuk-link brand__color" }) %></p>
```
andysellick, line 24: I think the current requirement is to remove this link. I'm reviewing this commit by commit, so it's possible it's addressed in a future commit, if so can you merge that change with this commit?

app/views/organisations/_latest_news.html.erb
```diff
@@ -0,0 +1,36 @@
...
+  </div>
+
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-one-half" data-module="ga4-link-tracker" data-ga4-link="<%= ga4_image_card_json %>" data-ga4-track-links-only>
+
+    <% latest_news_items = latest_news_items(@organisation, get_first_image: true) %>
+
+    <%= render "govuk_publishing_components/components/image_card", latest_news_items.shift %>
+
+    </div>
```
andysellick, line 27: Nitpick: indenting and extra spacing in the previous 5 lines seems off? I'd expect:

```
<div class="govuk-grid-column-one-half" data-module="ga4-link-tracker" data-ga4-link="<%= ga4_image_card_json %>" data-ga4-track-links-only>
  <% latest_news_items = latest_news_items(@organisation, get_first_image: true) %>
  <%= render "govuk_publishing_components/components/image_card", latest_news_items.shift %>
</div>
```

app/helpers/promotional_feature_helper.rb
```diff
@@ -0,0 +1,51 @@
+module PromotionalFeatureHelper
```
andysellick, line 1: This new file doesn't appear to have any tests - can some be added?

app/helpers/latest_news_helper.rb
```diff
@@ -0,0 +1,39 @@
+module LatestNewsHelper
```
andysellick, line 1: This new file doesn't seem to have any tests - can some be added?

andysellick (commented):
Some minor comments, mainly noting the lack of tests in some areas, but I appreciate that not everything might be testable, or covered by other kinds of tests, so happy to discuss.

app/views/organisations/_videos.html.erb
AshGDS: @andysellick I didn't know this - thanks for letting me know!

app/views/organisations/_videos.html.erb
AshGDS: @andysellick I can't find this requirement - I remember in the last meeting I was confused about a requirement that was added to remove "Summary text" called `Latest videos from the 10 Downing Street YouTube channel` - but I had never put that in the UI at all. In that same meeting I was asked to add spacing to this link, so I'm not sure this link would need to be removed?

But yep, Keith has been squashing commits 👍

app/views/organisations/_latest_news.html.erb
AshGDS: @andysellick Thanks, will fix

app/helpers/promotional_feature_helper.rb
AshGDS: I think @KludgeKML would be best placed for this as he wrote the code for this - but if he's busy I can try

app/helpers/latest_news_helper.rb
AshGDS: I think @KludgeKML would be best placed for this as he wrote the code for this as well - but if he's busy I can try

app/views/organisations/_latest_news.html.erb
AshGDS: I've pushed it as a separate commit as the first time I did it it got overwritten by accident, will squash later

KludgeKML: Should be okay now.

app/helpers/promotional_feature_helper.rb
KludgeKML: Added

andysellick (approved):
Frontend code looks good to me 👏

PR #3619, opened 2024-05-16
Ensure organisation contact sections use 19px font size

AshGDS:
## What
- Ensure organisation contact sections have our 19px font size
- I did not use `govuk-body` as that adds margin and there was already margin added to the elements, so I didn't want to affect that. Therefore I used a new `organisation__paragraph` class where necessary.

## Why
- https://trello.com/c/UHOaQNIl/129-contact-description-text-on-organisation-pages-is-too-small-m

### Before

image

### After

image

### DIFF
--- a/app/views/organisations/_contacts.html.erb
+++ b/app/views/organisations/_contacts.html.erb
@@ -27,13 +27,14 @@
           <% if contact[:email_addresses].any? %>
             <div class="organisation__margin-bottom">
               <% if I18n.exists?('organisations.contact.email', contact[:locale]) %>
-                <p><%= t('organisations.contact.email') %></p>
+                <p class="organisation__paragraph"><%= t('organisations.contact.email') %></p>
               <% else %>
-                <p lang="en"><%= t('organisations.contact.email') %></p>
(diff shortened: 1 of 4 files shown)


app/views/organisations/_contacts.html.erb
```diff
@@ -25,7 +25,7 @@
       <% if contact[:email_addresses].any? || contact[:links].any? || contact[:phone_numbers].any?  %>
         <div class="<%= contact[:post_addresses].any? ? "govuk-grid-column-one-half" : "govuk-grid-column-two-thirds" %>">
           <% if contact[:email_addresses].any? %>
-            <div class="organisation__margin-bottom">
+            <div class="organisation__margin-bottom organisation__paragraph">
```
andysellick, line 28: Feels a bit confusing to have a class name of `__paragraph` on a div element, could this be moved onto the `<p>` tag inside the element instead?

andysellick (approved):
Looks good, good approach. One small comment but non-blocking. Thanks for looking into this.

app/views/organisations/_contacts.html.erb
AshGDS: Good point, will do 👍

PR #3644, opened 2024-06-05
Conditionally style browse/covid page borders when global bar exists

AshGDS:
## What/Why
- The `/coronavirus` and `/browse` pages have custom border styles
- These borders clash with the globar bar when it is rendered
- Therefore, a `global-bar-present` class was added to the `<body>` whenever the global bar is rendered: https://github.com/alphagov/govuk_publishing_components/pull/4047
- We use that class in this app to conditionally style the borders on  `/coronavirus` and `/browse`

## Testing
- You can test this by:
    - Navigating to http://127.0.0.1:3070/browse and http://127.0.0.1:3070/coronavirus
    - Seeing that the border styles look correct with the global bar
    - Deleting the global bar from the DOM, and the `global-bar-present` class from the `<body>`, and seeing the border styles on these pages come back

### Coronavirus page
image

### Browse page
image

### Coronavirus page
image

### Browse page
image

### DIFF
--- a/app/assets/stylesheets/views/_browse.scss
+++ b/app/assets/stylesheets/views/_browse.scss
@@ -26,6 +26,17 @@ $browse-header-background-colour: #263135;
   overflow: auto;
 }
 
+.browse__breadcrumb-border {
+  background: govuk-colour("blue");
+  height: 10px;
+}
+
+.global-bar-present {
+  .browse__breadcrumb-border {
+    height: 0;
+  }
+}
+
 // The header spacings are a work in progress.
 .browse__header-wrapper {
   padding-bottom: govuk-spacing(7);
--- a/app/views/shared/_browse_breadcrumbs.html.erb
+++ b/app/views/shared/_browse_breadcrumbs.html.erb
@@ -7,6 +7,7 @@
 
 <div class="browse__breadcrumb-wrapper">
   <div class="govuk-width-container">
+    <div class="browse__breadcrumb-border"></div>
     <div class="govuk-grid-row">
       <div class="<%= column_classes.join(' ') %>">
         <%= render "application/breadcrumbs", {
--- a/app/assets/stylesheets/views/_covid.scss
+++ b/app/assets/stylesheets/views/_covid.scss
@@ -13,21 +13,28 @@ $c19-landing-page-header-background: govuk-colour("blue");
   }
 }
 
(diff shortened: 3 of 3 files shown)


app/assets/stylesheets/views/_browse.scss
```diff
@@ -26,6 +26,17 @@ $browse-header-background-colour: #263135;
   overflow: auto;
 }
 
+.browse__breadcrumb-border {
+  background: govuk-colour("blue");;
```
andysellick, line 30: Nitpick: need to remove the extra semi-colon.

app/assets/stylesheets/views/_browse.scss
```diff
@@ -26,6 +26,17 @@ $browse-header-background-colour: #263135;
...
+  height: 10px;
```
andysellick, line 31: My instinct with this kind of thing would be to use a border rather than setting the height and the background, but I think that's grounded in older browsers where regardless of content they'd often set a minimum height on elements based on the font size, so it's probably not a concern here. Interested in your thoughts though.

app/views/shared/_browse_breadcrumbs.html.erb
```diff
@@ -7,6 +7,7 @@
 
 <div class="browse__breadcrumb-wrapper">
   <div class="govuk-width-container">
+    <div class="browse__breadcrumb-border"></div>
```
andysellick, line 10: Shame to have to add another element here just for the border but I can't think of another solution, this seems sensible 👍

andysellick (approved):
I have a few comments, but only the extra semi-colon is a problem, so happy to approve.

app/assets/stylesheets/views/_browse.scss
AshGDS: @andysellick Thanks - I've taken all of this from existing CSS, so as to not rock the boat I didn't change it but happy to if you think it's worth it 👍

app/assets/stylesheets/views/_browse.scss
AshGDS: Thanks, will do!

app/views/shared/_browse_breadcrumbs.html.erb
AshGDS: Yep, this was the existing functionality so I assume it was the best solution for the requirement.

PR #3648, opened 2024-06-06
Upgrade to govuk frontend 5.1

MartinJJones:
## What

- Move components that rely on govuk-frontend modules to separate `es6-components.js` file
- Remove Sass variables
- Update polyfill file paths

### Move components that rely on govuk-frontend modules to seperate `es6-components.js` file

In the event that a browser below the target for `govuk-frontend` loads a page with JS on it, attempting to parse the JS from `govuk-frontend` will cause an error. To avoid this from happening, JS that contains `govuk-frontend` JS has been moved to seperate file which will be loaded in a script tag with `type="module"`. This will prevent the JS from being parsed and so prevent the error

### Remove Sass variables

- The `$legacy` attribute in `govuk-colour` has been deprecated and using it will have no effect (other than generating warnings on pre-compilation)

### Update polyfill file paths
Browser polyfills are no longer included in v5 of govuk-frontend, any Polyfills provided in v4 of govuk-frontend will be included in the new version of the govuk-publishing-components gem.

The intention is for this PR to include the bump to the govuk-frontend v5 version of the govuk-publishing-components gem, this will also fix the failing tests relating to the missing polyfill.

[Trello](https://trello.com/c/QtJuwSIH/2337-upgrade-collections-to-run-on-v51-of-govuk-frontend)

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -216,7 +216,7 @@ GEM
     govuk_personalisation (0.16.0)
       plek (>= 1.9.0)
       rails (>= 6, < 8)
-    govuk_publishing_components (39.2.5)
+    govuk_publishing_components (40.0.0)
       govuk_app_config
       govuk_personalisation (>= 0.7.0)
       kramdown
--- a/app/assets/config/manifest.js
+++ b/app/assets/config/manifest.js
@@ -1,5 +1,6 @@
 //= link_tree ../images
 //= link application.js
+//= link es6-components.js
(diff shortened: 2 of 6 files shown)


hannalaakso (commented):
Changes look good to me 👍 As discussed with Martin, I'm commenting instead of approving as we still have some failing tests on the PR. Once the new gem is released, these tests should pass and the PR can be approved at that point.

hannalaakso (approved):

PR #3673, opened 2024-06-26
Add guards to fix preview of Ministers page during a reshuffle

minhngocd:
Currently, the rendering view assumes there's only 2 views of the page - a fully assigned ministers index, or a banner that is displayed in case of a reshuffle. In reshuffle mode preview, however, means that the if statement in the view assumes wrongly about the data, since the reshuffle flag / message is not passed for the preview. We should cater for these cases during reshuffle where there could be empty lists of ministers.

PR required for tests to pass: https://github.com/alphagov/publishing-api/pull/2789

https://trello.com/c/iSsdS462/2697-fix-frontend-issues-preventing-preview-of-ministers-index-page

### DIFF
--- a/app/presenters/ministers_index_presenter.rb
+++ b/app/presenters/ministers_index_presenter.rb
@@ -16,19 +16,22 @@ class MinistersIndexPresenter
   end
 
   def cabinet_ministers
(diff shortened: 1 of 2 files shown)


ChrisBAshton (approved):
🚀 (Assuming the publishing API PR causes these tests to pass)

PR #3703, opened 2024-07-17
Use type=module for application.js

AshGDS:
## What / Why
- Use type=module for this application's JS file
- To coincide with https://github.com/alphagov/govuk_publishing_components/pull/4111
- Has [DO NOT MERGE] as we want the above PR live on `static` first before this is merged.
- We will combine `es6-components.js` again as a separate task unless you disagree with this
- To accurately test this you will need to use local static that is using a local version the `govuk_publishing_components` branch linked above. If you test without, this app's `application.js` seems to crash.
- Trello card: https://trello.com/c/Hoa45SMD/141-turn-off-javascript-in-legacy-browsers

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -36,7 +36,7 @@
 <head>
   <title><%= yield :title %> - <%= t('application.title.suffix')%></title>
   <%= javascript_include_tag "test-dependencies" if Rails.env.test? %>
-  <%= javascript_include_tag "application" %>
+  <%= javascript_include_tag "application", type: "module" %>
   <%= javascript_include_tag 'es6-components', type: "module" %>
   <%= csrf_meta_tags %>
   <%= yield :meta_tags %>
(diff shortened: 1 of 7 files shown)


MartinJJones (approved):

MartinJJones (commented):
The code changes look to me 👍 

The JS error in the browser relates to the closest polyfill, further info here - https://github.com/alphagov/smart-answers/pull/6846#pullrequestreview-2198973334

I'm happy to give this another look and approve when the new version of the gem is released and included in this PR as this should then resolve the issue.

app/views/layouts/application.html.erb
```diff
@@ -36,7 +36,7 @@
 <head>
   <title><%= yield :title %> - <%= t('application.title.suffix')%></title>
   <%= javascript_include_tag "test-dependencies" if Rails.env.test? %>
```
MartinJJones, line 38: I wonder if we should also add type="module" here as well. It appears that test-dependencies.js is used to pull in [JS used by static to replicate the approach used on dev/production](https://github.com/alphagov/frontend/blob/main/app/assets/javascripts/test-dependencies.js#L1-L2), not 100% sure if this would actually make any difference to the tests, but would just align the approach used (to be used) more closely.

MartinJJones (approved):

PR #3726, opened 2024-08-06
Do not force analytics labels for popular links to be in English and refactor browse helper

unoduetre:

## What

In preparation for adding the AB test configuration to collections, we can refactor the way the code is currently organised.

In addition to that, do not force the analytics labels of popular links to always be in English.

## Why

[Trello card](https://trello.com/c/SQT21D91/2761-refactor-browsehelper-in-preparation-for-ab-test-m)

### DIFF
--- a/app/helpers/browse_helper.rb
+++ b/app/helpers/browse_helper.rb
@@ -1,28 +1,27 @@
 module BrowseHelper
-  ACTION_LINK_DATA = {
-    benefits: [
-      { lang: "browse.check_benefits_and_financial_support", href: "/check-benefits-financial-support" },
-    ],
-    business: [
-      { lang: "browse.hmrc_online_services", href: "/log-in-register-hmrc-online-services" },
-      { lang: "browse.self_assessment_tax_returns", href: "/self-assessment-tax-returns" },
-      { lang: "browse.pay_employers_paye", href: "/pay-paye-tax" },
-    ],
-  }.freeze
-
-  def display_action_links_for_slug?(slug)
-    ACTION_LINK_DATA.key?(slug.to_sym)
+  def display_popular_links_for_slug?(slug)
+    I18n.exists?(slug.to_s, scope: "browse.popular_links")
   end
 
-  def action_link_data(slug)
-    links = ACTION_LINK_DATA[slug.to_sym]
-    return [] unless links
-
-    links.each_with_index do |link, index|
-      link[:text] = I18n.t(link[:lang])
-      link[:ga4_text] = I18n.t(link[:lang], locale: :en)
-      link[:index_link] = index + 1
-      link[:index_total] = links.length
+  def popular_links_for_slug(slug)
+    links = I18n.t(slug.to_s, scope: "browse.popular_links")
+    count = links.length
+    links.map.with_index(1) do |link, index|
+      {
+        text: link[:title],
+        href: link[:url],
+        data_attributes: {
+          module: "ga4-link-tracker",
+          ga4_track_links_only: "",
+          ga4_link: {
+            event_name: "navigation",
+            type: "action",
+            index_link: index,
+            index_total: count,
+            text: link[:title],
+          },
+        },
+      }
     end
   end
 end
--- a/app/views/browse/show.html.erb
+++ b/app/views/browse/show.html.erb
@@ -34,7 +34,7 @@
   } %>
 <% end %>
 
-<% if display_action_links_for_slug?(page.slug) %>
+<% if display_popular_links_for_slug?(page.slug) %>
   <div class="govuk-width-container">
     <div class="govuk-grid-row">
       <div class="govuk-grid-column-full">
@@ -45,9 +45,9 @@
             font_size: "m"
           } %>
           <ul class="govuk-list govuk-!-margin-bottom-7">
-            <% action_link_data(page.slug).each do |link| %>
+            <% popular_links_for_slug(page.slug).each do |link| %>
               <li>
-                <%= render "shared/browse_action_link", locals: link %>
+                <%= render partial: "shared/browse_action_link", locals: {link:} %>
               </li>
             <% end %>
           </ul>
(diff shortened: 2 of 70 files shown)


config/locales/en/browse.yml
```diff
@@ -4,9 +4,25 @@ en:
...
+        - title: "HMRC online services: sign in or set up an account"
+          url: /log-in-register-hmrc-online-services
+          analytics_label_key: hmrc_online_services
+        - title: Self Assessment tax returns
+          url: /self-assessment-tax-returns
+          analytics_label_key: self_assessment_tax_returns
+        - title: "Pay employers' PAYE"
+          url: /pay-paye-tax
+          analytics_label_key: pay_employers_paye
+    analytics_labels:
```
hannako, line 24: I think we could simplify here. The `analytics_label_key` is storing a reference to a field in `analytics_labels` which just stores the link title. But you've already got that string stored in the `title` field. So let's just use the `title` field and remove the duplication.

ie remove all the `analytics_label_key`'s and the `analytics_labels` hash. We only need two keys:`title` and `url`.

app/helpers/browse_helper.rb
```diff
@@ -1,28 +1,14 @@
...
-      link[:text] = I18n.t(link[:lang])
-      link[:ga4_text] = I18n.t(link[:lang], locale: :en)
-      link[:index_link] = index + 1
-      link[:index_total] = links.length
+  def each_popular_link_for_slug(slug)
+    links = I18n.t(slug.to_s, scope: "browse.popular_links")
+    count = links.length
+    links.each.with_index(1) do |link, index|
+      analytics_label = I18n.t(link[:analytics_label_key], scope: "browse.analytics_labels", locale: :en)
+      yield link, analytics_label, index, count
```
hannako, line 11: It looks like the action link component accepts a `data_attributes` option (docs [here](https://components.publishing.service.gov.uk/component-guide/action_link))

So I think it might be clearer to just pass a single `data_attributes` key to the view that contains all the values,  rather than having elements of the data attributes hash defined in different files. Something like:

```
def popular_link_for_slug(slug)
  links = I18n.t(slug.to_s, scope: "browse.popular_links")
  count = links.length
  links.map.with_index(1) do |link, index|
    {
      text: link[:title],
      href: link[:url],
      data_attributes: {
       module: "ga4-link-tracker",
       ga4_track_links_only: "",
       ga4_link: {
          event_name: "navigation",
          type: "action",
          index_link: index,
          index_total: count,
          text: link[:title],
        },
      }
    }
  end
end
```

What do you think? Similar to [this](https://github.com/alphagov/collections/blob/main/app/presenters/taxon_organisations_presenter.rb#L62-L83)

app/views/browse/show.html.erb
```diff
@@ -45,9 +45,12 @@
...
           <ul class="govuk-list govuk-!-margin-bottom-7">
-            <% action_link_data(page.slug).each do |link| %>
-              <li>
-                <%= render "shared/browse_action_link", locals: link %>
+            <% each_popular_link_for_slug(page.slug) do |link, analytics_label, link_index, link_count| %>
+              <li
+                  data-module="ga4-link-tracker"
+                  data-ga4-track-links-only=""
+                  data-ga4-link='{"event_name":"navigation","type":"action","index_link":<%= link_index %>,"index_total":<%= link_count %>,"text":"<%= analytics_label %>"}'>
+                <%= render partial: "shared/browse_action_link", locals: {link:} %>
```
hannako, line 53: If we create a `link` with a `data attributes` key this goes back to `<%= render partial: "shared/browse_action_link", locals: {link:} %>`

app/views/shared/_browse_action_link.erb
```diff
@@ -1,17 +1,6 @@
 <%= render "govuk_publishing_components/components/action_link", {
-  text: locals[:text],
-  href: locals[:href],
+  text: link[:title],
+  href: link[:url],
```
hannako, line 3: And then we just pass in `data_attributes: link[:data_attributes],`

hannako (commented):
Thanks for your work on this Matt - just one main suggestion to keep all the data attribute values in a single place - hopefully makes sense, let me know if not.

app/helpers/browse_helper.rb
unoduetre: The requirement of the ticket was:

`Move the[ data attributes](https://github.com/alphagov/collections/blob/main/app/views/browse/show.html.erb#L34-L51) for tracking into the view`

Has the requirement changed?

hannako: The intention of the requirement was to have all the data attributes defined in one place. Rather than some being defined in the helper, and some being defined in the view.

 But as there's a pattern already used in this app of passing through the `data_attributes` key I think this is a nicer way to meet the objective of all in one place.

unoduetre: I'll also remove the partial, as it's no longer necessary, as it only contains one line and is used in one place only.

unoduetre: Changed.

app/views/browse/show.html.erb
unoduetre: I've removed the partial as it's no longer necessary.

app/views/shared/_browse_action_link.erb
unoduetre: Please, see my other replies.

config/locales/en/browse.yml
unoduetre: Removed.

app/views/browse/show.html.erb
unoduetre: Please, see my other comments.

config/locales/ar/browse.yml
```diff
@@ -1,12 +1,19 @@
 ---
 ar:
   browse:
+    popular_links:
+      benefits:
+        - title:
+          url: /check-benefits-financial-support
```
hannako, line 7: Is there any point adding the URL in all these places? I think it will just default to the `en` locale if the values are missing? So I'd be inclined to delete them to avoid confusion.

hannako:
@andysellick We're doing a bit of a refactor before we add some AB test logic. I've checked the tracking isn't impacted by comparing the review app with live, but would you mind giving it a quick once over as well just from the tracking POV to make sure we've not broken anything? 

Browse business
- [review app](https://collections-pr-3726.herokuapp.com/browse/business)
- [live](https://www.gov.uk/browse/business)

Browse benefits
- [review app](https://collections-pr-3726.herokuapp.com/browse/benefits)
- [live](https://www.gov.uk/browse/benefits)

config/locales/ar/browse.yml
unoduetre: It only makes sense, if we use translations, but as discussed earlier that's unlikely, so I'll remove them.

hannako (approved):
Looks good, thank you!

PR #3736, opened 2024-08-14
Remove world and world news atom feeds

theseanything:
We no longer support atom feeds for this content. The link to atom feeds were removed in this [PR](https://github.com/alphagov/collections/pull/3238).

Another issue has meant that these routes have been inaccessible and returning 404 anyway.


hannako (approved):
Thanks Sean this looks good.

PR #3742, opened 2024-08-21
[DO NOT MERGE] Display GOV.UK Chat promo on certain pages

jackbot:
We want to show a promo to the new GOV.UK Chat application on certain pages on GOV.UK. It's a CTA pointing users to the chat app.

The list of URLs we want the promo to appear on is quite short and covers only L1 browse, L2 browse and step_by_step  content types from this codebase.

Following are screenshots of each page type on desktop and mobile, showing the before/after states.

### L1 Browse

| Before  | After |
| ------------- | ------------- |
| l1_browse_desktop_before | l1_browse_desktop_after |
| l1_browse_mobile_before | l1_browse_mobile_after |

### L2 browse

| Before  | After |
| ------------- | ------------- |
| l2_browse_desktop_before | l2_browse_desktop_after |
| l2_browse_mobile_before | l2_browse_mobile_after |

### step_by_step

| Before  | After |
| ------------- | ------------- |
| sbs_desktop_before |sbs_desktop_after |
| sbs_mobile_before | sbs_mobile_after |

### DIFF
--- a/app/assets/stylesheets/views/_browse.scss
+++ b/app/assets/stylesheets/views/_browse.scss
@@ -93,3 +93,16 @@ $browse-header-background-colour: #263135;
     margin-bottom: govuk-spacing(6);
   }
 }
+
+.browse__govuk-chat-promo {
+  @include govuk-media-query($until: "tablet") {
+    border-top: 4px solid govuk-colour("blue");
+    padding-top: govuk-spacing(5);
+  }
+}
+
+.browse__govuk-chat-promo--second-level {
+  @include govuk-media-query($until: "tablet") {
+    border-top-width: 2px;
+  }
+}
--- a/app/helpers/govuk_chat_promo_helper.rb
+++ b/app/helpers/govuk_chat_promo_helper.rb
@@ -1,0 +1,14 @@
+module GovukChatPromoHelper
+  GOVUK_CHAT_PROMO_BASE_PATHS = %w[
+    /browse/business
+    /browse/business/business-tax
+    /browse/business/setting-up
+    /browse/tax/self-assessment
+    /set-up-limited-company
+    /set-up-as-sole-trader
+  ].freeze
+
+  def show_govuk_chat_promo?(base_path)
+    ENV["GOVUK_CHAT_PROMO_ENABLED"] == "true" && GOVUK_CHAT_PROMO_BASE_PATHS.include?(base_path)
+  end
+end
--- a/app/views/browse/show.html.erb
+++ b/app/views/browse/show.html.erb
@@ -37,9 +37,9 @@
 
 <% if display_popular_links_for_slug?(page.slug) %>
   <div class="govuk-width-container">
-    <div class="govuk-grid-row">
-      <div class="govuk-grid-column-full">
-        <div class="browse__action-links">
+    <div class="browse__action-links">
+      <div class="govuk-grid-row">
+        <div class="<%= show_govuk_chat_promo?(page.base_path) ? "govuk-grid-column-one-half govuk-grid-column-two-thirds-from-desktop" : "govuk-grid-column-full" %>">
           <%= render "govuk_publishing_components/components/heading", {
             text: t("browse.popular_tasks"),
             margin_bottom: 6,
@@ -53,6 +53,14 @@
             <% end %>
           </ul>
         </div>
+
+        <% if show_govuk_chat_promo?(page.base_path) %>
+          <div class="govuk-grid-column-one-half govuk-grid-column-one-third-from-desktop">
+            <div class="browse__govuk-chat-promo">
+              <%= render "govuk_publishing_components/components/chat_entry" %>
+            </div>
+          </div>
+        <% end %>
       </div>
     </div>
   </div>
--- a/app/views/second_level_browse_page/show_curated.html.erb
+++ b/app/views/second_level_browse_page/show_curated.html.erb
@@ -34,8 +34,16 @@
 
 <div class="govuk-width-container">
   <div class="govuk-grid-row">
-    <div class="govuk-grid-column-two-thirds" data-module="ga4-link-tracker">
+    <div class="govuk-grid-column-one-half govuk-grid-column-two-thirds-from-desktop" data-module="ga4-link-tracker">
       <%= render partial: curated_partial, locals: { page: page } %>
     </div>
+
+    <% if show_govuk_chat_promo?(page.base_path) %>
+      <div class="govuk-grid-column-one-half govuk-grid-column-one-third-from-desktop">
+        <div class="browse__govuk-chat-promo browse__govuk-chat-promo--second-level">
+          <%= render "govuk_publishing_components/components/chat_entry" %>
+        </div>
+      </div>
+    <% end %>
   </div>
 </div>
--- a/app/views/step_nav/show.html.erb
+++ b/app/views/step_nav/show.html.erb
@@ -4,19 +4,27 @@
   <meta name="description" content="<%= step_by_step.description %>">
 <% end %>
 <div class="govuk-grid-row">
-  <div class="govuk-grid-column-two-thirds">
+  <div class="govuk-grid-column-one-half govuk-grid-column-two-thirds-from-desktop">
     <%= render 'govuk_publishing_components/components/title',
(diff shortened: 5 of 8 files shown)


kevindew:
Nice, this looks great - does look a bit more painful than government-frontend but seems necessary.

Yeah, it would seem strange to only test step_by_step and not browse. I imagine we'd probably just skip the step_by_step test if we can't find a reasonable place. 

Feature test wise, could probably go in: [mainstream_browsing_spec.rb](https://github.com/alphagov/collections/blob/main/spec/features/mainstream_browsing_spec.rb), although it's quite a different feature test than the step_by_step one.

Otherwise there's the controller tests, which would probably seem more semantically correct than a feature test, where it looks like there is ok precedent to do this type of test in the [browse_controller_spec.rb](https://github.com/alphagov/collections/blob/main/spec/controllers/browse_controller_spec.rb) file, but no precedence of anything beyond status code testing in [step_nav_controller_spec](https://github.com/alphagov/collections/blob/main/spec/controllers/step_nav_controller_spec.rb).

I imagine the worst case would be to have one document type tested by a feature test and the other by a controller test, so it's probably the route of least surprise to stick with feature test for both.

app/assets/stylesheets/views/_browse.scss
```diff
@@ -92,4 +92,18 @@ $browse-header-background-colour: #263135;
...
+    border-top: 4px solid govuk-colour("blue");
+    padding-top: govuk-spacing(5);
+  }
+}
+
+.browse__govuk-chat-promo .gem-c-chat-entry {
+  @include govuk-media-query($until: "tablet") {
+    border-top: 2px solid govuk-colour("blue");
+    padding-top: govuk-spacing(5);
+  }
```
kevindew, line 108: I think we are committing a sin by modifying the component directly because of the tight coupling this produces. (I tried to find a doc but best I got was: https://github.com/alphagov/govuk_publishing_components/blob/main/docs/component_principles.md which doesn't cover this point precisely).

I do think it makes sense to be doing this styling change in this app rather than the component though since this is specific to the browse page and wouldn't make sense decoupled.

I think instead we should use a wrapper div to apply the styles e.g:

```
.browse__govuk-chat-promo {
  @include govuk-media-query($until: "tablet") {
    border-top: 2px solid govuk-colour("blue");
    padding-top: govuk-spacing(5);
  }
```

and then we change the markup for both of the additions for HTML i.e:

```
<% if show_govuk_chat_promo?(page.base_path) %>
    <div class="browse__govuk-chat-promo">
      <%= render "govuk_publishing_components/components/chat_entry" %>
    </div>
  </div>
<% end %>
```

I expect we could use that same class for both browse usages.

kevindew (commented):
Great job, just a suggested change for the CSS and then I think we'll be good to reach out for a team member review

app/assets/stylesheets/views/_browse.scss
jackbot: Thanks @kevindew, appreciate the review.

I've actioned your point on not targeting the component directly and instead added the wrapper classes.

Unfortunately the two pages need different border widths - one 2px and one 4px - so I've had to keep in the two class names.

If we apply the same 2px border to both, the L1 browse page looks off, with a 2px border at the top and a 4px border at the bottom:

Screenshot 2024-08-29 at 15 20 08

kevindew: Argh 🙈 did not spot that minor difference. Thanks for spotting.

Change seems good, will have another suggestion though.

app/assets/stylesheets/views/_browse.scss
```diff
@@ -92,4 +92,18 @@ $browse-header-background-colour: #263135;
...
+}
+
+.browse__govuk-chat-promo {
+  @include govuk-media-query($until: "tablet") {
+    border-top: 4px solid govuk-colour("blue");
+    padding-top: govuk-spacing(5);
+  }
+}
+
+.browse_second_level__govuk-chat-promo {
```
kevindew, line 104: This class doesn't seem consistent with GOV.UK BEM as we don't have precedence for `.browse_second_level`

I think the most appropriate thing would be to have this as a modifier. So something like:

```
.browse__govuk-chat-promo {
  @include govuk-media-query($until: "tablet") {
    border-top: 4px solid govuk-colour("blue");
    padding-top: govuk-spacing(5);
  }
}

.browse__govuk-chat-promo--second-level {
  @include govuk-media-query($until: "tablet") {
    border-top-width: 2px;
  }
}
```

jackbot: Sounds good, I've updated the original commit.

kevindew (commented):
Great thank you @jackbot ⭐ 

Assuming the failing test is just something minor, I think this is good to be reviewed by the team.

leenagupte:
I'll take a look, but I'll tag @AshGDS and @andysellick for a review from a frontend perspective.

app/views/browse/show.html.erb
```diff
@@ -40,18 +40,30 @@
     <div class="govuk-grid-row">
       <div class="govuk-grid-column-full">
```
leenagupte, line 41: I think these two lines can be removed as they have been duplicated lines 43 and 44 and I don't think these classes need to be added twice?

app/helpers/govuk_chat_promo_helper.rb
```diff
@@ -0,0 +1,14 @@
+module GovukChatPromoHelper
+  GOVUK_CHAT_PROMO_BASE_PATHS = %w[
+    /browse/business
+    /browse/business/business-tax
+    /browse/business/setting-up
+    /browse/tax/self-assessment
+    /set-up-limited-company
+    /set-up-self-employed
```
leenagupte, line 8: In production this path redirects to https://www.gov.uk/set-up-as-sole-trader, so you should change this path to match. It's still a step-by-step.

leenagupte (commented):
I think the backend code looks fine. It's great that the helper follows the same pattern as the one in government-frontend. 🎉 

The only concern I have is about how intertwined the chat promo is with the popular links code is in the level one browse view. 
After looking at different ways to decouple them, all I can recommend is removing the duplicated divs. Still, I think that will make it look a little less gnarly.

I'll leave reviewing the new css classes to people who have a better idea of what the best practice should be.

As in the previous PR, I've recreated the [review app](https://collections-pr-3742.herokuapp.com/browse/business) and added the environment variable.

app/helpers/govuk_chat_promo_helper.rb
jackbot: Thanks @leenagupte, I've updated the URL now.

app/views/browse/show.html.erb
jackbot: Good spot! I've removed these two lines.

jon-kirwan:
> I'm seeing a horizontal scrollbar when resizing the page with this component. This is likely because I have scroll bars always enabled on my Mac, which some devs don't have enabled, so it's easy to miss. Windows machines have scrollbars enabled by default, so I think this should be fixed.

Thanks @AshGDS. Interesting issue. I've found it difficult to fix in the component itself (the issue seems to be with the word 'experimental' when space gets limited). Also, I don't think we can stack the elements since the designs have the logo and text side by side on mobile.

One option might be too modify the grid layout classes on this page e.g.
```
<div class="govuk-grid-row">
  <div class="govuk-grid-column-two-thirds-from-desktop"></div>
  <div class="govuk-grid-column-one-third-from-desktop"></div>
</div>
```
The media query on `browse__govuk-chat-promo` would also need to be changed.

Another option might be to use the `govuk-grid-column-one-half govuk-grid-column-two-thirds-from-desktop` grid classes together:
```
<div class="govuk-grid-row">
  <div class="govuk-grid-column-one-half govuk-grid-column-two-thirds-from-desktop"></div>
  <div class="govuk-grid-column-one-half govuk-grid-column-one-third-from-desktop"></div>
</div>
```

AshGDS:
Thanks @jon-kirwan, good solution - changing the grid class sounds good from my end. 👍

AshGDS:
Thanks @jackbot - I can't see the resizing issue anymore 👍

AshGDS (approved):
LGTM 👍

PR #3761, opened 2024-09-06
Spike generate popular tasks using BigQuery

beccapearce:
Trello card: https://trello.com/c/AGRVqS4S

This spike sets up a service that connects to BigQuery 

To run this locally you will need to set up some config following the steps in the [dev docs](https://docs.publishing.service.gov.uk/repos/content-data-api/google_analytics_setup.html). 

It collects the data from big query and outputs it on the benefits and business pages.

There is an initial implementation of caching and back up caching. This needs to be amended so it will run locally.

The links are currently displayed as a slug, this will need to be changed to show a title.

Still need to add a commit for the sentry logging.

| Before | After |
|--------|--------|
| image | image |


app/services/popular_tasks.rb
```diff
@@ -0,0 +1,57 @@
...
+      group by click,event_date,cleaned_page_referrer,search_term,link_url
+      order by cleaned_page_referrer,Rank asc)
+
+      SELECT
+      *
+      FROM CTE2
+      WHERE Rank <6
+    SQL
+
+    data = @fetch_data.query(query).all
```
hannako: How many results does this query fetch?

app/services/popular_tasks.rb
```diff
@@ -1,5 +1,6 @@
 class PopularTasks
   CACHE_EXPIRATION = 24.hours # Set the cache expiration time
+  BACKUP_CACHE_EXPIRATION = 7.days # Backup cache can have a longer expiration
```
hannako, line 3: I think we need to think a bit more about how this would work.

If the bigquery data was unavailable for more than 7 days then what happens? 

I can think of other ways to do it - but this feels like a problem that must have been solved many times before. i.e. Only expire the cache if fresh data is available to fill it.

hannako: I've caught up now... the cache will expire regardless of whether or not the API responds so I understand the need for a backup. And I like the idea of writing to the backup at the same time as you fetch the fresh data.

hannako:
Some things to consider for next steps (which I've started thinking about [here](https://github.com/alphagov/collections/pull/3773)

- Error handing. What happens if the api returns an error
- Some rake tasks so we can manually expire/refresh the cache if needed
- Where are we going to add logging and what kind
- Is memcached the best option?

app/services/popular_tasks.rb
beccapearce: this is the top 5 we can increase or lower it by changing the ranking here:
``` WHERE Rank <6```

PR #3763, opened 2024-09-09
Fix breadcrumbs a11y issue in the covid page

unoduetre:

## What

collections places breadcrumbs inside the <main> container in app/views/coronavirus_landing_page/components/landing_page/_page_header.html.erb, creating a WCAG fail on the [covid landing page](https://www.gov.uk/coronavirus).

## Why

Having the breadcrumbs inside the main container is a WCAG fail.

[Trello ticket](https://trello.com/c/goKwXDWK/2822-fix-breadcrumbs-wcag-fail-in-collections-s-m)


app/assets/stylesheets/views/_covid.scss
```diff
@@ -15,27 +15,40 @@ $c19-landing-page-header-background: govuk-colour("blue");
...
+  border-top: solid govuk-spacing(2) $covid-yellow;
+  overflow: auto; // become a block formatting context to prevent bleeding the margins
+}
+
 .covid__page-header {
   margin-bottom: govuk-spacing(8);
   background-color: $c19-landing-page-header-background;
-  border-top: solid govuk-spacing(2) $covid-yellow;
   color: govuk-colour("white");
+
```
unoduetre, line 44: Doing it in a similar way to how it's done for /browse pages.

app/assets/stylesheets/views/_covid.scss
```diff
@@ -15,27 +15,40 @@ $c19-landing-page-header-background: govuk-colour("blue");
 
 .global-bar-present {
   @include govuk-media-query($from: desktop) {
-    .covid__page-header {
+    .covid__breadcrumb-wrapper {
```
hannalaakso, line 18: Do we think that these three renamed classes are still being used? Their parent is something called `.global-bar-present` but there doesn't seem to be 
`.global-bar-present` on the page or the app.

app/assets/stylesheets/views/_covid.scss
```diff
@@ -15,27 +15,40 @@ $c19-landing-page-header-background: govuk-colour("blue");
...
+  overflow: auto; // become a block formatting context to prevent bleeding the margins
+}
+
 .covid__page-header {
   margin-bottom: govuk-spacing(8);
   background-color: $c19-landing-page-header-background;
-  border-top: solid govuk-spacing(2) $covid-yellow;
   color: govuk-colour("white");
+
+  @include govuk-media-query($until: "tablet") {
```
hannalaakso, line 45: Just a nit-pick: we could combine this media query and the following one into:
```
padding-top: govuk-spacing(1);
  
@include govuk-media-query($from: "tablet") {
    padding-top: govuk-spacing(4);
  }
```

hannalaakso (commented):
Looks good, just left a couple of small comments.

app/assets/stylesheets/views/_covid.scss
unoduetre: I didn't change anything just to stay safe, but I also don't think `.global-bar-present` is still used, so I'm happy to remove this part. It's also used for browse pages, so I can also remove it from there.

app/assets/stylesheets/views/_covid.scss
unoduetre: I'll apply your suggestion shortly.

unoduetre: Applied.

app/assets/stylesheets/views/_covid.scss
hannalaakso: I dug into this and realised this for the global banner coming from static: https://docs.publishing.service.gov.uk/manual/global-banner.html#activate-the-global-banner It might be good to double check that these styles work if the global banner is there although I think it's been a bit waffy to test with the global banner in the past since it's in a separate repo 🤔

unoduetre: I'll test it with a global banner enabled in static.

hannalaakso: I've just had a look at integration/coronavirus and it looks good to me with the global banner enabled 👍 I'm going to approve the PR but let me know if there was something else about the global banner I should be checking.

unoduetre: That's all I think.

PR #3775, opened 2024-09-23
WIP merge hardcoded bits content into content items

richardTowers:
Work in progress mechanism for extending content items with hardcoded content in collections.

This approach is being considered for a project that's currently `[OFFICIAL]`, so I can't give full details of the use case. If you know, then you know. If you don't know, then don't worry. If you worry, you just suffer twice.

Opening this early as a draft to facilitate discussion of the general approach. There's a bit of not-very-good explanation in the READMEs I've added.

| Before | After |
|-----|-----|
| image | image |


leenagupte:
I think that rather than hard coding the content into a rendering app, it should be put into a publishing app.
The presentation logic could be updated to pick up content from the YAML file to be passed to publishing-api. It would mean having to update the content schema, but that's something that will be required anyway.
Publishing the content via a publishing tool means that it will also be pushed to search without any extra effort.

The work in the rendering app could then be focused on rendering the extra fields from the content item as usual. This is similar to the approach we took for sign in pages. Though in that case, there wasn't an existing content item that needed to be extended.

There are things to work out, such as what the publishing workflow will be for any content changes, but it means that we'd be adding far fewer hacks to the system for content item changes that are expected to be permanent.

app/views/organisations/show.html.erb
```diff
@@ -12,6 +12,10 @@
            schema: :organisation,
            content_item: @organisation.content_item.content_item_data %>
 
+<% @organisation.content_extensions&.each do |content_extension| %>
+  <%= render partial: "content_extension_#{content_extension["document_type"]}", locals: { content_extension: } %>
+<% end %>
+
```
leenagupte, line 18: If there are multiple "hub" like pages that behave a bit like parts, it'd probably be better to render this new type of content in frontend, that already handles parts.

richardTowers: I'm assuming we're adorning the existing organisation pages with some extra components here. Appreciate that might not be the approach we end up taking though.

richardTowers:
Sorry about the unfinished sentences in my comment yesterday 😅  - Rory was barking in my face.

I've added some screenshots which show that this approach "works" (although obviously it's not a particularly good implementation of a hero image component, even if the image itself is very good).

leenagupte:
To be honest, I think this is all a bit premature as we don't have any designs yet and it's likely that we're going to have to render multiple linked pages.

Saying that, it's very easy to make things extremely flexible by hard-coding directly into the frontend, and then struggling later to apply that flexibility into the publishing app. We struggled with this on the COVID landing page, when we had to reduce flexibility of the sections to move it into collections publisher. I think timeline section was never moved for this reason.

I do think we should be using the content schemas to restrict how the data is presented so that we don't have to "take away" flexibility later.

richardTowers:
Closing this as it's just us scratching ideas together.

PR #3807, opened 2024-10-23
Fix markup for world-locations section

CodeSonia:
## What
[Trello card](https://trello.com/c/HsaHBtP3/2952-invalid-html-in-list-of-services-around-the-world)

The [Help and Services around the world](https://www.gov.uk/world) page has invalid HTML as the `<ol>` around the results has child elements that are not `<li>` elements.

When I ran the results page in W3 Validator, it states that`Element [div] not allowed as child of element [ol]: <div class="world-locations-group" data-filter="inner-block">`

## Why
As it's invalid HTML, it should be fixed. 

## Page Preview
https://collections-pr-3807.herokuapp.com/world

### DIFF
--- a/app/assets/stylesheets/views/_world_index.scss
+++ b/app/assets/stylesheets/views/_world_index.scss
@@ -6,6 +6,7 @@
   }
 
   .world-locations-groups {
+    list-style-type: none;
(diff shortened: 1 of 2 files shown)


unoduetre (approved):
Looks good to me.

PR #3815, opened 2024-10-25
Handle incorrectly UTF-8 encoded query and cookie url

unoduetre:
Use a different exception when this situation is detected:
* Sanitiser::Strategy::SanitisingError This error is not sent to Sentry.
See: https://github.com/alphagov/govuk_app_config/pull/402

## What

Stop sentry alerting on the “invalid byte sequence in UTF-8 errors”.

As discussed with @hannako, we will use a a solution based on an external gem called `rack-utf8_sanitizer` instead of developing our own gem. This solution will then be used in all our apps by copy-pasting the configuration of this gem.

## Why

When security scanning of our site takes place we are inundated with these errors in our slack channel. These errors are not serving any purpose since there is no issue to be solved. Therefore we should remove them to keep all our alerts actionable.

We often get exceptions in Sentry coming from a security checking bot that attempts some attacks by using strings with special characters. All of those "attacks" fail, but they produce exceptions which we would like to ignore, so they do not clutter Sentry. On the other hand we don’t want to ignore exceptions of the same class coming from other sources.

[Trello card](https://trello.com/c/FeOwxE2x/2938-silently-handle-invalid-byte-sequence-in-utf-8-errors-l)


unoduetre:
Closing this PR as [an alternative approach](https://github.com/alphagov/collections/pull/3819) will be used.

unoduetre:
I need to recreate this PR as there are some issues with the tests.

PR #3825, opened 2024-10-30
Use GraphQL for rendering World Index page

brucebolt:
This will be used to allow us to build a proof of concept for using GraphQL on GOV.UK.  It runs alongside Content Store and can be switched on by setting the `GRAPHQL_FEATURE_FLAG` environment variable.

There are no user facing changes.

[Trello card](https://trello.com/c/jDF1nSQ0)


app/presenters/world_index_graphql_presenter.rb
```diff
@@ -0,0 +1,15 @@
+class WorldIndexGraphqlPresenter < WorldIndexPresenter
```
JonathanHallam, line 1: Is this kind of inheritance okay long term? I'm just wondering what happens if we want to make graphql the default?

brucebolt: My theory was to reduce the deviation from what's already there as much as possible, so both behave the same.

When we want to make GraphQL the default, we can just move the methods from this class into `WorldIndexPresenter`.

spec/features/world_index_spec.rb
```diff
@@ -1,76 +1,157 @@
 require "integration_spec_helper"
 
 RSpec.feature "World index page" do
-  before do
-    stub_content_store_has_item("/world", GovukSchemas::Example.find("world_index", example_name: "world_index"))
-    visit "/world"
-  end
+  context "without the GraphQL feature flag" do
```
JonathanHallam, line 4: How gross would it be to do this iteratively? 

```
graphql_setup = {
   "on" = "with the GraphQL feature flag", 
   "off" = "without the GraphQL feature flag"
}

graphql_setup.each do |state, message|
 blah
  blah
   blah
```

JonathanHallam (commented):
Looking really good, couple of small questions

spec/features/world_index_spec.rb
brucebolt: I've moved the tests into a `shared_example` then used `it_behaves_like` in each of the contexts, which has removed all the duplication.

app/queries/graphql/world_index_query.rb
```diff
@@ -0,0 +1,31 @@
+class Graphql::WorldIndexQuery
+  def initialize(base_path)
```
yndajas, line 2: **suggestion:** simplification

Is the base path necessary? Won't it always be the same for this (and likewise on other index pages)? If it's not necessary, it could probably be removed from the model's `.find` method too

config/routes.rb
```diff
@@ -84,12 +84,12 @@
...
       as: :court
 
+  get "/world", to: "world#index"
+
   constraints DocumentTypeRoutingConstraint.new("step_by_step_nav") do
     get "/:slug", to: "step_nav#show"
   end
 
-  get "/world", to: "world#index"
-
```
yndajas, line 92: **question:** clarification

I'm not too sure what `DocumentTypeRoutingConstraint` does so I'm unclear what the need for this is, but it's not that important!

spec/controllers/world_controller_spec.rb
```diff
@@ -1,12 +1,30 @@
...
+        stub_publishing_api_graphql_query(
+          Graphql::WorldIndexQuery.new("/world").query,
+          fetch_graphql_fixture("world_index"),
+        )
+      end
+
+      it "has a success response" do
+        get :index
+        expect(response).to have_http_status(:success)
+      end
```
yndajas, line 27: **suggestion:** assertions

Is there a way that this test can be expanded to make the test fixture loading and stubbing more meaningful? Maybe this is just the style in Collections/frontend controller tests, but it feels like a lot of quite specific setup just to test that a stubbed response has a status code of 200

spec/features/world_index_spec.rb
```diff
@@ -1,76 +1,95 @@
...
 
 RSpec.feature "World index page" do
-  before do
-    stub_content_store_has_item("/world", GovukSchemas::Example.find("world_index", example_name: "world_index"))
-    visit "/world"
-  end
+  shared_examples "world index page" do
+    scenario "returns 200 when visiting world index page" do
+      expect(page.status_code).to eq(200)
+    end
```
yndajas, line 7: **question:** usefulness of tests

Again, maybe it's the style, but it feels like the feature and controller tests are doing almost the same thing: stubbing a response then checking the status code is 200. Presumably the under-the-hood setup will be a bit different between the two test types, but I wonder if one or the other could be more useful and/or require less setup

spec/presenters/world_index_graphql_presenter_spec.rb
```diff
@@ -0,0 +1,93 @@
+RSpec.describe WorldIndexGraphqlPresenter do
+  let(:fixture) { fetch_graphql_fixture("world_index").dig("data", "edition") }
+  let(:content_item) { WorldIndexGraphql.new(fixture) }
+  let(:world_index_presenter) { described_class.new(content_item) }
```
yndajas, line 4: **suggestion:** test performance

Can we memoise this somehow to avoid (re)loading the fixture on every test?

spec/presenters/world_index_graphql_presenter_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+
+    it "returns the joined name and slug" do
+      expect(subject).to eq "united-kingdom United Kingdom"
+    end
+  end
+
+  describe "#international_delegations_count" do
+    subject { world_index_presenter.international_delegations_count }
+
+    it "returns the number of world locations" do
```
yndajas, line 89: **chore:** typo

"international delegations"?

yndajas (approved):
Nothing blocking here. I have a few questions about the usefulness/performance of some of the tests, but they relate to broader Collections app decisions/patterns that are out of scope

app/queries/graphql/world_index_query.rb
brucebolt: I'm guessing the theory is that the same controller could be used to render another identically laid out page, but with different content.  It seems highly unlikely though.

Will leave it to the team who own Collections whether they'd like to make this change.

spec/controllers/world_controller_spec.rb
brucebolt: Agreed.  I added this because there was already a test like this, but we're already testing the exact same thing in the feature test.

I'm not sure how we'd make it any more useful though, so I'm inclined to say it should just be deleted.

spec/features/world_index_spec.rb
JonathanHallam: I don't have a chef kiss emoji on github 👨‍🍳 💋

spec/presenters/world_index_graphql_presenter_spec.rb
brucebolt: I'll fix that and also the original test that I coped it from.

spec/presenters/world_index_graphql_presenter_spec.rb
brucebolt: Good spot.  I'd forgotten that `let` only caches for the duration of each test.

I've updated this and also the other non-GraphQL test to behave the same way.

config/routes.rb
brucebolt: From what I could understand, it's being used here to determine the controller and method using the returned document type (from within the content item), rather than by path.

JonathanHallam (approved):
I'm happy with this from my comments point of view. Thanks Bruce :)

spec/features/world_index_spec.rb
mike3985: We've deleted the controller spec that was only testing the 200 response

JonathanHallam:
Merging this in Ynda and Bruces absence. I think all comments have been answered

PR #3835, opened 2024-11-04
Improve print styles for right-to-left (RTL) content

matthillco:
## What
Improve print styles for right-to-left (RTL) content:
- Force `world_location_news` layout to print full width.
- Enable RTL rendering for breadcrumbs and header

_Note: While the heading has now been flipped to read right-to-left in RTL mode, the layout remains the same: the heading is on the left of the screen with the language switch on the right. This follows the pattern set out on other pages, where the columns remain in their original direction._

This is part of the work to improve print styles more generally. [Trello](https://trello.com/c/6E5UoTiK/335-rtl-print-styles-have-been-considered-and-implemented-where-required)

_Note: This is a dependency of https://github.com/alphagov/government-frontend/pull/3404_

## Why
The components should render in the intended direction for RTL languages.

### Full width layouts
| Before    | After |
| -------- | ------- |
| image  |image |

### Breadcrumbs & Header
| Before    | After |
| -------- | ------- |
| image  | image |

### DIFF
--- a/app/views/application/_breadcrumbs.html.erb
+++ b/app/views/application/_breadcrumbs.html.erb
@@ -3,5 +3,6 @@
 <%= render 'govuk_publishing_components/components/breadcrumbs', {
   breadcrumbs: Breadcrumbs.new(@content_item).breadcrumbs,
   collapse_on_mobile: true,
-  inverse: inverse
+  inverse: inverse,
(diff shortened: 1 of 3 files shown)


AshGDS (approved):
👍

PR #3844, opened 2024-11-12
Remove taxon-list app component from collections

MartinJJones:
## What

- Remove the [taxon-list](https://govuk-collections.herokuapp.com/component-guide/taxon_list) app component
- Update taxon_list_helper.rb to work with the document-list component
- Replaced [taxon-list](https://govuk-collections.herokuapp.com/component-guide/taxon_list) app component with either the [list](https://components.publishing.service.gov.uk/component-guide/list) or [document-list](https://components.publishing.service.gov.uk/component-guide/document_list) gem components in several view templates
- Update tests for world location and world wide taxons to check that the "title", "link" and "description" for each search result/tagged content is included on the page

### The taxon-list component is no longer required

The taxon-list component was originally created for an AB Test on https://www.gov.uk/education/funding-and-finance-for-students#/education/student-grants-bursaries-scholarships. This page is now using the document-list component from the govuk_publishing_components gem

### Accessibility improvement

Since the document-list component doesn’t support headings for lists of links, we’ve decided not to try and add this option into the component. This is because wrapping links in headings is not a suitable option for loads of [GOV.UK](http://gov.uk/) pages since:
- It is technically possible for some links not to have a description (a `<p>` tag) after it. Having a heading or a mix of headings and plain links doesn’t seem to be a good option here as it could be confusing for users
- Using https://www.gov.uk/world/afghanistan as an example, it currently includes 50+ heading elements (inside the accordions)
  - This seems like it would be difficult to navigate through all of the heading elements on the page, especially if some links were headings and others were not
  - We checked with an accessibility specialist to get advice on removing the heading element and agreed that this change looks like a better option as it also adds an additional layer in the accessibility tree which is likely more time consuming to navigate

## Visual changes

Spacing is increased slightly for list items when using the document-list component

https://collections-pr-3844.herokuapp.com/world/afghanistan (accordion.html.erb)

| Before | After |
| --- | --- |
| accordion-before | accordion-after |

https://collections-pr-3844.herokuapp.com/world/living-in-afghanistan (tagged_content_list.html.erb)

| Before | After |
| --- | --- |
| Screenshot 2024-11-26 at 15 21 02 | Screenshot 2024-11-26 at 15 21 10 |

https://collections-pr-3844.herokuapp.com/government/ministers/secretary-of-state-for-foreign-commonwealth-and-development-affairs (past_role_holders.html.erb) - No visual changes

[Trello card](https://trello.com/c/cBHVSgZK/3021-remove-taxon-list-component-from-collections-s)


hannalaakso (approved):
LGTM 🎉  Swapping these out turned out to be a bit more complicated than we initially thought and this solution resolves it all quite cleanly without the need for adding extra options to the gem which is nice. I've compared the generated HTML of the relevant templates before and after and it all looks good. The extensive commit messages were also super helpful when reviewing the changes, thanks for doing that 🙏

PR #3855, opened 2024-11-18
Enable toggling of GraphQL with query string

brucebolt:
This will be used during load testing to allow us to make GraphQL queries without needing to update the environment variable.

[Trello card](https://trello.com/c/P7cdLbnK)

### DIFF
--- a/app/controllers/world_controller.rb
+++ b/app/controllers/world_controller.rb
@@ -1,6 +1,6 @@
 class WorldController < ApplicationController
   def index
-    if Features.graphql_feature_enabled?
+    if Features.graphql_feature_enabled? || params.include?(:graphql)
       index = WorldIndexGraphql.find!("/world")
       @presented_index = WorldIndexGraphqlPresenter.new(index)
     else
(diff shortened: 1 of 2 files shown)


app/controllers/world_controller.rb
```diff
@@ -1,6 +1,6 @@
 class WorldController < ApplicationController
   def index
-    if Features.graphql_feature_enabled?
+    if Features.graphql_feature_enabled? || params.include?(:graphql)
```
mike3985, line 3: Now that I've seen this for real, it's got me thinking how about
```suggestion
    if Features.graphql_feature_enabled? && params.include?(:graphql)
```

I just imagined, like, a second ago, what if the feature flag enables the feature to be used at all in the given env and then the query param tells Collections to actually use the GraphQL data source for this request... Would that be pointless?

brucebolt: That could be useful, but I wonder if the benefit of rolling out by environment is that we get the benefit of traffic replay where public GET requests are replicated in staging?

mike3985 (approved):

PR #3867, opened 2024-11-26
GOV.UK Chat promo: show on more pages

jackbot:
Displays the promo on a few more pages.

Some of the new URLs use the `show_a_to_z` view, so the logic to render
the promo needed adding there too.

Also updates a test which relies on the content item base path being a
mainstream browse page to use a hard-coded URL.


davidgisbey (approved):
looks great

PR #3872, opened 2024-11-28
Use GraphQL for rendering Ministers Index page

mike3985:
When the `GRAPHQL_FEATURE_FLAG` environment variable or the `graphql` query param is set, use Publishing API's GraphQL endpoint as the source of the data for serving the Ministers Index page. Otherwise use the standard Content Store data source.

(Corresponding backend changes: https://github.com/alphagov/publishing-api/pull/3009)

https://trello.com/c/PlR5Ken4/1452-querywinkle-the-ministers-index-page

### DIFF
--- a/app/controllers/ministers_controller.rb
+++ b/app/controllers/ministers_controller.rb
@@ -2,8 +2,15 @@ class MinistersController < ApplicationController
   around_action :switch_locale
 
   def index
-    ministers_index = MinistersIndex.find!(request.path)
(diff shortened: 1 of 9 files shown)


brucebolt (approved):

hannako (approved):
I'm not sure I understand enough about the publishingAPI changes to be able to review the code in app/queries/graphql/ministers_index_query.rb, but everything else makes sense and follows the same pattern from previous PR's. Thanks for reaching out for a review! Hope it's going well 👍

PR #3829, opened 2024-10-31
Dynamic popular tasks - by content id

hannako:
⚠️ This repo is Continuously Deployed: make sure you [follow the guidance](https://docs.publishing.service.gov.uk/manual/development-pipeline.html#merge-your-own-pull-request) ⚠️

Follow [these steps](https://guides.rubyonrails.org/upgrading_ruby_on_rails.html) if you are doing a Rails upgrade.

PR #3873, opened 2024-11-28
Add additional guard clause when we fetch popular tasks from SearchAPI

hannako:
Missing from https://github.com/alphagov/collections/pull/3829

### DIFF
--- a/app/models/popular_list_set.rb
+++ b/app/models/popular_list_set.rb
@@ -30,6 +30,8 @@ private
       fields: SearchApiFields::POPULAR_BROWSE_SEARCH_FIELDS,
     }
     Services.cached_search(params)
+  rescue GdsApi::HTTPServerError
+    {}
   end
 
   def second_level_browse_content_ids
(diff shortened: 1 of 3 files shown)


spec/support/search_api_helpers.rb
```diff
@@ -1,6 +1,10 @@
 module SearchApiHelpers
   include SearchApiFields
 
+  def stub_search_api_isnt_available
```
hannako, line 4: Wondering why this isn't available as a test helper already.

app/models/popular_list_set.rb
```diff
@@ -30,6 +30,8 @@ def search_response
       fields: SearchApiFields::POPULAR_BROWSE_SEARCH_FIELDS,
     }
     Services.cached_search(params)
+  rescue GdsApi::HTTPServerError
+    {}
```
unoduetre, line 34: I wonder, should the rescued error be sent to Sentry, so we know something is going on?

PR #3882, opened 2024-12-02
Add govuk_web_banners gem

KludgeKML:
- We don't include the partial in this commit, just the dependencies.
  Partials are a bit more complex in collections because there are some
  pages where the partial needs to go within the main block. The partial
  can be added later in the exact place it's needed. This makes collections
  a bit more work for the developer, but at least for each page type the
  partial will only have to be added once.

https://trello.com/c/aYXKcrCo/375-create-govukwebbanner-gem-with-support-for-recruitment-banners


hannako (commented):
I think we've had a good percentage of banner requests for browse pages so am wondering what best course of action is here... This was always the bit of the jigsaw I was worried about, as I'm aware that we add the banner in different files/locations depending on the document type!

We could just add dependency in these PR's and then when we need to add a banner we will have to figure out which file to add the partial?

KludgeKML:
> I think we've had a good percentage of banner requests for browse pages so am wondering what best course of action is here... This was always the bit of the jigsaw I was worried about, as I'm aware that we add the banner in different files/locations depending on the document type!
> 
> We could just add dependency in these PR's and then when we need to add a banner we will have to figure out which file to add the partial?

Yeah, that seems reasonable. It'll mean each time a banner gets added we'll have to work out whether or not the partial is included, but that should be relatively quick (and the partial can easily be added ahead of the go-live time for the banner).

leenagupte:
> I think we've had a good percentage of banner requests for browse pages so am wondering what best course of action is here... This was always the bit of the jigsaw I was worried about, as I'm aware that we add the banner in different files/locations depending on the document type!
> 
> We could just add dependency in these PR's and then when we need to add a banner we will have to figure out which file to add the partial?

Yes I think agree with this suggestion. Otherwise we run the risk of the banner appearing in odd places. Once the requests come in we can add the banner to the appropriate partials. On the brightside once they're in, they won't need to be removed from here again.

hannako (approved):
LGTM

PR #3894, opened 2024-12-09
Remove document navigation list component

MartinJJones:


hannalaakso (approved):
Thanks Martin for the IRL chat. As we discussed, a few very minor things in `_featured_links.html.erb` (which as Martin points out in the commit message was inherited from the existing component code) could perhaps be refactored to be a bit clearer. But as it's currently following the conventions of what we do in other components, we're going to raise this separately to discuss component level conventions. 

LGTM from a dev point of view, subject to a final design check.

PR #3896, opened 2024-12-09
Fix GA4 event data on taxon pages

MartinJJones:
## What

Set `disable_ga4` to true for the document-list component used to render each section on a taxon page.

## Why

This fixes an issue on taxon pages (/education for example), where link clicks would fire 2 events instead of 1, both of which were missing some data.

## Event data changes

The data below is sent to GA after clicking the "Student finance for undergraduates" link on this page - https://www.gov.uk/education/funding-and-finance-for-students#/education/student-grants-bursaries-scholarships

### Before

```json
{
 "event": "event_data",
 "event_data": {
  "event_name": "navigation",
  "external": "false",
  "index": {
   "index_link": "2"
  },
  "link_domain": "https://www.gov.uk",
  "method": "command/win click",
  "text": "Student finance for undergraduates",
  "type": "document list",
  "url": "/student-finance"
 },
 "govuk_gem_version": "46.2.0",
 "timestamp": "1733824170576"
}

{
 "event": "event_data",
 "event_data": {
  "event_name": "navigation",
  "external": "false",
  "index": {
   "index_link": "2"
  },
  "link_domain": "https://www.gov.uk",
  "method": "command/win click",
  "text": "Student finance for undergraduates",
  "type": "document list",
  "url": "/student-finance"
 },
 "govuk_gem_version": "46.2.0",
 "timestamp": "1733824170581"
}
```

### After

Only 1 event is fired and information for "index_section", "index_section_count" and "section" is now included.

```json
{
 "event": "event_data",
 "event_data": {
  "event_name": "navigation",
  "external": "false",
  "index": {
   "index_link": "2",
   "index_section": "1",
   "index_section_count": "8"
  },
  "index_total": "6",
  "link_domain": "https://www.gov.uk",
  "method": "command/win click",
  "section": "Services",
  "text": "Student finance for undergraduates",
  "type": "document list",
  "url": "/student-finance"
 },
 "govuk_gem_version": "46.2.0",
 "timestamp": "1733824112924"
}
```

### DIFF
--- a/app/views/taxons/sections/_guidance_and_regulation.html.erb
+++ b/app/views/taxons/sections/_guidance_and_regulation.html.erb
@@ -1,4 +1,5 @@
 <%= render 'govuk_publishing_components/components/document_list',
+  disable_ga4: true,
   items: section[:documents],
   margin_bottom: 7
 %>
--- a/app/views/taxons/sections/_news_and_communications.html.erb
+++ b/app/views/taxons/sections/_news_and_communications.html.erb
@@ -19,6 +19,7 @@
 
 <% if section[:documents].any? %>
   <%= render 'govuk_publishing_components/components/document_list',
+    disable_ga4: true,
     items: section[:documents_with_promoted],
     margin_bottom: 7
   %>
--- a/app/views/taxons/sections/_policy_and_engagement.html.erb
+++ b/app/views/taxons/sections/_policy_and_engagement.html.erb
@@ -1,4 +1,5 @@
 <%= render 'govuk_publishing_components/components/document_list',
+  disable_ga4: true,
   items: section[:documents],
   margin_bottom: 7
 %>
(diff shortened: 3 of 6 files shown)


app/views/taxons/show.html.erb
```diff
@@ -62,7 +62,7 @@
               }
             %>
             <%= content_tag(:div, data: ga4_data) do %>
-              <%= render(partial: section[:partial_template], locals: { section: section }) %>
+              <%= render(partial: section[:partial_template], locals: { section: section, disable_ga4: true }) %>
```
andysellick, line 65: Putting a `false` default into the partials seems sensible, but is there a situation where that isn't overridden by this line?

MartinJJones: I've not seen any other uses of the section partials, so not essential right now, my initial option was to hardcode `disable_ga4: true` in each section partial, happy to go with this approach if preferred

andysellick: I think that might be simpler - otherwise it implies that this is necessary because some other part of the code doesn't pass `true`, which isn't correct. We can always extend it later if needed 👍

MartinJJones: Agreed, thanks Andy, I've made this change

PR #3918, opened 2024-12-19
Update usage of the title component

andysellick:

## What

- moving towards a world where the title component has no margin_top option: https://github.com/alphagov/govuk_publishing_components/pull/4508
- this removes instances of the use of the margin_top option for the title component
- in some cases it was possible to replace the title component with the heading component for no visual change (where the margin_top: 0 option was applied)
- in other cases there is a visual change, but this will be minimal

## Why
We're updating the component spacing model to make it more consistent and having a margin_top option isn't part of the intended outcome.

## Visual changes
For the [coronavirus landing page](https://www.gov.uk/coronavirus) the title component used the `margin_top: 0` option, so we replace the component with the heading component, which means no visual change (app/views/coronavirus_landing_page/components/landing_page/_page_header.html.erb).

For the development index page we replace the title component with the heading component. This results in a small visual change, but this is a non-public facing page so isn't a concern (app/views/development/index.html.erb).

For the [taxon pages](https://www.gov.uk/transport/driving-and-motorcycle-tests) the title component used the `margin_top: 0` option, so we replace the component with the heading component, which means no visual change (app/views/taxons/_page_header.html.erb).

For the [world location news pages](https://www.gov.uk/world/ukraine/news) the title component used the `margin_top: 0` option, but requires the 'context' feature of the title component (not available in the heading component) so we keep it as it is, but remove the option (app/views/world_location_news/show.html.erb). This currently will not produce a visual difference until we merge https://github.com/alphagov/govuk_publishing_components/pull/4508 at which point it will change as shown below (only on desktop and tablet, mobile remains unchanged). I think this change is acceptable, as the title seems quite squashed against the breadcrumb in this instance.

Before | After
------ | -----
Screenshot 2024-12-19 at 16 20 19 | Screenshot 2024-12-19 at 16 20 26

### DIFF
--- a/app/views/coronavirus_landing_page/components/landing_page/_page_header.html.erb
+++ b/app/views/coronavirus_landing_page/components/landing_page/_page_header.html.erb
@@ -8,10 +8,12 @@
   <div class="govuk-width-container">
     <div class="govuk-grid-row">
       <div class="govuk-grid-column-two-thirds">
-        <%= render "govuk_publishing_components/components/title", {
-          title: raw(title),
+        <%= render "govuk_publishing_components/components/heading", {
+          text: title,
+          heading_level: 1,
+          margin_bottom: 8,
           inverse: true,
-          margin_top: 0
+          font_size: "xl",
         } %>
       </div>
     </div>
--- a/app/views/development/index.html.erb
+++ b/app/views/development/index.html.erb
@@ -8,7 +8,13 @@
 <body>
 <div id="wrapper">
   <main class="govspeak">
-    <%= render "govuk_publishing_components/components/title", title: "collections" %>
+    <%= render "govuk_publishing_components/components/heading", {
+      text: "collections",
+      heading_level: 1,
+      margin_bottom: 4,
+      font_size: "xl",
+      padding: true,
+    } %>
     <% html = capture do %>
       <p><%= t('development.header') %>.</p>
       <table>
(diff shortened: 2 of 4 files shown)


MartinJJones (commented):
The changes look good to me overall.

After looking further at the changes to the [gem and other apps mentioned in the PR description](https://github.com/alphagov/govuk_publishing_components/pull/4508), it appears that it was possible to swap most of the title components out with a heading component and the context option is only heavily used in collections.

I think this makes it a bit easier [avoid adding the padding-top CSS](https://github.com/alphagov/govuk_publishing_components/pull/4508/files#diff-48e170e38a72d0e94ffd7f325b407e9b4fdba7e308db0fc516d2e2a4fd3733e8R3-R8) to the title component for now as a workaround, as there are only a few apps left that may need to use the `govuk-maim-wrapper` class approach.

I've created a PR to do this in collections which also fixes several layout issues - https://github.com/alphagov/collections/pull/3944

If this could go live first, then it would mean that there are no visual changes in this PR and could make switching from the title component to the heading component, once the context variant has been added, a bit easier.

andysellick:
@MartinJJones happy for your PR to be merged first although I'd suggest if you're adding the `margin_top: 0` option on the title component there it would be better to switch to the heading component instead. Heading now includes the context option and has no margin top. Also we're trying to retire the title component in favour of the heading component. Happy to discuss.

MartinJJones:
@andysellick I've merged the other PR now which makes [better use of govuk-main-wrapper](https://github.com/alphagov/collections/pull/3944), replacing the title component with the heading component where required.

You should now be able to remove the change made to `app/views/world_location_news/show.html.erb` which would resolve the merge conflict and I think any visual changes in this PR

andysellick:
Closing in favour of https://github.com/alphagov/collections/pull/3963

PR #3936, opened 2024-12-31
Update GraphQL queries to use generic edition type

brucebolt:
Publishing API was updated in https://github.com/alphagov/publishing-api/pull/3063 to have a single edition type for most content, instead of a specific type for roles and the world index page.

Updating the query used by this application in making such requests to Publishing API.

The ministers index page retains a specific edition type, as optimisations exist for determining one of the links in that page.

[Trello card](https://trello.com/c/MSjuzc9M)


JonathanHallam (approved):
Nice one, thanks

PR #3945, opened 2025-01-08
Make the GraphQL overrides work consistently

brucebolt:
This makes the GraphQL override actually override the feature flag.

The behaviour is now consistent with Government Frontend.

[Trello card](https://trello.com/c/sVW4QFue)

### DIFF
--- a/app/controllers/ministers_controller.rb
+++ b/app/controllers/ministers_controller.rb
@@ -2,15 +2,25 @@ class MinistersController < ApplicationController
   around_action :switch_locale
 
   def index
-    if params.include?(:graphql)
-      ministers_index = Graphql::MinistersIndex.find!(request.path)
-      content_item_data = ministers_index.content_item
-    else
-      ministers_index = MinistersIndex.find!(request.path)
-      content_item_data = ministers_index.content_item.content_item_data
-    end
+    content_item_data = if params[:graphql] == "false"
+                          load_from_content_store
+                        elsif params[:graphql] == "true" || Features.graphql_feature_enabled?
+                          load_from_graphql
+                        else
+                          load_from_content_store
+                        end
 
     @presented_ministers = MinistersIndexPresenter.new(content_item_data)
-    setup_content_item_and_navigation_helpers(ministers_index)
+    setup_content_item_and_navigation_helpers(@ministers_index)
+  end
+
+  def load_from_graphql
+    @ministers_index = Graphql::MinistersIndex.find!(request.path)
+    @ministers_index.content_item
+  end
+
+  def load_from_content_store
+    @ministers_index = MinistersIndex.find!(request.path)
+    @ministers_index.content_item.content_item_data
   end
 end
--- a/spec/features/ministers_spec.rb
+++ b/spec/features/ministers_spec.rb
@@ -123,52 +123,124 @@ RSpec.feature "Ministers index page" do
     end
   end
 
-  context "without the graphql feature flag" do
+  context "without the GraphQL feature flag" do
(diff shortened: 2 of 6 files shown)


app/controllers/ministers_controller.rb
```diff
@@ -2,15 +2,27 @@ class MinistersController < ApplicationController
...
-    end
+    content_item_data = if params.include?(:graphql) && params[:graphql] == "true"
+                          load_from_graphql
+                        elsif params.include?(:graphql) && params[:graphql] == "false"
+                          load_from_content_store
+                        elsif Features.graphql_feature_enabled?
+                          load_from_graphql
+                        else
+                          load_from_content_store
+                        end
```
yndajas, line 13: **suggestion (non-blocking):** simplification

Is the `include?` necessary here? The square bracket access should return `nil` if the param is not present, which I think would produce the same results. I wonder if the `if` statement could therefore be simplified to:

```rb
if params[:graphql] == "false"
  load_from_content_store
elsif params[:graphql] == "true" || Features.graphql_feature_enabled?
  load_from_graphql
else
  load_from_content_store
end
```

Likewise in the other controllers

spec/features/ministers_spec.rb
```diff
@@ -128,47 +128,119 @@
 
     before do
       stub_content_store_has_item("/government/ministers", document)
-      visit "/government/ministers"
     end
 
-    it_behaves_like "ministers index page"
+    context "when the graphql parameter not set" do
```
yndajas, line 133: **nitpick:** missing word

"is not set"?

yndajas, line 133: **suggestion (non-blocking):** test order

I wonder if this test would be better placed adjacent to the other GraphQL-related ones? Currently they're separated by the reshuffle stuff

spec/features/ministers_spec.rb
```diff
@@ -128,47 +128,119 @@
...
+      let(:document) { GovukSchemas::Example.find("ministers_index", example_name: "ministers_index-reshuffle-mode-off") }
+
+      before do
+        stub_content_store_has_item("/government/ministers", document)
+        visit "/government/ministers?graphql=false"
+      end
+
+      it "does not get the data from GraphQL" do
+        expect(a_request(:post, "#{Plek.find('publishing-api')}/graphql")).not_to have_been_made
+      end
```
yndajas, line 243: **suggestion (non-blocking, if-minor):** test structure

Do you think it's necessary to test everything within each GraphQL feature flag context, or could we pare down and flatten these tests to something like this?

- a 'default' test, i.e. feature flag disabled, no query parameter
- a test with the feature flag disabled but the query parameter as `true`
- a test with the feature flag enabled and no query parameter
- a test with the feature flag enabled but the query parameter as `false`
- the two reshuffle tests

We'd lose coverage of the more edge case scenarios, but it might make the likely scenarios easier to read/compare at a glance

yndajas (approved):
A few suggestions, but nothing blocking

app/controllers/ministers_controller.rb
brucebolt: I've removed the `include?`, which definitely isn't required.  Also made the refactor, which has definitely simplified the logic.

spec/features/ministers_spec.rb
brucebolt: Good spot.  I've updated this.

spec/features/ministers_spec.rb
brucebolt: I'm not massively inclined to restructure the tests right now, especially as we're approaching a decision point of whether to continue with GraphQL work.  We can look into this if we decide to continue, and make the three sets of tests more consistent with each other.

PR #3963, opened 2025-01-22
Replace title components with headings

matthillco:
## What
Replace title components with headings.

## Why
We are attempting to retire the page title component in favour of the heading component. [Trello](https://trello.com/c/Z8t7nZrM/441-replace-title-with-heading-in-collections)

## Visual Changes
None. Components look the same.


MartinJJones (approved):
Changes look good to me 👍

PR #3966, opened 2025-01-24
Change inverse header padding top

andysellick:

## What
Change the padding top option call in the inverse header on taxon pages, such as https://www.gov.uk/entering-staying-uk

## Why
The option in the component has been changed to a scale, instead of true/false, see this PR: https://github.com/alphagov/govuk_publishing_components/pull/4590

Note that this change will not be merged until a new version of the components gem including this change is included here.

## Visual changes
None.

Trello card: https://trello.com/c/hRL2Nmyp/450-improve-inverse-header-component-padding


AshGDS (approved):

AshGDS:
Approved the code change itself - but the test is failing, looks like some changes to the global bar tests need to be made.

andysellick:
@AshGDS I've just pushed another commit if you're able to check - it's the JS for the covid page, which seems to relate to whether the global bar appears or not. Since this will now be handled by the banners gem I think it's safe to just delete all this code?

AshGDS (commented):
Looks good, can you remove `data-module="coronavirus-landing-page"` from `app/views/coronavirus_landing_page/show.html.erb` as well?

andysellick:
Thanks for spotting that @AshGDS - have removed.

PR #4038, opened 2025-04-01
Add A/B test for GraphQL

brucebolt:
Sets up a A/B test to divide traffic between GraphQL and Content Store for role pages and the ministers index page.

Each document type has been given a different test name, so we can enable/disable the tests for each document type separately.

[Trello card](https://trello.com/c/stxnESVI)

### DIFF
--- a/app/controllers/ministers_controller.rb
+++ b/app/controllers/ministers_controller.rb
@@ -2,9 +2,17 @@ class MinistersController < ApplicationController
   around_action :switch_locale
 
   def index
+    ab_test = GovukAbTesting::AbTest.new(
+      "GraphQLMinistersIndex",
+      allowed_variants: %w[A B Z],
+      control_variant: "Z",
+    )
+    @requested_variant = ab_test.requested_variant(request.headers)
+    @requested_variant.configure_response(response)
+
     content_item_data = if params[:graphql] == "false"
                           load_from_content_store
-                        elsif params[:graphql] == "true"
+                        elsif params[:graphql] == "true" || @requested_variant.variant?("B")
                           load_from_graphql
                         else
                           load_from_content_store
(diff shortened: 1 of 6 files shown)


JonathanHallam (approved):
LGTM, I'm assuming we'll need a similar PR for Frontend and Government FE?

brucebolt:
> LGTM, I'm assuming we'll need a similar PR for Frontend and Government FE?

Yes, I've just done this one for now to check the configuration works ok before adding it to another app.

app/controllers/ministers_controller.rb
```diff
@@ -2,9 +2,17 @@ class MinistersController < ApplicationController
   around_action :switch_locale
 
   def index
+    ab_test = GovukAbTesting::AbTest.new(
+      "GraphQLMinistersIndex",
+      allowed_variants: %w[ContentStore GraphQL],
+      control_variant: "ContentStore",
```
hannako, line 8: I think it's good practice to have three variants A,B,Z where Z is excluded from the test, and A and B are always equal in weight. 

So in this case A and Z might be both be set to ContentStore. 

Then in Fastly you can set 100% to Z to confirm that the cookie is being set, and all the data is coming through. And then scale up keeping A and B equal, so instead of https://github.com/alphagov/govuk-fastly/pull/144/files you probably would have 98% Z, 1%A and 1%B

hannako: See https://github.com/alphagov/govuk-fastly/pulls?q=is%3Apr+is%3Aclosed+vertex+ for prior art

brucebolt: Ooh, that's a great approach.  I've updated to have three variants: A, B and Z.  A and Z will use Content Store, and B will use GraphQL.  I'll update the Fastly config to enable to tests and point 100% of traffic to variant Z for now.

hannako (approved):
LGTM :)

PR #4039, opened 2025-04-01
Remove GraphQL feature flag

brucebolt:
This is not being used for anything at the moment, since we are manually selecting GraphQL by adding a query parameter to requests.

Removing in preparation for setting up an A/B test that switches between GraphQL and Content Store.  Having a feature flag in addition to this will cause complication.

[Trello card](https://trello.com/c/stxnESVI)


JonathanHallam (approved):
nice!

PR #4049, opened 2025-04-10
Add additional fields to role GraphQL query

brucebolt:
We need additional fields included in the GraphQL query in order to render meta tags on the page.

These tags are extracted from the response by [`govuk_publishing_components`](https://github.com/alphagov/govuk_publishing_components/blob/b2b64934c89db058bfb1d765ecfcbd6b6b90c683/lib/govuk_publishing_components/presenters/meta_tags.rb).

[Trello card](https://trello.com/c/ZOEv2NXr)

### DIFF
--- a/app/queries/graphql/role_query.rb
+++ b/app/queries/graphql/role_query.rb
@@ -9,8 +9,16 @@ class Graphql::RoleQuery
         edition(base_path: "#{@base_path}") {
           ... on Edition {
             base_path
+            content_id
+            document_type
+            first_published_at
             locale
+            public_updated_at
+            publishing_app
+            rendering_app
+            schema_name
             title
+            updated_at
 
             details {
               body
@@ -43,8 +51,13 @@ class Graphql::RoleQuery
               }
 
               ordered_parent_organisations {
+                analytics_identifier
                 base_path
                 title
+              }
+
+              organisations {
+                analytics_identifier
               }
             }
           }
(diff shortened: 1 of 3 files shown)


app/queries/graphql/role_query.rb
```diff
@@ -9,8 +9,16 @@ def query
         edition(base_path: "#{@base_path}") {
           ... on Edition {
             base_path
+            content_id
```
yndajas, line 12: **question:** how did you work out which of the fields accessed in the meta tags presenter in publishing components are needed for roles?

brucebolt: I just compared what was in the head of the live document and figured out what GraphQL fields to add based on the code in Publishing Components.

app/queries/graphql/role_query.rb
```diff
@@ -43,9 +51,14 @@ def query
               }
 
               ordered_parent_organisations {
+                analytics_identifier
                 base_path
                 title
               }
+
+              organisations {
```
mike3985, line 59: Are you sure that this would be used? The presenter seems to alias `ordered_parent_organisations` as `organisations`

brucebolt: It looks like `govuk_publishing_components` [uses the value for `organisations`](https://github.com/alphagov/govuk_publishing_components/blob/b2b64934c89db058bfb1d765ecfcbd6b6b90c683/lib/govuk_publishing_components/presenters/meta_tags.rb#L64) and the content item appears to be passed into that before reaching the presenter.

app/queries/graphql/role_query.rb
brucebolt: I've added these fields to the fixture file.

mike3985 (approved):
👍

PR #4052, opened 2025-04-14
Render meta tags and <main> `lang` attribute on GraphQL version of World Index page

mike3985:
https://trello.com/c/ezCkqbaq/1665-make-graphql-version-of-world-index-the-same-as-the-regular-version

Request some additional fields from GraphQL when serving the World Index page so that `<main>`'s `lang` attribute and a bunch of meta tags are included in the rendered page.

We're doing this to improve consistency between the GraphQL and the regular versions of this page.

### DIFF
--- a/spec/features/world_index_spec.rb
+++ b/spec/features/world_index_spec.rb
@@ -71,6 +71,10 @@ RSpec.feature "World index page" do
       expect(ga4_focus_loss_data["action"]).to eq "filter"
       expect(ga4_focus_loss_data["section"]).to eq "Help and services around the world"
     end
+
+    scenario "renders the locale in <main> element" do
+      expect(page).to have_css("main[lang='en']")
+    end
   end
 
   before do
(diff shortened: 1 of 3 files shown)


spec/features/world_index_spec.rb
```diff
@@ -75,6 +75,25 @@
...
+      meta_tag = page.first("meta[name='govuk:content-id']", visible: false)
+      expect(meta_tag["content"]).to be_present
+      meta_tag = page.first("meta[name='govuk:schema-name']", visible: false)
+      expect(meta_tag["content"]).to be_present
+      meta_tag = page.first("meta[name='govuk:rendering-app']", visible: false)
+      expect(meta_tag["content"]).to be_present
+      meta_tag = page.first("meta[name='govuk:publishing-app']", visible: false)
+      expect(meta_tag["content"]).to be_present
+      meta_tag = page.first("meta[name='govuk:format']", visible: false)
+      expect(meta_tag["content"]).to be_present
```
yndajas, line 95: **thought:** how about a loop to avoid the variable reassignment, assertion repetition, and wall of textness?

```rb
%w[
  public-updated-at
  updated-at
  ...
].each do |name|
  meta_tag = page.first("meta[name='govuk:#{name}']", visible: false)
  expect(meta_tag["content"]).to be_present
end

beccapearce (approved):
Makes sense to me! All looks good!

PR #4053, opened 2025-04-16
Add missing content to GraphQL version of ministers index page

brucebolt:
This adds missing content to the ministers index page when rendered using the response from GraphQL.

See each commit for details of what has been added.

Depends on https://github.com/alphagov/publishing-api/pull/3289.

[Trello card](https://trello.com/c/HcOHeixw)

### DIFF
--- a/app/queries/graphql/ministers_index_query.rb
+++ b/app/queries/graphql/ministers_index_query.rb
@@ -12,6 +12,7 @@ class Graphql::MinistersIndexQuery
             content_id
             document_type
             first_published_at
+            locale
             public_updated_at
             publishing_app
             rendering_app
--- a/app/queries/graphql/ministers_index_query.rb
+++ b/app/queries/graphql/ministers_index_query.rb
@@ -9,6 +9,14 @@ class Graphql::MinistersIndexQuery
         edition(base_path: "#{@base_path}") {
           ... on MinistersIndex {
             base_path
+            content_id
+            document_type
+            first_published_at
+            public_updated_at
+            publishing_app
+            rendering_app
+            schema_name
+            updated_at
 
             links {
               ordered_cabinet_ministers {
--- a/app/queries/graphql/ministers_index_query.rb
+++ b/app/queries/graphql/ministers_index_query.rb
@@ -19,6 +19,10 @@ class Graphql::MinistersIndexQuery
             schema_name
             updated_at
 
+            details {
+              body
+            }
+
             links {
               ordered_cabinet_ministers {
                 ...basePersonInfo
--- a/app/queries/graphql/ministers_index_query.rb
+++ b/app/queries/graphql/ministers_index_query.rb
@@ -21,6 +21,7 @@ class Graphql::MinistersIndexQuery
 
             details {
               body
+              reshuffle
             }
 
             links {
--- a/spec/features/ministers_spec.rb
+++ b/spec/features/ministers_spec.rb
@@ -16,6 +16,22 @@ RSpec.feature "Ministers index page" do
       expect(page).to have_selector(".gem-c-heading__text", text: I18n.t("ministers.title"))
     end
 
+    %w[
+      govuk:content-id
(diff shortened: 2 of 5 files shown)


spec/features/ministers_spec.rb
```diff
@@ -3,11 +3,7 @@
 RSpec.feature "Ministers index page" do
   include GovukAbTesting::RspecHelpers
 
-  shared_examples "ministers index page" do
-    scenario "returns 200 when visiting ministers page" do
-      expect(page.status_code).to eq(200)
-    end
```
yndajas, line 9: **question:** this scenario seems to be in both Content Store and GraphQL shared examples - is there a reason for not keeping it in the generic shared examples?

spec/features/ministers_spec.rb
```diff
@@ -100,59 +89,80 @@
...
-
-  context "when the GraphQL parameter is not set" do
-    before do
-      visit "/government/ministers"
+    it "does not get the data from GraphQL" do
+      expect(a_request(:post, "#{Plek.find('publishing-api')}/graphql")).not_to have_been_made
     end
 
-    it_behaves_like "ministers index page"
+    it_behaves_like "ministers index page rendered content"
```
yndajas, line 130: **nitpick:** not sure that the new shared example name lends itself to being read after "it behaves like"

app/queries/graphql/ministers_index_query.rb
```diff
@@ -9,6 +9,14 @@ def query
         edition(base_path: "#{@base_path}") {
           ... on MinistersIndex {
             base_path
+            content_id
+            document_type
+            first_published_at
+            public_updated_at
+            publishing_app
+            rendering_app
+            schema_name
```
yndajas: **note:** we also need `document_type` and `schema_name` for the Prometheus labels now (in case it's worth noting)

spec/features/ministers_spec.rb
```diff
@@ -171,6 +171,35 @@
     end
 
     it_behaves_like "ministers index page rendered content"
+
+    context "during a reshuffle" do
+      let(:document) { fetch_graphql_fixture("ministers_index-reshuffle-mode-on") }
```
yndajas, line 176: **suggestion:** given you set the document inside each of the source-specific shared examples (lines 120 and 163), couldn't you move this and the Content Store version into the generic shared example?

app/queries/graphql/ministers_index_query.rb
brucebolt: I've added a note about this to the commit message.

Maybe Publishing API should reject queries without them, or just append them to all GraphQL queries if they aren't included?

spec/features/ministers_spec.rb
brucebolt: There's no reason, so I've moved it into the shared examples.

spec/features/ministers_spec.rb
brucebolt: These override the default document, with something different each time, so I don't think they could be moved into any of the shared examples.

app/queries/graphql/ministers_index_query.rb
yndajas: The latter would be good, but [I'm not sure how easy it would be](https://github.com/alphagov/publishing-api/pull/3287#discussion_r2044562958)

spec/features/ministers_spec.rb
yndajas: Oh I meant the reshuffle scenarios in the reshuffle contexts, not the lets. So you'd have a shared example of "behaves like a ministers index during a reshuffle", then keep the current contexts and lets but then call the shared example inside

brucebolt: I see.  Turns out I actually removed the shared example for this in the first commit, so I've added this back in and used them in the reshuffle commit on this PR.

yndajas (approved):

PR #4077, opened 2025-05-09
Allow `ordered_parent_organisations` links to be an empty array

brucebolt:
The code is currently assuming the `ordered_parent_organisations` links will be nil if there are no parent organisations.

This assumption is not correct, as the link could theoretically be present but with an empty array.  This would lead to the organisations section appearing, but without any organisations listed.

Therefore updating to support both cases, so we only display the organisations section when they are any.

This is important for GraphQL, since we will always be including an empty array, even if there are no links.

[Trello card](https://trello.com/c/5ggdXNEB)


JonathanHallam (approved):
Sweet!

PR #4078, opened 2025-05-09
Include taxonomy meta tags for GraphQL roles

brucebolt:
Some role pages include meta tags representing the taxonomy that the role belongs to.

We therefore need to request the `taxons` links as part of the GraphQL query to retrieve the content for these meta tags.

Adding five levels of taxonomy, [as was done in Frontend](https://github.com/alphagov/frontend/pull/4741/commits/68b98ece36c05258f06fddef7c23e14ce97089f3), since this is the largest level of taxonomy we currently have.

[Trello card](https://trello.com/c/5ggdXNEB)

### DIFF
--- a/app/queries/graphql/role_query.rb
+++ b/app/queries/graphql/role_query.rb
@@ -59,9 +59,41 @@ class Graphql::RoleQuery
               organisations {
                 analytics_identifier
               }
+
+              taxons {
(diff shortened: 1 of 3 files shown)


JonathanHallam (approved):
Looks good to me!

PR #4091, opened 2025-05-23
Replace use of tint mixins with hex variables

jon-kirwan:
## What

Replace use of tint mixins with hex variables

## Why

Dart Sass 1.79.0 introduced support for CSS Color Level 4 color spaces. Consequently, the `govuk-tint` mixin now outputs rgb values with floating-point precision, leading to unsupported property value warnings in some browsers, such as Safari 11.1.1. To maintain compatibility, we're defining equivalent hex colour values as variables instead.

Screenshot 2025-05-14 at 14 08 19

### DIFF
--- a/app/assets/stylesheets/views/_browse.scss
+++ b/app/assets/stylesheets/views/_browse.scss
@@ -5,7 +5,7 @@
 // be run as an A/B test, we need to use a different class name for each view.
 .browse__breadcrumb-wrapper,
 .browse__header-wrapper {
-  background-color: govuk-tint(govuk-colour("blue"), 80%);
+  background-color: $govuk-blue-tint-80;
 }
 
 .browse__breadcrumb-wrapper {


MartinJJones (approved):

app/assets/stylesheets/views/_browse.scss
```diff
@@ -1,11 +1,18 @@
 @import "govuk_publishing_components/individual_component_support";
 
+// Dart Sass 1.79.0 introduced support for CSS Color Level 4 color spaces.
+// Consequently, the `govuk-tint` mixin now outputs rgb values with floating-point precision,
+// leading to unsupported property value warnings in some browsers, such as Safari 11.1.1.
+// To maintain compatibility, we're defining equivalent hex color values as variables.
+
+$govuk-blue-tint-80: #d2e2f1; // govuk-tint(govuk-colour("blue"), 80%)
```
MartinJJones, line 8: Just wondering if this section is needed here, it looks like the `$govuk-blue-tint-80` variable will be available through the `individual_component_support` import

MartinJJones (approved):

PR #4114, opened 2025-06-16
Remove cookies from GraphQL A/B test

brucebolt:
We are currently using the GOV.UK A/B testing framework to run an A/B test for GraphQL.

This relies of users opting into analytics cookies, so we are seeing very little traffic. Especially as these pages represent very small levels of traffic in the first place.

We don't actually need cookies for this A/B test, since the changes are all behind the scenes, so users won't see any difference.

Therefore updating the GraphQL A/B testing code to assign a variant for each request, without needing any cookies.

The initial traffic rates have been selected by using the rate of traffic seen to each document type's GraphQL variant in the last 3 days.  This will be adjusted once the new A/B test is in place.

[Trello card](https://trello.com/c/GRDVWkv1)

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -108,4 +108,9 @@ private
   def render_partial(partial_name, locals = {})
     render_to_string(partial_name, formats: [:html], layout: false, locals:)
   end
+
+  def graphql_ab_test?(graphql_traffic_rate)
+    random_number = Random.rand(1.0)
+    random_number < graphql_traffic_rate
+  end
 end
--- a/app/controllers/roles_controller.rb
+++ b/app/controllers/roles_controller.rb
@@ -1,18 +1,12 @@
 class RolesController < ApplicationController
   around_action :switch_locale
 
-  def show
-    ab_test = GovukAbTesting::AbTest.new(
-      "GraphQLRoles",
-      allowed_variants: %w[A B Z],
-      control_variant: "Z",
-    )
-    @requested_variant = ab_test.requested_variant(request.headers)
-    @requested_variant.configure_response(response)
+  GRAPHQL_TRAFFIC_RATE = 0.02207862 # This is a decimal version of a percentage, so can be between 0 and 1
 
+  def show
     content_item_data = if params[:graphql] == "false"
                           load_from_content_store
-                        elsif params[:graphql] == "true" || @requested_variant.variant?("B")
+                        elsif params[:graphql] == "true" || graphql_ab_test?(GRAPHQL_TRAFFIC_RATE)
                           load_from_graphql
                         else
                           load_from_content_store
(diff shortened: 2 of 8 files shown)


app/controllers/application_controller.rb
```diff
@@ -108,4 +108,9 @@ def breadcrumb_content
   def render_partial(partial_name, locals = {})
     render_to_string(partial_name, formats: [:html], layout: false, locals:)
   end
+
+  def graphql_ab_test?(graphql_traffic_rate)
+    random_number = Random.new.rand(1.0)
```
mike3985, line 113: If we used the class method version of rand, `Random.rand(1.0)`, here I guess we could use standard `allow`s in the specs instead of the sometimes contentious `allow_any_instance_of`

spec/features/world_index_spec.rb
```diff
@@ -99,6 +99,7 @@
 
   context "when the GraphQL parameter is not set" do
     before do
+      allow_any_instance_of(Random).to receive(:rand).with(1.0).and_return(1)
```
mike3985, line 102: Can we remove the `GovukAbTesting::RspecHelpers` mixins from these specs now?

app/controllers/roles_controller.rb
```diff
@@ -1,6 +1,8 @@
 class RolesController < ApplicationController
   around_action :switch_locale
 
+  GRAPHQL_TRAFFIC_RATE = 0.02207862 # This is a decimal version of a percentage, so can be between 0 and 1
+
   def show
     ab_test = GovukAbTesting::AbTest.new(
```
mike3985, line 7: You've left the `GovukAbTesting` setup in this controller?

mike3985 (commented):
You added a line to the application layout when you were first introducing this A/B testing https://github.com/alphagov/collections/commit/888429b7e688a50c7f8350a5e2d04f0ac7b91477#diff-f43fe075643e681b2c01c2f853bb0c4299d135b47fcbd4da96890d521c49e3ebR46

Can that be removed now, too?

app/controllers/application_controller.rb
brucebolt: Yes, I hadn't spotted that there was a `rand` class method on `Random`.

spec/features/world_index_spec.rb
brucebolt: I'll remove this.

brucebolt:
> You added a line to the application layout when you were first introducing this A/B testing [888429b#diff-f43fe075643e681b2c01c2f853bb0c4299d135b47fcbd4da96890d521c49e3ebR46](https://github.com/alphagov/collections/commit/888429b7e688a50c7f8350a5e2d04f0ac7b91477#diff-f43fe075643e681b2c01c2f853bb0c4299d135b47fcbd4da96890d521c49e3ebR46)
> 
> Can that be removed now, too?

I think it can be removed.

hannako (approved):
I think this can go? https://github.com/alphagov/collections/blob/main/app/views/layouts/application.html.erb#L46

PR #4136, opened 2025-07-02
[PP-7232] Increase GraphQL traffic level for roles

brucebolt:
We have been serving 15% of traffic to roles pages using GraphQL for the last 3 hours. There has been no noticeable reduction in performance compared to the previous level.

Therefore increasing the level of traffic served by GraphQL to 50%, to match the Ministers Index and World Index pages.


JonathanHallam (approved):
Ship it!

PR #4179, opened 2025-08-08
[PP-7284] Switch GraphQL method for retrieving content

brucebolt:
We have moved the queries for GraphQL from frontend applications into Publishing API, and instead exposed an API that behaves exactly like Content Store.

Therefore switching to use this new API method.

Depends on https://github.com/alphagov/gds-api-adapters/pull/1360.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -64,13 +64,6 @@ def fetch_fixture(filename)
   JSON.parse(json)
 end
 
-def fetch_graphql_fixture(filename)
-  json = File.read(
-    Rails.root.join("spec", "fixtures", "graphql", "#{filename}.json"),
-  )
-  JSON.parse(json)
-end
-
 def search_api_response(titles_and_links_hash)
   results_array = titles_and_links_hash.to_a.map do |title, link|
     {


app/models/graphql/content_item.rb
```diff
@@ -0,0 +1,11 @@
+class Graphql::ContentItem < ContentItem
+  def self.find!(base_path)
+    content_item = Services.publishing_api.graphql_live_content_item(base_path)
+    content_item_hash = content_item.to_h
+    content_item_hash["cache_control"] = {
+      "max_age" => content_item.cache_control["max-age"],
+      "public" => !content_item.cache_control.private?,
+    }
```
richardTowers, line 8: I'm curious why this is necessary

richardTowers (approved):
Approved with one comment which is more of a curiosity than a request for changes.

app/models/graphql/content_item.rb
brucebolt: It appears to be possible to override the maximum caching time in content items, through the `max_age`.

Although I [haven't added support](https://github.com/alphagov/publishing-api/blob/5f39effbfff806cce6da5580ee2a402c9c13ac28/app/controllers/graphql_controller.rb#L157) for this to Publishing API yet as none of the content types we've migrated use this, so probably don't need it here yet.  But decided to leave it in, so it's not forgotten when we add additional content types and to remain consistent with the equivalent code for Content Store.

Similar reasons for leaving in the `public` key, even though we're not yet serving drafts.

brucebolt: The same happens for Content Store too:
https://github.com/alphagov/collections/blob/8342d039e0a94ba07eede6b3f70f6e2b3f6b7929/app/models/content_item.rb#L7-L10

I'm not quite sure why Collections works like this, but the [Content Store side of things](https://github.com/alphagov/content-store/blob/029bcc31d5f1efcb2696b637acfd80b1fc98026b/app/controllers/content_items_controller.rb#L159-L163) either uses default values or (if set) the `max_cache_time` from the content item to set this header.

PR #4304, opened 2025-12-03
[PP-7471] Send 1% traffic for Topical Events to GraphQL

brucebolt:
We have confirmed that the HTML rendered version of Topical Events is the same for both GraphQL and Content Store.

Therefore allowing 1% of traffic to be served from GraphQL. This will be increased in later commits, once we confirm the response times in production are reasonable.

Note: the tests have been copied from the `RolesController` with almost no changes.

### DIFF
--- a/app/controllers/topical_events_controller.rb
+++ b/app/controllers/topical_events_controller.rb
@@ -1,9 +1,18 @@
 class TopicalEventsController < ApplicationController
+  include PrometheusSupport
+
   enable_request_formats show: :atom
 
+  GRAPHQL_TRAFFIC_RATE = 0.01 # This is a decimal version of a percentage, so can be between 0 and 1
+
   def show
-    path = topical_event_path(params[:name])
-    @topical_event = TopicalEvent.find!(path)
+    if params[:graphql] == "false"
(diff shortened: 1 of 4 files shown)


app/controllers/topical_events_controller.rb
```diff
@@ -1,9 +1,18 @@
 class TopicalEventsController < ApplicationController
+  include PrometheusSupport
+
   enable_request_formats show: :atom
 
+  GRAPHQL_TRAFFIC_RATE = 0.01 # This is a decimal version of a percentage, so can be between 0 and 1
```
AgaDufrat: I realise this was copied over for consistency within this code base. 
I was wondering if it would be better to make those strings to prepare to use env vars. Example: https://github.com/alphagov/govuk-helm-charts/blob/e99e8a384bbe86bc62e738a3c7778374dd898c10/charts/app-config/values-integration.yaml#L1394-L1395

brucebolt: I'm going to raise another PR to switch this app to use env vars, so I'll switch to using strings in that PR.

PR #4307, opened 2025-12-04
[PP-7471] Retrieve GraphQL traffic rates from environment variables

brucebolt:
The rate of GraphQL traffic for each schema is currently hardcoded into the application. This means we have the same traffic rate in every environment.

Switching to use environment variables, so we can adjust traffic rates to be different across the environments.

This also makes changing the traffic levels faster, as we don't need to deploy a new version of the application.

Depends on https://github.com/alphagov/govuk-helm-charts/pull/3838.

This mirrors the same change that was made in Frontend: https://github.com/alphagov/frontend/pull/5158.

### DIFF
--- a/app/controllers/topical_events_controller.rb
+++ b/app/controllers/topical_events_controller.rb
@@ -3,7 +3,7 @@ class TopicalEventsController < ApplicationController
 
   enable_request_formats show: :atom
 
-  GRAPHQL_TRAFFIC_RATE = 0.01 # This is a decimal version of a percentage, so can be between 0 and 1
+  GRAPHQL_TRAFFIC_RATE = Rails.application.config.graphql_traffic_rates.fetch("topical_event", 0)
 
   def show
     if params[:graphql] == "false"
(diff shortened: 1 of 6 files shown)


app/controllers/topical_events_controller.rb
```diff
@@ -3,12 +3,10 @@ class TopicalEventsController < ApplicationController
 
   enable_request_formats show: :atom
 
-  GRAPHQL_TRAFFIC_RATE = 0.01 # This is a decimal version of a percentage, so can be between 0 and 1
+  GRAPHQL_TRAFFIC_RATE = Rails.application.config.graphql_traffic_rates.fetch("topical_event", 0)
 
   def show
-    if params[:graphql] == "false"
```
mike3985: 👀

config/initializers/graphql_traffic_rates.rb
```diff
@@ -0,0 +1,4 @@
+Rails.application.config.graphql_traffic_rates = ENV
+  .select { |key, _| key.starts_with?("GRAPHQL_RATE_") }
+  .transform_keys { |key| key.gsub(/^GRAPHQL_RATE_/, "").downcase }
```
mike3985: We can use just a normal `sub` here instead of `gsub`, right?

app/controllers/topical_events_controller.rb
brucebolt: Oops.  Have restored that back into the controller.

PR #4507, opened 2026-05-11
Fix missing default image on organisation pages

jon-kirwan:
## What

Fix missing default image on organisation pages

## Why

https://gds.slack.com/archives/C052X8S2P8A/p1778247553902209

### Before and after

Screenshot 2026-05-11 at 13 43 30

PR #4508, opened 2026-05-11
Add a test for multiple image sources in latest news on No. 10 page

jon-kirwan:
## What

Add a test for multiple image sources in latest news on No. 10 page

## Why

https://github.com/alphagov/collections/pull/4507


KludgeKML (approved):
LGTM!

PR #4522, opened 2026-05-20
Handle content store errors on organsiation pages

catalinailie:
Organisation pages return 404's, despite having a content item. This happens when a recent document showing on the page is returning 404. 

The GDS API call for content store is raising an error when the page is not found, instead of returning a http response. We need to catch this error in order for the org page to work even if the linked pages are not available.

Jira card: https://gov-uk.atlassian.net/browse/NAV-18939


beccapearce (approved):
Gorgeous! ✨

PR #4587, opened 2026-07-17
Apply consistent spacing for taxon page breadcrumbs

jon-kirwan:
## What

The spacing between breadcrumbs on taxon pages is larger than the spacing used elsewhere. This change updates the spacing to make it consistent with other pages e.g. https://www.gov.uk/browse.

## Why

To address a style inconsistency.

### Before, after (desktop)

https://www.gov.uk/transport/driving-and-motorcycle-tests

Screenshot 2026-07-17 at 11 31 21

### Before, after (mobile)

Screenshot 2026-07-17 at 11 30 41

### Comparison with browse page (after changes)

https://www.gov.uk/transport/driving-and-motorcycle-tests and https://www.gov.uk/browse

Screenshot 2026-07-17 at 11 28 33

## Anything else

The taxon and browse page breadcrumb styles could likely be consolidated. However, the correct spacing should be confirmed with design, so that change should be handled separately.

### DIFF
--- a/app/assets/stylesheets/views/_taxons.scss
+++ b/app/assets/stylesheets/views/_taxons.scss
@@ -10,7 +10,10 @@
 
 .taxon-page__breadcrumbs-wrapper {
   background-color: $govuk-blue-tint-80;
-  padding: govuk-spacing(3) 0 govuk-spacing(4);
+  // Setting the overflow to auto is required to prevent the breadcrumb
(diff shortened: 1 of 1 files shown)


luke-booker (approved):
Nice work 👍 Checked out locally and confirmed 'Driving and motorcycle tests' breadcrumb consistency against other pages.

PR #4600, opened 2026-07-22
Update the Civil Service organisation page to use its official brand colour

luke-booker:
## What
In this PR we're updating the [Civil Service organisation page](https://www.gov.uk/government/organisations/civil-service) to use its [official brand colour](https://www.brand.gov.uk/department/civil-service/) (#b2292e) for borders, removing any hardcoded overrides. This brings the page in line with the brand guidelines and the Design System.

## Why
Historically, an override was put in place to force Civil Service page borders to black because the anchor links on the page were also red, causing a visual clash.

Jira card: https://gov-uk.atlassian.net/browse/NAV-19016

## Visual Changes

| Before | After |
|--------|--------|
| Screenshot 2026-07-22 at 15 49 58 | Screenshot 2026-07-22 at 15 50 24 | 
| Screenshot 2026-07-23 at 14 54 00 | Screenshot 2026-07-23 at 14 53 53 |

### DIFF
--- a/app/views/organisations/_corporate_information.html.erb
+++ b/app/views/organisations/_corporate_information.html.erb
@@ -9,7 +9,7 @@
       } %>
     <% end %>
   <% if @show.corporate_information[:corporate_information_links][:items].any? && !@organisation.is_no_10? %>
-    <div class="<%= "organisation__section-border-top" if @organisation.is_civil_service? %>">
+    <div class="<%= "brand--civil-service brand__border-color organisation__brand-border-top govuk-!-static-padding-top-3" if @organisation.is_civil_service? %>">
       <%= render "shared/featured_links", @show.corporate_information[:corporate_information_links].merge({ ga4_data: { section: t('organisations.corporate_information', locale: :en) } }) %>
     </div>
   <% end %>
--- a/app/assets/stylesheets/views/_organisation.scss
+++ b/app/assets/stylesheets/views/_organisation.scss
@@ -9,14 +9,6 @@
   overflow-wrap: break-word;
 }
 
-// This is only used to apply to promotional orgs where the border colour is always black
-// regardless of the link colour. This can be removed if we add the black border
-// colour as branding to the organisation.
-.organisation__section-border-top {
-  border-top: 5px solid govuk-colour("black");
-  padding-top: govuk-spacing(3);
-}
-
 .organisation__corporate-information {
   .organisation__float-section {
     @include govuk-media-query($from: tablet) {


app/views/organisations/_corporate_information.html.erb
```diff
@@ -9,9 +9,7 @@
       } %>
     <% end %>
   <% if @show.corporate_information[:corporate_information_links][:items].any? && !@organisation.is_no_10? %>
-    <div class="<%= "organisation__section-border-top" if @organisation.is_civil_service? %>">
-      <%= render "shared/featured_links", @show.corporate_information[:corporate_information_links].merge({ ga4_data: { section: t('organisations.corporate_information', locale: :en) } }) %>
-    </div>
+    <%= render "shared/featured_links", @show.corporate_information[:corporate_information_links].merge({ ga4_data: { section: t('organisations.corporate_information', locale: :en) } }) %>
```
MartinJJones: After looking at this further, I think we need to keep the top border above the "About us" link, but ensure the colour is used is the Civil Service brand colour, one way to do this would be to keep the div element used but update the CSS classes from:

```<div class="<%= "organisation__section-border-top" if @organisation.is_civil_service? %>">```

to:

```
<div class="<%= "brand--civil-service brand__border-color organisation__brand-border-top govuk-!-static-padding-top-3" if @organisation.is_civil_service? %>">
```

This approach is consistent with the top borders added on other sections of the organisation page, the news section for example - https://github.com/alphagov/collections/blob/main/app/views/organisations/_latest_news.html.erb#L10.

The `govuk-!-static-padding-top-3` added is to ensure the previous padding-top: 15px remains in all screen sizes, further info on this in the design system - https://design-system.service.gov.uk/styles/spacing/#static-spacing-override-classes

From looking at the template itself, I do wonder if this can be simplified, I think this is outside the scope of this ticket for now and likely something we will visit when merging collections with frontend in the future.

MartinJJones (commented):
The removal of the `organisation__section-border-top` CSS class looks good to me, I've just added one comment on keeping the border and one approach for doing this, happy to catch up further on it any time.

app/views/organisations/_corporate_information.html.erb
luke-booker: Thank you for the review @MartinJJones. 

As discussed, in this PR I'll maintain the top border for the 'Corporate information' links section and colour that according to the organisation brand. 

I'll raise the various template improvements we spoke about in a separate ticket.

MartinJJones: Nice, thanks for creating the ticket as well 👍

MartinJJones (approved):

PR #4620, opened 2026-08-05
Fix ARIA list item hierarchy

### LINKED ISSUE #4502: HTML Listitem Errors
When validating the HTML on the historical people pages the following error occurs multiple times. 

Error: An element with role=listitem must be contained in, or owned by, an element with the role value list, or group.

See the [errors here ](https://validator.w3.org/nu/?level=warning&doc=https%3A%2F%2Fwww.gov.uk%2Fgovernment%2Fhistory%2Fpast-chancellors) when checking [Past Chancellors page](https://www.gov.uk/government/history/past-chancellors).

The error can be located to one of the [ app template found here ](https://github.com/alphagov/collections/blob/main/app/views/historic_appointments/_grouped_historical_people.html.erb#L17).

luke-booker:
## What

- Added role="list" to the outer wrapper container of list items in the section
- Added role="presentation" to the row element within that section
- Fixed missing quotes around the dynamic ID attribute on the wrapper div

## Why

W3C / HTML validators fail with:
`An element with role=listitem must be contained in, or owned by, an element with the role value list, or group.`
The grid layout requires intermediate row divs, without explicit ARIA roles, these intermediate divs break the expected parent-child hierarchy in the accessibility tree between the list container and the role="listitem" elements.

Adding role="presentation" instructs screen readers and assistive technologies to ignore the purely presentational grid row wrappers while preserving the CSS grid visual layout. This allows assistive tools to map the items directly to the parent role="list".

Jira card: https://gov-uk.atlassian.net/browse/NAV-18949

Fixes: https://github.com/alphagov/collections/issues/4502

### Before

Screenshot 2026-08-05 at 16 49 26

### After

Screenshot 2026-08-05 at 16 49 43

### DIFF
--- a/app/views/historic_appointments/_grouped_historical_people.html.erb
+++ b/app/views/historic_appointments/_grouped_historical_people.html.erb
@@ -9,9 +9,9 @@
     <hr class="govuk-section-break govuk-section-break--m govuk-section-break--visible">
   </div>
 </div>
-<div id=<%= "historical-people-#{heading.downcase.gsub(' ', '-')}" %>>
+<div id="<%= "historical-people-#{heading.downcase.gsub(' ', '-')}" %>" role="list">
   <% people.each_slice(4) do |row| %>
-    <div class="govuk-grid-row historic-people-index__section-row">
+    <div class="govuk-grid-row historic-people-index__section-row" role="presentation">
       <% row.map(&:deep_symbolize_keys).each do |info| %>
         <% if include_images %>
           <div role="listitem" class="govuk-grid-column-one-quarter">


MartinJJones (commented):
Nice work, the changes look good to me and looks like it will fix the immediate issue

One thing I'm wondering is if we should create a follow up card to try that looks to remove the need for `role="listitem"` altogether by using semantic markup, further info here - https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/listitem_role#best_practices.

I think it would be worth checking with @calumryan on this to see if we should make any further changes now, or best to capture on a new card

luke-booker:
> Nice work, the changes look good to me and looks like it will fix the immediate issue
> 
> One thing I'm wondering is if we should create a follow up card to try that looks to remove the need for `role="listitem"` altogether by using semantic markup, further info here - https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/listitem_role#best_practices.
> 
> I think it would be worth checking with @calumryan on this to see if we should make any further changes now, or best to capture on a new card

Thanks for the review @MartinJJones. I agree that native ul and li markup should be the ideal long term fix. This will involve replacing the ARIA roles and refactoring the .govuk-grid-row layout.

calumryan:
If there's any other instances then it might make sense to add this to another card. Agree overall it will be better in the long term if everything uses semantic HTML instead and prevent more opportunities for mistakes in future implementations of this pattern.

MartinJJones:
Thanks @calumryan, I've searched alphagov repos for any other mention of `role="listitem"` and this only appears to be used in this application in the 2 templates to be updated as part of this PR which is reassuring - https://github.com/search?q=org%3Aalphagov+role%3D%22listitem%22&type=code

@luke-booker nice work on the changes, I'll approve this PR, just wanted to check if you are OK to create a new card to add to the backlog to look at the longer term solution of using semantic elements instead of the role attributes to help avoid this being copied and used elsewhere, thanks

MartinJJones (approved):

PR #4661, opened 2026-09-09
Migrate from standardx to standard

MartinJJones:
## What

Migrate from standardx to standard

## Why

Standardx has now been updated for several years and is constantly needs security updates and patches. Standard has better support and is now safe to migrate to as we can now use const and let in our JavaScript code

Jira card: https://gov-uk.atlassian.net/browse/NAV-19101


jon-kirwan (approved):
Looks good to me!

